PR #1, opened 2022-06-29
Setup publishing for conductor on mavenCentral

amanjeetsingh150:


Leland-Takamine (approved):

Leland-Takamine:
This only publishes a snapshot for now is that intended?

dmitry-zaitsev:
Nice!

As a nice-to-have, can we try `jreleaser` for Maven as well? I use it for homebrew, so it would be great to unify deployment pipeline

dmitry-zaitsev (approved):

PR #2, opened 2022-06-29
Homebrew deployment

dmitry-zaitsev:


amanjeetsingh150 (approved):

amanjeetsingh150:
Should we also setup yaml for normal stable release as @Leland-Takamine suggested [here](https://github.com/mobile-dev-inc/conductor/pull/1#issuecomment-1170184399). Should we do snapshot first and then a stable 0.0.1

PR #13, opened 2022-07-08
Cleanup task for conductor 1.0.1

amanjeetsingh150:


build.gradle.kts
```diff
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath("com.android.tools.build:gradle:7.0.4")
+        classpath("com.android.tools.build:gradle:4.2.1")
     }
```
amanjeetsingh150, line 9: Downgrading since this 7+ version has Java 11 validations.

.github/workflows/publish-release.yml
```diff
@@ -19,11 +19,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install JDK 11
+      - name: Install JDK 8
```
amanjeetsingh150, line 22: This would mean we release conductor artifacts on JDK8 and it would readily be available on java 8 projects. Should not be a problem for java 8+ also.

examples/samples/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+
+repositories {
+    mavenCentral()
+    mavenLocal()
+}
+
+dependencies {
+    implementation 'dev.mobile:conductor-client:1.0.0-SNAPSHOT'
+    implementation 'dev.mobile:conductor-orchestra:1.0.0-SNAPSHOT'
+    implementation 'dev.mobile:conductor-ios:1.0.0-SNAPSHOT'
```
amanjeetsingh150, line 13: These artifacts are needed to consume conductor in Kotlin/java.

PS: PR is breaking because this 1.0.0-SNAPSHOT was locally published. I will fix it by releasing as soon as PR is approved.

examples/samples/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,76 @@
...
+
+fun main() {
+    executeAndroidCommands()
+    executeIOSCommands()
+}
+
+private fun executeAndroidCommands() {
+    val dadb = Dadb.create("localhost", 5555)
+    val conductor = Conductor.android(dadb)
+    val launchAppCommand = ConductorCommand(launchAppCommand = LaunchAppCommand("dev.mobile.sample"))
```
amanjeetsingh150, line 38: I have tested these commands in the android app and iOS in examples in module. Should we include the APK and .app in the resources of the sample project.

dmitry-zaitsev (approved):

PR #14, opened 2022-07-12
Wait until view is unobsucred by any other views before tapping on it

dmitry-zaitsev:


conductor-client/src/main/java/conductor/Conductor.kt
```diff
@@ -71,6 +74,18 @@ class Conductor(private val driver: Driver) : AutoCloseable {
         }
     }
 
+    private fun waitUntilVisible(element: UiElement) {
+        repeat(10) {
+            if (!ViewUtils.isVisible(viewHierarchy(), element.treeNode)) {
+                LOGGER.info("Element is not visible yet. Waiting.")
+                Thread.sleep(1000)
```
artem888, line 81: nit : could use coroutines here to avoid blocking the thread

artem888 (approved):
👍

conductor-client/src/main/java/conductor/Conductor.kt
```diff
@@ -71,6 +74,18 @@ class Conductor(private val driver: Driver) : AutoCloseable {
         }
     }
 
+    private fun waitUntilVisible(element: UiElement) {
+        repeat(10) {
```
artem888, line 78: nit : could be a const

conductor-client/src/main/java/conductor/Conductor.kt
dmitry-zaitsev: We would only benefit from that if the rest of Conductor code would use the same coroutine scope everywhere. As it stands, our thread usage is pretty minimal, so this won't give us any gains

conductor-client/src/main/java/conductor/utils/ViewUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private fun getElementAt(
+        node: TreeNode,
+        x: Int,
+        y: Int
+    ): TreeNode? {
+        return node
+            .children
+            .asReversed()
```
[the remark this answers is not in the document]
dmitry-zaitsev: Views in `children` are listed in the order they are drawn on the screen. When tapping on the screen, normally the top-most view is the first one to respond so to reproduce that behaviour we are traversing views backwards

conductor-client/src/main/java/conductor/utils/ViewUtils.kt
```diff
@@ -0,0 +1,52 @@
+package conductor.utils
```
amanjeetsingh150, line 1: nit: Should we add copyright here. I added a setting in IntelliJ to produce copyright to each new file in the conductor project. Can we share the setting in some form? Not blocker for this PR though.

dmitry-zaitsev: @amanjeetsingh150 sounds good, could you add this setting to the repo? Also, are there perhaps any Gradle plugins which do the same?

conductor-client/src/main/java/conductor/Conductor.kt
dmitry-zaitsev: It would be a bit tricky to name that constant as it is used in combination with 1000ms in `Thread.sleep`, so I would rather not do it in this specific case

PR #24, opened 2022-07-14
Removing InstrumentationThread. Keeping an open shell session instead of reading endlessly

dmitry-zaitsev:
We were using `dadb.shell()` which reads from socket non-stop, hogging resources


Leland-Takamine:
lgtm

PR #27, opened 2022-07-18
Add swipe command

axelniklasson:


conductor-orchestra-models/build.gradle
```diff
@@ -9,3 +9,7 @@ plugins.withId("com.vanniktech.maven.publish") {
         sonatypeHost = "S01"
     }
 }
+
+dependencies {
+    implementation project(path: ':conductor-client')
```
amanjeetsingh150, line 14: Why is client needed in models?

axelniklasson: I need to use the `Point` model which is defined in` Point.kt` in the client module -- did not want to change too much, but maybe it is worth pulling that out of client and define it in the models module instead?

amanjeetsingh150: Hmm, I think we might need to change the name of orchestra-models module because it will also have client models that orchestrator do not directly use. Let's keep it as is for now and scope this out in separate PR.

PR #41, opened 2022-07-26
Feature: launchApp now allows to clear app state

dmitry-zaitsev:


maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
```diff
@@ -193,6 +195,18 @@ class IdbIOSDevice(
         }
     }
 
+    override fun clearAppState(id: String): Result<Unit, Throwable> {
```
artem888, line 198: we already have `clearAppData`, which looks pretty similar to me

artem888 (approved):

maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
dmitry-zaitsev: @artem888 you mean in internal repo, right?

artem888: ah yeah, i mixed it up, never mind

PR #42, opened 2022-08-01
Add Android utilities for app file restoration

Leland-Takamine:


maestro-client/src/main/java/maestro/android/AndroidBuildToolsDirectory.kt
```diff
@@ -0,0 +1,60 @@
+package maestro.android
+
+import java.io.File
+import java.util.regex.Pattern
+
+internal object AndroidBuildToolsDirectory {
+
+    fun findBuildToolsDir(androidHome: File): File {
+        val buildToolsParent = File(androidHome, "build-tools")
```
dmitry-zaitsev, line 9: This makes Maestro dependent on Android SDK, contrary to what we've been saying. Let's make sure this is not a required step for running a simple flow on an already debuggable app

maestro-client/src/main/java/maestro/android/ApkDebuggable.kt
```diff
@@ -0,0 +1,110 @@
...
+                    override fun child(ns: String?, name: String?): NodeVisitor {
+                        if (name != "application") return super.child(ns, name)
+
+                        return object : NodeVisitor(super.child(ns, name)) {
+
+                            override fun end() {
+                                super.attr(
+                                    "http://schemas.android.com/apk/res/android",
+                                    "debuggable",
+                                    16842767,
```
dmitry-zaitsev, line 66: Some comment on what those number mean would be nice

maestro-client/src/main/java/maestro/utils/TemporaryDirectory.kt
```diff
@@ -0,0 +1,16 @@
+package maestro.utils
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+internal object TemporaryDirectory {
+
+    fun use(block: (tmpDir: Path) -> Unit) {
```
dmitry-zaitsev, line 8: Nice!

dmitry-zaitsev (approved):

maestro-client/src/main/java/maestro/android/AndroidBuildToolsDirectory.kt
Leland-Takamine: Good point. I'll add some logic that detects whether the app is already debuggable and if so, `ApkDebuggable.enable` can just copy the src apk to dst location.

dmitry-zaitsev: Awesome! 👍

PR #44, opened 2022-08-02
Add support for initFlow config

Leland-Takamine:
Support the `initState` config key


maestro-cli/src/main/java/maestro/cli/runner/FileWatcher.kt
```diff
@@ -0,0 +1,54 @@
+package maestro.cli.runner
+
+import com.sun.nio.file.SensitivityWatchEventModifier
+import java.io.File
+import java.nio.file.FileSystems
+import java.nio.file.Path
+import java.nio.file.StandardWatchEventKinds
+import java.nio.file.WatchKey
+
+class FileWatcher {
```
dmitry-zaitsev, line 10: @Leland-Takamine should this be an `object` too perhaps?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCommandReader.kt
```diff
@@ -38,25 +42,73 @@ class YamlCommandReader : CommandReader {
...
-            )
+    // Files to watch for changes. Includes any referenced files.
+    fun getWatchFiles(flowFile: File): List<File> {
+        return listOfNotNull(
+            flowFile,
+            getInitFlowFile(flowFile)
+        ).filter { it.parentFile.isDirectory }
+    }
+
+    fun readCommands(flowFile: File): Pair<List<MaestroCommand>, List<MaestroCommand>> {
```
dmitry-zaitsev, line 53: It should be a responsibility of the caller to extract init commands. This class is only concerned with parsing commands and returning them back

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -0,0 +1,77 @@
...
+        val view = ResultView()
+
+        fun runFlow(): Int {
+            val result = runCatching(view) {
+                MaestroCommandRunner.run(maestro, view, flowFile)
+            }
+            return if (result == true) 0 else 1
+        }
+
+        if (!continuous) {
```
dmitry-zaitsev, line 31: I would've preferred to have single and continuous runners in separate files as they have a fair number of differences. I.e. in Single runner, we would actually not need to erase the screen as it will be used for running in CI (so the `view` might end up being different)

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -0,0 +1,77 @@
...
+import kotlin.concurrent.thread
+
+object TestRunner {
+
+    fun run(
+        maestro: Maestro,
+        flowFile: File,
+        continuous: Boolean,
+    ): Int {
+        AnsiConsole.systemInstall()
```
dmitry-zaitsev, line 19: Thinking about it, those 2 lines should probably be a part of the `ResultView`

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -0,0 +1,77 @@
...
+            return runFlow()
+        }
+
+        val fileWatcher = FileWatcher()
+
+        var ongoingTest: Thread? = null
+        do {
+            if (ongoingTest != null) {
+                ongoingTest.interrupt()
+                ongoingTest.join()
```
dmitry-zaitsev, line 41: Nice, forgot that `join` was a thing

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCommandReader.kt
```diff
@@ -38,25 +42,73 @@ class YamlCommandReader : CommandReader {
...
             }
         }
+    }
+
+    private fun readCommandsUnsafe(source: Source) = mapper.readValue(
+        source.buffer().inputStream(),
+        object : TypeReference<List<YamlFluentCommand>>() {}
+    ).map { it.toCommand() }
+
+    private fun getInitCommands(
```
dmitry-zaitsev, line 78: Could you make it public? I would benefit from having access to this function on the backend

maestro-cli/src/main/java/maestro/cli/runner/FileWatcher.kt
Leland-Takamine: This class holds state so it needs to be an instantiable class

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCommandReader.kt
Leland-Takamine: Updated

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
Leland-Takamine: Updated

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCommandReader.kt
Leland-Takamine: Updated

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
Leland-Takamine: Split into two separate methods

dmitry-zaitsev (approved):

PR #46, opened 2022-08-03
Proper initFlow state restoration

Leland-Takamine:
Also adds a nice convenience for restarting a test in continuous mode by pressing ENTER.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -23,6 +23,7 @@ import maestro.Maestro
 import maestro.orchestra.MaestroCommand
 import maestro.orchestra.MaestroInitFlow
 import maestro.orchestra.Orchestra
+import java.io.File
 import java.util.IdentityHashMap
 
 object MaestroCommandRunner {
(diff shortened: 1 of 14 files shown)


maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -82,23 +83,39 @@ object MaestroCommandRunner {
...
         }
 
-        if (skipInitFlow) {
-            initCommands.forEach { initCommandStatuses[it] = CommandStatus.COMPLETED }
+        val initAppState: File? = if (initFlow == null) {
+            null
         } else {
-            executeCommands(initCommands, initCommandStatuses)
-        }
+            val state = if (cachedAppState == null) {
```
dmitry-zaitsev, line 92: How could we reuse that logic in our backend runner?

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
```diff
@@ -76,6 +77,12 @@ class FakeDriver : Driver {
         events.add(Event.LaunchApp(appId))
     }
 
+    override fun stopApp(appId: String) {
```
dmitry-zaitsev, line 80: Would be great to have an integration test that goes through init flow

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
Leland-Takamine: As discussed offline - This logic is actually dependent on the context. Not much can be reused.

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
Leland-Takamine: Actually the test doesn't really fit into our current integration tests because the responsibility of running of the init commands are left up to the caller. The thing that needs to be tested is whether the init commands are parsed properly which is covered in existing tests. Running the init commands from there is covered by existing integration tests.

PR #57, opened 2022-08-17
Allow cli user to provide host and port

berikv:


maestro-cli/src/main/java/maestro/cli/util/MaestroFactory.kt
```diff
@@ -26,21 +26,26 @@ import dadb.Dadb
...
-        return when (target) {
-            "android" -> {
-                val dadb = Dadb.discover("localhost")
-                if (dadb == null) {
-                    println("No Android devices found")
-                    throw IllegalStateException()
-                }
+        if (target == null) {
+            findDevice()
+        } else if (target.startsWith("android")) {
```
dmitry-zaitsev, line 31: I believe `target` argument is not widely used at the moment. How about instead of using `android` or `ios`, we will provide `host:port` and try to infer platform based on that?

berikv: Wouldn't it be very confusing if you accidentally connect to the wrong device platform?

berikv: Otherwise, I'm happy to adjust. So then it'd be:
`maestro --target localhost:10882` for iOS
and
`maestro --target localhost` for android

berikv: How about spelling it all out:
`maestro --platform ios --host localhost --port 10882`

dmitry-zaitsev: Yeah, I like the explicit approach. If port is not provided, we could check known ports implicitly (i.e. 10882, 10883 and then simply `Dadb.discover()` if those are not available)

dmitry-zaitsev (approved):
Nice 👍

Issue #56, opened 2022-08-16
TapOn toolbar item clicks on notification instead


sunilson:
Directly after starting our app a notification is shown because a foreground service is started. But also directly after starting the app I want to click on an item in the toolbar. Unfortunately most of the time maestro clicks on the notification instead of waiting until the notification is gone and the toolbar item is not obscured. Can I manually add some waiting time here or how can I handle this? Thanks!

dmitry-zaitsev:
Thanks for sharing the context! We have some ideas on how to implement it and will address it in the next few days.

dmitry-zaitsev:
@sunilson we have published a new Maestro version (1.3.0) that introduces an `assertNotVisible` command (https://maestro.mobile.dev/reference/assertions)

It will wait for a view to disappear within a certain timeout before moving on to the next command

PR #61, opened 2022-08-17
assertNotVisible command

dmitry-zaitsev:
Addresses #56

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -153,7 +153,7 @@ class Orchestra(
     private fun assertNotVisible(selector: ElementSelector, timeoutMs: Long = lookupTimeoutMs) {
         val result = MaestroTimer.withTimeout(timeoutMs) {
             try {
-                findElement(selector)
+                findElement(selector, timeoutMs = 2000L)
 
                 // If we got to that point, the element is still visible.
                 // Returning null to keep waiting.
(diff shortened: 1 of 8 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -141,6 +141,22 @@ class Orchestra(
 
     private fun assertCommand(command: AssertCommand) {
         command.visible?.let { assertVisible(it) }
+        command.notVisible?.let { assertNotVisible(it) }
+    }
+
+    private fun assertNotVisible(selector: ElementSelector) {
+        try {
+            assertVisible(selector)
```
Leland-Takamine, line 149: With this implementation:

* If the view does not exist at the time of calling the method, we will wait for assertVisible to time out before returning
* If the view does exist at the time of calling the method, we will throw an assertion failure immediately

I believe we want the inverse behavior

dmitry-zaitsev: Oh snap, yes you are right

dmitry-zaitsev: Addressed that. Now the caveat is - we will wait _either way_ whenever view is visible or not. The rationale is:

- If view is not currently visible, what if it will show up soon?
- If view is currently visible, what if it will disappear soon?

Leland-Takamine: This is going to introduce a ton of lag in almost every interaction won't it?

dmitry-zaitsev: Reduced waiting time in case if view is currently not visible to 2 seconds

Leland-Takamine (approved):

PR #66, opened 2022-08-18
Improve logic for autoselecting platform

berikv:
This aligns better with the logic that was in v1.2.3

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/util/MaestroFactory.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/MaestroFactory.kt
@@ -24,37 +24,48 @@ import maestro.utils.SocketUtils.isPortOpen
 import dadb.Dadb
 
 object MaestroFactory {
-    @Suppress("NAME_SHADOWING")
+    private const val defaultHost = "localhost"
+    private const val idbPort = 10882
+
     fun createMaestro(platform: String?, host: String?, port: Int?): Maestro {
-        val host = host ?: "localhost"
-        val port = port ?: 10882
 
         return when (platform) {
-            null -> {
-                val dadb = Dadb.discover(host)
-
-                if (dadb != null) {
-                    if (isPortOpen(host, port)) {
-                        throw IllegalStateException("Multiple devices found. Specify one with --platform <android|ios> --host and/or --port")
-                    }
-                    return Maestro.android(dadb)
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/util/MaestroFactory.kt
```diff
@@ -24,43 +24,38 @@ import maestro.utils.SocketUtils.isPortOpen
...
         val host = host ?: "localhost"
         val port = port ?: 10882
 
         return when (platform) {
             null -> {
-                try {
-                    connectAndroid(host)
-                } catch (e: Exception) {
-                    connectIos(host, port)
+                val dadb = Dadb.discover(host)
```
dmitry-zaitsev, line 34: We need to respect a `port` provided by user (if any)

maestro-cli/src/main/java/maestro/cli/util/MaestroFactory.kt
```diff
@@ -24,43 +24,38 @@ import maestro.utils.SocketUtils.isPortOpen
...
+
+                if (dadb != null) {
+                    if (isPortOpen(host, port)) {
```
dmitry-zaitsev, line 37: If user provided their own port for Android device - this feel fail, which is not what we want

berikv:
In 1.2.3 the auto-detect-platform logic was:
```
if (discover adb) {
  device = Device(adb)
}

if (portIsOpen(idbPort) {
  if (device != null) { throw "Multiple devices available" }
  device = Device(iOS(idbPort))
}

if (portIsOpen(idbPort + 1) {
  if (device != null) { throw "Multiple devices available" }
  device = Device(iOS(idbPort + 1))
}

return device ?: throw "No device found"
```

Now that we support passing the `host` and `port` to maestro, this logic can't be used anymore.

I suggest:
```
try { return createAndroid(host, port) }
catch (androidException) {
  try { return createIos(host, port) }
  catch (iosException) { throw "No devices found" }
}
```
createAndroid() will use Dad.create if the port is not null and Dadb.discover otherwise, and throw on null.

The downside is that in "autodetect" mode, it'll swallow the errors.
To solve for that, we can have the error message ask the user to pass "--platform <android|ios>".

@Leland-Takamine Does this logic work for your use case?

Leland-Takamine (approved):
Tested and works when running an Android emulator

Issue #83, opened 2022-08-23
Support for long press


aitorvs:
We've been trying maestro intertnally and it's great btw!

Some of our tests require long press actions but I don't think that's currently supported. Is it something we missed in the documentation? 
Or if indeed not supported, any plans for it?

dmitry-zaitsev:
It is indeed not supported yet, but it is an easy feature to add. You can expect it around this week

aitorvs:
You're awesome! thanks a lot

dmitry-zaitsev:
Since version 1.3.6 you can use a `longPressOn` command. [Documentation](https://maestro.mobile.dev/reference/tap-on-view#long-press)

```
- longPressOn: Hello World
```

aitorvs:
This is awesome!!

PR #87, opened 2022-08-24
Feture: long press

dmitry-zaitsev:
Addresses #83

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -158,6 +158,10 @@ class AndroidDriver(
         ) ?: throw IllegalStateException("Response can't be null")
     }
 
+    override fun longPress(point: Point) {
+        dadb.shell("input swipe ${point.x} ${point.y} ${point.x} ${point.y} 3000")
+    }
+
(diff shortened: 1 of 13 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -158,6 +158,10 @@ class AndroidDriver(
         ) ?: throw IllegalStateException("Response can't be null")
     }
 
+    override fun longPress(point: Point) {
+        dadb.shell("input swipe ${point.x} ${point.y} ${point.x} ${point.y} 3000")
```
Leland-Takamine, line 162: FYI 4.4 and above supports this:
https://stackoverflow.com/a/27630157

Leland-Takamine (approved):

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
dmitry-zaitsev: Saw it, but it is for key events, not taps

Leland-Takamine: ah makes sense

PR #95, opened 2022-08-29
Fix: Path.parent can be null

dmitry-zaitsev:


maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -107,7 +107,7 @@ object TestRunner {
                 is SyntaxError -> "Could not parse Flow file:\n\n${e.message}"
                 is NoInputException -> "No commands found in Flow file"
                 is InvalidInitFlowFile -> "initFlow file is invalid: ${e.initFlowPath}"
-                else -> e.stackTraceToString()
+                else -> throw e
```
dmitry-zaitsev, line 110: Throwing an exception here, otherwise it gets erased and we will never know what went wrong

dmitry-zaitsev: Nope, it gets erased

Leland-Takamine: If this doesn't work then do these error messages get erased as well? "No commands found in Flow file", etc
I didn't think so, and if so then why does this behave differently?

dmitry-zaitsev: I believe the reason is that we are not stopping execution at that step and move forward, which causes the result to be overwritten by the next update to the view. That is not the case with other errors

Leland-Takamine: An exception is thrown and caught here so I'm not following why this would be the case

dmitry-zaitsev: When we catch an exception, we return `null` that is then converted to a default value (`listOf(flowFile)`) and execution contunues

Leland-Takamine: But isn't that the case for all errors caught here not just the else case?

Leland-Takamine (approved):

PR #112, opened 2022-09-02
fix: only save state of its own package name

igorsmotto:
Currently, the `pull` function for the Android driver zips all files that the adb user has access to. Afterward it zips in a `/data/data/package.name` folder structure, this could cause issues since this user should only have access to the folder `package.name` within the `/data/data`.

This PR inverts this logic to zip as a simple  `package.name/` folder structure


dmitry-zaitsev (approved):

igorsmotto:
Closing because if the user have other files in other directories than /data/data the zip file should be consistent even tho the current implementation doesn't support this yet

PR #139, opened 2022-09-07
Feature: eraseText command

dmitry-zaitsev:


Leland-Takamine (commented):
Defaulting to 50 backspaces feels arbitrary. What if we go with `backspace` (single backspace) and `backspace: N` (N backspaces)? This is more precise of an API. "eraseText" feels a little weird to me semantically since it's not immediately clear what this maps to

dmitry-zaitsev:
@Leland-Takamine I would like to stick to a non-eng friendly notation to make it more approachable. `backspace` feels like exposing internal implementation details

Leland-Takamine:
I think backspace is a well-understood concept in general and still think the backspace API makes more sense. But leaving to you for the final call

Leland-Takamine (approved):

dmitry-zaitsev:
@Leland-Takamine we could introduce an alias as well, but perhaps it could make it more confusion. What do you think?

PR #148, opened 2022-09-09
feature: AndroidTV-Remote navigation

depapp:
hi all,

I've added AndroidTV Remote navigation to `pressKey()`
- `Remote Dpad Up`
- `Remote Dpad Down`
- `Remote Dpad Left`
- `Remote Dpad Right`
- `Remote Dpad Center`
- `Remote Media Play Pause`
- `Remote Media Stop`
- `Remote Media Next`
- `Remote Media Previous`
- `Remote Media Rewind`
- `Remote Media Fast Forward`

please let me know if I miss anything 🙇🏻

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -938,6 +938,17 @@ class IntegrationTest {
         driver.assertHasEvent(Event.PressKey(KeyCode.VOLUME_UP))
         driver.assertHasEvent(Event.PressKey(KeyCode.VOLUME_DOWN))
         driver.assertHasEvent(Event.PressKey(KeyCode.LOCK))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_UP))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_DOWN))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_LEFT))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_RIGHT))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_CENTER))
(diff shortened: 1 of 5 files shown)


maestro-test/src/test/resources/034_press_key.yaml
```diff
@@ -3,7 +3,7 @@ appId: com.example.app
 - pressKey: Enter
 - pressKey: Backspace
 - pressKey: Home
-- pressKey: back
+- pressKey: Back
 - pressKey: Volume Up
-- pressKey: volume Down
-- pressKey: lOcK
```
dmitry-zaitsev, line 9: Those were actually intentional, to validate that input works regardless of the case

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -938,6 +938,17 @@ class IntegrationTest {
         driver.assertHasEvent(Event.PressKey(KeyCode.VOLUME_UP))
         driver.assertHasEvent(Event.PressKey(KeyCode.VOLUME_DOWN))
         driver.assertHasEvent(Event.PressKey(KeyCode.LOCK))
+        driver.assertHasEvent(Event.PressKey(KeyCode.REMOTE_UP))
```
dmitry-zaitsev, line 941: Since you added those new assertions, you also need to update `034_press_key.yaml` to actually simulate key press events in this test, otherwise the assertion will fail

dmitry-zaitsev (changes requested):
Thank you for the submission - you are almost there, just need to update the test

depapp:
> Thank you for the submission - you are almost there, just need to update the test

hi @dmitry-zaitsev 

thanks for the review, I've updated on https://github.com/mobile-dev-inc/maestro/pull/148/commits/ac69848e3abc60483c02da9e84ee7d1c6017d5d5 

please let me know if I miss anything 🙇🏻

dmitry-zaitsev (approved):
Looks good! Thank you for the contribution

Issue #105, opened 2022-09-01
Command to set a Geo location


dmitry-zaitsev:
This appears to be doable:

- [Android](https://stackoverflow.com/a/2587369/926907)
- [iOS](https://fbidb.io/docs/commands#set-a-simulators-location)

ragunathjawahar:
Can you help with how the DSL should look like for this?

**A**
```
location: -82.411629 28.054553
```

**B**
```
location:
    lat: -82.411629
    long: 28.054553
```

**Something else** 
```
?
```

Once we have a consensus on this, I can pick it up.

dmitry-zaitsev:
I would suggest we start with option B. We could add option A in addition to option B later on, but I think this is not super important.

ragunathjawahar:
I'll pick this up later today, if anyone is already working on this please update?

amanjeetsingh150:
Added this command in latest release.

Issue #100, opened 2022-08-31
Command to take screenshots


bobatsar:
Is there any command planned to take screenshots?
This would be very handy to document changes between different versions of an app.

dmitry-zaitsev:
That would be relatively easy to add, so we will address this soon 👍

ca-ke:
@dmitry-zaitsev we can use dadb to take screenshots via adb. I do think that we can go with something like this

`adb shell screencasts -p /sdcard/$name`
`adb pull /sdcard/$name $opt`
`adb shell em /sdcard/$name`

dmitry-zaitsev:
@ca-ke indeed, that is more or less how it would go. If you like to give it a try, I can guide you through the implementation steps

ragunathjawahar:
@dmitry-zaitsev and @ca-ke do you folks have a DSL in mind?

dmitry-zaitsev:
Something like:

```
- takeScreenshot: {file name}
```

dmitry-zaitsev:
Good news - this is now available in 1.7.0: https://maestro.mobile.dev/reference/take-screenshot

Credit for Android implementation goes to @ragunathjawahar , thank you!

PR #150, opened 2022-09-09
[Feedback PR] Prepare maestro to make it easier to add new commands

ragunathjawahar:
Folks, I was looking at the issue list and was interested in implementing https://github.com/mobile-dev-inc/maestro/issues/100 and https://github.com/mobile-dev-inc/maestro/issues/105. This is an initial refactor PR towards that direction. If this looks good to you folks, we can explore simplifying `Orchestra.executeCommand`. After that, extending Maestro should be relatively easier. Let me know what you think.

Thank you and great job pulling this off.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -21,6 +21,7 @@ package maestro.cli.runner
 
 import maestro.Maestro
 import maestro.MaestroException
+import maestro.orchestra.ApplyConfigurationCommand
 import maestro.orchestra.MaestroCommand
 import maestro.orchestra.Orchestra
 import maestro.orchestra.OrchestraAppState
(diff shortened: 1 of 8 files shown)


dmitry-zaitsev (changes requested):
I know that this seems like a straightforward thing to do, but there is one caveat that unfortunately makes this change not suitable for us (and I realise we should've left a comment about that) - we actually serialize `MaestroCommand` as JSON when using Maestro as part of our backend product (mobile.dev).

Thank you for taking your time and submitting the suggestion!

ragunathjawahar:
Interesting, this is new information to me. I was wondering what the `MaestroCommand` is for. Now it makes sense. The goal I was aiming was this—how do we keep the system open for extension but closed for modification along the axis for adding or removing commands.

As you know, the compiler's feedback is something that we can use as a checklist to educate folks that want to add more commands. Do you mind if I take another stab at this? I will also add the required tests for the JSON serialisation. Anyways this is a feedback PR and I don't mind if the next set of changes don't make it either. I'd be a learning exercise for me.

ragunathjawahar:
> we actually serialize MaestroCommand as JSON when using Maestro as part of our backend product (mobile.dev)

Ah... overlooked this bit. Closing this PR.

dmitry-zaitsev:
> Do you mind if I take another stab at this? I will also add the required tests for the JSON serialisation. Anyways this is a feedback PR and I don't mind if the next set of changes don't make it either. I'd be a learning exercise for me.

As long as tests for serialization are in place - sure, go for it!

PR #151, opened 2022-09-09
Replace `ViewUtils` with a `ViewHierarchy` value class

ragunathjawahar:
This PR doesn't have a dependency on 150.


dmitry-zaitsev (approved):
That makes sense to me, thank you!

PR #157, opened 2022-09-10
Prepare maestro to make it easier to add new commands

ragunathjawahar:
This is the same set of changes from https://github.com/mobile-dev-inc/maestro/pull/150 but with added tests for `MaestroCommand` serialisation for the backend.

**Things to note**
1. The `MaestroCommandSerializer` has two exceptions with property names `tapOnPoint` and `tapOnPoint`. The rest of the commands have a `Command` suffix (e.g. `swipeCommand`). We preserve this behavior so that the contract with the backend remains the same after the refactor.
2. The `MaestroCommandSerializationTest` has duplication. I found IntelliJ's syntax highlighting from language injection (`@Language("json")`) useful in the IDE. We could go with a JUnit parameterised test or a combination approval test if you find it too verbose.
3. I have added a comment to `MaestroCommand` on why we need the custom serializer.

If this PR gets accepted, the next steps are the same as discussed in #150 (starting with `Orchestra.executeCommand`).

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
@@ -19,88 +19,44 @@
 
 package maestro.orchestra
 
+import com.fasterxml.jackson.core.JsonGenerator
+import com.fasterxml.jackson.databind.JsonSerializer
+import com.fasterxml.jackson.databind.SerializerProvider
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+
+@JsonSerialize(using = MaestroCommandSerializer::class)
 data class MaestroCommand(
-    val tapOnElement: TapOnElementCommand? = null,
-    val tapOnPoint: TapOnPointCommand? = null,
-    val scrollCommand: ScrollCommand? = null,
-    val swipeCommand: SwipeCommand? = null,
-    val backPressCommand: BackPressCommand? = null,
-    val assertCommand: AssertCommand? = null,
-    val inputTextCommand: InputTextCommand? = null,
-    val launchAppCommand: LaunchAppCommand? = null,
-    val applyConfigurationCommand: ApplyConfigurationCommand? = null,
-    val openLinkCommand: OpenLinkCommand? = null,
-    val pressKeyCommand: PressKeyCommand? = null,
-    val eraseTextCommand: EraseTextCommand? = null,
+    val command: Command?,
 ) {
 
     fun injectEnv(envParameters: Map<String, String>): MaestroCommand {
-        return copy(
-            tapOnElement = tapOnElement?.injectEnv(envParameters),
-            tapOnPoint = tapOnPoint?.injectEnv(envParameters),
-            scrollCommand = scrollCommand?.injectEnv(envParameters),
-            swipeCommand = swipeCommand?.injectEnv(envParameters),
-            backPressCommand = backPressCommand?.injectEnv(envParameters),
-            assertCommand = assertCommand?.injectEnv(envParameters),
-            inputTextCommand = inputTextCommand?.injectEnv(envParameters),
-            launchAppCommand = launchAppCommand?.injectEnv(envParameters),
-            applyConfigurationCommand = applyConfigurationCommand?.injectEnv(envParameters),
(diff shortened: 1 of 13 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -114,39 +114,21 @@ class Orchestra(
...
-            command.launchAppCommand != null -> command.launchAppCommand
-                ?.let { launchAppCommand(it) }
-            command.openLinkCommand != null -> command.openLinkCommand
-                ?.let { openLinkCommand(it) }
-            command.pressKeyCommand != null -> command.pressKeyCommand
-                ?.let { pressKeyCommand(it) }
-            command.eraseTextCommand != null -> command.eraseTextCommand
-                ?.let { eraseTextCommand(it) }
+    private fun executeCommand(command: Command?) {
+        when (command) {
```
dmitry-zaitsev, line 118: nit: let's do `return when (command) {}` to enforce `when` being an exhaustive list

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
```diff
@@ -19,88 +19,50 @@
...
-    val swipeCommand: SwipeCommand? = null,
-    val backPressCommand: BackPressCommand? = null,
-    val assertCommand: AssertCommand? = null,
-    val inputTextCommand: InputTextCommand? = null,
-    val launchAppCommand: LaunchAppCommand? = null,
-    val applyConfigurationCommand: ApplyConfigurationCommand? = null,
-    val openLinkCommand: OpenLinkCommand? = null,
-    val pressKeyCommand: PressKeyCommand? = null,
-    val eraseTextCommand: EraseTextCommand? = null,
+    val command: Command?,
```
dmitry-zaitsev, line 35: Perhaps we can make `Command` a `sealed class` now?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ragunathjawahar: We could do that. I believe `ApplyConfigurationCommand` would be a no-op in that case?

Also, can we make the `command` parameter non-null? Are there situations where we need to deal with `null` values?

dmitry-zaitsev: > I believe ApplyConfigurationCommand would be a no-op in that case?

Exactly

> Also, can we make the command parameter non-null? Are there situations where we need to deal with null values?

I thought about that, but I think it makes sense to keep it nullable for backwards compatibility (i.e. if we encounter an unknown command in JSON)

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
```diff
@@ -19,88 +19,50 @@
...
-        }
-
-        applyConfigurationCommand?.let {
-            return it.description()
-        }
-
-        openLinkCommand?.let {
-            return it.description()
-        }
+        return command?.description() ?: "No op"
```
ragunathjawahar, line 43: The only reason `command` is nullable here is because of this function returning a `"No op"` value. Is this value used in the backend? If not, we could make `command` a non-null property member in `MaestroCommand` and also get rid of the associated test from `MaestroCommandTest`.

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
ragunathjawahar: We could. I have left a couple of comments related to this change as well.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ragunathjawahar: Got it, then I'll go ahead and make the change and keep the nullability as is.

ragunathjawahar: This change is done 👍🏽

dmitry-zaitsev (approved):

PR #160, opened 2022-09-11
[Feedback PR] Replace `Orchestra.doExecute` with a polymorphic call

ragunathjawahar:
change

**Overview**
The PR aims to simplify `Orchestra.executeCommand` by,
1. Introducing a new abstract function `execute(Context)` to the `Command` interface.
2. Move functionality for each command from `executeCommand` to their respective implementations.
3. Eliminate the `executeCommand` and replace it with a polymorphic call `Command.execute(Context)`.

**Notes to the reviewer**
* Moved `UnicodeNotSupportedError` from the _**maestro-orchestra**_ to _**maestro-orchestra-models**_ because it is used by `InputTextCommand`.
* The `AssertCommand` and the `TapOnElementCommand` rely on the flakiness tolerance mechanism and the rest of the commands don't. So, I had to abstract this idea and move it to `OrchestraContext`. Not sure if this is the right name or the right abstraction.
* I'd like if someone took a second look at this commit (https://github.com/mobile-dev-inc/maestro/commit/53c76bf59f2ce5c88f23ea151c63a8f32e1197ae, it is also tagged with `refactor!!` instead of `refactor`). Because this was the largest step in the entire refactoring and I don't think we have a test that covers all the conditions for this change.
* For every move method refactor, I replaced the actual code with `TODO()`, ensured the test failed before replacing it with the equivalent `command.execute(Context)` call.

**After split**
post-refactor
*_Please note_: The files on the right do not represent the absolute size of the file.


dmitry-zaitsev:
While the change does seem clean in isolation, I think it loses one important component (that we are not fully utilising yet) - Orchestra knows about the history of commands as well as the planned commands and can make decisions based on that (i.e. accumulate state by copying text on the screen, skip `optional` commands if it knows that it can proceed with non-optional commands, etc.). This information could be passed in a context, but in a way I think we are actually breaking separation of concerns.

ragunathjawahar:
> Orchestra knows about the history of commands as well as the planned commands and can make decisions based on that (i.e. accumulate state by copying text on the screen, skip optional commands if it knows that it can proceed with non-optional commands, etc.).

This is interesting, can you help me understand this correctly. Let's say the UI has an optional checkbox that says "Send me latest offers to my email address" on the login screen and the test contains an instruction to select this checkbox. But, Maestro somehow knows this step is not needed for the test to run and just logs in with the credentials without selecting the checkbox. Is my understanding correct?

> This information could be passed in a context, but in a way I think we are actually breaking separation of concerns.

Agreed, thank you for bringing this up. I wasn't entirely sure about the `Context` abstraction either and now from what you say, I agree that it is not the right abstraction 🤔 however, the first piece of information helps to explore alternatives.

PR #165, opened 2022-09-12
feat: add hideKeyboard event

depapp:
hi @dmitry-zaitsev 

## description

I have an issue when try to `tapOn` a button that blocked by device keyboard.

- the UI before keyboard is showing
image

- the UI after keyboard is showing
image

## solution

I've tried to add `hideKeyboard()` event to avoid that failed case (before `tapOn()` desired button)

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -220,6 +220,8 @@ class IOSDriver(
 
     override fun backPress() {}
 
+    override fun hideKeyboard() {}
+
     override fun inputText(text: String) {
         iosDevice.input(text).expect {}
     }
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -220,7 +220,9 @@ class IOSDriver(
 
     override fun backPress() {}
 
-    override fun hideKeyboard() {}
+    override fun hideKeyboard() {
+        iosDevice.pressKey(41).expect {}
+    }
 
     override fun inputText(text: String) {
         iosDevice.input(text).expect {}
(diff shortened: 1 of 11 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -220,6 +220,8 @@ class IOSDriver(
 
     override fun backPress() {}
 
+    override fun hideKeyboard() {}
```
dmitry-zaitsev, line 223: I just checked and it seems that pressing Escape key (key code `41`) on iOS is hiding the keyboard (at least in Safari). Let's use that

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
```diff
@@ -162,6 +162,12 @@ class FakeDriver : Driver {
         events += Event.BackPress
     }
 
+    override fun hideKeyboard() {
```
dmitry-zaitsev, line 165: Could you please add a new test case in `IntegrationTest`?

Same pattern as with other tests there:
- Create a `yaml` file where you invoke `hide keyboard`
- Load this file in a test
- Run it on a `FakeDriver`
- Assert that `HideKeyboard` event was received

dmitry-zaitsev (changes requested):
Looks good, but please also add:

- iOS support as mentioned in a comment
- Integration test

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
depapp: updated on [27539f2](https://github.com/mobile-dev-inc/maestro/pull/165/commits/27539f2b48a65a80b12a6e3a2cff4b452ef92650)

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
depapp: updated on this commit [fbc5007](https://github.com/mobile-dev-inc/maestro/pull/165/commits/fbc50075fbf271eb763cb3daf07cb44d47a574f0)

depapp:
> Looks good, but please also add:
> 
> * iOS support as mentioned in a comment
> * Integration test

hi @dmitry-zaitsev 

I've updated the PR. could you please help to review it again? 🙇🏻

dmitry-zaitsev (approved):
Looks good! Thank you

PR #169, opened 2022-09-12
Fix: dumping accessibility info on Android ourselves

dmitry-zaitsev:
It seems that there is some bug in Android accessibility service that sometimes reports contents of a `WebView` as not visible to users. 

This fix is inspired by Stackoverflow answer: https://stackoverflow.com/a/66716800/926907. Instead of relying on built-in dump method, we are iterating through accessibility nodes ourselves.


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -32,10 +33,12 @@ class MaestroDriverService {
             .setWaitForIdleTimeout(0L)
             .setWaitForSelectorTimeout(0L)
 
-        val uiDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        val instrumentation = InstrumentationRegistry.getInstrumentation()
```
amanjeetsingh150, line 36: This only needs to update server APK right? Not the other one?

amanjeetsingh150 (approved):

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
dmitry-zaitsev: Yeah, the other one did not have any changes in it since it is just a dummy app

PR #175, opened 2022-09-14
fix: hideKeyboard didn't work

depapp:
hi @dmitry-zaitsev 

just realized that `escape` key didn't work on latest android API.
I've updated the `keyevent` both on `android` and `ios`.

please let me know if I miss anything 🙇🏻


dmitry-zaitsev (approved):
Why updating it on iOS though? Did it also not work?

PR #185, opened 2022-09-15
(Feedback PR) Migrate tests to JUnit 5

ragunathjawahar:
Folks, the JUnit 4 → 5 migration for most tests was minimally invasive.
1. Updated dependencies
2. Replaced `org.junit.*` imports with `org.junit.jupiter.api.*` imports.
3. Replace `@Test(expected = MaestroException.ElementNotFound::class)` with JUnit 5's `assertThrows<MaestroException.ElementNotFound>` assert statement (JUnit 5 `@Test` annotation does not allow us to specify the exception within the annotation).

### YamlCommandReaderTest
The special case was `YamlCommandReaderTest` because it used the JUnit 4 `TestName` rule. On deeper inspection, I noticed the test name was coupled to the resource file name. So, `T002_launchApp` also meant that there should be a file named `resources/YamlCommandReaderTest/002_launchApp.yaml`. If the test name or the file name changed, then a test would fail. I did not find out this coupling until I started looking into the assertions.

Therefore, I made changes to decouple the test name from the YAML file name and made it more explicit using JUnit 5 extensions and a `@YamlFile` annotation. After this change, the test name is no longer _implicitly_ coupled to the file name.

### Parsing failure test
**Before**
```
@Test
fun T004_config_empty() = expectException<SyntaxError> { e ->
    assertThat(e.message).contains("Flow files must contain a config section and a commands section")
}
```

**After**
```
@Test
fun config_empty(
    @YamlFile("004_config_empty.yaml") e: SyntaxError,
) {
    assertThat(e.message).contains("Flow files must contain a config section and a commands section")
}
```

If the type of exception received is different from the one specified in the parameter, the test would fail with an assertion error.

### Parsing commands test
**Before**
```
@Test
fun T002_launchApp() = expectCommands(
    ApplyConfigurationCommand(MaestroConfig(
        appId = "com.example.app",
    )),
    LaunchAppCommand(
        appId = "com.example.app"
    ),
)
```

**After**
```
@Test
fun launchApp(
    @YamlFile("002_launchApp.yaml") commands: List<Command>,
) {
    assertThat(commands).containsExactly(
        ApplyConfigurationCommand(MaestroConfig(
            appId = "com.example.app"
        )),
        LaunchAppCommand(
            appId = "com.example.app"
        ),
    )
}
```

### Refactoring approach
```
test(⛔️): prepare tests for refactor                   -> 1. Modify YAML files to make all the tests fail (no changes were made to the tests).
test: migrate expect commands tests to JUnit 5         -> 2. Refactor tests and keep running them to see if ALL tests are still failing.

  > Revert "test(⛔️): prepare tests for refactor"      -> 3. To verify if ONLY expect commands tests are passing, this commit was dropped after verification

test: migrate syntax error tests to JUnit 5            -> 4. Refactor tests and keep running them to see if ALL test are still failing.
Revert "test(🟢): prepare tests for refactor"          -> 5. Verify that ALL tests are passing and restore changes to YAML files
```

### Final verification
The test run showed 76 passing tests, 6 ignored tests, and a total of 82 tests before and after the migration.


maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlCommandReaderTest.kt
```diff
@@ -1,202 +1,264 @@
...
 internal class YamlCommandReaderTest {
 
-    @JvmField
-    @Rule
-    val name: TestName = TestName()
-
     @Test
-    fun T001_empty() = expectException<SyntaxError> { e ->
+    fun empty(
+        @YamlFile("001_empty.yaml") e: SyntaxError,
```
ragunathjawahar, line 28: Should we call this `@Dsl` or `@FlowFile` instead of `@YamlFile`?

dmitry-zaitsev: I think `@YamlFile` is great, it is explicit

dmitry-zaitsev (approved):
Thanks, this looks good 👍

PR #186, opened 2022-09-16
Add maestro upload command

Leland-Takamine:


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -0,0 +1,142 @@
...
+        appFile: Path,
+        workspaceZip: Path,
+        uploadName: String,
+        repoOwner: String?,
+        repoName: String?,
+        branch: String?,
+        pullRequestId: String?,
+    ): UploadResponse {
+        if (!appFile.exists()) throw CliError(appFile.absolutePathString())
+        if (!workspaceZip.exists()) throw CliError(workspaceZip.absolutePathString())
```
ragunathjawahar, line 82: These throw a `CliError` with the path, shouldn't they also tell the reason why? For instance, `App file does not exist: ${appFile.absolutePathString()}`?

Leland-Takamine: Good catch! These used to be more descriptive exceptions and after converting to a CLIError I forgot to update the message

maestro-cli/src/main/java/maestro/cli/command/UploadCommand.kt
```diff
@@ -0,0 +1,191 @@
...
+            }
+            Thread.sleep(1000)
+        }
+    }
+
+    companion object {
+
+        init {
+            AnsiConsole.systemInstall()
+        }
```
ragunathjawahar, line 176: Curious, can this happen in the parent command or should it happen in every subcommand?

Leland-Takamine: Good suggestion! It can probably be moved to App.kt. Updated!

maestro-cli/src/main/java/maestro/cli/command/UploadCommand.kt
```diff
@@ -0,0 +1,221 @@
...
+    companion object {
+
+        private const val AUTH_SUCCESS_REDIRECT_URL = "https://console.mobile.dev/auth/success"
+
+        private fun message(message: String) {
+            println(ansi().render("@|cyan \n$message |@"))
+        }
+
+        private fun prompt(message: String): String {
+            print(ansi().render("\n@|yellow,bold $message\n> |@"))
```
dmitry-zaitsev, line 213: minor nit: `ansi` has nice DSL for setting colours

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/command/UploadCommand.kt
Leland-Takamine: This is actually using the jansi convenience syntax already. The actual ansi codes are much uglier 😄

PR #187, opened 2022-09-16
Take screenshot feature (need help with iOS implementation)

ragunathjawahar:
This PR addresses #100.

**Flow DSL**
```
appId: com.example.app
---
- takeScreenshot: /User/adamtornhill/screenshot.png
```

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -288,6 +288,12 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.close()
     }
 
+    fun takeScreenshot(outFile: File) {
+        LOGGER.info("Taking screenshot: $outFile")
+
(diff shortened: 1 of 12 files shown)


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -288,6 +288,12 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.close()
     }
 
+    fun takeScreenshot(outFile: File) {
```
dmitry-zaitsev, line 291: Would rather have an `OutputStream` or Okio `Sink` here, there might be use cases where we would operate on images in memory

dmitry-zaitsev:
Thanks! I will address my own comments and add iOS implementation

dmitry-zaitsev:
I fetched your changes in my local branch and merged them in as part of #193 . Thank you!

Issue #176, opened 2022-09-14
Custom wait duration


dmitry-zaitsev:
In some special cases we need to be able to wait longer than usual (i.e. waiting for a video to play, waiting for some CPU processing to finish). It would be handly to have some sort of `wait` command to do that

lalon:
Would also be helpful to have a wait until a view/text is displayed command

jschisler:
Wait until a control is enabled would be excellent.

simon-gilmurray:
Agreed, I have certain app actions that require a minimum duration (e.g. video identification that requires at least x seconds) so this would be needed for some of my tests

dmitry-zaitsev:
This is now available in 1.7.0: https://maestro.mobile.dev/reference/wait

nisrulz:
Adding for anyone else looking for this.

In case you would like to add an arbitrary wait (not dependent on a UI element becoming visible/not visible) i.e time between app restart, this is how I did it:

```
# HACK: Extra tap attempt to wait for app relaunch
- repeat:
      times: 2
      commands:
          - tapOn:
                point: -1,-1
```

You can all other [options of tapOn](https://maestro.mobile.dev/reference/tap-on-view). I am using to tap on outside the visible area.

I have repurposed the hack that was [suggested in mobile.dev Slack by Manish Verma](https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1667996545976799?thread_ts=1667992991.490819&cid=C041FU72T54)

PR #197, opened 2022-09-16
Feature: extendedWait command

dmitry-zaitsev:
Using `extendedWait` as a name to be very clear that this is a command you would not normally need to use

Requested in #176


Leland-Takamine (approved):
Approving assuming the updates we discussed offline: Require a selector ie. timeout cannot be used by itself

PR #200, opened 2022-09-17
Feature: clearKeychain command

dmitry-zaitsev:


Leland-Takamine (approved):

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -152,6 +152,10 @@ class AndroidDriver(
         shell("pm clear $appId")
     }
 
+    override fun clearKeychain() {
```
Leland-Takamine, line 155: Users may expect this to be cleared on Android: https://developer.android.com/reference/android/security/KeyChain

Issue #183, opened 2022-09-15
Maestro does not move on with the next step as the hierarchy is still changing even though in some cases it's what we want.


StylianosGakis:
My use case is described exactly in the slack channel in this discussion https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1663244249203259?thread_ts=1663243109.313569&cid=C041FU72T54
The gist is that in some step of the test, there's an audio recording and a timer is updating every second to show how many seconds long the recording is. This however made the test in one case take 20 seconds to commit to do the next action. There is an assumption that the constantly changing hierarchy is contributing to this delay as it's still waiting for the hierarchy to settle before going on the next step.
This may be true for various UIs where there might be some constantly changing view state that should not obstruct the test from continuing on.

mollyhartwick3:
We put in a fix that we believe will fix your issue in version 1.11.4. Closing this issue, please feel free to reopen if this continues to be an issue 🙂

PR #206, opened 2022-09-19
Tweak: Comparing screenshots after tapping on a view

dmitry-zaitsev:
One of the common sources of confusion when using Maestro is - why sometimes we tap on a view multiple times?

This PR helps to mitigate that by _also_ comparing screenshots before and after user interaction to verify that indeed something have visibly changed in the hierarchy.

To facilitate the testing, `FakeDriver` was updated to actually draw a screen onto an image, just like a real device would.

### DIFF
--- a/maestro-client/build.gradle
+++ b/maestro-client/build.gradle
@@ -50,11 +50,11 @@ dependencies {
     api "dev.mobile:dadb:1.1.0"
     api "org.slf4j:slf4j-simple:1.7.36"
     api 'com.squareup.okio:okio:3.2.0'
+    api 'com.github.romankh3:image-comparison:4.4.0'
 
     implementation project(':maestro-ios')
     implementation 'com.google.code.findbugs:jsr305:3.0.2'
     implementation 'de.upb.cs.swt:axml:2.1.2'
-    implementation 'com.github.romankh3:image-comparison:4.4.0'
 
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.9.0'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.9.0'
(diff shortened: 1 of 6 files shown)


amanjeetsingh150:
Hey @dmitry-zaitsev CMIIW, I think this may fix this issue: https://github.com/mobile-dev-inc/maestro/issues/183. Lets link it if it does.

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -177,10 +182,31 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+                screenshotAfterTap != null &&
+                screenshotBeforeTap.width == screenshotAfterTap.width &&
+                screenshotBeforeTap.height == screenshotAfterTap.height
+            ) {
+                val imageDiff = ImageComparison(
+                    screenshotBeforeTap,
+                    screenshotAfterTap
+                ).compareImages().differencePercent
+
+                if (imageDiff > 0.005) {
```
amanjeetsingh150, line 200: Is it like a difference of more than 5%? 

nit: If yes maybe put that in the constant field name.

maestro-client/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     implementation project(':maestro-ios')
     implementation 'com.google.code.findbugs:jsr305:3.0.2'
     implementation 'de.upb.cs.swt:axml:2.1.2'
+    implementation 'com.github.romankh3:image-comparison:4.4.0'
```
amanjeetsingh150, line 57: I think we need to do api in this right? Folks implementing maestro-client otherwise will not get this artifact and we will face no class def found errors

maestro-client/src/main/java/maestro/Maestro.kt
dmitry-zaitsev: 0.5% in fact

PR #209, opened 2022-09-20
docs: update README

PierreCapo:
### Before this PR
Following the installation process in the documentation gives an error:
image

### Problem
There are 2 taps with the same name which causes an installation conflict: 
- `mobile-dev-inc/tap/idb-companion`
- `facebook/fb/idb-companion`

I have imagined that the one from `facebook` is for now the correct one, and I have opened a pull request in order that the installation process doesn't throw an error anymore. Let me know if the one from `mobile-dev-inc` is the correct one and I can change my PR.

Have a good day and thanks for this library 🙌

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Install [Facebook IDB](https://fbidb.io/) tool
 
 ```
 brew tap facebook/fb
-brew install idb-companion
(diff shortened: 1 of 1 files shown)


dmitry-zaitsev (approved):
Thanks :+1:

Issue #172, opened 2022-09-13
Race when launching an app with LeakCanary


prashanOS:
When running the `launchApp` command with our `debug` application, about half the time it launches the LeakCanary activity instead which prevents the test from continuing. Guessing there is some sort of race here since it doesn't happen all the time. When hitting the `X` to close the LeakCanary activity, it doesn't go back to our app's activity, so maybe it hasn't been launched at all?

Our app uses the default LeakCanary configuration. (Just including LeakCanary as a debug dependency.)

StylianosGakis:
Can confirm this feels like it's the last big blocker from using maestro. I've gotten some test flows to work perfectly so I can share with the rest of the team but while this problem persists there simply is no way for us to reliably use it unfortunately.
If there's any temp workarounds or such I'd also be very interested in hearing about it.
One I can think of is instead of doing `launchApp` add a test only deep link which opens the app directly so there's no ambiguity regarding which activity to start, but still would rather not have to do this.

dmitry-zaitsev:
We hear you folks, we will think of some workaround to make it work

amanjeetsingh150:
Fix is going to be part of the next release 🎉

fchristysen:
Hi, I found a potential issue. I have try version 1.8.0, but it still sometimes launch leakcanary instead of my main activity. On my app setup, the resulting manifest when LeakCanary enabled is as follows
``` .xml
<activity android:name="leakcanary.internal.activity.LeakActivity" ... > 
   <!-- Without main action and launcher category --> 
</.activity>

<activity-alias android:name="leakcanary.internal.activity.LeakLauncherActivity" ... > 
   <intent-filter>
                <action
                    android:name="android.intent.action.MAIN" />
                <category
                    android:name="android.intent.category.LAUNCHER" />
                <category
                    android:name="android.intent.category.LEANBACK_LAUNCHER" />
            </intent-filter>
</.activity-alias>
```
I have tried to manually run `adb shell am ..` command and it does consistently launch main activity. So I suspect that this check 
``` .kt 
// in maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
if (hasThirdPartyLauncherConfigured(manifest, appId) && ...
```
fails and fallback to monkey which intermittently launch LeakCanary instead.

Barryrowe:
> Fix is going to be part of the next release 🎉

[...]

PR #214, opened 2022-09-20
Opens only app based launcher activity in case of 3rd party launcher

amanjeetsingh150:
Fixes: https://github.com/mobile-dev-inc/maestro/issues/172

## Reproducing the issue

1. Implementing leakcanary
2. Write down the following test:
```
val dadb =  Dadb.create("localhost", 5555)
val maestro = Maestro.android(dadb)
repeat(10) {
    maestro.launchApp("com.example.name")
    Thread.sleep(3000)
}
```
Was able to observe that out of 10 majority times leak canary launcher opens

## Fix
Check for launcher activities and opens the one which starts with the app id passed. Since all the third-party ones will start with different package id.

### DIFF
--- a/maestro-client/src/main/java/maestro/android/AndroidAppFiles.kt
+++ b/maestro-client/src/main/java/maestro/android/AndroidAppFiles.kt
@@ -37,6 +37,15 @@ object AndroidAppFiles {
         }
     }
 
+    fun getApkFile(dadb: Dadb, appId: String): File {
+        val apkPath = dadb.shell("pm list packages -f | grep $appId")
+            .output.substringAfterLast("package:").substringBefore("=$appId")
+        apkPath.substringBefore("=$appId")
+        val dst = File.createTempFile("tmp", ".apk")
+        dadb.pull(dst, apkPath)
+        return dst
+    }
+
     fun push(dadb: Dadb, packageName: String, appFilesZip: File) {
         val remoteZip = "/data/local/tmp/app.zip"
         dadb.push(appFilesZip, remoteZip)
--- a/maestro-client/build.gradle
+++ b/maestro-client/build.gradle
@@ -51,6 +51,8 @@ dependencies {
     api "org.slf4j:slf4j-simple:1.7.36"
     api 'com.squareup.okio:okio:3.2.0'
     api 'com.github.romankh3:image-comparison:4.4.0'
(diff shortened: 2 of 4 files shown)


maestro-client/src/main/java/maestro/android/AndroidAppFiles.kt
```diff
@@ -37,6 +37,15 @@ object AndroidAppFiles {
         }
     }
 
+    fun getApkFile(dadb: Dadb, appId: String): File {
+        val apkPath = dadb.shell("pm list packages -f | grep $appId")
+            .output.substringAfterLast("package:").substringBefore("=$appId")
+        apkPath.substringBefore("=$appId")
+        val dst = File.createTempFile("tmp", ".apk")
+        dadb.pull(dst, apkPath)
```
dmitry-zaitsev, line 45: Just to confirm, does this work on non-root devices and for non-debuggable apps as well?

ArthurSav: > dadb.pull(dst, apkPath)

This requires a debuggable app.

dmitry-zaitsev: If so, I am afraid this is a no-go. It needs to work with any kind of app

amanjeetsingh150: I checked it on few apps from apk mirror and it was working 🤔

Leland-Takamine: I have tried this as well with non-debuggable apps and I think I remember it working

dmitry-zaitsev (approved):

maestro-client/src/main/java/maestro/android/AndroidAppFiles.kt
ArthurSav: I stand corrected, seems to be a case only when trying to pull the app cache (probably with `run-as`)

amanjeetsingh150: Tested this out on few other apps:

1. We are going to fallback on monkey command in case we have exception
2. Also fixed logic to detect which activity app belongs to.

PR #231, opened 2022-09-27
Fix: hide keyboard on Android

### LINKED ISSUE #229: Hide keyboard does not work on Android
Maestro: 1.8.1

dmitry-zaitsev:
## Testing
Tested on an emulator

## Issues Fixed

`hideKeyboard` was not always working

Fixes #229

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -261,6 +261,7 @@ class AndroidDriver(
 
     override fun hideKeyboard() {
         dadb.shell("input keyevent 66")
+        dadb.shell("input keyevent 111")
     }
 
     override fun takeScreenshot(out: Sink) {


Leland-Takamine (approved):

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -261,6 +261,7 @@ class AndroidDriver(
 
     override fun hideKeyboard() {
         dadb.shell("input keyevent 66")
+        dadb.shell("input keyevent 111")
```
amanjeetsingh150, line 264: Is this because this key is mapped differently 🤔 ?

dmitry-zaitsev: Might be different OS versions or due to different vendors. I was never able to reproduce the first case but another dev reported that code 66 worked for them

PR #236, opened 2022-09-27
Fix: properly mapping top-level syntax errors

dmitry-zaitsev:


Leland-Takamine (approved):

Leland-Takamine:
Does this hide any information that would be useful for debugging? A before and after would be helpful

dmitry-zaitsev:
From Maestro perspective, it pretty much returns the same error just with a different type. This change is more important for mobile.dev to properly map HTTP return codes

Leland-Takamine:
👍

PR #238, opened 2022-09-28
Fix: step() input shouldn't be 0

dmitry-zaitsev:


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -158,6 +158,7 @@ class IOSDriver(
         // We are locating such a view and then scan it for children by probing individual points on the screen
         val groupChildren = accessibilityNodes
             .filter { it.type == "Group" }
+            .filter { (it.frame?.width ?: 0F) > 50F }
```
amanjeetsingh150, line 161: 50 pixel?

amanjeetsingh150 (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
dmitry-zaitsev: 50 units, depends on screen density I believe. Either way, we are not interested in tiny views so this is a small optimisation

PR #245, opened 2022-09-30
Conditional flow execution

dmitry-zaitsev:
## Proposed Changes

- `runFlow` takes an optional `when` parameter that allows to restrict flow execution to a certain condition (i.e. text is visible)
- Introducing a concept of composite commands that will come in handy once we will introduce a `repeat` command
- Introducing a `SKIPPED` command status

## Testing
- Integration test
- Locally on an iOS Simulator

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -135,9 +135,14 @@ class Orchestra(
             is StopAppCommand -> maestro.stopApp(command.appId)
             is ClearStateCommand -> maestro.clearAppState(command.appId)
             is ClearKeychainCommand -> maestro.clearKeychain()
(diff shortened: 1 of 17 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -135,11 +139,73 @@ class Orchestra(
...
+            // Element was actually visible
+            if (result != true) {
+                return false
+            }
+        }
+
+        return true
+    }
+
+    private fun runSubFlow(commands: List<MaestroCommand>) {
```
dmitry-zaitsev, line 192: This command is _almost_ like `executeCommands` but what makes it hard to merge the two is the fact that `onCommandFailed` callback can produce exception of its own, making it an awkward branching

Leland-Takamine (approved):

aakash1313:
excited :)

PR #253, opened 2022-10-03
feat: set geolocation support (iOS only, pending manual testing)

ragunathjawahar:
## Proposed Changes
Add support to set location on iOS.

## Testing
- Integration tests to test commands and driver integration.
- @amanjeetsingh150 is helping with testing this on iOS until we setup iOS dev environment on my machine this week.

## Issues Fixed
This PR addresses #105.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -412,3 +412,17 @@ data class RunFlowCommand(
     }
 
 }
+
+data class SetLocationCommand(
+    val latitude: Double,
+    val longitude: Double,
+) : Command {
+
+    override fun description(): String {
+        return "Set location (${latitude}, ${longitude})"
+    }
+
+    override fun injectEnv(env: Map<String, String>): SetLocationCommand {
+        return this
+    }
+}
--- a/maestro-test/src/test/resources/051_set_location.yaml
+++ b/maestro-test/src/test/resources/051_set_location.yaml
@@ -1,0 +1,7 @@
+appId: com.example.app
+---
+- launchApp:
+    appId: com.example.app
+- setLocation:
+    latitude: 12.5266
+    longitude: 78.2150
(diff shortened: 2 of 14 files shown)


maestro-test/src/test/resources/051_set_location.yaml
```diff
@@ -0,0 +1,7 @@
+appId: com.example.app
+---
+- launchApp:
+    appId: com.example.app
+- setLocation:
+    lat: 12.5266
```
dmitry-zaitsev, line 6: Let's use `latitude` and `longitude` instead of shortened `lat`, `long`

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -412,3 +412,27 @@ data class RunFlowCommand(
...
+
+    override fun description(): String {
+        return "Set location (${latitude}, ${longitude})"
+    }
+
+    override fun injectEnv(env: Map<String, String>): SetLocationCommand {
+        return this
+    }
+
+    override fun equals(other: Any?): Boolean {
```
dmitry-zaitsev, line 429: `equals` and `hashCode` are not needed for data classes, we only use it for commands that don't have any fields

maestro-test/src/test/resources/051_set_location.yaml
```diff
@@ -0,0 +1,7 @@
+appId: com.example.app
+---
+- launchApp:
+    appId: com.example.app
+- setLocation:
```
dmitry-zaitsev, line 5: Totally optional, but it would be great to have an additonal alias version that just accepts 2 numbers separated by comma: `setLocation: 13.5266,78.2150`

maestro-test/src/test/resources/051_set_location.yaml
ragunathjawahar: Done

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
ragunathjawahar: Done

maestro-test/src/test/resources/051_set_location.yaml
ragunathjawahar: Tracking this feature on https://github.com/mobile-dev-inc/maestro/issues/254. We can pick this up after adding Android support.

dmitry-zaitsev (approved):

PR #255, opened 2022-10-03
Feature/repeat

dmitry-zaitsev:
## Proposed Changes

Introduce a `repeat` command. There are 4 modes of operation:

**Repeat N times**
```
- repeat:
     times: 3
     commands:
         - inputText: Hello World
```

**Repeat while**
```
- repeat:
     while:
          notVisible: Some text
     commands:
         - inputText: Hello World
```

And some more exotic uses

***Repeat while but not more than N times**
```
- repeat:
       times: 3
       while:
            notVisible: Some text
       commands:
            - inputText: Hello World
```

**Repeat indefinitely**
```
- repeat:
       commands:
           - inputText: Hello World
```

## Testing

In the middle of running a command
image

If command have completed 0 times:
image


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -1431,6 +1431,46 @@ class IntegrationTest {
         )
     }
 
+    @Test
+    fun `Case 051 - Repeat N times`() {
```
ragunathjawahar, line 1435: Should we consider restructuring tests and command files separately? The current setup prevents parallel development while adding new commands. For instance, if a new command PR gets merged and if someone works on a command in parallel, then they end up resolving conflicts across several files.

dmitry-zaitsev: Open to suggestions on that. Having those sequential numbers has the advantage of allowing to easily correlate text file to the test number, as there are quite a lot of them. My personal opinion - reading experience is more important than a writing experience, i.e. in this case it would not be a big deal for me to rebase and fix the number :)

ragunathjawahar: Agree on the rebase part, it is not much work when it happens with just the test. This happens across other production classes as well, depending upon where the edit was made (drivers, command, yaml command, maestro, orchestra, etc.,). Especially when adding new commands.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -1456,6 +1456,46 @@ class IntegrationTest {
...
+            }
+
+            element {
+                text = "Button"
+                bounds = Bounds(0, 0, 100, 100)
+                onClick = {
+                    counter++
+                    indicator.text = counter.toString()
+                }
+            }
```
ragunathjawahar, line 1479: Nice `driver` setup.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -426,3 +426,37 @@ data class SetLocationCommand(
...
+
+    override fun description(): String {
+        val timesInt = times?.toIntOrNull() ?: 1
+
+        return when {
+            condition != null && timesInt > 1 -> {
+                "Repeat while ${condition.description()} (up to $timesInt times)"
+            }
+            condition != null -> {
+                "Repeat while ${condition.description()}"
```
ragunathjawahar, line 448: Curious, do we have complementary tests for repeat under a certain `condition`?

dmitry-zaitsev: We decided to remove `while` condition for now (removed [here](https://github.com/mobile-dev-inc/maestro/pull/255/commits/3b8ae1ad670eb471deabc38107560d1f630a01f6)) until we are absolutely sure that it is needed

ragunathjawahar: Got it, thank you.

Leland-Takamine (approved):

danielstr:
Is "Repeat while" already implemented? It would be very useful for some tests I need to do :)

PR #314, opened 2022-10-24
Tweak: more robust device-selection logic. Embedding idb_companion

dmitry-zaitsev:
## Proposed Changes

There are 4 major changes:

- We no longer assume that `idb_companion` is running and launching it ourselves. If it is already running, we are restarting it (has shown to improve stability).
- If there are multiple devices running, we ask the user to pick one explicitly.
- For that reason, `--platform` option is deprecated.
- If no devices are running, user can pick one to start up.

## Testing
Tested manually.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -1,0 +1,160 @@
+package maestro.cli.device
+
+import com.github.michaelbull.result.expect
+import dadb.Dadb
+import io.grpc.ManagedChannelBuilder
+import ios.idb.IdbIOSDevice
+import maestro.MaestroTimer
+import maestro.cli.CliError
+import maestro.cli.device.ios.Simctl
+import maestro.cli.util.EnvUtils
+import java.io.File
+
+object DeviceService {
+
+    fun startDevice(device: Device): Device {
+        when (device.platform) {
+            Device.Platform.IOS -> {
+                Simctl.launchSimulator(device.id)
+                Simctl.awaitLaunch(device.id)
+                return device
+            }
+            Device.Platform.ANDROID -> {
+                val androidHome = EnvUtils.androidHome()
+                    ?: throw CliError("ANDROID_HOME environment variable is not set. Unable to find 'ANDROID_HOME/tools/emulator' executable.")
+
+                ProcessBuilder(
+                    File(androidHome, "tools/emulator").absolutePath,
+                    "@${device.id}",
+                ).start()
+
+                val dadb = MaestroTimer.withTimeout(30000) {
+                    try {
+                        Dadb.discover()
+                    } catch (ignored: Exception) {
+                        Thread.sleep(100)
+                        null
+                    }
+                } ?: throw CliError("Unable to start device: ${device.id}")
+
+                return device.copy(id = dadb.toString())
+            }
+        }
+    }
+
+    fun prepareDevice(device: Device) {
+        if (device.platform == Device.Platform.IOS) {
+            killIdbCompanion()
+            startIdbCompanion(device)
+        }
+    }
+
+    private fun startIdbCompanion(device: Device) {
+        ProcessBuilder("idb_companion", "--udid", device.id)
+            .redirectError(ProcessBuilder.Redirect.DISCARD)
+            .redirectOutput(ProcessBuilder.Redirect.DISCARD)
+            .start()
+
+        val iosDevice = IdbIOSDevice(
+            ManagedChannelBuilder.forAddress("localhost", 10882)
+                .usePlaintext()
+                .build()
+        )
+        try {
+            // Try to connect to the device repeatedly
+            MaestroTimer.withTimeout(10000) {
+                try {
+                    iosDevice.deviceInfo().expect {}
+                } catch (ignored: Exception) {
+                    // Ignore
+                    null
+                }
+            } ?: error("idb_companion did not start in time")
+        } finally {
+            iosDevice.close()
+        }
+    }
+
+    private fun killIdbCompanion() {
+        ProcessBuilder("killall", "idb_companion")
+            .redirectError(ProcessBuilder.Redirect.PIPE)
+            .start()
+            .waitFor()
+    }
+
+    fun listConnectedDevices(): List<Device> {
+        return listAvailableDevices()
+            .filter { it.connected }
+    }
+
+    fun listAvailableDevices(): List<Device> {
+        return listAndroidDevices() + listIOSDevices()
+    }
+
+    private fun listAndroidDevices(): List<Device> {
+        val connectedDevices = Dadb.list()
+            .map {
+                Device(
+                    id = it.toString(),
+                    description = it.toString(),
+                    platform = Device.Platform.ANDROID,
+                    connected = true
+                )
+            }
+
+        // Note that there is a possibility that AVD is actually already connected and is present in
+        // connectedDevices.
+        val avds = try {
+            ProcessBuilder("emulator", "-list-avds")
+                .start()
+                .inputStream
+                .bufferedReader()
+                .useLines { lines ->
+                    lines
+                        .map {
+                            Device(
+                                id = it,
+                                description = it,
+                                platform = Device.Platform.ANDROID,
+                                connected = false
+                            )
+                        }
+                        .toList()
+                }
+        } catch (ignored: Exception) {
+            emptyList()
+        }
+
+        return connectedDevices + avds
+    }
+
+    private fun listIOSDevices(): List<Device> {
+        val simctlList = Simctl.list()
+
+        val runtimes = simctlList
+            .runtimes
+
+        return runtimes
+            .flatMap { runtime ->
+                runtime.supportedDeviceTypes
+                    .flatMap { supportedDeviceType ->
+                        simctlList.devices
+                            .values
+                            .flatten()
+                            .filter {
+                                it.isAvailable &&
+                                    it.deviceTypeIdentifier == supportedDeviceType.identifier
+                            }
+                    }
+                    .map { device ->
+                        Device(
+                            id = device.udid,
+                            description = "${device.name} - ${runtime.name} - ${device.udid}",
+                            platform = Device.Platform.IOS,
+                            connected = device.state == "Booted"
+                        )
+                    }
+            }
+    }
+
+}
--- a/maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
@@ -1,0 +1,61 @@
+package maestro.cli.device
+
+object PickDeviceView {
+
+    fun showRunOnDevice(device: Device) {
+        println("Running on ${device.description}")
+    }
+
+    fun pickDeviceToStart(devices: List<Device>): Device {
+        printIndexedDevices(devices)
+
+        println("No running devices detected. Choose a device to boot and run on.")
+        println()
+        println("Enter a number from the list above:")
+
+        return pickIndex(devices)
+    }
+
+    fun pickRunningDevice(devices: List<Device>): Device {
+        printIndexedDevices(devices)
+
+        println("Multiple running devices detected. Choose a device to run on.")
+        println()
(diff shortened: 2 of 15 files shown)


maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -0,0 +1,164 @@
...
+    fun startDevice(device: Device): Device {
+        when (device.platform) {
+            Device.Platform.IOS -> {
+                Simctl.launchSimulator(device.id)
+                Simctl.awaitLaunch(device.id)
+                return device
+            }
+            Device.Platform.ANDROID -> {
+                val androidHome = EnvUtils.androidHome()
+                    ?: throw CliError("ANDROID_HOME environment variable is not set. Unable to find 'ANDROID_HOME/tools/emulator' executable.")
```
Leland-Takamine, line 24: Even if ANDROID_HOME is set, that doesn't not mean the emulator is installed necessarily. We should also check for the existence of the emulator binar

Leland-Takamine:
Seems like Maestro starts running before the simulator is fully booted

Leland-Takamine:
Should we add this to the top of the Maestro output so it's visible while the flow is running?

```
Running on iPhone 11 Pro Max - iOS 15.5 - 4A4BB058-96A0-40D4-B8AB-0F1B564341FE
```

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -0,0 +1,164 @@
...
+
+                ProcessBuilder(
+                    File(androidHome, "tools/emulator").absolutePath,
```
Leland-Takamine, line 27: This should be `emulator/emulator`

Leland-Takamine:
This is usually what I use to run the Android emulator:

```
emulator -avd Pixel_4_API_30 -netdelay none -netspeed full
```

Leland-Takamine:
These prompts should leave the cursor at the end of the line:
```
Enter a number from the list above:
```

Leland-Takamine:
After printing this error message we should re-prompt the user for input:

```
Invalid index
```

Leland-Takamine:
After launching an emulator, it's takes a while for the Maestro flow to actually start executing. Not a blocker

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -0,0 +1,164 @@
...
+            .start()
+            .waitFor()
+    }
+
+    fun listConnectedDevices(): List<Device> {
+        return listAvailableDevices()
+            .filter { it.connected }
+    }
+
+    fun listAvailableDevices(): List<Device> {
```
Leland-Takamine, line 90: If this is used to list devices available to start, it should not include devices listed by Dadb

maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
```diff
@@ -0,0 +1,61 @@
+package maestro.cli.device
+
+object PickDeviceView {
+
+    fun showRunOnDevice(device: Device) {
+        println("Running on ${device.description}")
+    }
+
+    fun pickDeviceToStart(devices: List<Device>): Device {
```
Leland-Takamine, line 9: We should handle the situation where there are no devices available (eg. When the user has not set up any Android emulators and doesn't have XCode tools installed

maestro-cli/src/main/java/maestro/cli/device/ios/Simctl.kt
```diff
@@ -0,0 +1,50 @@
...
+                    .values
+                    .flatMap { it }
+                    .find { it.udid == deviceId }
+                    ?.state == "Booted"
+            ) true else null
+        } ?: throw CliError("Device $deviceId did not boot in time")
+    }
+
+    fun launchSimulator(deviceId: String) {
+        runCommand("xcrun simctl boot $deviceId")
```
lochness42, line 32:
```suggestion
        runCommand("idb_companion --boot $deviceId")
```

`idb_companion` terminates only after sim is booted up

lochness42: Example output of the command as it's booting up the sim. We could also parse last line and assert whether the sim got booted up successfully or not.

```
idb_companion --boot 726166E0-3766-459B-AAEB-971C0E1D326B

IDB Companion Built at Aug 12 2022 08:41:50
IDB Companion architecture arm64
Booting 726166E0-3766-459B-AAEB-971C0E1D326B
CoreSimulator: Loading from /Library/Developer/PrivateFrameworks/CoreSimulator.framework
CoreSimulator: Successfully loaded
CoreSimulator: SimDevice has correct path of /Library/Developer
AccessibilityPlatformTranslation: Already loaded, skipping
AccessibilityPlatformTranslation: AXPTranslationObject has correct path of /System/Library
Loaded All Private Frameworks [CoreSimulator, AccessibilityPlatformTranslation]
Booting 726166E0-3766-459B-AAEB-971C0E1D326B normally
Booting 726166E0-3766-459B-AAEB-971C0E1D326B with verification
[simulator_set] boot: called with: [Boot Environment {} | Options []]
[simulator_set] disconnectWithTimeout:logger: called with: [30.000000, <FBIDBLogger: 0x6000013a0060>]
[726166E0-3766-459B-AAEB-971C0E1D326B] Simulator connections torn down in 0.000134 seconds
[simulator_set] disconnectWithTimeout:logger: succeeded
[simulator_set] disconnectWithTimeout:logger: called with: [30.000000, <FBIDBLogger: 0x6000013a0060>]
[726166E0-3766-459B-AAEB-971C0E1D326B] Simulator connections torn down in 0.000126 seconds
[simulator_set] disconnectWithTimeout:logger: succeeded
[simulator_set] resolveState: called with: [3]
[simulator_set] resolveState: succeeded
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: Booting | Elapsed 0.000000
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: WaitingOnSystemApp | Elapsed 1.000000
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status has not changed from 'WaitingOnSystemApp | Elapsed 1.000000' for 1.578630 seconds
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: WaitingOnSystemApp | Elapsed 3.000000
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: WaitingOnSystemApp | Elapsed 3.000000
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status has not changed from 'WaitingOnSystemApp | Elapsed 3.000000' for 1.590303 seconds
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: WaitingOnSystemApp | Elapsed 5.000000
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status has not changed from 'WaitingOnSystemApp | Elapsed 5.000000' for 1.564724 seconds
[726166E0-3766-459B-AAEB-971C0E1D326B] Boot Status Changed: Finished | Elapsed 8.000000
[simulator_set] boot: succeeded
{"model":"iPhone 13 mini","os_version":"iOS 16.0","udid":"726166E0-3766-459B-AAEB-971C0E1D326B","architecture":"x86_64","type":"Simulator","name":"iPhone 13 mini","state":"Booted"}
```

dmitry-zaitsev:
> This is usually what I use to run the Android emulator:
> 
> ```
> emulator -avd Pixel_4_API_30 -netdelay none -netspeed full
> ```

Let's stick to a default option to avoid any discrepancies in behaviour when running against an existing emulator. If someone would want to customize those parameters they could just launch an emulator by themselves.

dmitry-zaitsev:
> After printing this error message we should re-prompt the user for input:
> 
> ```
> Invalid index
> ```

We already do that. Or did you mean something else?

dmitry-zaitsev:
> After launching an emulator, it's takes a while for the Maestro flow to actually start executing. Not a blocker

I noticed that too. It seems that adb server takes some time to launch?

maestro-cli/src/main/java/maestro/cli/device/ios/Simctl.kt
dmitry-zaitsev: Thanks for the suggestion. Though that is valid, we are trying to minimise our dependency on `idb_companion` and get by with Apple-provided solutions wherever we can.

lochness42: Sure, I've assumed that if you have to use `idb_companion` anyway, it's worth using it + you don't have to have your own workaround waiting for the sim device to be booted up.

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
dmitry-zaitsev: Tweaked it in a way that it contains a sealed class. Splitting this logic into different methods would do more harm than good, I feel

Leland-Takamine:
> > This is usually what I use to run the Android emulator:
> > ```
> > emulator -avd Pixel_4_API_30 -netdelay none -netspeed full
> > ```
> 
> Let's stick to a default option to avoid any discrepancies in behaviour when running against an existing emulator. If someone would want to customize those parameters they could just launch an emulator by themselves.

I believe this is similar to what android studio does by default

Leland-Takamine:
> > After printing this error message we should re-prompt the user for input:
> > ```
> > Invalid index
> > ```
> 
> We already do that. Or did you mean something else?

I meant re-printing the prompt

dmitry-zaitsev:
> > > After printing this error message we should re-prompt the user for input:
> > > ```
> > > Invalid index
> > > ```
> > 
> > 
> > We already do that. Or did you mean something else?
> 
> I meant re-printing the prompt

The prompt is quite huge and will erase the warning/error message. I would leave it as is

dmitry-zaitsev:
> > > This is usually what I use to run the Android emulator:
> > > ```
> > > emulator -avd Pixel_4_API_30 -netdelay none -netspeed full
> > > ```
> > 
> > 
> > Let's stick to a default option to avoid any discrepancies in behaviour when running against an existing emulator. If someone would want to customize those parameters they could just launch an emulator by themselves.
> 
> I believe this is similar to what android studio does by default

Ah, yes you are right: https://developer.android.com/studio/run/emulator-commandline

Made the change

Leland-Takamine:
> > > > After printing this error message we should re-prompt the user for input:
> > > > ```
> > > > Invalid index
> > > > ```
> > > 
> > > 
> > > We already do that. Or did you mean something else?
> > 
> > 
> > I meant re-printing the prompt
> 
> The prompt is quite huge and will erase the warning/error message. I would leave it as is

Just the one-line prompt. It's not clear that you should enter a number still after the error. It feels like it's just hanging

dmitry-zaitsev:
> > > > > After printing this error message we should re-prompt the user for input:
> > > > > ```
> > > > > Invalid index
> > > > > ```
> > > > 
> > > > 
> > > > We already do that. Or did you mean something else?
> > > 
> > > 
> > > I meant re-printing the prompt
> > 
> > 
> > The prompt is quite huge and will erase the warning/error message. I would leave it as is
> 
> Just the one-line prompt. It's not clear that you should enter a number still after the error. It feels like it's just hanging

Done

Leland-Takamine (approved):

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -0,0 +1,199 @@
...
+                    platform = device.platform,
+                )
+            }
+            Platform.ANDROID -> {
+                val androidHome = EnvUtils.androidHome()
+                    ?: throw CliError("ANDROID_HOME environment variable is not set")
+
+                val emulatorBinary = File(androidHome, "emulator/emulator")
+                    .takeIf { it.exists() }
+                    ?: File(androidHome, "tools/emulator")
```
Leland-Takamine, line 32: This should be `emulator/emulator` I believe

dmitry-zaitsev: See line 30. I use this as a fallback

PR #316, opened 2022-10-25
Tweak: tripled Android Driver launch timeout

dmitry-zaitsev:


amanjeetsingh150 (approved):

amanjeetsingh150:
Do we also want to track GRPC errors and instrumentation test debugging events for keeping a check on timeout?

PR #319, opened 2022-10-26
Fix #279 opening leak canary instead of app launcher

### LINKED ISSUE #279: maestro opens Leak companion app instead of the main app when using a real device
Hello!

**Maestro version**: 1.10.1

When running a flow on a real device it always opens the Leak companion app instead of the main app. Tested it on a Samsung s10e.

This issue was fixed in v1.8.0 with this PR: https://github.com/mobile-dev-inc/maestro/pull/214/files

Back then maestro didn’t support real devices. Looking at the PR, I'm not sure why this approach wouldn't work on real devices as well.

alishari:
## Proposed Changes
There was an issue when I used maestro for an Android app with LeakCanary. I fixed the issue in this PR.

## Issues Fixed
#279 #172

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -159,12 +159,12 @@ class AndroidDriver(
                 }
                 shell("am start-activity -n $appId/${activity.name}")
             } else {
-                shell("monkey --pct-syskeys 0 -p $appId 1")
+                shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
             }
         } catch (ioException: IOException) {
-            shell("monkey --pct-syskeys 0 -p $appId 1")
+            shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
         } catch (saxException: SAXException) {
-            shell("monkey --pct-syskeys 0 -p $appId 1")
+            shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
         }
     }
 


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -159,12 +159,12 @@ class AndroidDriver(
...
             } else {
-                shell("monkey --pct-syskeys 0 -p $appId 1")
+                shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
             }
         } catch (ioException: IOException) {
-            shell("monkey --pct-syskeys 0 -p $appId 1")
+            shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
         } catch (saxException: SAXException) {
-            shell("monkey --pct-syskeys 0 -p $appId 1")
+            shell("monkey --pct-syskeys 0 -p $appId -c android.intent.category.LAUNCHER 1")
```
amanjeetsingh150, line 167: Hey, @alishari thanks for your contribution. To ensure, if I am not wrong this selects activity with the launcher category, right?

The `LeakLauncherActivity` also has launcher category which means it could still launch leak canary activity.

```
<activity-alias android:name="leakcanary.internal.activity.LeakLauncherActivity" ... > 
   <intent-filter>
                <action
                    android:name="android.intent.action.MAIN" />
                <category
                    android:name="android.intent.category.LAUNCHER" />
                <category
                    android:name="android.intent.category.LEANBACK_LAUNCHER" />
            </intent-filter>
</.activity-alias>
```

alishari: Yes, you are right, I will look into it to find a solution.

alishari:
I was unable to find a solution to it. Maybe we should raise the issue on LeakCanary repo and try to figure it out there.

PR #326, opened 2022-10-27
Fix #279 opening leak canary instead of app launcher

### LINKED ISSUE #279: maestro opens Leak companion app instead of the main app when using a real device

alishari:
## Proposed Changes
Using the `ManifestData.launcherActivity` to run the app. I think it should fix #279 

## Issues Fixed

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -159,7 +159,11 @@ class AndroidDriver(
                 }
                 shell("am start-activity -n $appId/${activity.name}")
             } else {
-                shell("monkey --pct-syskeys 0 -p $appId 1")
+                try {
+                    shell("am start-activity -n $appId/${manifest.launcherActivity.name}")
+                } catch (exception: Exception) {
+                    shell("monkey --pct-syskeys 0 -p $appId 1")
+                }
             }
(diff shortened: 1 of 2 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -159,7 +159,7 @@ class AndroidDriver(
                 }
                 shell("am start-activity -n $appId/${activity.name}")
             } else {
-                shell("monkey --pct-syskeys 0 -p $appId 1")
+                shell("am start-activity -n $appId/${manifest.launcherActivity.name}")
```
dmitry-zaitsev, line 162: We would like to keep `monkey` option as a fallback in case if somehow this code doesn't work. Could you perhaps wrap it in a try/catch?

alishari: Done

dmitry-zaitsev:
Thank you!

PR #341, opened 2022-10-31
feat: pass agent: ci in maestro in upload request when known CI env var is set

axelniklasson:


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -86,6 +89,17 @@ class ApiClient(
         }
     }
 
+    private fun getAgent(): String {
+        for (ciVar in knownCIEnvVars) {
+            try {
+                val value = System.getenv(ciVar)
+                if (value.isNotEmpty()) return "ci"
```
Leland-Takamine, line 96: What if we have `CI=false`?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -34,6 +35,8 @@ class ApiClient(
         .protocols(listOf(Protocol.HTTP_1_1))
         .build()
 
+    private val knownCIEnvVars = listOf("CI", "JENKINS_HOME")
```
dmitry-zaitsev, line 38: What about Bitrise?

dmitry-zaitsev: Also Circle CI

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
axelniklasson: Good call, didn't think about that case. Do you think it makes more sense to consider it being ci iff `CI=true` or should we consider it being ci if `CI != false`?

Leland-Takamine: That and also if set to "0"

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
axelniklasson: According to their [docs](https://circleci.com/docs/variables/#built-in-environment-variables) they set `CI=true` so we should be good. realized that Bitrise only sets `CI=true` when running the CLI in continuous integration mode so will update to also check for the `BITRISE_IO` variable

axelniklasson:
@dmitry-zaitsev @Leland-Takamine comments are addressed, let me know what you think!

Leland-Takamine (approved):

dmitry-zaitsev:
@axelniklasson what about Circle CI?

axelniklasson:
> @axelniklasson what about Circle CI?

@dmitry-zaitsev seems GitHub scrapped that comment but it is handled through checking for `CI=true` as that is a [default env variable](https://circleci.com/docs/variables/#built-in-environment-variables) in Circle CI

dmitry-zaitsev (approved):

PR #345, opened 2022-11-01
maestro cloud command

dmitry-zaitsev:
https://user-images.githubusercontent.com/2990722/199242184-e37448a8-7932-48c8-bbfb-e3e9d11d84d9.mov


ArthurSav:
@dmitry-zaitsev Can we have a loading state for the current pending test? So that it doesn't look like it's stuck

dmitry-zaitsev:
The idea is that it runs on CI, so having a loading state wouldn't quite work

lochness42:
Just a quick sanity check question - does this command fail if there is any failing check (exit code 0 vs non-zero one) so CI can easily evaluate whether command was fully successful?

dmitry-zaitsev:
@lochness42 right, I actually forgot to connect the return code from the Interactor. Now it works as you described.

lochness42:
@dmitry-zaitsev thank you, I wasn't sure whether I was overlooking it somewhere as I've never coded in Kotlin. This feature looks really good

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -0,0 +1,203 @@
...
+
+        PrintUtils.warn("Upload did not complete in time due to an issue on mobile.dev side. Follow the results of your upload below:")
+        println(uploadUrl(uploadId, teamId, appId))
+
+        return if (failOnTimeout) {
+            PrintUtils.err("FAIL")
+
+            1
+        } else {
+            PrintUtils.warn("SKIPPED")
```
Leland-Takamine, line 168: What does skipped mean exactly?

maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
```diff
@@ -0,0 +1,88 @@
...
+
+import maestro.cli.api.ApiClient
+import maestro.cli.cloud.CloudInteractor
+import picocli.CommandLine
+import picocli.CommandLine.Option
+import java.io.File
+import java.util.concurrent.Callable
+
+@CommandLine.Command(
+    name = "cloud",
```
Leland-Takamine, line 30: Add a description for the command. Should be pretty detailed since this is how a lot of folks will learn about the command. Maybe we can link to docs as well

maestro-cli/src/main/java/maestro/cli/command/UploadCommand.kt
```diff
@@ -73,76 +65,29 @@ class UploadCommand : Callable<Int> {
...
-            ansi.cursorToColumn(0)
-            ansi.fgCyan()
-            repeat(progressWidth) { ansi.a("█") }
-            repeat(width - progressWidth) { ansi.a("░") }
-            print(ansi)
-        }
+        println(
+            ansi()
+                .fgRed()
+                .render("'maestro upload' command is deprecated and is going to be removed in one of the future releases. Use 'maestro cloud --async' instead.")
```
Leland-Takamine, line 72: Nice

Leland-Takamine (approved):

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
dmitry-zaitsev: This is a rare case where our backend did not complete the upload in 30 minutes. In such case, we are not going to fail the CLI command and instead report it as skipped (exit code 0)

Leland-Takamine: Got it. At first glance, the logic feels like "skipped" is a catch all and I just want to make sure that we aren't letting real issues fall through the cracks by bucketing errors into this skipped state

dmitry-zaitsev: We are going to retry  a few times and then eventually throw an error in case of server errors: https://github.com/mobile-dev-inc/maestro/pull/345/files#diff-c0b9059374eeada60a2aa7a5d016bcbd26ce2fee9e77ce674648c4d9f08550c3R134

Leland-Takamine: Ok reading the code more carefully now this makes perfect sense. Thanks!

PR #355, opened 2022-11-04
Better check ios simulator status when starting tests

berikv:
## Proposed Changes

Check if
* idb port is open
* sim boot status is booted
* accessibility info returns frames with non-zero width

## Testing

Manual testing

## Issues Fixed

When running `maestro test` on an iOS Simulator, the error `java.lang.IllegalStateException: idb_companion did not start in time` can mean several things:
* idb did not start correctly
* iOS Simulator boot time too long
* Issue with Accessibility frameworks SDK / Simulator runtime

This change checks for these causes and reports with a different error message for each of them.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -1,7 +1,6 @@
 package maestro.cli.device
 
 import com.github.michaelbull.result.get
-import com.github.michaelbull.result.runCatching
 import dadb.Dadb
 import io.grpc.ManagedChannelBuilder
 import ios.idb.IdbIOSDevice
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -96,7 +96,6 @@ object DeviceService {
 
             // The first time a simulator boots up, it can
             // take 10's of seconds to complete.
-
             MaestroTimer.retryUntilTrue(timeoutMs = 60000) {
                 val process = ProcessBuilder("xcrun", "simctl", "bootstatus", device.instanceId)
                     .start()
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -79,27 +84,35 @@ object DeviceService {
...
+
+        IdbIOSDevice(channel).use { iosDevice ->
+            MaestroTimer.retryUntilTrue(timeoutMs = 2000) {
+                Socket(idbHost, idbPort).use { true }
+            } || error("idb_companion did not start in time")
+
+            // The first time a simulator boots up, it can
+            // take 10's of seconds to complete.
+            MaestroTimer.retryUntilTrue(timeoutMs = 60000) {
+                val process = ProcessBuilder("xcrun", "simctl", "bootstatus", device.instanceId)
```
artem888, line 99: is is fine to call `xcrun simctl bootstatus` several times, from my perspective it should be just called once and then wait until timeout

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -79,27 +84,35 @@ object DeviceService {
...
+                    .start()
+                process
+                    .waitFor(1000, TimeUnit.MILLISECONDS)
+                process.exitValue() == 0
+            } || error("Simulator failed to boot")
+
+            // Test if idb can get accessibility info elements with non-zero frame with
+            MaestroTimer.retryUntilTrue(timeoutMs = 20000) {
+                val nodes = iosDevice
+                    .contentDescriptor()
```
artem888, line 109: would be nice to have some sort of retryInterval here to not to spam idb_companion with too many requests

artem888 (approved):

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
berikv: It'll return "Booting" as a status when the simulator is not finished booting

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
berikv: I don't think it is much of an issue. The network roundtrip and request encoding / decoding will slow everything down anyway.

Issue #251, opened 2022-10-03
Control swipe speed


dmitry-zaitsev:

mollyhartwick3:
this has been implemented in the linked pull request ^

PR #358, opened 2022-11-07
Adds directional swipe, makes swipe speed configurable through duration and fix iOS scroll

amanjeetsingh150:
### Adding directional swipe

Adds directional swipe: RIGHT, LEFT, DOWN, UP. This will help in use cases like swiping pagers in onboarding irrespective of device width height. API:
```
- swipe:
      direction: LEFT | RIGHT | UP | DOWN
```
PS: start and end coordinates still remain but are optional now.

### Controlling swipe speed
Swipe speed can be controlled by supplying duration. More the duration slower the swipe speed. The lesser the duration the more swipe speed. API:

```
- swipe:
     direction: LEFT
     duration: 300
```

### Fix iOS scrolling
Accounting scroll delta and duration which fixes iOS vertical scrolling as well

## Testing

https://user-images.githubusercontent.com/12881364/200290424-32c19707-3bbf-4eef-8658-7dfd175e947a.mp4

https://user-images.githubusercontent.com/12881364/200289793-119f3a61-56f5-4d06-b8d5-ab98b445bbb9.mp4

https://user-images.githubusercontent.com/12881364/200299337-6fe688dc-38a7-4feb-85a9-2243ab42ea20.mp4

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/251

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -252,7 +252,43 @@ class IOSDriver(
     }
 
     override fun swipe(swipeDirection: SwipeDirection) {
-        TODO("Not yet implemented")
+        val width = widthPixels ?: throw IllegalStateException("Device width not available")
+        val height = heightPixels ?: throw IllegalStateException("Device height not available")
+
+        when(swipeDirection) {
+            SwipeDirection.UP -> {
+                iosDevice.scroll(
+                    xStart = width / 4,
+                    yStart = height,
+                    xEnd = width / 4 ,
+                    yEnd = height / 4
+                ).expect {}
+            }
+            SwipeDirection.DOWN -> {
+                iosDevice.scroll(
+                    xStart = width / 4,
+                    yStart = 0,
+                    xEnd = width,
+                    yEnd = height / 4
+                ).expect {}
+            }
+            SwipeDirection.RIGHT -> {
+                iosDevice.scroll(
+                    xStart =  0,
+                    yStart = height / 4,
+                    xEnd =  width / 4 ,
+                    yEnd = height / 4
+                ).expect {}
+            }
+            SwipeDirection.LEFT -> {
+                iosDevice.scroll(
+                    xStart = width / 4 ,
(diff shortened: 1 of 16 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -239,20 +243,70 @@ class IOSDriver(
...
+                    IdbIOSDevice.ScrollType.SWIPE
+                ).expect {}
+            }
+        }
+    }
+
     override fun backPress() {}
 
-    override fun hideKeyboard() {
+    override fun hideKeyboard() {3
```
Leland-Takamine, line 309: Typo

Leland-Takamine:
What happens when direction and start/end are set? What about when only start is set but not end?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Fixed ✅

amanjeetsingh150:
> What happens when direction and start/end are set? What about when only start is set but not end?

Nice catch I have added a YamlSwipeDeserializer, which will look at fields. If cases like this happened I am throwing IllegalStateException:

```
java.lang.IllegalArgumentException: Provide swipe direction UP, DOWN, RIGHT, OR LEFT or by giving explicit start and end coordinates.
```

Let me know if you think the error message is not explicit enough or should be improved.

amanjeetsingh150:
> I think the error message should be clearer

Hey @Leland-Takamine , have addressed the error messages. Have addressed all the common errors and raised different messages. Considered cases:

You can check error messages in this commit: https://github.com/mobile-dev-inc/maestro/pull/358/commits/d69476d37ffb1fd7992ee9c053597d038cf437bb

```
     1. Typo in direction command
     2. Typo in the start command
     3. Typo in the end command
     4. Providing direction start
     5. Providing direction end
     6. Providing direction start end
     7. Providing end only
     8. Providing start only
```

Leland-Takamine (approved):

PR #361, opened 2022-11-07
Wait longer for idb and simulator, print wait status to cli

berikv:
## Proposed Changes

* Timeout idb_companion starting if socket is not open within one minute
* Timeout Simulator booting if bootstatus is not positive within two minutes
* Print the to the cli what is being waited on

Screenshot 2022-11-07 at 19 53 48

## Testing

Tested locally

## Issues Fixed

Launching idb_companion can take up to a minute, Below a minute makes maestro runs unstable (sometimes idb is fast enough, sometimes it isn't).
Booting the simulator on an older, less capable Mac for the first time will take over a minute to complete.
Users waiting for maestro to start idb and the simulator are informed via cli messages.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -90,13 +90,15 @@ object DeviceService {
             .build()
 
         IdbIOSDevice(channel).use { iosDevice ->
-            MaestroTimer.retryUntilTrue(timeoutMs = 2000) {
+            println("Waiting for idb service to start..")
+            MaestroTimer.retryUntilTrue(timeoutMs = 60000) {
                 Socket(idbHost, idbPort).use { true }
             } || error("idb_companion did not start in time")
 
             // The first time a simulator boots up, it can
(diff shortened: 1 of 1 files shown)


Leland-Takamine (approved):

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -90,13 +90,15 @@ object DeviceService {
             .build()
 
         IdbIOSDevice(channel).use { iosDevice ->
-            MaestroTimer.retryUntilTrue(timeoutMs = 2000) {
+            println("Waiting for idb service to start..")
```
artem888, line 93: would be also useful to have date timestamps  in those logs

artem888 (approved):

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
berikv: This is not a log, this is a message to the user so that they know why maestro isn't showing a result yet.
I agree we should build some logging facility. But this is not the right place.

PR #364, opened 2022-11-08
Fix an issue when clearState operation had no effect, when repeating flows

artem888:
## Proposed Changes
- [x] adds a stopApp command before doing the launch, turned out to be necessary for clearState operation to have a full effect

## Testing
Tested locally

https://user-images.githubusercontent.com/961175/200639412-9f0fd6bb-61f9-4cbc-8198-3d8f6f00e5cb.mov

### DIFF
--- a/maestro-client/src/main/java/maestro/MaestroTimer.kt
+++ b/maestro-client/src/main/java/maestro/MaestroTimer.kt
@@ -23,10 +23,13 @@ object MaestroTimer {
         return null
     }
 
-    fun retryUntilTrue(timeoutMs: Long, block: () -> Boolean): Boolean {
+    fun retryUntilTrue(timeoutMs: Long, delayMs: Long? = null, block: () -> Boolean): Boolean {
         val endTime = System.currentTimeMillis() + timeoutMs
         do {
             try {
+                delayMs?.let {
+                    sleep(Reason.BUFFER, delayMs)
+                }
                 if (block()) {
                     return true
                 }
--- a/maestro-client/src/main/java/maestro/MaestroTimer.kt
+++ b/maestro-client/src/main/java/maestro/MaestroTimer.kt
@@ -23,13 +23,10 @@ object MaestroTimer {
         return null
     }
 
-    fun retryUntilTrue(timeoutMs: Long, delayMs: Long? = null, block: () -> Boolean): Boolean {
+    fun retryUntilTrue(timeoutMs: Long, block: () -> Boolean): Boolean {
         val endTime = System.currentTimeMillis() + timeoutMs
         do {
             try {
-                delayMs?.let {
-                    sleep(Reason.BUFFER, delayMs)
-                }
                 if (block()) {
                     return true
                 }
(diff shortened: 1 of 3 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -283,6 +284,10 @@ class Orchestra(
     }
 
     private fun launchAppCommand(it: LaunchAppCommand) {
+        // stop app before launching it, so clear operations have an effect
+        maestro.stopApp(it.appId)
+        MaestroTimer.sleep(MaestroTimer.Reason.BUFFER, 3000)
```
Leland-Takamine, line 289: We should try to avoid arbitrary sleeps like this - what is the reason for the sleep here?

maestro-client/src/main/java/maestro/MaestroTimer.kt
```diff
@@ -23,10 +23,13 @@ object MaestroTimer {
...
     }
 
-    fun retryUntilTrue(timeoutMs: Long, block: () -> Boolean): Boolean {
+    fun retryUntilTrue(timeoutMs: Long, delayMs: Long? = null, block: () -> Boolean): Boolean {
         val endTime = System.currentTimeMillis() + timeoutMs
         do {
             try {
+                delayMs?.let {
+                    sleep(Reason.BUFFER, delayMs)
+                }
```
berikv, line 32: Is this needed to fix clearState? If not, then please add this in a separate PR.

One issue I have with this approach is that the user is waiting 3 seconds for no reason. If we really want to put a delay there (which I'm not convinced that we need that), it would be better placed after the block has been evaluated once.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
artem888: stopApp idb_companion operation finishes almost immediately (same thing for `xcrun simctl terminate`), while it probably takes longer to kill all the simulator app processes, if I remove the delay clearState won't work again.

can experiment with polling a PID of the running app, but I'd say can be done as an improvement.

Leland-Takamine: > can experiment with polling a PID of the running app

@artem888 Can you create a task and follow up on this?

maestro-client/src/main/java/maestro/MaestroTimer.kt
artem888: sounds good, the thing is without this delay it was hanging almost every second time for me and might happen to other users as well. also smth more robust like RxJava or Coroutines would be a better idea here.

I'll create a separate PR and issue, lets discuss there.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
berikv: Sleeps without a validation check at the end make the testing platform flakey.
If it is possible to check the state that's waited on, then we should build that check. (waitloop + exception on timeout). Checking the pid shouldn't be too difficult (and is quite reliable: osses prevent reuse of pids as much as possible).

Leland-Takamine: If it's not too difficult I agree with doing the PID check in this change

dmitry-zaitsev (changes requested):
Let's indeed avoid the sleep and do a PID check

PR #370, opened 2022-11-09
Fix idb service start hanging issue

artem888:
## Proposed Changes
Currently idb_companion start often hangs infinitely with the message `Waiting for idb service to start...`, this PR fixes that.

## Testing
Tested locally

### DIFF
--- a/maestro-client/src/main/java/maestro/MaestroTimer.kt
+++ b/maestro-client/src/main/java/maestro/MaestroTimer.kt
@@ -23,10 +23,13 @@ object MaestroTimer {
         return null
     }
 
-    fun retryUntilTrue(timeoutMs: Long, block: () -> Boolean): Boolean {
+    fun retryUntilTrue(timeoutMs: Long, delayMs: Long? = null, block: () -> Boolean): Boolean {
         val endTime = System.currentTimeMillis() + timeoutMs
         do {
             try {
+                delayMs?.let {
+                    sleep(Reason.BUFFER, delayMs)
+                }
                 if (block()) {
                     return true
                 }
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -91,14 +91,14 @@ object DeviceService {
 
         IdbIOSDevice(channel).use { iosDevice ->
             println("Waiting for idb service to start..")
-            MaestroTimer.retryUntilTrue(timeoutMs = 60000) {
+            MaestroTimer.retryUntilTrue(timeoutMs = 60000, delayMs = 1000) {
                 Socket(idbHost, idbPort).use { true }
             } || error("idb_companion did not start in time")
 
             // The first time a simulator boots up, it can
             // take 10's of seconds to complete.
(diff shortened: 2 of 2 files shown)


maestro-client/src/main/java/maestro/MaestroTimer.kt
```diff
@@ -23,10 +23,13 @@ object MaestroTimer {
...
     }
 
-    fun retryUntilTrue(timeoutMs: Long, block: () -> Boolean): Boolean {
+    fun retryUntilTrue(timeoutMs: Long, delayMs: Long? = null, block: () -> Boolean): Boolean {
         val endTime = System.currentTimeMillis() + timeoutMs
         do {
             try {
+                delayMs?.let {
+                    sleep(Reason.BUFFER, delayMs)
+                }
```
amanjeetsingh150, line 32: Curious how delay is preventing the hang we are facing?

artem888: it seems that there are two issues:
1. polling `Socket(idbHost, idbPort)` too fast, adding some delay here is healthy, may be decreasing it to 500ms or smth would be fine. same for further operations like `iosDevice.contentDescriptor()`
2. this do->while thing hangs infinitely even after 60s timeout is passed, so I'm not sure if we could rely on this wrapper function long-term at all, I'd rather use RxJava or Coroutines instead here.

as a quickfix I'd suggest to merge this now, because many users can be involved. and then work on more robust approach meanwhile.

amanjeetsingh150 (approved):

maestro-client/src/main/java/maestro/MaestroTimer.kt
berikv: > as a quickfix I'd suggest to merge this now
Agree, to have this merged today (before the launch event)

Can the delay be inserted after the first run of `block()`?

berikv: Lets also set the delay to something like 100-250ms. 100ms is a really long pause compared to local-socket validation time.

artem888: yes, it doesn't timeout while hanging for me, does it for you? 

I think such utility classes for things like retry, retryWithDelay, poll and relying on `System.currentTimeMillis()` is not the best practice, feels like reinventing the wheel a bit, because there will be always some edge cases. In frameworks like RxJava / Coroutines such things are implemented in a more robust way and are better tested.

berikv (approved):

maestro-client/src/main/java/maestro/MaestroTimer.kt
artem888: found this one https://github.com/michaelbull/kotlin-retry, which is pretty lightweight, but works on top of coroutines 🤔

berikv: No objection to adding a dependency to kotlin-retry

berikv: Tested the branch locally, looks to be stable and fast!

PR #374, opened 2022-11-09
Prepare next development version

artem888:


gradle.properties
```diff
@@ -2,7 +2,7 @@ android.useAndroidX=true
 android.enableJetifier=true
 kotlin.code.style=official
 GROUP=dev.mobile
-VERSION_NAME=1.13.1
+VERSION_NAME=1.13.1-SNAPSHOT
```
dmitry-zaitsev, line 5: It should be 1.13.2-SNAPSHOT

artem888: fixed

dmitry-zaitsev (approved):

Issue #365, opened 2022-11-08
not able to input underscore with inputText command on iOS


tom-daggett-wd:
on iOS, I'm not able to input an underscore in a text field.

- inputText: "foo_bar"

seems to get inputed in to an actual text field as "foobar".

The problem appears to be this file:
https://github.com/mobile-dev-inc/maestro/blob/main/maestro-ios/src/main/java/ios/idb/TextInputUtil.kt
a double-quote character invokes 
`'"' to keyPressShiftedToEvents(45)`   -- see here: https://github.com/mobile-dev-inc/maestro/blob/db57d3f79c76b9a1c83556c6fba8eeda95f1a224/maestro-ios/src/main/java/ios/idb/TextInputUtil.kt#L198
and 
`'"' to keyPressShiftedToEvents(52)` -- see here: https://github.com/mobile-dev-inc/maestro/blob/db57d3f79c76b9a1c83556c6fba8eeda95f1a224/maestro-ios/src/main/java/ios/idb/TextInputUtil.kt#L203

I'm pretty sure that this line https://github.com/mobile-dev-inc/maestro/blob/db57d3f79c76b9a1c83556c6fba8eeda95f1a224/maestro-ios/src/main/java/ios/idb/TextInputUtil.kt#L198
should be 
`'_' to keyPressShiftedToEvents(45)`
since 45 is the raw DEC usage code for the hyphen key ( https://developer.apple.com/documentation/uikit/uikeyboardhidusage/keyboardhyphen ) which when the shifted key is pressed turns into an underscore.

franciscomorais:
up

PR #387, opened 2022-11-11
Fix not able to input underscore.

franciscomorais:
## Proposed Changes
Change the underscore unicode in keyMap.

## Testing
- tapOn: input
-  inputText: foo_bar

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/365


maestro-ios/src/main/java/ios/idb/TextInputUtil.kt
```diff
@@ -195,7 +195,7 @@ internal object TextInputUtil {
         '*' to keyPressShiftedToEvents(37),
         '(' to keyPressShiftedToEvents(38),
         ')' to keyPressShiftedToEvents(39),
-        '"' to keyPressShiftedToEvents(45),
+        '_' to keyPressShiftedToEvents(45),
```
dmitry-zaitsev, line 198: isn't underscore code 95?

franciscomorais: Sorry @dmitry-zaitsev .. I can't test. The truth is that I was doing a flow and I noticed that the underscore is replaced by "". I found this issue and decided to PR it.

If you want I can change it to 95 but I can't really test it.

franciscomorais: @dmitry-zaitsev Okay, I ended up change it. Please, take another look. Thanks.

dmitry-zaitsev: I checked your changes and your original version was correct. It should be 45

Sorry for the confusion

dmitry-zaitsev:
Could you please exclude version number changes from this PR?

jrelorcasa (approved):

dmitry-zaitsev:
Thanks for pointing out the solution! Since you did not update PR I just pushed the fix myself

tom-daggett-wd:
Note that what got pushed in this PR is this change:
`'_' to keyPressShiftedToEvents(95),`
but from the comments, I  thought people agreed that the change *should* be:
`'_' to keyPressShiftedToEvents(45),`

Leland-Takamine:
@tom-daggett-wd This is the PR that ended getting merged: https://github.com/mobile-dev-inc/maestro/pull/411

PR #396, opened 2022-11-14
Do not kill idb_companion, exit with an error instead

berikv:
## Proposed Changes

Do not kill idb_companion when running maestro. Maestro should not kill an unrelated process.
Instead, warn the user that idb_companion is running and exit.

## Testing
Tested locally

## Issues Fixed
MOB-1551

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
@@ -73,7 +73,6 @@ object DeviceService {
 
     fun prepareDevice(device: Device.Connected) {
         if (device.platform == Platform.IOS) {
-            killIdbCompanion()
             startIdbCompanion(device)
         }
     }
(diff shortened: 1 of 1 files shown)


artem888 (approved):

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -82,6 +81,14 @@ object DeviceService {
         val idbHost = "localhost"
         val idbPort = 10882
 
+        val isIdbCompanionRunning =
+            try { Socket(idbHost, idbPort).use { true } }
+            catch(_: Exception) { false }
+
+        if (isIdbCompanionRunning) {
```
dmitry-zaitsev, line 88: Should we perhaps reuse existing idb_companion instance if it is already running?

berikv: I considered this possibility, and it'd probably work in the majority of cases.
Issues are:
- user might not notice that maestro is using their running idb, which can lead to surprises
- maestro asks the user to select a device, then continues to use an already started idb which is configured to use a different device

dmitry-zaitsev: That is a fair point

dmitry-zaitsev (approved):

PR #397, opened 2022-11-14
Feature: Run all tests from a folder

dmitry-zaitsev:
Running all tests in the folder (excluding subfolders)

image

There are some TODOs in the PR that are there in preparation for test reporting


Leland-Takamine (approved):

Leland-Takamine:
Just confirming - when running against a single file the output is the same as before correct?

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -0,0 +1,143 @@
...
+        PrintUtils.message("Waiting for flows to complete...")
+        println()
+
+        var passed = true
+        flows.forEach { flowFile ->
+            val result = runFlow(flowFile, env, maestro)
+
+            if (result.status == FlowStatus.ERROR) {
+                passed = false
+            }
```
berikv, line 58: This feels a bit fragile, though currently not an issue. The code assumes passed = true, then has a condition to set passed to false.
There is no way that the runFlow can except out so that passed is true but the test failed. Still it would be nice to tight this code down a bit more. For example by having a `successCount = 0` then adding 1 for each successful test. Then success will be `successCount == flows.count()`

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -0,0 +1,143 @@
...
+            runTestSuite(
+                maestro, device, listOf(input), env
+            )
+        } else {
+            val flowFiles = input
+                .listFiles()
+                .filter {
+                    it.isFile
+                        && it.extension in setOf("yaml", "yml")
+                        && it.nameWithoutExtension != "config"
```
berikv, line 31: Let's also ignore files starting with underscore. That way users can quickly disable tests.

berikv (approved):

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
dmitry-zaitsev: I find the counter to be quite less intuitive than a boolean as it simply has many more possible states and in fact is not going to be used as a counter

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
dmitry-zaitsev: I am planning to introduce an `ignore: true` in flows later on so that we can have an explicit support for that

PR #403, opened 2022-11-15
adds copy paste command

ArthurSav:
## Proposed Changes
- Adds command to copy text from an element 
- Adds command to paste that same text into an input field later

## Testing
- Added integration test
- Tested manually on iOS/Android contacts app

### DIFF
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -69,4 +69,10 @@ interface Driver {
     fun takeScreenshot(out: Sink)
 
     fun setLocation(latitude: Double, longitude: Double)
+
+    fun copyText(text: String)
+
+    fun pasteText()
+
+    fun copiedText(): String?
 }
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -69,10 +69,4 @@ interface Driver {
     fun takeScreenshot(out: Sink)
 
     fun setLocation(latitude: Double, longitude: Double)
-
-    fun copyText(text: String)
-
-    fun pasteText()
-
-    fun copiedText(): String?
 }
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -41,8 +41,6 @@ import javax.imageio.ImageIO
 @Suppress("unused", "MemberVisibilityCanBePrivate")
 class Maestro(private val driver: Driver) : AutoCloseable {
 
-    private var savedElementText: String = ""
-
     fun deviceName(): String {
         return driver.name()
     }
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -105,7 +105,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.hideKeyboard()
         waitForAppToSettle()
     }
-
+    
     fun swipe(swipeDirection: SwipeDirection? = null, start: Point? = null, end: Point? = null, duration: Long) {
         when {
             swipeDirection != null -> driver.swipe(swipeDirection, duration)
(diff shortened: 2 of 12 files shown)


maestro-test/src/test/resources/062_copy_paste_text.yaml
```diff
@@ -0,0 +1,5 @@
+appId: com.example.app
+---
+- copyText:
```
dmitry-zaitsev, line 3: Let's call it `copyTextFrom` to not create an illusion that it will copy from a currently focused view

maestro-test/src/test/resources/062_copy_paste_text.yaml
```diff
@@ -0,0 +1,5 @@
...
+    id: "myId"
+- pasteText
```
dmitry-zaitsev, line 5: Let's either reuse `clipboardPaste` or remove `clipboardPaste`

maestro-client/src/main/java/maestro/Driver.kt
```diff
@@ -71,4 +71,10 @@ interface Driver {
     fun takeScreenshot(out: Sink)
 
     fun setLocation(latitude: Double, longitude: Double)
+
+    fun copyText(text: String)
```
dmitry-zaitsev, line 75: This shouldn't live in `Driver` as it is not device-specific. In fact, this should probably be isolated to `Orchestra`

dmitry-zaitsev: This applies to all 3 new methods

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -168,6 +168,30 @@ class ClipboardPasteCommand : Command {
...
+class CopyTextCommand(
+    val selector: ElementSelector,
+): Command {
+
+    override fun description(): String {
+        return "Copy text from element with ${selector.description()}"
+    }
+
+    override fun injectEnv(env: Map<String, String>): CopyTextCommand {
+        return this
```
dmitry-zaitsev, line 180: You need to inject `env` into `selector` - this would allow users to use `env` parameters in a selector

dmitry-zaitsev: Can be done with `selector.injectEnv(env)`

maestro-client/src/main/java/maestro/Driver.kt
ArthurSav: There are 2 reasons i added it to the driver:
- Made testing easier
- Potential to expand functionality per device

With that said, i have no strong opinion on it.

maestro-test/src/test/resources/062_copy_paste_text.yaml
dmitry-zaitsev: Sounds good 👍

maestro-client/src/main/java/maestro/Driver.kt
dmitry-zaitsev: - I believe your current tests will cover this logic even if it would leave in `Orchestra`
- Let's add something to Driver only once we have a concrete use case where functionality could indeed be expanded

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -41,6 +41,8 @@ import javax.imageio.ImageIO
 @Suppress("unused", "MemberVisibilityCanBePrivate")
 class Maestro(private val driver: Driver) : AutoCloseable {
 
+    private var copiedText: String? = null
```
dmitry-zaitsev, line 44: Not used?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -67,6 +66,8 @@ class AndroidDriver(
 
     private var instrumentationSession: AdbShellStream? = null
 
+    private var copiedText: String? = null
```
dmitry-zaitsev, line 69: Not used?

dmitry-zaitsev (approved):

PR #408, opened 2022-11-16
Store idb_companion logs for later inspection

berikv:
## Proposed Changes

- DebugLogStore stores log files in $HOME/Library/Logs/maestro/ (similar but different dirs for Linux and Windows).
- Add method to DebugLogStore to capture all output of a subprocess
- DebugLogStore deletes any logs except the last 5 (this run included)
- Store idb_companion logs using DebugLogStore

## Testing
Tested locally

### DIFF
--- a/maestro-cli/build.gradle
+++ b/maestro-cli/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     implementation 'io.ktor:ktor-client-core:2.1.2'
     implementation 'io.ktor:ktor-client-cio:2.1.2'
     implementation 'org.rauschig:jarchivelib:1.2.0'
+    implementation 'net.harawata:appdirs:1.2.1'
 }
 
 java {
--- a/maestro-cli/src/main/java/maestro/cli/App.kt
+++ b/maestro-cli/src/main/java/maestro/cli/App.kt
@@ -27,6 +27,7 @@ import maestro.cli.command.PrintHierarchyCommand
 import maestro.cli.command.QueryCommand
 import maestro.cli.command.TestCommand
 import maestro.cli.command.UploadCommand
+import maestro.cli.debuglog.DebugLogStore
 import org.fusesource.jansi.AnsiConsole
 import picocli.CommandLine
 import picocli.CommandLine.Command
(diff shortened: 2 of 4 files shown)


maestro-cli/build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
     implementation 'io.ktor:ktor-client-core:2.1.2'
     implementation 'io.ktor:ktor-client-cio:2.1.2'
     implementation 'org.rauschig:jarchivelib:1.2.0'
+    implementation 'net.harawata:appdirs:1.2.1'
```
artem888, line 40: do we need to use a 3rd party for that? the logic there seems to be quite easy to reimplement on our own, Windows can be tricky though 🙂

artem888 (approved):

maestro-cli/build.gradle
berikv: I'm fine either way the logic is simple and should be *mostly* correct for a given platform.

berikv: Do you feel strongly about this @artem888? We can at any moment choose to refactor and write it ourselves.
The Appdirs library is tiny in code size, well written and most importantly, maintained. I'd rather use this than having to deal with feature requests for other platforms because their file system changed.

artem888: sounds fine for me , especially if it is well-maintained.

berikv:
Added system_info.log with some basic information:
```
Maestro version: 1.14.0
OS: Mac OS X
OS version: 11.5
OS Architecture: x86_64
```

Zip the logs at the end of the run to save storage (80MB becomes 3MB in the runs below):
```
 11M	2022-11-16_191259
 29M	2022-11-16_191427
 80M	2022-11-16_191703
 80M	2022-11-16_192004
2.9M	2022-11-16_192846.zip
```

berikv:
Added `loggerFor(class:)`:

```
private val logger = DebugLogStore.loggerFor(DeviceService::class.java)
logger.info("This log line will appear in the \$HOME/Library/Logs/maestro/<date>/maestro.log file")
```

PR #412, opened 2022-11-17
Tidy up DebugLogStore

berikv:


maestro-cli/src/main/java/maestro/cli/debuglog/DebugLogStore.kt
```diff
@@ -23,15 +23,13 @@ object DebugLogStore {
...
-        System.setProperty("java.util.logging.SimpleFormatter.format", "[%1\$tF %1\$tT] [%4$-7s] %5\$s %n")
-
         val dateFormatter = DateTimeFormatter.ofPattern(LOG_DIR_DATE_FORMAT)
         val date = dateFormatter.format(LocalDateTime.now())
         val baseDir = File(AppDirsFactory.getInstance().getUserLogDir(APP_NAME, null, APP_AUTHOR))
         logDirectory = File(baseDir, date)
         logDirectory.mkdirs()
         removeOldLogs(baseDir)
-        println("Debug logs are stored in $logDirectory")
+        println("Debug log store $logDirectory.zip")
```
dmitry-zaitsev, line 32: Could we remove this from the output btw? To keep it clean for happy cases

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/debuglog/DebugLogStore.kt
berikv: Let's remove it once there is a `maestro bugreport` command available.

dmitry-zaitsev: People will be confused if we add it now and then remove it later. We can instead instruct users on where to look for those logs in Slack and in our docs

dmitry-zaitsev: > Otherwise, we can add maestro bugreport now and make it print out the log directory and ask the user to sent the zip files in there.

That sounds like a good start!

berikv: Done in https://github.com/mobile-dev-inc/maestro/pull/414

PR #413, opened 2022-11-17
Use logger.warning to send a message to the console

berikv:
## Proposed Changes

Use `logger.warning()` to message both to the console as well as to the debug log.
If this is accepted I'll update all the println() lines to use `logger.warning()` instead.

`logger.severe()` will print "Error: <message>" to the console.


dmitry-zaitsev:
Could you share a screenshot of a terminal output?

berikv:
> Could you share a screenshot of a terminal output?

Maestro clears the terminal, so the printed lines are hidden:
Screenshot 2022-11-17 at 12 35 56

If you scroll up (and before the terminal clear happens) the output looks like this:
Screenshot 2022-11-17 at 12 37 54

dmitry-zaitsev (approved):

PR #428, opened 2022-11-22
Feature: Run Javascript files

dmitry-zaitsev:
## Proposed Changes

Run Javascript files and record the output.

### Run script

```
# flow.yaml
- runScript: script.js
```

### Record output

Each Flow file has its own wrapper `output` object defined that other scripts can populate. Those objects are unique to each flow file and are not shared (i.e. scripts that run under `runFlow` will not interfere with outputs of their parent flow)

```
// script.js
output.myOutputName = 'Hello World'
```

```
# flow.yaml
- runScript: script.js
- inputText: ${output.myOutputName}
```

### Global variables

Sometimes, one might need to populate a variable outside of their flow's scope. To do that, user can update a `ext` variable that is a singleton defined at the root level:

```
// It doesn't matter where this script is located, the variable is going to be visible to all other flows / scripts
ext.myGlobalVariable = 'Hello World!'
```


Leland-Takamine (approved):
Approved pending change to remove ext as discussed

PR #434, opened 2022-11-24
Retry upload request up to 3 times on 5XX

axelniklasson:
This PR updates `ApiClient.upload` to automatically retry upload requests up to 3 times when a 5XX is returned back from the Maestro Cloud API.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -37,6 +37,7 @@ class ApiClient(
         .build()
 
     private val knownCIEnvVars = listOf("CI", "JENKINS_HOME", "BITRISE_IO")
+    private val BASE_RETRY_DELAY_MS = 3000L
 
     fun magicLinkLogin(email: String, redirectUrl: String): Result<String, Response> {
         return post<Map<String, Any>>(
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -179,7 +183,29 @@ class ApiClient(
 
         response.use {
             if (!response.isSuccessful) {
-                throw CliError("Upload request failed (${response.code}): ${response.body?.string()}")
+                if (response.code >= 500 && completedRetries < maxRetryCount) {
+                    PrintUtils.message("Request failed, retrying...")
```
axelniklasson, line 187: @dmitry-zaitsev the progress bar is present during the retries and is reset upon the next request, so this message is not showing up when testing locally. Do you know of a neat way to get this info out to the user? Worst case we can drop it as the progress bar resetting might a good enough of an indicator that it is being retired. Let me know what you think!

dmitry-zaitsev: I think that is fine, the message will show up in CI output anyway (as it does not erase the progress bar there)

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
axelniklasson: Great, thanks! 👍

PR #435, opened 2022-11-24
Feature: Wait for animation end command

artem888:
## Proposed Changes

Implement a command to wait until an animation has fully finished.

Some suitable use-cases: splash screen animation, animation is triggered as a side-effect of some action (i.e. scroll or swipe)

yaml example:
```
appId: org.wikimedia.wikipedia
---
- launchApp
- waitForAnimationToEnd
```

## Testing
Tested locally with apps that have splash screen animations

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -388,10 +388,11 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.eraseText(charactersToErase)
     }
     
-    fun waitForAnimationToEnd() {
-        LOGGER.info("Waiting for animation to end")
+    fun waitForAnimationToEnd(timeout: Long?) {
+        val timeout = timeout ?: ANIMATION_TIMEOUT_MS
+        LOGGER.info("Waiting for animation to end with timeout $timeout")
 
-        MaestroTimer.retryUntilTrue(ANIMATION_TIMEOUT_MS) {
+        MaestroTimer.retryUntilTrue(timeout) {
             val startScreenshot: BufferedImage? = tryTakingScreenshot()
             val endScreenshot: BufferedImage? = tryTakingScreenshot()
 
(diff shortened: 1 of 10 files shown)


dmitry-zaitsev:
Could you think of a way to implement an `IntegrationTest` for that?

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -379,10 +382,35 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.setLocation(latitude, longitude)
     }
 
+    fun waitForAnimationToEnd() {
+        LOGGER.info("Waiting for animation to end")
+
+        MaestroTimer.retryUntilTrue(ANIMATION_TIMEOUT_MS) {
```
dmitry-zaitsev, line 388: Let's make it configurable by the user

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -379,10 +382,35 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+                endScreenshot != null &&
+                startScreenshot.width == endScreenshot.width &&
+                startScreenshot.height == endScreenshot.height
+            ) {
+                val imageDiff = ImageComparison(
+                    startScreenshot,
+                    endScreenshot
+                ).compareImages().differencePercent
+
+                return@retryUntilTrue imageDiff <= SCREENSHOT_DIFF_THRESHOLD
```
dmitry-zaitsev, line 402: Perhaps we could make threshold configurable as well. What do you think?

dmitry-zaitsev (approved):
Nice!

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -379,10 +382,35 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+
+            return@retryUntilTrue false
+        } || throw MaestroException.AnimationDidTimeout("Animation did timeout, took longer than $ANIMATION_TIMEOUT_MS ms")
+    }
+
     companion object {
 
         private val LOGGER = LoggerFactory.getLogger(Maestro::class.java)
         private const val SCREENSHOT_DIFF_THRESHOLD = 0.005 // 0.5%
+        private const val ANIMATION_TIMEOUT_MS: Long = 5000
```
dmitry-zaitsev, line 413: This might be a bit short, let's set it to 15s

maestro-client/src/main/java/maestro/Maestro.kt
artem888: this one not sure, I think it'd not be that clear for a user

dmitry-zaitsev: Fair, let's not add it for now

dmitry-zaitsev:
As discussed, let's make sure this command does not fail the test (at least not by default) if the animation did not complete

PR #452, opened 2022-12-01
[Feature] - Tap with percents

### LINKED ISSUE #363: [FeatureRequest] Point format
Today:

```
- tapOn:
    point: 100,200    # This command will tap on point x:100 y:200 on the screen (in pixels)
```

We can only type in pixels, which is very device dependent. How can we make this device-independent?

Proposal:

```
- tapOn:
    point: 100,200    # This command will tap on point x:100 y:200 on the screen (in pixels)
- tapOn:
    point: -100,200    # This command will tap on point x:(screen width - 100) y:200 on the screen (in pixels), allowing to target elements on the right or bottom without knowing the screen size
- tapOn:
    point: "80%,20%"    # This command will tap on point x:80% of screen width (so 960 pixel is screen is 1200px) y:20% on the screen height (so 720 pixel is height is 3600 px)
```

mslalith:
## Testing
I ran a local build on my other OSS project. I enabled `Pointer location` from Developer Options (Android) for testing to be easy.
```yaml
- launchApp
- tapOn:
    point: 100,100
- tapOn:
    point: 10%,70%
```

## Issues Fixed
#363

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -206,6 +206,20 @@ data class TapOnPointCommand(
     }
 }
 
+data class TapOnPercentCommand(
+    val percentX: Int,
+    val percentY: Int,
+    val retryIfNoChange: Boolean? = null,
+    val waitUntilVisible: Boolean? = null,
+    val longPress: Boolean? = null,
+) : Command {
+    override fun description(): String {
+        return "Top on percent ($percentX%, $percentY%)"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command = this
+}
+
 // Do not delete this class. It might have been already serialized in the past and stored in DB.
 @Deprecated("Use AssertConditionCommand instead")
 data class AssertCommand(
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -214,7 +214,7 @@ data class TapOnPercentCommand(
     val longPress: Boolean? = null,
 ) : Command {
     override fun description(): String {
-        return "Top on percent ($percentX%, $percentY%)"
+        return "Tap on percent ($percentX%, $percentY%)"
     }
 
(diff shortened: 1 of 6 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -206,6 +206,20 @@ data class TapOnPointCommand(
     }
 }
 
+data class TapOnPercentCommand(
+    val percentX: Int,
```
dmitry-zaitsev, line 210: I will make some tweaks here myself - those need to be `String` values to allow users to inject Javascript in there (in `evaluateScripts`)

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
```diff
@@ -294,20 +294,33 @@ data class YamlFluentCommand(
...
+                    .map { it.trim().toInt() }
 
-            MaestroCommand(
-                TapOnPointCommand(
-                    x = points[0],
-                    y = points[1],
-                    retryIfNoChange = retryIfNoChange,
-                    waitUntilVisible = waitUntilVisible,
-                    longPress = longPress,
+                // TODO: handle values outside 0-100
```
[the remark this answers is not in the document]
dmitry-zaitsev: Yes, we can throw a `SyntaxError`. I will tweak it myself

maestro-orchestra/src/test/java/maestro/orchestra/MaestroCommandSerializationTest.kt
```diff
@@ -84,6 +84,42 @@ internal class MaestroCommandSerializationTest {
             .isEqualTo(command)
     }
 
+    @Test
+    fun `serialize TapOnPercentCommand`() {
```
dmitry-zaitsev, line 88: It would be great to also have an entry in `IntegrationTests`

dmitry-zaitsev:
@mslalith thanks a lot for the contribution!

There are some minor tweaks that need to be done (i.e. injecting JS or adding integration test), but I am willing to do those myself to move this along faster. Will link a follow up PR here

dmitry-zaitsev (approved):

dmitry-zaitsev:
Here is the PR: #475

PR #463, opened 2022-12-05
Scaffolding for Studio command

Leland-Takamine:
Uncomment the line in App.kt to enable `maestro studio`. For now, this just opens up a web app that hits a sample `/api/hierarchy` endpoint which returns the raw maestro hierarchy output.


maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
```diff
@@ -0,0 +1,49 @@
...
+        val (maestro, _) = MaestroFactory.createMaestro(parent?.host, parent?.port, parent?.deviceId)
+
+        val port = getFreePort()
+        MaestroStudio.start(port, maestro)
+
+        val studioUrl = "http://localhost:${port}"
+        val message = ("Maestro Studio".bold() + " is running at " + studioUrl.blue()).box()
+        println()
+        println(message)
+        if (Desktop.isDesktopSupported()) {
```
dmitry-zaitsev, line 38: And if not?

maestro-studio/server/src/main/java/maestro/studio/MaestroStudio.kt
```diff
@@ -0,0 +1,35 @@
+package maestro.studio
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import maestro.Maestro
+import org.http4k.core.Method.GET
```
dmitry-zaitsev, line 6: I suggest we use Ktor instead. It is more mature and team is more familiar with it

Leland-Takamine: updated

maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
Leland-Takamine: Updated the message to be clearer

PR #468, opened 2022-12-07
Studio

Leland-Takamine:
Implemented UI to select element bounds overlayed on a screenshot. This will be the primary way users will select an element to inspect. Search is the other discovery mechanism which will be implemented next.

https://user-images.githubusercontent.com/847683/206075162-a09d39ff-2562-4c91-9755-73eaa3e130ab.mov


maestro-studio/web/src/App.tsx
```diff
@@ -1,40 +1,24 @@
...
-  const [hierarchy, setHierarchy] = useState<Hierarchy>()
+  const [deviceScreen, setDeviceScreen] = useState<DeviceScreen>()
   useEffect(() => {
     (async () => {
-      const response = await fetch('/api/hierarchy')
-      const hierarchy: Hierarchy = await response.json()
-      setHierarchy(hierarchy)
+      const response = await fetch('/api/device-screen')
+      const hierarchy: DeviceScreen = await response.json()
+      setDeviceScreen(hierarchy)
```
axelniklasson, line 12: Super Nit as we're moving away from hierarchy to device screen terminology
```suggestion
      const deviceScreenData: DeviceScreen = await response.json()
      setDeviceScreen(deviceScreenData)
```

dmitry-zaitsev (approved):

maestro-studio/web/src/AnnotatedScreenshot.tsx
```diff
@@ -0,0 +1,135 @@
...
+      boundsY / deviceScreen.height,
+      boundsWidth / deviceScreen.width,
+      boundsHeight / deviceScreen.height,
+      mouseX / elementWidth,
+      mouseY / elementHeight,
+    )
+  })
+  if (hoveredList.length === 0) return null
+  return hoveredList.sort((a, b) => {
+    if (!a.bounds && !a.bounds) return 0
```
axelniklasson, line 76: Do we want to check the bounds of both `a` and `b` here?
```suggestion
    if (!a.bounds && !b.bounds) return 0
```

axelniklasson (approved):
This is amazing, great stuff 🎉

PR #470, opened 2022-12-07
[MOB 1756] Creates http server for XCUITest driver

amanjeetsingh150:
## Proposed Changes
Instead of getting view hierarchy from idb. We are now going to directly depend on XCUITest to obtain view hierarchy. This will remove problems such as:
1. iOS hierarchy not showing bottom tabs
2. iOS hierarchy not available for iOS 16
3. Supporting identifiers which will also work for Swift UI

Upcoming PRs:
- [x] Deducing appId through running foreground app
- [x] Integrate HTTP server on iOS driver to get view hierarchy in maestro

## Testing
Tested on the Wikipedia app which failed earlier if we try to click on the bottom navigation bars

https://user-images.githubusercontent.com/12881364/206286883-e9037fcd-5735-4976-8dbb-0bdea8b6139e.mp4


dmitry-zaitsev (approved):

maestro-driver-ios/maestro-driver-ios/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  AppDelegate.swift
+//  maestro-driver-ios
+//
+//  Created by Amanjeet Singh on 28/11/22.
```
artem888, line 5: this headers can be removed :)

artem888 (approved):

maestro-driver-ios/maestro-driver-ios/AppDelegate.swift
amanjeetsingh150: Done ✅

maestro-client/src/main/java/maestro/Filters.kt
```diff
@@ -209,4 +209,16 @@ object Filters {
         }
     }
 
+    fun deepestMatchingElement(filter: ElementFilter): ElementFilter {
```
Leland-Takamine, line 212: @amanjeetsingh150 @dmitry-zaitsev Won't this return duplicate nodes in the list since the initial list can contain both parent and child nodes?

Leland-Takamine: @amanjeetsingh150 Can you explain (1) what this filter is supposed to do and (2) why we need this?

amanjeetsingh150: There were instances, especially in iOS, where we found that the accessibility like text was same for the element and its children. Earlier we used to select the first match which resulted in clicking the wrong element where should have clicked the deepest child. That would actually result in the click.

PR #471, opened 2022-12-07
Support setting ansi output yes / no, default to isTTY(STDOUT)

berikv:
## Proposed Changes

Support no-color and no-ansi flags on the cli.
Default to no-color and no-ansi if STDOUT is not a TTY.
Adjust ProgressBar and ResultsView so the format makes sense without ansi characters.

Note that two commands will ignore the --no-ansi and --no-color parameters:

- `maestro test --continuous`
- `maestro record`

## Testing

Tested commands locally with and without ansi / colors.

## Issues Fixed

Output looking bad when running `maestro test` on CI or when piping to a file or an editor.

## Detail

Example output:
```
> ./maestro --no-color test ~/samples/ios-flow.yaml
Running on MobileDevTesterSim - iOS 15.5 - 61D1FB84-96D3-45FC-962F-30FCC6DFDD4F
Waiting for idb service to start..
Waiting for Simulator to boot..
Waiting for Accessibility info to become available..
Simulator ready
Running on MobileDevTesterSim - iOS 15.5 - 61D1FB84-96D3-45FC-962F-30FCC6DFDD4F
 > Flow
Launch app "org.wikimedia.wikipedia" with clear state... COMPLETED
Tap on "Next"... COMPLETED
Tap on "Next"... COMPLETED
Tap on "Next"... COMPLETED
Tap on "Get started"... COMPLETED
Tap on "Search Wikipedia"... COMPLETED
Input text Automated testing... COMPLETED
Tap on "Software testing.*"... COMPLETED
```

```
> ./maestro --no-color cloud ~/samples/Wikipedia.app  ~/samples/ios-flow.yaml

Uploading Flow(s)...

....................

Visit the web console for more details about the upload: https://console.mobile.dev/uploads/bd6d098e-46a9-4136-828c-a0d532368f9b?teamId=1c1291f2-578f-46bc-8ecc-f61477b1fefa&appId=f565f644-b693-4d45-a0aa-3efc0980ad5e

Waiting for analyses to complete...

[Passed] ios-flow

1/1 Flow Passed

==== View details in the console ====

https://console.mobile.dev/uploads/bd6d098e-46a9-4136-828c-a0d532368f9b?teamId=1c1291f2-578f-46bc-8ecc-f61477b1fefa&appId=f565f644-b693-4d45-a0aa-3efc0980ad5e

```


maestro-cli/src/main/java/maestro/cli/App.kt
```diff
@@ -72,52 +73,50 @@ class App {
...
-            1
+            println(props["version"])
         }
 
-    val exitCode = commandLine
-        .execute(*args)
-
-    DebugLogStore.finalizeRun()
+        @JvmStatic
+        fun main(args: Array<String>) {
```
Leland-Takamine, line 85: Is moving the main method into the App class necessary?

Leland-Takamine: Yeah let's keep as it was before so it's clear what the important changes are in this PR

berikv: It seems strange to have both an App class and a AppKt class. The AppKt class of course magically inserted by the compiler to capture the free methods.

berikv: Alright. I'll move it back

maestro-cli/src/main/java/maestro/cli/command/BugReportCommand.kt
```diff
@@ -14,6 +13,9 @@ import java.util.concurrent.Callable
 )
 class BugReportCommand : Callable<Int> {
 
+    @CommandLine.Mixin
+    var disableANSIMixin: DisableAnsiMixin? = null
```
Leland-Takamine, line 17: Does this need to go on every subcommand or can it just be at the app level?

berikv: This needs to be at every subcommand. Picocli does not support setting parameters on every subcommand without changing the subcommand definition. By using a mixin, the impact is minimal.

Leland-Takamine: I think what we want is the ability to do `maestro --no-ansi <anysubcommand>` as opposed to `maestro <anysubcommand> --no-ansi`. Do we still need to add to every subcommand in this case?

berikv: In that case no, but users will be confused that `maestro test --no-ansi` does not work

Leland-Takamine: I think this tradeoff is fine since we have this issue with other options as well. I agree we should consider fixing this, but i think the proper approach would be to add a subclass that all our commands inherit from so all our base option can benefit from this. For now I don't think it's worth the duplication of code to fix for only --no-ansi as this problem already exists for other options

berikv: I know I will be confused. I know because we have the same argument / command requirement with our internal cli tools.

Leland-Takamine: Yep agree we should address - but I think we should fix this holistically instead of just for a single option

Leland-Takamine: And I believe it makes sense to address that as a follow up

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/runner/PlainTextResultView.kt
```diff
@@ -0,0 +1,107 @@
+package maestro.cli.runner
+
+class PlainTextResultView {
```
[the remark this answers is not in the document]
berikv: I have updated the PR according to your suggestion. Could you re-review?

maestro-cli/src/main/java/maestro/cli/runner/resultview/AnsiResultView.kt
```diff
@@ -17,27 +17,30 @@
...
-) {
+): ResultView {
 
     private val startTimestamp = System.currentTimeMillis()
 
     private val frames = mutableListOf<Frame>()
 
     private var previousFrame: String? = null
 
+    private val plainTextResultView = PlainTextResultView()
```
dmitry-zaitsev, line 37: Is it used for anything here?

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/runner/resultview/AnsiResultView.kt
berikv: Good catch

PR #477, opened 2022-12-09
[MOB-1757] Integrate HTTP server on IOS driver works locally

amanjeetsingh150:
## Proposed Changes
1. Integrate HTTP server requests in `IOSDriver`
2. Refactor `Maestro.ios` to take `IOSDriver` as input
3. Ensure xcuitest artifacts install while setting up IOSDriver
4. Adds logging so that we can root cause issues
5. Search elements from node children depth first 

## Testing

https://user-images.githubusercontent.com/12881364/206658944-f0cc5c51-d626-40b5-a3b9-3fe209600ca9.mp4


maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -122,19 +115,7 @@ object DeviceService {
                 process
                     .waitFor(1000, TimeUnit.MILLISECONDS)
                 process.exitValue() == 0
-            } || error("Simulator failed to boot")
```
dmitry-zaitsev, line 125: This should stay, I believe

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -122,19 +115,7 @@ object DeviceService {
...
-
-            // Test if idb can get accessibility info elements with non-zero frame with
-            logger.warning("Waiting for Accessibility info to become available..")
-            MaestroTimer.retryUntilTrue(timeoutMs = 20000, delayMs = 100) {
```
dmitry-zaitsev, line 129: We still need to wait for IDB to boot, otherwise there is a risk that `tap` command will fail because IDB is not ready. We can't use `contentDescriptor` for that anymore, so we need some other way to deduce that

maestro-client/build.gradle
```diff
@@ -65,6 +65,13 @@ dependencies {
     api 'org.mozilla:rhino:1.7.14'
     api 'org.jsoup:jsoup:1.15.3'
     api 'com.squareup.okhttp3:okhttp:4.10.0'
+    api 'org.rauschig:jarchivelib:1.2.0'
+    api 'com.fasterxml.jackson.core:jackson-databind:2.13.4'
+    api 'com.fasterxml.jackson.module:jackson-module-kotlin:2.13.4'
+    api 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.4'
+    api 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.13.4'
+    api 'commons-io:commons-io:2.11.0'
```
dmitry-zaitsev, line 73: What do we need commons-io for? Kotlin typically has many of the common IO utils, so just double checking whether we really need it

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -437,16 +437,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
-            val channel = ManagedChannelBuilder.forAddress(host, port)
-                .usePlaintext()
-                .build()
-
-            return ios(channel)
-        }
-
-        fun ios(channel: ManagedChannel): Maestro {
-            val driver = IOSDriver(IdbIOSDevice(channel))
+        fun ios(driver: Driver): Maestro {
```
dmitry-zaitsev, line 440: Maestro already have a constructor that takes `Driver` as a parameter. This is not much different except that it calls `driver.open()`

Why not have `ios()` method actually create a driver internally?

maestro-client/src/main/java/maestro/MaestroTimer.kt
```diff
@@ -1,5 +1,7 @@
 package maestro
 
+import java.util.concurrent.TimeoutException
```
dmitry-zaitsev, line 3: Does not seem to be used?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -45,12 +51,39 @@ class IOSDriver(
...
+
+        logger.info("[Start] Ensuring ui test runner app is launched on ${iosDevice.deviceId}")
+        IOSUiTestRunner.ensureOpen()
+        logger.info("[Done] Ensuring ui test runner app is launched on ${iosDevice.deviceId}")
+
+        logger.info("[Start] Trying to view hierarchy for ${iosDevice.deviceId}")
+        val nodes = iosDevice
+            .contentDescriptor(appId = IOSUiTestRunner.UI_TEST_RUNNER_APP_BUNDLE_ID)
+            .get()
+        MaestroTimer.retryUntilTrue(5000) {
```
dmitry-zaitsev, line 80: You are retrying a static piece of code here, should you call `contentDescriptor` inside of the `retry` block?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -59,9 +92,12 @@ class IOSDriver(
 
     override fun close() {
         iosDevice.close()
+        iosDevice.uninstall(IOSUiTestRunner.UI_TEST_RUNNER_APP_BUNDLE_ID)
```
dmitry-zaitsev, line 95: I think we need to uninstall it before installing it again. There is no guarantee that `close()` will always be called

maestro-client/src/main/java/maestro/ios/IOSUiTestRunner.kt
```diff
@@ -0,0 +1,88 @@
...
+        val processOutput = ProcessBuilder(
+            "bash",
+            "-c",
+            "xcrun simctl spawn booted launchctl print system | grep $UI_TEST_RUNNER_APP_BUNDLE_ID | awk '/$UI_TEST_RUNNER_APP_BUNDLE_ID/ {print \$3}'"
+        ).start().inputStream.source().buffer().readUtf8().trim()
+
+        if (!processOutput.contains(UI_TEST_RUNNER_APP_BUNDLE_ID)) {
+            logger.info("Not able to find ui test runner app, going to install now")
+            val tempDir = System.getenv("TMPDIR")
+            val xctestRunFile = File("$tempDir/maestro-driver-ios-config.xctestrun")
```
dmitry-zaitsev, line 35: nit: you can do `File(tempDir, "maestro-driver-ios-config.xctestrun")` instead

maestro-ios/src/main/java/ios/hierarchy/Frame.kt
```diff
@@ -0,0 +1,10 @@
+package ios.hierarchy
+
+import com.google.gson.annotations.SerializedName
+
+data class Frame(
+    @SerializedName("Width") val Width: Float,
```
dmitry-zaitsev, line 6: Since you are already using `SerializedName`, let's keep actual field names canonical (camelCase)

maestro-test/src/test/resources/038_partial_id.yaml
```diff
@@ -4,5 +4,5 @@ appId: com.example.app
     id: "keyboard_area"
     retryTapIfNoChange: false
 - tapOn:
-    id: ".*keyboard_area"
+    id: ".*another_keyboard_area"
```
dmitry-zaitsev, line 7: Why was this changed?

maestro-client/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
artem888, line 1: long-term should those files be under the source control? for me they can be stored elsewhere, i.e. gcloud

maestro-driver-ios/build-maestro-runner.sh
```diff
@@ -0,0 +1,20 @@
+## Build the UI test
```
artem888, line 1: whey there are two files:
- maestro-driver-ios/build-maestro-runner.sh
- maestro-driver-ios/build-maestro-ios-runner.sh

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
amanjeetsingh150: True that I have used tap result to check if idb is ready. Dispatching tap event on 0,0 coordinates and see if it's Ok.

maestro-client/build.gradle
amanjeetsingh150: Removed this and used File.deleteRecursively

maestro-client/src/main/java/maestro/MaestroTimer.kt
amanjeetsingh150: Removed ✅

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Good catch have replaced with call to contentdescriptor

maestro-ios/src/main/java/ios/hierarchy/Frame.kt
amanjeetsingh150: Done, changed!

maestro-client/src/main/java/maestro/ios/Simctl.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import maestro.MaestroTimer
+import ios.commands.CommandLineUtils
+import okio.buffer
+import okio.source
+import java.util.concurrent.TimeUnit
+
+object Simctl {
```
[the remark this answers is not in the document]
amanjeetsingh150: Yup have changed path of the same file to be inside maestro-client

maestro-test/src/test/resources/038_partial_id.yaml
amanjeetsingh150: This was changed because we now match from depth first due to this change: https://github.com/mobile-dev-inc/maestro/pull/477/commits/bdad3e52090e95820adcfb4737331eb8b9552e87

maestro-driver-ios/build-maestro-runner.sh
amanjeetsingh150: Removed one

maestro-client/src/main/resources/maestro-driver-ios-config.xctestrun
amanjeetsingh150: I think its fine to keep it here since these are intentionally packaged in jar.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Done

maestro-client/src/main/java/maestro/Maestro.kt
dmitry-zaitsev: `Maestro` already has a constructor that takes `Driver` as a parameter, we can just use that constructor for Cloud

maestro-test/src/test/resources/038_partial_id.yaml
dmitry-zaitsev: This doesn't seem right in this test case:

```
val driver = driver {
            element {
                id = "com.google.android.inputmethod.latin:id/another_keyboard_area"
                bounds = Bounds(0, 0, 100, 100)
            }

            element {
                id = "com.google.android.inputmethod.latin:id/keyboard_area"
                bounds = Bounds(0, 100, 100, 200)
            }
}
```

Those views are siblings, so I would've expected `id: .*keyboard_area` to match the first view

amanjeetsingh150: FIxed ✅

PR #479, opened 2022-12-09
Implement endpoint to fetch bundle id of the app that is currently running (iOS)

artem888:
## Proposed Changes

This PR implements a new request for xctest maestro driver to fetch the app that is currently running on foreground. If no app is detected - default fallback is com.apple.springboard.

## Testing

Tested locally:

https://user-images.githubusercontent.com/961175/206758593-76102c67-1c60-4984-af51-6472ead3e5cc.mov


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -191,8 +191,10 @@ class IOSDriver(
     }
 
     override fun contentDescriptor(): TreeNode {
-        logger.info("Getting view hierarchy for $appId")
-        val result = iosDevice.contentDescriptor(appId)
+        val getRunningAppId = GetRunningAppIdResolver().getRunningAppId()
+        logger.info("Getting view hierarchy for $getRunningAppId")
+
+        val result = iosDevice.contentDescriptor(appId ?: getRunningAppId)
```
amanjeetsingh150, line 197: 1. Let's change the var name to `runningAppId`
2. Are you keeping the value returned from `getRunningAppId` nullable? If yes then let's first depend on app id from the API response and then `appId` here which is initialized by what app is launched. (see the on success in launchApp)

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+import okhttp3.RequestBody.Companion.toRequestBody
+import java.util.concurrent.TimeUnit
+
+class GetRunningAppIdResolver {
+    private val okHttpClient by lazy {
+        OkHttpClient.Builder()
+            .connectTimeout(10, TimeUnit.SECONDS)
+            .readTimeout(30, TimeUnit.SECONDS)
+            .build()
+    }
```
amanjeetsingh150, line 19: Can you unify okhttp client for the iOS module lets make it a singleton which can be used for this and view hierarchy calls both?

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    private val logger = DebugLogStore.loggerFor(GetRunningAppIdResolver::class.java)
+
+    fun getRunningAppId(): String {
+        val gson = Gson()
+        val appIds = GetRunningAppRequest(Simctl.listApps())
+
+        logger.info("installed apps: $appIds")
```
amanjeetsingh150, line 27: Why not print value of `Simctl.listApps()`

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+            .host("localhost")
+            .addPathSegment("runningApp")
+            .port(9080)
+            .build()
+        val request = Request.Builder()
+            .addHeader("contentType", "application/json")
+            .url(httpUrl)
+            .post(body)
+            .build()
+        val response = okHttpClient.newCall(request).execute()
```
amanjeetsingh150, line 43: Lets close the response by `use` to close the response body. See the view hierarchy call for implementation.

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+        val runningAppBundleId = if (response.isSuccessful) {
+            response.body?.let {
+                val type = object : TypeToken<Map<String, Any>>() {}.type
+                val responseBody: Map<String, Any> = gson.fromJson(String(it.bytes()), type)
+
+                responseBody["runningAppBundleId"] as? String
+            } ?: throw IllegalStateException("View Hierarchy not available, response body is null")
+        } else {
+            throw IllegalArgumentException("Maestro was not able to capture view hierarchy. Run maestro bugreport command and submit " +
```
amanjeetsingh150, line 52: Should we rather fallback on null and on null use the appId from inferencing what app is launched?

maestro-client/src/main/java/maestro/ios/GetRunningAppRequest.kt
```diff
@@ -0,0 +1,5 @@
+package maestro.ios
+
+import com.google.gson.annotations.SerializedName
+
+    data class GetRunningAppRequest(@SerializedName("appIds") val appIds: Set<String>)
```
amanjeetsingh150, line 5: Redundant intentation

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/RunningAppRouteHandler.swift
```diff
@@ -0,0 +1,32 @@
...
+import XCTest
+
+class RunningAppRouteHandler: RouteHandler {
+    private static let springboardBundleId = "com.apple.springboard"
+    
+    func handle(request: HTTPRequest) async throws -> HTTPResponse {
+        let decoder = JSONDecoder()
+        
+        guard let requestBody = try? decoder.decode(RunningAppRequest.self, from: request.body) else {
+            throw ServerError.RunningAppRequestSerializeFailure
```
amanjeetsingh150, line 11: Lets not throw errors and respond back with server codes 4xx in these cases. I think we should treat these as client errors

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/RunningAppRouteHandler.swift
```diff
@@ -0,0 +1,32 @@
...
+        }
+        
+        let response = ["runningAppBundleId": runningAppId ?? RunningAppRouteHandler.springboardBundleId]
+        
+        guard let responseData = try? JSONSerialization.data(
+            withJSONObject: response,
+            options: .prettyPrinted
+        ) else {
+            print("Serialization of runningAppBundleId failed")
+            throw ServerError.RunningAppResponseSerializeFailure
```
amanjeetsingh150, line 27: Same here about not throwing error

maestro-driver-ios/maestro-driver-iosUITests/Routes/ServerError.swift
```diff
@@ -0,0 +1,6 @@
+enum ServerError: Error {
```
amanjeetsingh150, line 1: Lets remove this completely I have also stopped throwing errors since all of them should be client errors. See the hierarchy handler for implementation

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.google.gson.reflect.TypeToken
+import maestro.debuglog.DebugLogStore
+import okhttp3.HttpUrl
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import java.util.concurrent.TimeUnit
+
+class GetRunningAppIdResolver {
```
dmitry-zaitsev, line 13: This can be an `object`

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+        val body = gson.toJson(appIds).toRequestBody(mediaType)
+
+        val httpUrl = HttpUrl.Builder()
+            .scheme("http")
+            .host("localhost")
+            .addPathSegment("runningApp")
+            .port(9080)
+            .build()
+        val request = Request.Builder()
+            .addHeader("contentType", "application/json")
```
dmitry-zaitsev, line 39: I believe the header name should be `Content-Type`

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+            .url(httpUrl)
+            .post(body)
+            .build()
+        val response = okHttpClient.newCall(request).execute()
+        val runningAppBundleId = if (response.isSuccessful) {
+            response.body?.let {
+                val type = object : TypeToken<Map<String, Any>>() {}.type
```
dmitry-zaitsev, line 46: Why not introduce a dedicated type for the response?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
artem888: @amanjeetsingh150 it is not nullable, because we use default value com.apple.springboard if nothing is returned

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+        OkHttpClient.Builder()
+            .connectTimeout(10, TimeUnit.SECONDS)
+            .readTimeout(30, TimeUnit.SECONDS)
+            .build()
+    }
+
+    private val logger = DebugLogStore.loggerFor(GetRunningAppIdResolver::class.java)
+
+    fun getRunningAppId(): String {
+        val gson = Gson()
```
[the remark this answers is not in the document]
artem888: is there any specific we prefer jackson over gson? the latter one looks more elegant and less verbose to me

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
```diff
@@ -0,0 +1,60 @@
...
+        val appIds = GetRunningAppRequest(Simctl.listApps())
+
+        logger.info("installed apps: $appIds")
+
+        val mediaType = "application/json; charset=utf-8".toMediaType()
+        val body = gson.toJson(appIds).toRequestBody(mediaType)
+
+        val httpUrl = HttpUrl.Builder()
+            .scheme("http")
+            .host("localhost")
```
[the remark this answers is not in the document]
artem888: right now xctest driver is local only, but in the second iteration we'll modify it

maestro-client/src/main/java/maestro/ios/GetRunningAppIdResolver.kt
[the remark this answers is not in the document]
dmitry-zaitsev: Jackson is generally more performant and also supports other formats like Yaml and Xml (we use both). We already started using one library, I would avoid introducing another one that serves essentially the same purpose.

amanjeetsingh150 (approved):

PR #481, opened 2022-12-11
Add update check

Leland-Takamine:


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -349,3 +381,32 @@ data class RenderState(
     val error: String?,
     val downloadUrl: String?,
 )
+
+data class CliVersion(
+    val major: Int,
+    val minor: Int,
+    val patch: Int,
```
dmitry-zaitsev, line 388: `patch` is not guaranteed to be an `Int`, i.e. this is possible: `16.0.1-RC`

axelniklasson (approved):

maestro-cli/src/main/java/maestro/cli/update/Updates.kt
```diff
@@ -0,0 +1,102 @@
...
+            false
+        } else {
+            uuidPath.parent.createDirectories()
+            uuidPath.writeText(UUID.randomUUID().toString())
+            true
+        }
+        DEVICE_UUID = uuidPath.readText()
+    }
+
+    fun fetchUpdatesAsync() {
```
dmitry-zaitsev, line 48: Logic seems correct but just to double check - none of those operations print an error if you are not connected to the Internet?

Leland-Takamine: Yep that's right

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
Leland-Takamine: As discussed offline - in this case we will not report any updates available

PR #499, opened 2022-12-13
iOS: Reset app permissions on reset app state

berikv:
## Proposed Changes

Call `xcrun simctl privacy <udid> reset all <appId>` on clearState.


dmitry-zaitsev:
Changes look good to me, let's just sort out how this will work for Cloud

PR #505, opened 2022-12-14
Feature: allow to install a specific Maestro version

dmitry-zaitsev:
## Proposed Changes

Once deployed to GCP, user would be able to run scripts like that to install a specific version:

```
export MAESTRO_VERSION=1.16.0
```

or one-liner

```
export MAESTRO_VERSION=1.16.0; curl -Ls "https://get.maestro.mobile.dev" | bash
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -88,7 +88,13 @@ esac
 echo "* Create distribution directories..."
 mkdir -p "$maestro_tmp_folder"
 
-download_url="https://github.com/mobile-dev-inc/maestro/releases/latest/download/maestro.zip"
+
+if [ -z "$MAESTRO_VERSION" ]; then
+    download_url="https://github.com/mobile-dev-inc/maestro/releases/latest/download/maestro.zip"
+else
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -88,7 +88,13 @@ esac
 echo "* Create distribution directories..."
 mkdir -p "$maestro_tmp_folder"
 
-download_url="https://github.com/mobile-dev-inc/maestro/releases/latest/download/maestro.zip"
+
+if [ -z "$MAESTRO_VERSION" ]; then
+    download_url="https://github.com/mobile-dev-inc/maestro/releases/latest/download/maestro.zip"
+else
+    download_url="https://github.com/mobile-dev-inc/maestro/releases/download/cli-$MAESTRO_VERSION/maestro.zip"
```
dmitry-zaitsev, line 95: Anticipating a question of why I don't just assign a default value to `MAESTRO_VERSION` - the URLs have swapped order depending on whether you are fetching a latest version or specific version:

```
.../latest/download/...
```

and

```
.../download/{version}/...
```

🤷

Leland-Takamine (approved):

PR #519, opened 2022-12-16
Tweak Swipe Command Start/End Points

jrexd:
## Proposed Changes
Tweaking the start and end points for the swipe command.

@Leland-Takamine @dmitry-zaitsev 

## Testing
Performed swipe command on both Android and iOS applications.

## Issues Fixed

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -272,14 +272,14 @@ class AndroidDriver(
         when (swipeDirection) {
             SwipeDirection.UP -> {
                 val startX = deviceInfo.widthGrid / 2
-                val startY = deviceInfo.heightGrid / 9
-                val endY = deviceInfo.heightGrid / 3
+                val startY = deviceInfo.heightGrid / 3
+                val endY = deviceInfo.heightGrid / 9
                 dadb.shell("input swipe $startX $startY $startX $endY $durationMs")
             }
             SwipeDirection.DOWN -> {
                 val startX = deviceInfo.widthGrid / 2
-                val startY = deviceInfo.heightGrid / 3
-                val endY = deviceInfo.heightGrid / 9
+                val startY = deviceInfo.heightGrid / 9
+                val endY = deviceInfo.heightGrid / 3
                 dadb.shell("input swipe $startX $startY $startX $endY $durationMs")
             }
             SwipeDirection.RIGHT -> {
@@ -289,7 +289,7 @@ class AndroidDriver(
                 dadb.shell("input swipe $startX $startY $endX $startY $durationMs")
             }
             SwipeDirection.LEFT -> {
-                val startX = deviceInfo.widthPixels / 2
+                val startX = deviceInfo.widthGrid / 2
                 val startY = deviceInfo.heightGrid / 2
                 val endX = 0
                 dadb.shell("input swipe $startX $startY $endX $startY $durationMs")
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -272,14 +272,14 @@ class AndroidDriver(
         when (swipeDirection) {
             SwipeDirection.UP -> {
                 val startX = deviceInfo.widthGrid / 2
-                val startY = deviceInfo.heightGrid / 1
-                val endY = deviceInfo.heightGrid / 2
+                val startY = deviceInfo.heightGrid / 9
+                val endY = deviceInfo.heightGrid / 3
                 dadb.shell("input swipe $startX $startY $startX $endY $durationMs")
             }
             SwipeDirection.DOWN -> {
                 val startX = deviceInfo.widthGrid / 2
-                val startY = deviceInfo.widthGrid / 2
-                val endY = deviceInfo.heightGrid / 8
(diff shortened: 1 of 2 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -272,14 +272,14 @@ class AndroidDriver(
         when (swipeDirection) {
             SwipeDirection.UP -> {
                 val startX = deviceInfo.widthPixels / 2
-                val startY = deviceInfo.heightPixels
+                val startY = deviceInfo.heightPixels / 8
```
Leland-Takamine, line 275: I think this is meant to be 7/8 instead of 1/8 starting point right?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -272,14 +272,14 @@ class AndroidDriver(
...
                 val endY = deviceInfo.heightPixels / 2
                 dadb.shell("input swipe $startX $startY $startX $endY $durationMs")
             }
             SwipeDirection.DOWN -> {
                 val startX = deviceInfo.widthPixels / 2
-                val startY = 0
-                val endY = deviceInfo.heightPixels / 2
+                val startY = deviceInfo.widthPixels / 2
```
dmitry-zaitsev, line 281: While you are on that change, could you change `widthPixels` and `heightPixels` to `widthGrid` and `heightGrid` (both from `deviceInfo` model). iOS actually expects us to send coordinates in points (which is what `grid` returns) whereas we are using pixels here

dmitry-zaitsev:
Thank you for contributing!

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -271,26 +271,26 @@ class AndroidDriver(
         val deviceInfo = deviceInfo()
         when (swipeDirection) {
             SwipeDirection.UP -> {
-                val startX = deviceInfo.widthPixels / 2
-                val startY = deviceInfo.heightPixels
-                val endY = deviceInfo.heightPixels / 2
+                val startX = deviceInfo.widthGrid / 2
+                val startY = deviceInfo.heightGrid / 1
```
Leland-Takamine, line 275: Hm, doesn't this mean it will swipe from the very bottom of the screen still? (`deviceInfo.heightGrid / 1 == deviceInfo.heightGrid `)

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -271,26 +271,26 @@ class AndroidDriver(
...
+                val endY = deviceInfo.heightGrid / 2
                 dadb.shell("input swipe $startX $startY $startX $endY $durationMs")
             }
             SwipeDirection.DOWN -> {
-                val startX = deviceInfo.widthPixels / 2
-                val startY = 0
-                val endY = deviceInfo.heightPixels / 2
+                val startX = deviceInfo.widthGrid / 2
+                val startY = deviceInfo.widthGrid / 2
```
Leland-Takamine, line 281: Shouldn't this be based on the `heightGrid`?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -251,20 +251,20 @@ class IOSDriver(
         when (swipeDirection) {
             SwipeDirection.UP -> {
                 iosDevice.scroll(
-                    xStart = width / 4,
-                    yStart = height,
-                    xEnd = width / 4,
-                    yEnd = height / 4,
```
Leland-Takamine, line 257: Looks like there are some inconsistencies with the Android implementation. Let's keep them consistent unless there are reasons to have different behaviors between platforms

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
jrexd: Yes 🤦 fixing.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -271,26 +271,26 @@ class AndroidDriver(
...
-                val startX = deviceInfo.widthPixels / 2
-                val startY = deviceInfo.heightPixels / 2
-                val endX = deviceInfo.widthPixels
+                val startX = deviceInfo.widthGrid / 2
+                val startY = deviceInfo.heightGrid / 2
+                val endX = deviceInfo.widthGrid
                 dadb.shell("input swipe $startX $startY $endX $startY $durationMs")
             }
             SwipeDirection.LEFT -> {
                 val startX = deviceInfo.widthPixels / 2
```
Leland-Takamine, line 292: Switch to grid here as well

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -251,20 +251,20 @@ class IOSDriver(
...
+                    xStart = width / 2,
```
Leland-Takamine, line 254: @dmitry-zaitsev I think we need to use the grid here also right?

dmitry-zaitsev: Correct

dmitry-zaitsev (approved):
I will merge this and tweak the comments in a separate PR.

Thanks for the contribution!

PR #524, opened 2022-12-20
Repl

Leland-Takamine:
Minimal repl functionality complete. This will be available in Maestro Studio at the `/interact` path for now - otherwise it is hidden as it's still a WIP.

https://user-images.githubusercontent.com/847683/208822722-24c3cf00-290d-4fa2-8659-d71a6b27990a.mp4


maestro-studio/web/package.json
```diff
@@ -70,9 +72,13 @@
     "@storybook/testing-library": "^0.0.13",
     "autoprefixer": "^10.4.13",
     "babel-plugin-named-exports-order": "^0.0.2",
+    "msw": "^0.49.2",
```
axelniklasson, line 75: Love msw!

maestro-studio/web/src/ReplView.tsx
```diff
@@ -0,0 +1,352 @@
...
+          className="w-5 h-5 text-slate-900 hover:text-slate-600 active:text-slate-400"
+          onClick={onClick}
+        >
+          <rect width="20" height="20" transform="translate(0 0.5)" fill="currentColor"/>
+          <path d="M7 10.75L9.6 13.35L13.5 7.5" stroke="white" strokeLinecap="round" strokeLinejoin="round"/>
+        </svg>
+      )
+    } else {
+      return (
+        <svg
```
axelniklasson, line 91: There are a few icons in Studio now. Can we break them all out into a single component that renders the right icon based on a prop like we do internally? i.e. `<Icon id="<id>" />`

maestro-studio/web/src/api.ts
```diff
@@ -0,0 +1,79 @@
+import { DeviceScreen, Repl, ReplCommand } from './models';
+import useSWR, { mutate } from 'swr';
+
+export type ReplResponse = {
+  repl?: Repl | undefined
```
axelniklasson, line 5: Nit

```suggestion
  repl?: Repl
```

maestro-studio/web/src/api.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+  if (!response.ok) {
+    const body = await response.text()
+    throw new Error(`Request failed ${response.status}: ${body}`)
+  }
+  return await response.json()
+}
+
+const useRepl = (): ReplResponse => {
+  const {data: repl, error} = useSWR<Repl>('/api/repl/watch')
```
axelniklasson, line 30: I think we need to pass `?currentVersion=<replVersion>` here for the backend not to throw

maestro-studio/web/src/api.ts
```diff
@@ -0,0 +1,79 @@
...
+    },
+    reorderCommands: async (ids: string[]) => {
+      makeRequest('POST', '/api/repl/command/reorder', { ids })
+      mutate<Repl>('/api/repl/watch', (repl) => {
+        if (!repl) return undefined
+        const newCommands: ReplCommand[] = []
+        ids.forEach(id => {
+          const command = repl.commands.find(c => c.id === id)
+          command && newCommands.push(command)
+        })
```
axelniklasson, line 57:
```suggestion
       const idsAsSet = new Set(ids)
       repl.commands.filter(c => !!c && idsAsSet.has(c.id)).forEach(c => newCommands.push(c))
```

maestro-studio/server/src/main/java/maestro/studio/ReplService.kt
```diff
@@ -0,0 +1,215 @@
...
+        val response = jacksonObjectMapper().writeValueAsString(repl)
+        respondText(response)
+    }
+
+    private fun createEntries(yaml: String): List<ReplEntry> {
+        val newEntries = try {
+            readNodes(yaml).map { node ->
+                val yamlCommand = YamlCommandReader.MAPPER.convertValue(node, YamlFluentCommand::class.java)
+                val yamlString = YamlCommandReader.MAPPER.writeValueAsString(listOf(node))
+                val commands = yamlCommand.toCommands(Paths.get(""), "com.example.app")
```
axelniklasson, line 166: Should `com.example.app` be here?

axelniklasson (approved):

maestro-studio/web/src/api.ts
Leland-Takamine: This is just the key actually. It's not actually hitting the endpoint under the hood since there is not fetcher supplied here. I'm essentially just using swr as a cache for long polling results

maestro-studio/web/src/ReplView.tsx
Leland-Takamine: These icons are more single-use and are scoped to the file that uses them. I don't know if the proposed change would be an improvement tbh

maestro-studio/web/package.json
Leland-Takamine: 🙌

maestro-studio/server/src/main/java/maestro/studio/ReplService.kt
Leland-Takamine: Let me update to an empty string

maestro-studio/web/src/ReplView.tsx
Leland-Takamine: Let's revisit if we find ourselves wanting to reuse any of these icons

PR #525, opened 2022-12-21
Make logger generic for `XcUITestDriver`

amanjeetsingh150:
- Renames IOSUiTestRunner to `XCUITestDriver` as an interface
- Moves away from using `DebugLogStore` directly in driver. Makes logger interface instead


maestro-client/src/main/java/maestro/debuglog/IOSDriverLogger.kt
```diff
@@ -0,0 +1,13 @@
+package maestro.debuglog
+
+import ios.logger.Logger
+import maestro.drivers.IOSDriver
+
+class IOSDriverLogger : Logger {
+
+    private val logger by lazy { DebugLogStore.loggerFor(IOSDriver::class.java) }
```
axelniklasson, line 8: Should this be `IOSDriverLogger`?
```suggestion
    private val logger by lazy { DebugLogStore.loggerFor(IOSDriverLogger::class.java) }
```

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -70,18 +73,13 @@ class IOSDriver(
 
     private fun ensureXCUITestChannel() {
         logger.info("[Start] Uninstalling xctest ui runner app on ${iosDevice.deviceId}")
-        IOSUiTestRunner.uninstall()
+        xcUiTestDriver.uninstall()
```
axelniklasson, line 76: Do we need any logs in this method like we had before?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -212,7 +209,7 @@ class IOSDriver(
 
         val contentDescriptorResult = iosDevice.contentDescriptor(
             resolvedAppId ?: throw IllegalStateException("Failed to get view hierarchy, app id was not resolvedGetRunningAppRequest.kt")
-        )
+        ) { errorMessage -> logger.info(errorMessage) }
```
axelniklasson, line 212: Should we instead add an `error` function and call `logger.error` here?

axelniklasson (approved):

maestro-client/src/main/java/maestro/debuglog/IOSDriverLogger.kt
amanjeetsingh150: Nice catch! Done ✅

amanjeetsingh150: Nice catch! Done

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: The logs are still there they are just shifted to the `XcUiTestDriver` class

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Unfortunately, there is no logger error for the `DebugLogStore` class. I'll start a thread on this and let's solve that separately.

PR #528, opened 2022-12-22
Enable device interactions in Maestro Device interactive mode

Leland-Takamine:
Adding device interaction to interact page in Maestro Studio


maestro-studio/web/src/InteractPage.tsx
```diff
@@ -0,0 +1,43 @@
...
+          const deviceScreen = await API.getDeviceScreen(true)
+          setDeviceScreen(deviceScreen)
+        } catch (e) {
+          console.error(e)
+          await wait(1000)
+        }
+      }
+    })()
+    return () => { running = false }
+  }, [setDeviceScreen])
```
axelniklasson, line 24: Should this be an interval instead of a blocking while loop?

maestro-studio/web/src/InteractPage.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  if (!deviceScreen) return null
```
axelniklasson, line 26: Do we want a fallback message here when there is no device screen available?

maestro-studio/web/src/api.ts
```diff
@@ -32,8 +32,12 @@ const useRepl = (): ReplResponse => {
...
 
 export const API = {
-  getDeviceScreen: async (): Promise<DeviceScreen> => {
-    return makeRequest('GET', '/api/device-screen')
+  useDeviceScreen: (config?: SWRConfiguration<DeviceScreen>): SWRResponse<DeviceScreen> => {
+    return useSWR('/api/device-screen?deletePrevious=true', (url) => makeRequest('GET', url), config)
+  },
+  getDeviceScreen: async (deletePrevious?: boolean): Promise<DeviceScreen> => {
+    const query = deletePrevious ? '?deletePrevious=true' : ''
+    return makeRequest('GET', `/api/device-screen${query}`)
```
axelniklasson, line 40: Curious - any reason for doing data fetching through SWR only for `useDeviceScreen` and not for `getDeviceScreen`? Could we update the names of these methods to make it more clear how they are different?

axelniklasson (approved):
🙌

maestro-studio/web/src/InteractPage.tsx
Leland-Takamine: This isn't blocking actually, it using a async wait function that uses timeout under the hood. This is essentially what third party interval hooks use (I don't know of a built-in hook that does this)

maestro-studio/web/src/InteractPage.tsx
Leland-Takamine: Yeah error / loading states, etc are still wip

maestro-studio/web/src/api.ts
Leland-Takamine: I think `use*` and `get*` are the conventions for a hook vs async method right? That's essentially the difference here

axelniklasson: Yep, [hooks always starts with `use`](https://reactjs.org/docs/hooks-rules.html) but I guess the API is just a little bit confusing (to me) as to why both functions there. Is the `getDeviceScreen` there just so we can make the request outside of a hook/component where the `useDeviceScreen` hook can't be called?

maestro-studio/web/src/InteractPage.tsx
axelniklasson: Ah sorry I meant within the async anonymous function that essentially handles the ticking of the interval. It just seems a bit odd to me to not use `setInterval` and then `clearInterval` upon unmounting and instead handle this manually but not a big deal so I'm fine to leave as is 👍

maestro-studio/web/src/api.ts
Leland-Takamine: > Is the getDeviceScreen there just so we can make the request outside of a hook/component where the useDeviceScreen hook can't be called?

Yes that's exactly right. Let me know if there's a better way to name this?

maestro-studio/web/src/InteractPage.tsx
Leland-Takamine: Ah ok - I thought about setInterval, but the difference is that setInterval runs a function on a set time interval whereas this logic refreshes the device screen sequentially as frequently as possible. If used setInterval, (1) we wouldn't be refreshing as quickly as possible and/or (2) we'd encounter race conditions with simultaneous async http requests.

PR #532, opened 2022-12-23
Mob 1807 clicking system dialogs

amanjeetsingh150:
## Proposed Changes

1. Include springboard hierarchy in the XCUITest driver
2. System dialogs are part of the springboard and not the app itself. Not only system dialogs but things like wifi icon, cellular data
3. This will help to click the dialogs through maestro flows

## Testing

After including springboard

Screenshot 2022-12-23 at 5 49 00 PM

Before: 

Screenshot 2022-12-23 at 5 56 50 PM

## Issues Fixed


maestro-client/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -69,7 +69,7 @@
 			<key>DYLD_FRAMEWORK_PATH</key>
 			<string>__TESTROOT__/Debug-iphonesimulator:__TESTROOT__/Debug-iphonesimulator/PackageFrameworks:__PLATFORMS__/iPhoneSimulator.platform/Developer/Library/Frameworks</string>
 			<key>DYLD_INSERT_LIBRARIES</key>
-			<string>__PLATFORMS__/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libMainThreadChecker.dylib</string>
+			<string>/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 15.5.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libMainThreadChecker.dylib</string>
```
artem888, line 72: interesting why this was changed 🤔 the previous version looks more generic

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
```diff
@@ -14,15 +14,25 @@ class SubTreeRouteHandler: RouteHandler {
...
+                let springboardApplication = XCUIApplication(bundleIdentifier: springboardBundleId)
+                
+                logger.info("[Start] Now trying hierarchy for: \(appId)")
+                var viewHierarchyDictionary = try xcuiApplication.snapshot().dictionaryRepresentation
+                logger.info("[Done] Now trying hierarchy for: \(appId)")
+                logger.info("[Start] Now trying hierarchy for: \(springboardBundleId)")
+                let springboardHierarchyDictionary = try springboardApplication.snapshot().dictionaryRepresentation
+                logger.info("[Done] Now trying hierarchy for: \(springboardBundleId)")
+                
+                let children = viewHierarchyDictionary[XCUIElement.AttributeName(rawValue: "children")] as! Array<[XCUIElement.AttributeName: Any]>
```
artem888, line 30: is force type cast fine here, what if one of them returns an empty response of smth ?

amanjeetsingh150: Can you suggest how can we handle that here?

maestro-client/src/main/resources/maestro-driver-ios-config.xctestrun
amanjeetsingh150: Not sure as well I checked with one of our client flows locally the artifacts work

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
artem888: make both optional, soft-fail if one of them is missing

maestro-client/src/main/resources/maestro-driver-ios-config.xctestrun
artem888: would it still work if you revert this change?

amanjeetsingh150: I don't think so because this path was from the machine we have in script and the current one is from my local simulator. I would say lets fix this by having artifacts from staging machine

artem888 (approved):

PR #545, opened 2022-12-28
Add capacity of filtering flows based on the --include-tags and --exclude-tags for maestro cloud command

felipevolpone:
This PR adds the `--include-tags` and `--exclude-tags` parameters for the `cloud` command [1]. With these options, the users are able to filter the Flows containing (or not) those tags.

Usage example:
The `--include-tags` should look for all flows that contain the provided tag and it doesn't matter if those Flows also contain other tags. On the other hand, the `--exclude-tags` would remove from the list of Flows to run any Flow that contains the provided tags.

Let's say a user has two flows:

```
# flowA.yaml
appId: com.example.app
tags: dev,pull-request
```

```
# flowB.yaml
appId: com.example.app
tags: dev
```

In the scenario above:
- If they use `--include-tags=dev` we would run flowA and flowB. 
- If they use `--exclude-tags=pull-request` we would run only flowB.
- If they use `--exclude-tags=dev` we wouldn't run any Flow.
- If they use `--include-tags=dev --exclude-tags=pull-request` we would run only flowB

[1] I didn't add to the `upload` command too to push the users to not use it so we can deprecate it soon.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -41,6 +41,8 @@ class CloudInteractor(
         env: Map<String, String> = emptyMap(),
         androidApiLevel: Int? = null,
         failOnCancellation: Boolean = false,
+        includeTags: List<String> = emptyList(),
+        excludeTags: List<String> = emptyList(),
     ): Int {
         if (!flowFile.exists()) throw CliError("File does not exist: ${flowFile.absolutePath}")
         if (mapping?.exists() == false) throw CliError("File does not exist: ${mapping.absolutePath}")
@@ -53,7 +55,7 @@ class CloudInteractor(
 
         TemporaryDirectory.use { tmpDir ->
             val workspaceZip = tmpDir.resolve("workspace.zip")
-            WorkspaceUtils.createWorkspaceZip(flowFile.toPath().absolute(), workspaceZip)
+            WorkspaceUtils.createWorkspaceZip(flowFile.toPath().absolute(), workspaceZip, includeTags, excludeTags)
             println()
             val progressBar = ProgressBar(20)
 
--- a/maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
@@ -79,6 +79,20 @@ class CloudCommand : Callable<Int> {
     @Option(order = 10, names = ["--android-api-level"], description = ["Android API level to run your flow against"])
     private var androidApiLevel: Int? = null
 
+    @Option(
+        order = 11,
+        names = ["--include-tags"],
+        description = ["List of tags that will remove the Flows that does not have the provided tags"]
+    )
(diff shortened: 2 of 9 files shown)


Leland-Takamine:
Let's rename these options to `include-tags` and `exclude-tags`. Also let's implement this for `maestro test` as well to maintain parity between test and cloud

felipevolpone:
Parameter renamed. I won't implement the `test` command for now. Once it's done in maestro cloud, I'll get back and add the support to the local env/`test` command.

Leland-Takamine:
Sounds good 👍

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -41,13 +41,13 @@ class CloudInteractor(
...
         failOnCancellation: Boolean = false,
-        withTags: List<String> = emptyList(),
-        withoutTags: List<String>,
+        includeTags: List<String> = emptyList(),
+        excludeTags: List<String>,
     ): Int {
         if (!flowFile.exists()) throw CliError("File does not exist: ${flowFile.absolutePath}")
         if (mapping?.exists() == false) throw CliError("File does not exist: ${mapping.absolutePath}")
 
-        if (withTags.isNotEmpty() && withoutTags.isNotEmpty()) throw CliError("You can't use --withTags and --withoutTags together")
```
Leland-Takamine: Any particular reason for this? Semantically, I believe the convention would be to include all the `includeTags` then exclude the `excludeTags`

Leland-Takamine (approved):

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
felipevolpone: Let's clarify what should be the behavior without combining the parameters.

The `--include-tags` should look for all flows that contain the provided tag and it doesn't matter if those Flows also contain other tags. On the other hand, the `--exclude-tags` would remove from the list of Flows to run any Flow that contains the provided tags.

Let's say a user has two flows:

```
# flowA.yaml
appId: com.example.app
tags: dev,pull-request
```

```
# flowB.yaml
appId: com.example.app
tags: dev
```

In the scenario above:
- If they use `--include-tags=dev` we would run flowA and flowB. 
- If they use `--exclude-tags=pull-request` we would run only flowB.
- If they use `--exclude-tags=dev` we wouldn't run any Flow.

Is that right?

Leland-Takamine: Yeah that sounds right to me. And if we want to support both at the same time:

* `--include-tags=dev --exclude-tags=pull-request` we would run only flowB

felipevolpone: Sounds good 👍

felipevolpone:
@Leland-Takamine I changed the PR's approach, could you please review it again? Now, the Flow filtering happens all in the client and the feature is done from the CLI point of view.

Leland-Takamine (approved):
Lgtm!

PR #555, opened 2022-12-30
Take screenshot from xcrun commands instead of idb

amanjeetsingh150:
## Proposed Changes
Move away from taking screenshots from IDB and do it via xcrun commands

## Testing
- `takeScreenshot` command in flow

## Issues Fixed
Failed to decode image errors from idb


maestro-xcuitest-driver/src/main/kotlin/util/Simctl.kt
```diff
@@ -36,4 +36,8 @@ object Simctl {
             .start()
             .waitFor()
     }
+
+    fun screenshot(path: String) {
+        CommandLineUtils.runCommand("xcrun simctl io booted screenshot $path", waitForCompletion = true)
```
berikv, line 41: Please do split arguments on command execution. eg: `run("xcrun", "simctl", "io", "booted", "screenshot", path)`
String interpolation is a security risk

amanjeetsingh150: The abstraction `CommandLineUtils ` already does that implicitly

berikv: It should not do that, because string interpolation is a security risk

amanjeetsingh150: Can you elaborate where should this exactly happen and what security risk you are highlighting here?

berikv: Like so: `CommandLineUtils.runCommand("xcrun", "simctl", "io", "booted", "screenshot", path)`
The issue is mostly when using `run("sh foo bar $baz")`. If baz is user provided, it could be set to "; rm -rf /".
Not sure if there is an exploit in this line, I'll have to think about that, but why not keep it simple and avoid string interpolation in the first place?

berikv: Think of it like SQL, You wouldn't do `query.run("SELECT count(*) FROM users WHERE name = $name and password = $password")`

amanjeetsingh150: Couldn't that happen:

```
CommandLineUtils.runCommand("xcrun", "simctl", "io", "booted", "screenshot", path)
```
In this case as well 🤔

berikv: Nope, if you'd call `ProcessBuilder("sh", "foo", "bar", "; rm -rf /")`, the foo process would receive "; rm -rf /" as its third argument in argv. Which is different from running `ProcessBuilder("sh foo bar; rm -rf /")` which would spawn a shell that executes both `foo` and `rm`.

berikv: > Nope, if you'd call `ProcessBuilder("sh", "foo", "bar", "; rm -rf /")`, the foo process would receive "; rm -rf /" as its third argument in argv. Which is different from running `ProcessBuilder("sh foo bar; rm -rf /")` which would spawn a shell that executes both `foo` and `rm`.

Actually this is not entirely correct. I will make a better example that actually shows the problem.

amanjeetsingh150: I don't understand if that is even passed to CommanLineUtils it's already splitting into different arguments for ProcessBuilder. What difference it's going to make 😅 ? Maybe I don't understand the problem you are explaining

berikv: One difference is that spaces in the $path will cause $path to be split over several arguments. This may break with paths that contain spaces.

amanjeetsingh150: I can remove the white space in the path if any.

berikv: The point is to avoid string interpolation. Then the user input is not split and passed as a complete argument to the xcrun process. This will add support for paths with whitespaces and it will avoid security traps due to string interpolation. Double win. The only downside is that you'd have to split all arguments into strings in the call-site.

berikv (approved):

PR #556, opened 2022-12-30
Deprecate string splitting runCommand

berikv:
## Proposed Changes

Deprecate the runCommand() that slits up the single string argument. Refer to the runCommand() accepts a list of arguments.

### DIFF
--- a/maestro-xcuitest-driver/src/main/kotlin/util/CommandLineUtils.kt
+++ b/maestro-xcuitest-driver/src/main/kotlin/util/CommandLineUtils.kt
@@ -15,6 +15,7 @@ object CommandLineUtils {
         ) "NUL" else "/dev/null"
     )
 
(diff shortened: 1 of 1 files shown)


maestro-xcuitest-driver/src/main/kotlin/util/CommandLineUtils.kt
```diff
@@ -15,6 +15,7 @@ object CommandLineUtils {
         ) "NUL" else "/dev/null"
     )
 
+    @Deprecated("use runCommand(listOf(...)) instead")
```
amanjeetsingh150, line 18: Lets maybe add why it's deprecated. A small description, if not here then on internal slack

amanjeetsingh150 (approved):

dmitry-zaitsev (approved):

PR #575, opened 2023-01-06
Tweak: allow multiple parallel Maestro sessions

dmitry-zaitsev:
## Proposed Changes

Keep track of running Maestro sessions to decide whether a new Maestro driver should be started or an existing one should be reused.


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -0,0 +1,265 @@
...
+                if (!SessionStore.hasActiveSessions(sessionId, selectedDevice.platform)) {
+                    session.close()
+                }
+            }
+        })
+
+        return try {
+            block(session)
+        } finally {
+            heartbeatFuture.cancel(true)
```
berikv, line 80: Why not call `SessionStore.delete(sessionId, selectedDevice.platform)` here instead of secretly pruning sessions on a missing heartbeat?

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -0,0 +1,265 @@
...
+import maestro.cli.device.Platform
+import maestro.drivers.IOSDriver
+import java.util.UUID
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import kotlin.concurrent.thread
+
+object MaestroSessionManager {
+    private const val defaultHost = "localhost"
+    private const val idbPort = 10882
```
berikv, line 38: I believe this won't work for multiple iOS sessions yet because there needs to be multiple IDB instances with different port numbers available.

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
dmitry-zaitsev: I actually just removed this `finally` section and moved everything into a shutdown hook. There is no strong guarantee that this block will be called if thread is interrupted.

dmitry-zaitsev: furthermore, pruning is needed in cases where even shutdown-hook have failed. We can't count on program to always execute this code successfully, so prunning is a last resort

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
dmitry-zaitsev: The idea is that it should work for a single iOS device, in that case we can reuse the same idb instance

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
berikv: I get it, there is a separate thread that will make sure sessions are pruned -even if there is some unknown error in the application-

PR #578, opened 2023-01-09
Sort files by name before running

xo0ps:
## Proposed Changes

This PR is to sort files passed to `maestro test` by name before running.
I found it needed to make sure our flows run sequentially as expected.
## Testing

Before:
Screenshot 2023-01-09 at 16 54 52

After:
Screenshot 2023-01-09 at 16 55 25


dmitry-zaitsev:
By design, we highly discourage making tests dependent on each other as it will create scalability problems. I.e. eventually, some users would have too many tests to run sequentially and will need to parallelise them. Relying on this sorting behaviour will bite them back.

For that reason, unfortunately I have to decline this change, but thank you for taking your time to open a PR

PR #584, opened 2023-01-10
Feat: add error report mechanism

igorsmotto:
## Proposed Changes
- Add an error report mechanism: Every time an exception is detected by `picocli` it will send the exception details + the command executed to `api.mobile.dev`. 
    - This mechanism tries to not collect any PII information.
    - `-e/--env` can have PII information, so this PR also hash the values with SHA-256.
    - Also, there is a timeout of 1 second to complete the request to make sure that the user experience is not affected.
- Add `SystemInformationInterceptor` to `ApiClient` to simplify adding system information for each request.

### Testing
Tested manually:
- Modified an existing maestro command to always throw a CLIError
- Verified if the request went through

### Further possible improvements
- Send a _category_ of error. This could be made easier if there are custom exceptions for each _category_.
- Prompt the user to send all logs (saved by `DebugLogStore`) and flow file/workspace.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/util/ErrorReporter.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/ErrorReporter.kt
@@ -1,6 +1,7 @@
 package maestro.cli.util
 
 import maestro.cli.api.ApiClient
+import maestro.cli.update.Updates.BASE_API_URL
 import picocli.CommandLine
 import java.security.MessageDigest
 import java.util.concurrent.Executors
@@ -9,7 +10,6 @@ import kotlin.Exception
 
 object ErrorReporter {
 
-    private const val BASE_API_URL = "https://api.mobile.dev"
     private val executor = Executors.newCachedThreadPool {
         Executors.defaultThreadFactory().newThread(it).apply { isDaemon = true }
     }
@@ -23,12 +23,14 @@ object ErrorReporter {
             } else arg
         }
 
-        executor.submit {
+        val task = executor.submit {
             ApiClient(BASE_API_URL).sendErrorReport(
                 exception,
                 scrubbedArgs.joinToString(" ")
             )
-        }.get(1, TimeUnit.SECONDS)
+        }
+
+        runCatching { task.get(1, TimeUnit.SECONDS) }
     }
 
     private fun hashString(input: String): String {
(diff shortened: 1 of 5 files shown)


maestro-cli/src/main/java/maestro/cli/util/ErrorReporter.kt
```diff
@@ -0,0 +1,43 @@
...
+import maestro.cli.api.ApiClient
+import picocli.CommandLine
+import java.security.MessageDigest
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import kotlin.Exception
+
+object ErrorReporter {
+
+    private const val BASE_API_URL = "https://api.mobile.dev"
```
axelniklasson, line 12: Nit: we're defining `BASE_API_URL` here and also in `Updates.kt`. Shall we define it once somewhere and use it everywhere instead?

maestro-cli/src/main/java/maestro/cli/util/ErrorReporter.kt
```diff
@@ -0,0 +1,43 @@
...
+                key + "=" + hashString(value)
+            } else arg
+        }
+
+        executor.submit {
+            ApiClient(BASE_API_URL).sendErrorReport(
+                exception,
+                scrubbedArgs.joinToString(" ")
+            )
+        }.get(1, TimeUnit.SECONDS)
```
axelniklasson, line 31: What happens if this request fails? Will there be an error printed to users or will it be swallowed and hidden from them?

axelniklasson (approved):

maestro-cli/src/main/java/maestro/cli/util/ErrorReporter.kt
igorsmotto: I believe all of this reporting stuff should be invisible to the user, as this is not the primary concern of the user when running maestro. Fix here: https://github.com/mobile-dev-inc/maestro/pull/584/commits/979370c1cd94d25720f5fc88ee2b4b11db1f18d9

axelniklasson: Excellent, I completely agree 🙌

axelniklasson (approved):

PR #589, opened 2023-01-11
Fix openLink command to support extras

xo0ps:
## Proposed Changes

This PR fixes an issue with opening links that contain extras (example://deeplink?a=100&b=200)
Solution originated from [here](https://stackoverflow.com/questions/28802115/i-am-trying-to-test-android-deep-link-urls-through-adb-to-launch-my-app#comment118439409_58742435)

## Testing

Before:
Screenshot 2023-01-11 at 12 21 13 (1)

After:
Screenshot 2023-01-11 at 12 20 31 (1)


dmitry-zaitsev (approved):
Good catch 👍

PR #592, opened 2023-01-11
Reset permissions on clear app state

berikv:
## Proposed Changes

Reset permissions when launchApp is called with `clearState`.
This is similar to how Android works.

### DIFF
--- a/maestro-ios/src/main/java/ios/xcrun/Simctl.kt
+++ b/maestro-ios/src/main/java/ios/xcrun/Simctl.kt
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import maestro.utils.MaestroTimer
 import util.CommandLineUtils
+import util.CommandLineUtils.runCommand
 import java.io.File
 
 object Simctl {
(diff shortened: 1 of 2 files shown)


igorsmotto (approved):

amanjeetsingh150 (approved):

maestro-ios/src/main/java/ios/xcrun/Simctl.kt
```diff
@@ -106,4 +107,63 @@ object Simctl {
...
+                    Permission.REVOKE -> "NO"
+                    Permission.RESET -> "unset"
+                }
+            }
+            .map { (key, value) -> "$key=$value" }
+            .joinToString { "," }
+
+        runCommand(
+            listOf(
+                "applesimutils",
```
dmitry-zaitsev, line 150: Can we somehow auto-install `applesimutils` if they are not installed automatically?

dmitry-zaitsev (changes requested):
Let's avoid merging this until we address the `applesimutils` dependency concerns

maestro-ios/src/main/java/ios/xcrun/Simctl.kt
berikv: Only if the user already has homebrew installed. I think it wouldn't be acceptable to install homebrew (which takes >5 minutes) without checking in with the user.

berikv: One trick would be to package the binary with maestro, or download it directly from homebrew kegs.

Leland-Takamine: Is applesimutils installed by default with Xcode?

dmitry-zaitsev: No, it is a 3rd party tool

Leland-Takamine: Hm yeah this poses a pretty serious usability issue. What does applesimutils do under the hood can we mimic what it's doing manually?

dmitry-zaitsev: I think the logic in applesimutils is rather straightforward and could be copied over (i.e. reading from SQLite DB), though it will take a bit of time to do

Leland-Takamine: Sounds like the right approach to me

berikv:
Closed: will rewrite applesimutils in kotlin

berikv:
There is no rewrite of applesimutils taking place in the coming weeks. Users need this functionality. I propose to introduce the applesimutils dependency.

PR #604, opened 2023-01-13
Fix Maestro Studio Indexes

Leland-Takamine:
Update Maestro Studio logic to ensure that filtering logic is consistent with Orchestra. This inconsistency was causing incorrect indexes to be suggested in Maestro Studio.

### DIFF
--- a/maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
+++ b/maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
@@ -9,13 +9,16 @@ import io.ktor.server.http.content.staticRootFolder
 import io.ktor.server.response.respondText
 import io.ktor.server.routing.Routing
 import io.ktor.server.routing.get
+import maestro.ElementFilter
 import maestro.Filters
+import maestro.Filters.asFilter
 import maestro.Maestro
 import maestro.TreeNode
+import maestro.orchestra.Orchestra
 import java.io.File
 import java.nio.file.Path
+import java.util.IdentityHashMap
 import java.util.UUID
-import java.util.function.BiFunction
 import java.util.regex.Pattern
 import kotlin.io.path.Path
 import kotlin.io.path.createDirectories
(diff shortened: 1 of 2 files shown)


maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
```diff
@@ -78,20 +81,23 @@ object DeviceScreenService {
...
             .sortedWith(Filters.INDEX_COMPARATOR)
 
-        fun getIndex(element: TreeNode, attribute: String): Int? {
-            val value = element.attribute(attribute) ?: return null
-            val matchingElements = elements.filter { value == it.attributes[attribute] }
+        fun getIndex(filter: ElementFilter, element: TreeNode): Int? {
+            val identityHashMap = IdentityHashMap<TreeNode, Unit>()
+            val matchingElements = Filters.deepestMatchingElement(filter)(elements).filter {
+                // There are duplicate elements for some reason (likely due to unintended behavior in Filter.deepestMatchingElement) - filter them out
+                identityHashMap.put(it, Unit) == null
```
Leland-Takamine, line 88: We need to use object-equality here since TreeNode is a data class

maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
```diff
@@ -78,20 +81,23 @@ object DeviceScreenService {
         val elements = gatherElements(tree, mutableListOf())
             .sortedWith(Filters.INDEX_COMPARATOR)
 
-        fun getIndex(element: TreeNode, attribute: String): Int? {
-            val value = element.attribute(attribute) ?: return null
-            val matchingElements = elements.filter { value == it.attributes[attribute] }
+        fun getIndex(filter: ElementFilter, element: TreeNode): Int? {
+            val identityHashMap = IdentityHashMap<TreeNode, Unit>()
+            val matchingElements = Filters.deepestMatchingElement(filter)(elements).filter {
+                // There are duplicate elements for some reason (likely due to unintended behavior in Filter.deepestMatchingElement) - filter them out
```
dmitry-zaitsev, line 87: Duplicate elements are sometimes coming with the hierarchy, especially on iOS

dmitry-zaitsev (approved):

maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
Leland-Takamine: @dmitry-zaitsev The _objects_ are actually being duplicated by the filter so this is a bug in our code

PR #605, opened 2023-01-13
Add more advanced flows to samples

axelniklasson:
Adds two new advanced flows (one for iOS and one for Android) that incorporates swiping, subflows and javascript as well as assertions and input text.


dmitry-zaitsev (approved):

dmitry-zaitsev:
Heads up - this will need to be packaged and uploaded to GCP, there is currently no automatic deployment mechanism for that

PR #610, opened 2023-01-13
Refactor: make XCTestDriver configurable

dmitry-zaitsev:
## Proposed Changes

Communication with XCTestDriver was spread throughout several classes and abstraction layers, making it impossible to configure (i.e. to specify a different host name/port or a different way to install XCTest). Splitting the logic up based on `IOSDevice` interface to cleanly separate IDB-dependent logic from XCTest dependent logic and allow customization.


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -140,8 +146,29 @@ object MaestroSessionManager {
                         val channel = ManagedChannelBuilder.forAddress(defaultHost, idbPort)
                             .usePlaintext()
                             .build()
+
+                        val xcTestDriverClient = XCTestDriverClient(defaultHost, xcTestPort)
```
amanjeetsingh150, line 150: Can we test with the firewall that this is not giving the permission dialog for external network communication?

maestro-ios/src/main/java/ios/CompositeIOSDevice.kt
```diff
@@ -0,0 +1,127 @@
...
+    override fun stop(id: String): Result<Unit, Throwable> {
+        return idbIOSDevice.stop(id)
+    }
+
+    override fun openLink(link: String): Result<Unit, Throwable> {
+        return idbIOSDevice.openLink(link)
+    }
+
+    override fun takeScreenshot(out: Sink): Result<Unit, Throwable> {
+        return idbIOSDevice.takeScreenshot(out)
```
amanjeetsingh150, line 107: This was not happening with idb 🤔 implementation communicates with `XCRunnerSimctl` which directly issues simctl commands. Can we remove this idbIOSDevice call completely and directly use the XCRunnerSimctl.

dmitry-zaitsev: Whoops, will fix that by introducing a `XCRunDevice` and calling that one

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
dmitry-zaitsev: This change does not affect anything fundamentally as XCUITest server is still started on `.loopback` address. This is only needed for Cloud where we will do port forwarding

PR #641, opened 2023-01-19
Relatively swipe with percentage based start and end coordinates

amanjeetsingh150:
## Proposed Changes
Support percentage-based start and end coordinates that could help to be dimension agnostic and prevent giving absolute value.

Sample:
```
- swipe:
   start: 10%,50%
   end: 90%,50% 
```

## Testing

https://user-images.githubusercontent.com/12881364/213431391-f2f6ea5c-a095-426f-b52c-61a05f2c95be.mov

https://user-images.githubusercontent.com/12881364/213431409-c6989a8f-2d86-4764-a7ce-463fa262d466.mov

## Issues Fixed

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -110,10 +110,32 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         waitForAppToSettle()
     }
 
-    fun swipe(swipeDirection: SwipeDirection? = null, start: Point? = null, end: Point? = null, duration: Long) {
+    fun swipe(
+        swipeDirection: SwipeDirection? = null,
+        startPoint: Point? = null,
+        endPoint: Point? = null,
+        startRelative: String? = null,
+        endRelative: String? = null,
+        duration: Long
+    ) {
         when {
             swipeDirection != null -> driver.swipe(swipeDirection, duration)
-            start != null && end != null -> driver.swipe(start, end, duration)
+            startPoint != null && endPoint != null -> driver.swipe(startPoint, endPoint, duration)
+            startRelative != null && endRelative != null -> {
+                val startPoints = startRelative.replace("%", "")
+                    .split(",").map { it.trim().toInt() }
+                val startX = cachedDeviceInfo.widthGrid * startPoints[0] / 100
+                val startY = cachedDeviceInfo.heightGrid * startPoints[1] / 100
+                val start = Point(startX, startY)
+
+                val endPoints = endRelative.replace("%", "")
+                    .split(",").map { it.trim().toInt() }
+                val endX = cachedDeviceInfo.widthGrid * endPoints[0] / 100
+                val endY = cachedDeviceInfo.heightGrid * endPoints[1] / 100
+                val end = Point(endX, endY)
+
+                driver.swipe(start, end, duration)
+            }
         }
 
         waitForAppToSettle()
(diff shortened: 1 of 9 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -47,6 +47,8 @@ data class SwipeCommand(
     val startPoint: Point? = null,
     val endPoint: Point? = null,
     val elementSelector: ElementSelector? = null,
+    val startRelative: String? = null,
+    val endRelative: String? = null,
```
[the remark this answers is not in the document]
dmitry-zaitsev: I would go for V2 and soft-deprecating that one (marking as deprecated but not removing and not using in Yaml mapping)

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -126,6 +126,24 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         waitForAppToSettle()
     }
 
+    fun swipe(startRelative: String, endRelative: String, durationMs: Long) {
```
dmitry-zaitsev, line 129: I think this should be outside of `Maestro` class as Maestro is providing a strict/deterministic API. This can be done in Orchestra instead

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -47,6 +47,8 @@ data class SwipeCommand(
     val startPoint: Point? = null,
     val endPoint: Point? = null,
     val elementSelector: ElementSelector? = null,
+    val startRelative: String? = null,
```
dmitry-zaitsev, line 50: This those are strings, you can also update `evaluateScripts` to support JS in there

dmitry-zaitsev (approved):

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSwipe.kt
```diff
@@ -79,6 +77,47 @@ class YamlSwipeDeserializer : JsonDeserializer<YamlSwipe>() {
         }
     }
 
+    private fun resolveCoordinateSwipe(root: TreeNode, duration: Long): YamlSwipe {
+        when {
+            isRelativeSwipe(root) -> {
```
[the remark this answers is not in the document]
amanjeetsingh150: 💯 I'll note this and pick this while adding V2 command

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSwipe.kt
```diff
@@ -79,6 +77,47 @@ class YamlSwipeDeserializer : JsonDeserializer<YamlSwipe>() {
...
+                val start = root.path("start").toString().replace("\"", "")
+                val end = root.path("end").toString().replace("\"", "")
+                check(start.contains("%") && end.contains("%")) {
+                    "You need to provide start and end coordinates with %, Found: (${start}, ${end})"
+                }
+                val startPoints = start
+                    .replace("%", "")
+                    .split(",")
+                    .map { it.trim().toInt() }
```
Leland-Takamine, line 91: @amanjeetsingh150 Won't this crash when this isn't an integer? Also, if we are parsing the percentages here - why do we re-parse later upon executing the swipe?

PR #643, opened 2023-01-19
Feature scroll until view is visible

ArthurSav:
## Proposed Changes
Adds new command to scroll until an element is visible.

Example
```yaml
- scrollUntilVisible:
    element:
      text: "MyElementText"
    direction: DOWN
    timeout: 50000
```

## Testing
- Manual testing with Wikipedia app (Android/iOS)
- Integration test

### DIFF
--- a/maestro-client/src/main/java/maestro/ScrollDirection.kt
+++ b/maestro-client/src/main/java/maestro/ScrollDirection.kt
@@ -1,0 +1,15 @@
+package maestro
+
+enum class ScrollDirection {
+    UP,
+    DOWN,
+    RIGHT,
+    LEFT
+}
+
+fun ScrollDirection.toSwipeDirection(): SwipeDirection = when (this) {
+    ScrollDirection.DOWN -> SwipeDirection.UP
+    ScrollDirection.UP -> SwipeDirection.DOWN
+    ScrollDirection.LEFT -> SwipeDirection.RIGHT
+    ScrollDirection.RIGHT -> SwipeDirection.LEFT
+}
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -35,6 +35,7 @@ import maestro.orchestra.filter.FilterWithDescription
 import maestro.orchestra.filter.TraitFilters
 import maestro.orchestra.util.Env.evaluateScripts
 import maestro.orchestra.yaml.YamlCommandReader
+import maestro.toSwipeDirection
 import maestro.utils.MaestroTimer
 import maestro.utils.StringUtils.toRegexSafe
 import java.io.File
@@ -168,6 +169,7 @@ class Orchestra(
             is HideKeyboardCommand -> hideKeyboardCommand()
             is ScrollCommand -> scrollVerticalCommand()
             is CopyTextFromCommand -> copyTextFromCommand(command)
+            is ScrollUntilVisibleCommand -> scrollUntilVisible(command)
             is PasteTextCommand -> pasteText()
             is SwipeCommand -> swipeCommand(command)
             is AssertCommand -> assertCommand(command)
@@ -275,6 +277,27 @@ class Orchestra(
     private fun scrollVerticalCommand(): Boolean {
         maestro.scrollVertical()
         return true
+    }
+
+    private fun scrollUntilVisible(command: ScrollUntilVisibleCommand): Boolean {
+        val endTime = System.currentTimeMillis() + command.timeout
+        var error: MaestroException.ElementNotFound? = null
+        do {
+            try {
+                maestro.waitForAppToSettle()
+                val e =  findElement(command.selector, 500)
+                println("Bounds: ${e.bounds}")
+                return true
+            } catch (ignored: MaestroException.ElementNotFound) {
+                error = ignored
+            }
+
+            maestro.swipe(command.direction.toSwipeDirection(), duration = 600L)
+        } while (System.currentTimeMillis() < endTime)
+
+        if (error != null) throw error
+
+        return false
     }
 
     private fun hideKeyboardCommand(): Boolean {
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -285,8 +285,7 @@ class Orchestra(
         do {
             try {
                 maestro.waitForAppToSettle()
-                val e =  findElement(command.selector, 500)
-                println("Bounds: ${e.bounds}")
+                findElement(command.selector, 500)
                 return true
             } catch (ignored: MaestroException.ElementNotFound) {
                 error = ignored
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
@@ -1,0 +1,9 @@
+package maestro.orchestra.yaml
+
+import maestro.ScrollDirection
+
+data class YamlScrollUntilVisible(
+    val direction: ScrollDirection ?= null,
+    val element: YamlElementSelectorUnion? = null,
+    val timeout: Long? = null,
+)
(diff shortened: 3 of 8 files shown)


maestro-test/src/test/resources/078_scroll_until_visible.yaml
```diff
@@ -0,0 +1,7 @@
+appId: com.example.app
+---
+- scrollUntilVisible:
+    element:
+      text: "Test"
+    direction: DOWN
```
dmitry-zaitsev, line 6: Let's make sure lowercase values are also accepted

maestro-client/src/main/java/maestro/ScrollDirection.kt
```diff
@@ -0,0 +1,15 @@
+package maestro
+
+enum class ScrollDirection {
```
dmitry-zaitsev, line 3: Just a though, what if we call those: `ABOVE`, `BELOW`, `RIGHT`, `LEFT`? To avoid confusion with up/down scrolling

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -277,6 +279,27 @@ class Orchestra(
...
     }
 
+    private fun scrollUntilVisible(command: ScrollUntilVisibleCommand): Boolean {
+        val endTime = System.currentTimeMillis() + command.timeout
+        var error: MaestroException.ElementNotFound? = null
+        do {
+            try {
+                maestro.waitForAppToSettle()
+                val e =  findElement(command.selector, 500)
+                println("Bounds: ${e.bounds}")
```
dmitry-zaitsev, line 289: Remove `println` calls

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
```diff
@@ -0,0 +1,9 @@
+package maestro.orchestra.yaml
+
+import maestro.ScrollDirection
+
+data class YamlScrollUntilVisible(
+    val direction: ScrollDirection ?= null,
+    val element: YamlElementSelectorUnion? = null,
```
dmitry-zaitsev, line 7: It isn't really nullable, I think

maestro-client/src/main/java/maestro/ScrollDirection.kt
Leland-Takamine: This is indeed pretty confusing

ArthurSav: > ABOVE, BELOW, RIGHT, LEFT

This seems more confusing.

It is Up/Down scrolling, however it's the opposite of swipe direction. i.e `Swipe->UP` is the same as `Scroll->Down`

To me it makes sense that "Scroll Down" means it will start scrolling towards the bottom. Open to revise if users find it confusing.

maestro-test/src/test/resources/078_scroll_until_visible.yaml
ArthurSav: Done

maestro-client/src/main/java/maestro/ScrollDirection.kt
edgarfroes: What about naming them:

```
FROM_CENTER_LEFT_TO_CENTER_RIGHT
FROM_CENTER_RIGHT_TO_CENTER_LEFT
FROM_CENTER_TOP_TO_CENTER_BOTTOM
FROM_CENTER_BOTTOM_TO_CENTER_TOP
```

maestro-client/src/main/java/maestro/ScrollDirection.kt
```diff
@@ -0,0 +1,15 @@
...
+    UP,
+    DOWN,
+    RIGHT,
+    LEFT
+}
+
+fun ScrollDirection.toSwipeDirection(): SwipeDirection = when (this) {
```
amanjeetsingh150, line 10: Why do we need a different enum for this though?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -277,6 +279,26 @@ class Orchestra(
...
+    private fun scrollUntilVisible(command: ScrollUntilVisibleCommand): Boolean {
+        val endTime = System.currentTimeMillis() + command.timeout
+        var error: MaestroException.ElementNotFound? = null
+        do {
+            try {
+                maestro.waitForAppToSettle()
+                findElement(command.selector, 500)
+                return true
+            } catch (ignored: MaestroException.ElementNotFound) {
+                error = ignored
```
amanjeetsingh150, line 291: Wouldn't this pass the command if the element is not found? Or command would fail?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -277,6 +279,26 @@ class Orchestra(
...
 
+    private fun scrollUntilVisible(command: ScrollUntilVisibleCommand): Boolean {
+        val endTime = System.currentTimeMillis() + command.timeout
+        var error: MaestroException.ElementNotFound? = null
+        do {
+            try {
+                maestro.waitForAppToSettle()
+                findElement(command.selector, 500)
+                return true
+            } catch (ignored: MaestroException.ElementNotFound) {
```
amanjeetsingh150, line 290: Should we also be catching timeout exception. There could be a case where timeout configured is not sufficient and we should let user know if timeout needs to be adjusted

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
```diff
@@ -0,0 +1,11 @@
+package maestro.orchestra.yaml
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import maestro.ScrollDirection
+
+data class YamlScrollUntilVisible(
+    @JsonFormat(with = [JsonFormat.Feature.ACCEPT_CASE_INSENSITIVE_PROPERTIES])
+    val direction: ScrollDirection ?= null,
```
amanjeetsingh150, line 8: Why nullable 🤔

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
```diff
@@ -0,0 +1,11 @@
...
+    val element: YamlElementSelectorUnion,
+    val timeout: Long? = null,
```
amanjeetsingh150, line 10: There is a default timeout of 20s right we can keep that value if I am not wrong. Is nullable required here?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ArthurSav: Not sure i understand. If the element is not found within the timeout, an exception is thrown

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ArthurSav: There's no timeout exception. If the element is not found, only `ElementNotFound` is thrown

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt
ArthurSav: Good point

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
amanjeetsingh150: Yup so I was confused would it be Timeout or Element not found?

amanjeetsingh150: Got the answer below 👍

amanjeetsingh150 (approved):

maestro-client/src/main/java/maestro/ScrollDirection.kt
ArthurSav: Thanks @edgarfroes. We'll keep it as is for now and re-evaluate if necessary.

PR #647, opened 2023-01-20
Improve Android Screenshot Internal Logic

Leland-Takamine:
* Remove dependency on `adb shell screencap` and instead use `UIAutomation.takeScreenshot()`
* Use buffer of 10 screenshot files instead of deleting all previous screenshots


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -170,6 +173,22 @@ class Service(
         responseObserver.onCompleted()
     }
 
+    override fun screenshot(
+        request: MaestroAndroid.ScreenshotRequest,
+        responseObserver: StreamObserver<MaestroAndroid.ScreenshotResponse>
+    ) {
+        val outputStream = ByteString.newOutput()
+        val bitmap = uiAutomation.takeScreenshot()
```
amanjeetsingh150, line 181: What advantage this has over adb?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -170,6 +173,22 @@ class Service(
...
+        if (bitmap.compress(Bitmap.CompressFormat.PNG, 100, outputStream)) {
+            responseObserver.onNext(screenshotResponse {
+                bytes = outputStream.toByteString()
+            })
+            responseObserver.onCompleted()
+        } else {
+            responseObserver.onError(Throwable("Failed to compress bitmap"))
```
amanjeetsingh150, line 188: Lets log here as well since we can use it debug from logs of maestro server apk

maestro-studio/server/src/main/java/maestro/studio/DeviceScreenService.kt
```diff
@@ -27,22 +27,22 @@ import kotlin.io.path.createTempDirectory
...
             val screenshotFile: File
             synchronized(DeviceScreenService) {
                 tree = maestro.viewHierarchy().root
                 screenshotFile = takeScreenshot(maestro)
-                if (deletePrevious) {
-                    previousScreenshot?.delete()
-                    previousScreenshot = screenshotFile
+                savedScreenshots.add(screenshotFile)
+                while (savedScreenshots.size > MAX_SCREENSHOTS) {
+                    savedScreenshots.removeFirst().delete()
```
axelniklasson, line 45: I like this approach, good stuff 👍

axelniklasson (approved):

dmitry-zaitsev (approved):
Looks good, though what was the reasoning here?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
Leland-Takamine: This handles the screenshot all in memory and doesn't require writing to a file on the device. Also removes a dependency on the screencap tool

Leland-Takamine:
@dmitry-zaitsev I was trying to see if I could improve the performance of screenshotting to get better Maestro Studio rendering performance. This didn't really improve the performance much after all, but it is still a better way to handle screenshotting since it avoids writing to an intermediate file. I also updated the screenshotting logic within Maestro Studio to hopefully resolve some issues folks are having with screenshots not rendering.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
Leland-Takamine: Updated

PR #657, opened 2023-01-23
Tweak: report warning if we had to fall back to idb on inputText

dmitry-zaitsev:
## Proposed Changes

A new Insights mechanism to accumulate various warnings and tips throughout the test flow and report them at the end


dmitry-zaitsev:
If #673 works well, this PR won't really be needed

PR #661, opened 2023-01-23
feat: enable commitSha on cloud command

igorsmotto:
## Proposed Changes
- Adds a `commitSha` parameter to the `maestro cloud` command
- Send parameter to upload HTTP request

## Issues Fixed
- Today, GitHub PR checks always update the last commit of a PR. This is not the correct behavior, as a commit can have a check eternally on IN_PROGRESS state if a new one appears before the previous run finishes.


maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
```diff
@@ -65,6 +65,9 @@ class CloudCommand : Callable<Int> {
     @Option(order = 5, names = ["--branch"], description = ["The branch this upload originated from"])
     private var branch: String? = null
 
+    @Option(order = 6, names = ["--commitSha"], description = ["The commit SHA from this upload originated from"])
```
felipevolpone, line 68: The order number is wrong, is conflicting with the pullReuqestId parameter.

felipevolpone (approved):

maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
felipevolpone, line 68: Also, the description is confusing
I propose something like
```
The commit SHA of this upload
```

igorsmotto: https://github.com/mobile-dev-inc/maestro/pull/661/commits/95b8a7dfafd7556cb4e04f89d9cf26e5fab7a845

felipevolpone:
I'm not sure if changing the order breaks the compatibility for older clients, because of the order. So I suggest we just append at the end

maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
felipevolpone: I'm not sure if changing the order breaks the compatibility for older clients, because of the order. So I suggest we just append at the end

igorsmotto: Following the [docs](https://picocli.info/apidocs/picocli/CommandLine.Option.html), it should not be a problem

> When [@Command(sortOptions = false)](https://picocli.info/apidocs/picocli/CommandLine.Command.html#sortOptions--) is specified, this attribute can be used to control the order in which options are listed in the usage help message

dmitry-zaitsev (approved):

PR #664, opened 2023-01-24
Prepare for release 1.20.0

ArthurSav:


gradle.properties
```diff
@@ -2,7 +2,7 @@ android.useAndroidX=true
 android.enableJetifier=true
 kotlin.code.style=official
 GROUP=dev.mobile
-VERSION_NAME=1.19.6-SNAPSHOT
+VERSION_NAME=1.20.0-SNAPSHOT
```
artem888, line 5: according to documentation - should it be changed to non-snapshot version instead?

ArthurSav: Yep. Fixed it.

artem888 (approved):

PR #673, opened 2023-01-25
Use XCTest private API to input text on iOS

berikv:
## Proposed Changes

Use XCTest private API "_XCT_sendString:maximumFrequency:completion:" instead of looking up the element and typing to the element.
In complex UI's the element lookup can fail. By using the private API the element lookup is no longer needed.

## Testing

* git clone git@github.com:curlyCoder15/AwesomeProjectMaestro.git
* npm install -g npx
* Install cocoa pods
* Change the innermost <Text/> element in App.js to <TextInput />
* npx react-native start
* create a flow file with text input

```
appId: org.reactjs.native.example.AwesomeProject
---
- inputText: "Hello wørld 🙃"
```

* maestro test flow.yaml


dmitry-zaitsev:
Something I've been thinking about - if a call to private API fails, do we want to fallback to a previous behaviour? I am a bit torn on that as in case of a fallback we won't really know whether that happened or not, possibly contributing to flakiness

berikv:
> Something I've been thinking about - if a call to private API fails, do we want to fallback to a previous behaviour? I am a bit torn on that as in case of a fallback we won't really know whether that happened or not, possibly contributing to flakiness

Internally XCTest will call the same method. I don't think we gain anything by falling back to the public API in case of an error.

The only case I can think of when the public API would work but the private fails, is when an Xcode update changes the private API. This is something we should test for when a new major beta release of Xcode is available.

maestro-driver-ios/maestro-driver-iosUITests/maestro-driver-iosUITests-Bridging-Header.h
```diff
@@ -0,0 +1,23 @@
+//
```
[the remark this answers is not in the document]
berikv: Nope, I'll remove it

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
```diff
@@ -7,39 +7,48 @@ class InputTextRouteHandler : RouteHandler {
...
+            //            })
+
+            typealias sendStringMethod = @convention(c) (NSObject, Selector, NSString, Int, @escaping (Error?) -> ()) -> ()
+
+            let selector = NSSelectorFromString("_XCT_sendString:maximumFrequency:completion:")
+            let methodIMP = proxy.method(for: selector)
+
+            let method = unsafeBitCast(methodIMP, to: sendStringMethod.self)
+            method(proxy, selector, requestBody.text as NSString, 1, { error in
+                if let error = error {
```
amanjeetsingh150, line 43: Do we want to log here if we see any errors? Can we use `logger` to log it so that it gets written to the `xctest_runner_logs`

amanjeetsingh150: Awesome thanks @berikv, don't forget to update binary when you update driver code

Leland-Takamine (approved):

Leland-Takamine:
> > Something I've been thinking about - if a call to private API fails, do we want to fallback to a previous behaviour? I am a bit torn on that as in case of a fallback we won't really know whether that happened or not, possibly contributing to flakiness
> 
> Internally XCTest will call the same method. I don't think we gain anything by falling back to the public API in case of an error.
> 
> The only case I can think of when the public API would work but the private fails, is when an Xcode update changes the private API. This is something we should test for when a new major beta release of Xcode is available.

Agree with everything here - sounds like we should just be diligent about proactively testing Xcode updates

PR #691, opened 2023-01-26
Tweak: wait for IDB tap to respond

dmitry-zaitsev:


amanjeetsingh150 (approved):

Leland-Takamine (approved):
Anywhere else we're missing this?

dmitry-zaitsev:
@Leland-Takamine I don't think so, most of other calls are synchronous via a blockingStub or done via XCUITest (HTTP)

PR #698, opened 2023-01-27
Updates scroll until visible to work within viewport bounds

ArthurSav:
## Proposed Changes
Improves ScrollUntilVisible & fixes issues with iOS / Webviews

## Testing
- iOS Safari: www.duckduckgo.com
- iOS Maps

## Issues Fixed
 - **Fix**: ScrollUntilVisible now checks if an element is visible within viewport bounds.
 - **Tweak**: Scrolling will now start from the center of the screen. Avoids side effects like dragging bottom drawers.
 - **Tweak**: Better element alignment towards the center i.e it will scroll a bit more if the element is barely visible at a corner.

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -149,6 +149,13 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         waitForAppToSettle()
     }
 
+    fun swipeFromCenter(swipeDirection: SwipeDirection, durationMs: Long) {
+        LOGGER.info("Swiping ${swipeDirection.name} from center")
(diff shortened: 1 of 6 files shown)


dmitry-zaitsev (changes requested):
I believe we actually need to prune those views from view hierarchy, otherwise it starts to get confusing with commands such as `assertVisible`

ArthurSav:
> I believe we actually need to prune those views from view hierarchy, otherwise it starts to get confusing with commands such as `assertVisible`

@dmitry-zaitsev I was thinking of making a separate PR for this that would also fix `assertVisible`. This PR can work with or without hierarchy pruning.

dmitry-zaitsev (approved):
@ArthurSav if that is planned in another PR, then that is fair 👍

PR #710, opened 2023-01-30
Removes out of bound hierarchy elements

ArthurSav:
## Proposed Changes
Currently view hierarchy might return out of bound elements. This change will filter elements out of screen bounds.

## Testing
iOS Webview pages

## Issues Fixed
`assertVisible`

### DIFF
--- a/maestro-client/src/main/java/maestro/ViewHierarchy.kt
+++ b/maestro-client/src/main/java/maestro/ViewHierarchy.kt
@@ -97,11 +97,9 @@ value class ViewHierarchy(val root: TreeNode) {
 }
 
 fun TreeNode.filterOutOfBounds(width: Int, height: Int): TreeNode? {
-    val element = toUiElement()
-    val isXWithinBounds = element.bounds.x in 0..width
-    val isYWithinBounds = element.bounds.y in 0..height
+    val visiblePercentage = toUiElement().getVisiblePercentage(width, height);
 
-    if (!isXWithinBounds || !isYWithinBounds) {
+    if (visiblePercentage < 0.1) {
         return null
     }
 
@@ -119,3 +117,5 @@ fun TreeNode.filterOutOfBounds(width: Int, height: Int): TreeNode? {
         selected = selected
     )
 }
+
+
--- a/maestro-client/src/main/java/maestro/ViewHierarchy.kt
+++ b/maestro-client/src/main/java/maestro/ViewHierarchy.kt
@@ -26,11 +26,14 @@ value class ViewHierarchy(val root: TreeNode) {
     companion object {
         fun from(driver: Driver): ViewHierarchy {
             val deviceInfo = driver.deviceInfo()
-            val filtered = driver.contentDescriptor().filterOutOfBounds(
-                width = deviceInfo.widthGrid,
-                height = deviceInfo.heightGrid
-            ) ?: throw IllegalArgumentException("Root should not be null")
-            return ViewHierarchy(filtered)
+            val root = driver.contentDescriptor().let {
+                val filtered = it.filterOutOfBounds(
+                    width = deviceInfo.widthGrid,
+                    height = deviceInfo.heightGrid
+                )
+                filtered ?: it
+            }
+            return ViewHierarchy(root)
         }
     }
 
--- a/maestro-client/src/main/java/maestro/ViewHierarchy.kt
+++ b/maestro-client/src/main/java/maestro/ViewHierarchy.kt
@@ -97,15 +97,18 @@ value class ViewHierarchy(val root: TreeNode) {
 }
 
 fun TreeNode.filterOutOfBounds(width: Int, height: Int): TreeNode? {
-    val visiblePercentage = toUiElement().getVisiblePercentage(width, height);
-
-    if (visiblePercentage < 0.1) {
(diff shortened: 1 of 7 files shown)


maestro-client/src/main/java/maestro/ViewHierarchy.kt
```diff
@@ -89,3 +95,27 @@ value class ViewHierarchy(val root: TreeNode) {
         return root.aggregate()
     }
 }
+
+fun TreeNode.filterOutOfBounds(width: Int, height: Int): TreeNode? {
+    val element = toUiElement()
+    val isXWithinBounds = element.bounds.x in 0..width
+    val isYWithinBounds = element.bounds.y in 0..height
```
dmitry-zaitsev, line 102: You are checking for top left corner, however it could be that it is partially out of bounds while the bottom of the view is still in the viewport and could be visible

dmitry-zaitsev:
Could you add a test for that?

maestro-client/src/main/java/maestro/ViewHierarchy.kt
berikv: We'd need something along the lines of:
```
let isOverlapping = abs(screen.origin.x - view.origin.x) < 1 &&
                   abs(screen.origin.y - view.origin.y) < 1 &&
                   abs(screen.size.width - view.size.width) < 1 &&
                   abs(screen.size.height - view.size.height) < 1
```

ArthurSav: >You are checking for top left corner, however it could be that it is partially out of bounds while the bottom of the view is still in the viewport and could be visible

Should we consider partial elements as "visible"? We could check if any of the corners are within bounds and consider it "visible".

dmitry-zaitsev: _Or_ we could say that the view is visible if >10% of its area is in viewport, just to remove edge cases where only a small piece of the view is on the screen.

berikv: assertVisible and scrollToVisible should have the same **sensible** definition. Something to consider when going for a more or less heuristic definition like '>10%'.

dmitry-zaitsev: Agree, if we apply that criteria it should apply to both

ArthurSav: Screenshot 2023-01-31 at 18 14 54

Here's a valid case:

Both texts (top/bottom) that you see cut off are considered 100% visible since _technically_ they are within screen bounds. However most users would not consider this visible since the browser is hiding the texts.

A "hack" i did for scrollUntilVisible was to add a 10% element padding. Not sure we want the same for assertVisible.

Any thoughts?

dmitry-zaitsev: I would actually consider those to be visible technically speaking. As in, I would not expect my `assertVisible` to fail

ArthurSav: Right, for `assertVisible` it makes sense. For `scrollUntilVisible` i'd like to add a bit more control when the scroll stops (in another PR).

Conclusion for this PR. Consider elements visible, if 10% is visible, otherwise they remove from hierarchy.

(Will add tests for this)

maestro-client/src/test/java/maestro/UiElementTest.kt
```diff
@@ -0,0 +1,72 @@
+package maestro
+
+import org.junit.jupiter.api.Test
+import com.google.common.truth.Truth.assertThat
+
+internal class UiElementTest {
```
dmitry-zaitsev, line 6: Could you also add a case to `IntegrationTests` with an `assertVisible` command?

dmitry-zaitsev (approved):

ArthurSav:
@dmitry-zaitsev There are some weird edge cases that i'm looking into, so i'll delay merging this.
For example, in iOS webview parents are "out of bounds" bug children are not which causes side effects. 
Will update soon.

ArthurSav:
Updated the logic to include some edge cases.

I tested on:
- iOS webviews. [Works as expected]
- iOS apps (wikipedia) [Works as expected]
- android apps (reddit, Parallax) [Works as expected]
- android webviews [Does not work as expected]

> android webviews [Does not work as expected]
Bounds for android webviews are a bit off and don't seem to produce correct info. However this seems to be an existing issue - this PR won't change much there.
@dmitry-zaitsev 

A small illustration on how the pruning will look like.

maestro-hierarchy

maestro-client/src/main/java/maestro/ViewHierarchy.kt
dmitry-zaitsev: That makes sense

dmitry-zaitsev:
@ArthurSav thanks for the visualization, that makes total sense. Also ack on Android issue

maestro-client/src/test/java/maestro/UiElementTest.kt
ArthurSav: Done

Issue #395, opened 2022-11-14
inputText is typing too fast and skipping characters when app stutters/jank


edgarlealbd:
When running with Android emulator and calling `inputText`, it appears that the app suffers some stutter or jank and the command misses some characters. For example:

When executing `inputText` to type `Lorem ipsum`, sometimes it'll actually type `Loemipsu`. The skipped characters are random, so it does not appear to be a keyboard auto-correction. Sometimes, the same test will type characters correctly and tests will succeed.

I've tested that on debug builds, which are not optimized and the app can suffer from stutters or jank, and the problem is more frequent, but the same also works with an optimized release build.

My (hopefully) temporary workaround for this is to have one command per character, like this:

```
- inputText: "l"
- inputText: "o"
- inputText: "r"
- inputText: "e"
- inputText: "m"
```

Is there any way to set typing or command speed? If not, this could be a possible implementation for cases like this. E2E should have some breathing space when typing to avoid errors like this.

pwicherski:
@edgarlealbd is your problem maybe happening inside an Android webview, like some webform or login screen inside the app?

edgarlealbd:
@pwicherski the typing in question is not happening inside a WebView. The app is built with Flutter.

edgarlealbd:
Update: I can't reproduce the error locally with an Android 10 emulator, only with an Android 12 (API 31).

Leland-Takamine:
We've opened up a fix for this. The fix should be included in the next release of Maestro once this is landed #417

amanjeetsingh150:
Hey @edgarlealbd @pwicherski 

I have just merged the fix can you upgrade the local maestro to head and try these commands to check if this is fixed. Would recommend to do it multiple times as this was inconsistent.  To upgrade:

```
 brew uninstall maestro
 brew install maestro-head --HEAD
```
Let me know if this works will release this in the new version

To go back on stable after checking


```
 brew uninstall maestro
 brew install maestro
```

edgarlealbd:
@amanjeetsingh150 I'm getting java-related errors when running with the head version from homebrew:

I have a Mac Pro M1 Max.

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: maestro_android/MaestroDriverGrpc has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
[...]

PR #720, opened 2023-02-01
Fix input text skipping characters

artem888:
## Proposed Changes
Set `maximumFrequency` to a fixed value of 15 characters per second. Resolves flakiness issues 

## Testing
- [x] tested locally with RN app (that was previously flaky)
- [x] tested locally with UIKit app

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/395#issuecomment-1411549630


maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
```diff
@@ -62,12 +62,6 @@ class InputTextRouteHandler : RouteHandler {
         }
     }
 
-    private func typingFrequencyFor(text: String) -> Int {
-        let count = max(text.count, 5)
-        return count * 2
-    }
-
-
```
berikv, line 70: We decided to make typing frequency dependent on the text length because some users enter long strings of text.

dmitry-zaitsev: I think we should always type with maximum reliable frequency. If typing faster causes characters to be skipped - that would be a big problem.

berikv (approved):

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
berikv: How certain are we that the typing frequency causes the skipping?

dmitry-zaitsev: I have high confidence that it does have an impact, for two reasons:

- We had a similar problem before with idb and adjusting the frequency helped to solve it
- This problem resurfaced after moving to the new driver, meaning it is highly likely related

dmitry-zaitsev (approved):

Issue #646, opened 2023-01-19
"Unable to obtain active app id" error on maestro commands for iOS


edgarfroes:
I get the error below when running commands on an iOS simulator on the latest maestro version (`1.19.5`):

Commands that I've tested so far and that are failing: `test`, `studio`, and `hierarchy`.

<details><summary>Click to see the complete error log</summary>
<p>

``` bash
java.lang.IllegalStateException: Unable to obtain active app id                      
        at ios.xctest.XCTestIOSDevice.contentDescriptor(XCTestIOSDevice.kt:48)       
        at ios.LocalIOSDevice.contentDescriptor(LocalIOSDevice.kt:32)                
        at maestro.drivers.IOSDriver.contentDescriptor(IOSDriver.kt:175)             
        at maestro.Maestro$findElementWithTimeout$1.invoke(Maestro.kt:344)           
        at maestro.Maestro$findElementWithTimeout$1.invoke(Maestro.kt:343)           
        at maestro.utils.MaestroTimer.withTimeout(MaestroTimer.kt:16)                
        at maestro.Maestro.findElementWithTimeout(Maestro.kt:343)                    
        at maestro.orchestra.Orchestra.findElement(Orchestra.kt:646)                 
        at maestro.orchestra.Orchestra.findElement$default(Orchestra.kt:628)         
        at maestro.orchestra.Orchestra.tapOnElement(Orchestra.kt:558)                
        at maestro.orchestra.Orchestra.executeCommand(Orchestra.kt:159)              
        at maestro.orchestra.Orchestra.runSubFlow(Orchestra.kt:443)                  
        at maestro.orchestra.Orchestra.runFlowCommand(Orchestra.kt:358)              
        at maestro.orchestra.Orchestra.executeCommand(Orchestra.kt:185)              
        at maestro.orchestra.Orchestra.runSubFlow(Orchestra.kt:443)                  
        at maestro.orchestra.Orchestra.runFlowCommand(Orchestra.kt:358)              
        at maestro.orchestra.Orchestra.executeCommand(Orchestra.kt:185)              
        at maestro.orchestra.Orchestra.executeCommands(Orchestra.kt:137)             
        at maestro.orchestra.Orchestra.runFlow(Orchestra.kt:91)                      
        at maestro.cli.runner.MaestroCommandRunner.runCommands(MaestroCommandRunner.kt:120)
[...]

Issue #628, opened 2023-01-17
java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:9080 on iOS Simulator test run


edgarfroes:
Similarly to #595 and https://github.com/mobile-dev-inc/dadb/issues/48, this error occurs intermittently when trying to run flows with `maestro test`. In my local environment, this tends to happen when I have another terminal tab running `maestro studio` at the same time that I try to run `maestro test`, whether the `maestro studio` process is running or if the process had been closed a few seconds ago (conflicting background process?).

Maestro version 1.19.2 (installed via curl)
Environment: Mac Pro M1 Max

<details><summary>Click to see the complete log</summary>
<p>

``` bash
java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:9080
	at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.kt:297)
	at okhttp3.internal.connection.RealConnection.connect(RealConnection.kt:207)
	at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.kt:226)
	at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:106)
	at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:74)
	at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:255)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
[...]

PR #722, opened 2023-02-01
[MOB 1882] Recover from connect exception in iOS

amanjeetsingh150:
## Proposed Changes

1. Session Store corrupting state, and sometimes the driver is not opened. For example, if someone stops the setting up studio it becomes unusable with ConnectException in next tries. This PR guarantees to open the driver by checking the presence of UI test runner app and connection with the HTTP server.

2. Exiting studio by Ctrl +C, produces ConnectException and other logs which is misleading and not a good experience.

3. In case the execution of test is stopped it crashes the studio in the other terminal since it also requests for hierarchy. Avoid closing the driver if a session is active also restoring the session by installing the driver.

4. Throw `XCUITestUnreachableException` instead of ConnectException in cases where the communication cannot be established for example failure of a xcuitest.
 
5. In case of continuous mode in one session and studio in another, canceling the continuous mode would crash studio. Restoration of session also prevent the studio session to crash.

6. Meanwhile changes port number to some uncommon one. Finding available ports to be done soon in future improvements. 

## Testing

https://user-images.githubusercontent.com/12881364/216123887-0080eee4-c25d-4360-8151-cf6edb4abd61.mp4

https://user-images.githubusercontent.com/12881364/216125062-95fa6c3b-02e6-47db-92de-eb5215b13cfb.mp4

https://user-images.githubusercontent.com/12881364/216126657-eb404254-5a50-4ece-bf7f-9e9be6617197.mp4

## Issues Fixed

https://github.com/mobile-dev-inc/maestro/issues/646
https://github.com/mobile-dev-inc/maestro/issues/628


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -43,7 +44,7 @@ import kotlin.concurrent.thread
 object MaestroSessionManager {
     private const val defaultHost = "localhost"
     private const val idbPort = 10882
-    private const val xcTestPort = 9080
+    private const val xcTestPort = 22087
```
artem888, line 47: PS we need to update documentation and notify https://github.com/jpudysz/react-native-is-maestro about the change

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -156,4 +198,15 @@ class XCTestDriverClient(
             .port(port)
     }
 
+    fun cancelOngoingRequests() {
+        okHttpClient.dispatcher.runningCalls().filter { it.request().tag() == TAG_RUNNING_APP_CALL }
```
artem888, line 202: why are we only cancelling those two? what about other requests?

artem888 (approved):
LGTM after comments addressed 👏

hunzai:
It would be nice to pass the following as env vars
```
 private const val defaultHost = "localhost"
    private const val idbPort = 10882
    private const val xcTestPort = 9080
    private const val xcTestPort = 22087
```

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: Actually, this is not needed anymore, since we pass a dummy response when everything is shutting down. Removed this now

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
amanjeetsingh150: Will be updating documentation and sending PR to this project ✅

dmitry-zaitsev: Note that we need to update `didb` as well to forward traffic to this port

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -87,7 +88,7 @@ object MaestroSessionManager {
                 heartbeatFuture.cancel(true)
                 SessionStore.delete(sessionId, selectedDevice.platform)
 
-                if (!SessionStore.hasActiveSessions(sessionId, selectedDevice.platform)) {
+                if (SessionStore.activeSessions().isEmpty()) {
```
dmitry-zaitsev, line 91: This will always be `false` since _current_ session is present in this list

amanjeetsingh150: Current session actually gets deleted, in line number 89. I validated this from logs.

amanjeetsingh150:
Hey, @hunzai thanks for the suggestion. We are going to eventually be looking for available ports so as a stopgap, we have changed this to something not so common.

dmitry-zaitsev (approved):

PR #733, opened 2023-02-03
Remove unused return value for onCommandFailed

berikv:
## Proposed Changes

`ErrorResolution.CONTINUE` is never used. So it is safe to remove the condition and use the `ErrorResolution.FAIL` path.
I believe that the return code from onCommandFailed was used to make Orchestra continue on failure. If we need that feature then let's not use onCommandFailed for deciding that (a better is to pass a constructor argument `Orchestra(continueOnFailure:)`).

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -53,7 +53,7 @@ class Orchestra(
     private val onFlowStart: (List<MaestroCommand>) -> Unit = {},
     private val onCommandStart: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandComplete: (Int, MaestroCommand) -> Unit = { _, _ -> },
-    private val onCommandFailed: (Int, MaestroCommand, Throwable) -> ErrorResolution = { _, _, e -> throw e },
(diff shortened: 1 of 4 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -469,13 +465,8 @@ class Orchestra(
...
                     } catch (e: Throwable) {
-                        when (onCommandFailed(index, command, e)) {
-                            ErrorResolution.FAIL -> throw e
-                            ErrorResolution.CONTINUE -> {
-                                // Do nothing
-                                false
-                            }
-                        }
+                        onCommandFailed(index, command, e)
+                        throw e
```
berikv, line 469: @dmitry-zaitsev do you know why `runSubFlow` throws here, while `executeCommands` does not?

dmitry-zaitsev (changes requested):
I believe this code path is used in Cloud

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
dmitry-zaitsev: As discussed in person, I do not remember _exactly_ why it was needed but I believe there was a need for that. Something possibly related to CLI output or to special exception types that we need to propagate up (i.e. when command is skipped)

PR #734, opened 2023-02-03
Improve tapOn command execution times on iOS

artem888:
## Proposed Changes

This PR makes waitForAppToSettle logic to rely on newly implemented endpoint that compares two screenshots on iOS driver side.

Example of performance wins of one of [Food Truck app flows](https://github.com/artem888/sample-food-truck/blob/main/.mobiledev/first_flow.yaml):

Screenshot 2023-02-03 at 17 16 47

## Testing
- [x] Food Truck
- [x] Wikipedia (UIKit)
- [x] ReactNative


maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/ScreenDiffHandler.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import XCTest
+import CryptoKit
+import FlyingFox
+import os
+
+class ScreenDiffHandler : RouteHandler {
```
dmitry-zaitsev, line 7: We already have screenshot comparison logic within Maestro (that also allows to configure threshold), can we reuse that one instead of rebuilding it in Swift?

artem888: Had some issues with image-comparison library there, even when sampling / resizing images it wasn't performant enough comparing to this approach, hash function in Swift is much faster.

dmitry-zaitsev: Ok, let’s discuss this in person, I doubt that the performance difference is that large. The reason we are moving to screneshot comparison is to eliminate delay in viewHierarchy calls, so perhaps the compression speed is not really an important factor here?

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -393,21 +396,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
-                    return hierarchyAfter
-                }
-            }
-            latestHierarchy = hierarchyAfter
-
-            MaestroTimer.sleep(MaestroTimer.Reason.WAIT_TO_SETTLE, 200)
-        }
-
-        return latestHierarchy
+        return if(isIOS) screenshotWaitForAppToSettle(initialHierarchy) else genericWaitForAppToSettle(initialHierarchy)
```
amanjeetsingh150, line 399: Can we ask the driver to do waitForAppToSettle using polymorphism? The method calls and method names has become confusing 🤔

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -536,11 +507,67 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+        }
+    }
+    private fun screenshotWaitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy {
+        val timeout = ANIMATION_TIMEOUT_MS
+        LOGGER.info("Waiting for animation to end with timeout $timeout")
+        val didFinishOnTime = waitUntilScreenIsStatic(SCREEN_SETTLE_TIMEOUT_MS)
+
+        return if (didFinishOnTime) viewHierarchy() else genericWaitForAppToSettle(initialHierarchy)
+    }
+    private fun genericWaitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy {
```
amanjeetsingh150, line 547: Can we create one method for waitForAppSettle and pass comparing method as an input. Maybe we can have enum Hierarchy and Screenshot being two members

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/ScreenDiffHandler.swift
```diff
@@ -0,0 +1,38 @@
...
+    private let logger = Logger(subsystem: Bundle.main.bundleIdentifier!, category: "ScreenDiffHandler")
+    
+    func handle(request: FlyingFox.HTTPRequest) async throws -> FlyingFox.HTTPResponse {
+        let screenshot1 = XCUIScreen.main.screenshot()
```
amanjeetsingh150, line 11: Can this method fail? If yes lets have log around it. Also can you test if this works well with screens which were failing due to being long (feed screens)

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
```diff
@@ -212,6 +213,27 @@ class XCTestIOSDevice(
...
+                    if (response.isSuccessful) {
+                        val responseBody: GetScreenDiffResponse = mapper.readValue(
+                            String(body.bytes()),
+                            GetScreenDiffResponse::class.java
+                        )
+                        val screenChanged = responseBody.screenChanged
+                        logger.info("Screen diff request finished with changed = $screenChanged")
+                        screenChanged
+                    } else {
+                        val errorResponse = String(body.bytes()).trim()
```
amanjeetsingh150, line 229: Let's log this

dmitry-zaitsev:
I would like to see a comparison with Kotlin's image comparison library. Is there any way we could optimize that one instead?

dmitry-zaitsev (approved):
LGTM as long as comments from @amanjeetsingh150 are addressed

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/ScreenDiffHandler.swift
artem888: this one seems to be non-failable, result is not optional and the func doesn't throw

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -534,11 +505,66 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+                return@retryUntilTrue !screenChanged
+            }
+        } else {
+            return MaestroTimer.retryUntilTrue(timeoutMs) {
+                val startScreenshot: BufferedImage? = tryTakingScreenshot()
+                val endScreenshot: BufferedImage? = tryTakingScreenshot()
+
+                if (startScreenshot != null &&
+                    endScreenshot != null &&
+                    startScreenshot.width == endScreenshot.width &&
```
amanjeetsingh150, line 523: Why are we comparing width and height? Will not that always be same?

maestro-client/src/main/java/maestro/Driver.kt
```diff
@@ -84,4 +84,5 @@ interface Driver {
 
     fun isUnicodeInputSupported(): Boolean
 
+    fun isScreenChanged(): Boolean
```
berikv, line 87: Lets call this `isScreenStatic`, also `screenDiff` is better named `isScreenStatic`.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/ScreenDiffHandler.swift
```diff
@@ -0,0 +1,38 @@
...
+        let screenshot2 = XCUIScreen.main.screenshot()
+        let hash1 = SHA256.hash(data: screenshot1.pngRepresentation)
+        let hash2 = SHA256.hash(data: screenshot2.pngRepresentation)
+        
+        let screenChanged = hash1 != hash2
```
berikv, line 16: Note that iOS may emit two screens that are equal, even during an animation.
To reduce the risk of early a premature trigger, we can do:
```
let period = request.period
let start = Date()
var screenshot = XCUIScreen.main.screenshot()
while (Date().durationSince(start) < period) {
  val compare = XCUIScreen.main.screenshot()
  if hash(screenshot) != hash(compare) { return false }
}
return true
```

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -534,11 +505,66 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         return driver.isUnicodeInputSupported()
     }
 
+    private fun waitUntilScreenIsStatic(timeoutMs: Long): Boolean {
+        if (isIOS) {
+            return MaestroTimer.retryUntilTrue(timeoutMs) {
+                val screenChanged = driver.isScreenChanged()
```
berikv, line 511: It'd make sense to apply a heuristic matching algorithm when screenChanged returns false. This way we would have both the speedup of hash compare and the heuristic screen matching feature.

maestro-client/src/main/java/maestro/Maestro.kt
artem888: the logic was there already, I think it is because ImageComparison library expects it to be the same, we just catch this

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/ScreenDiffHandler.swift
artem888: discussed in private

PR #742, opened 2023-02-07
Add maestro mockserver scaffold command

axelniklasson:
## Proposed Changes
Adds a `maestro mockserver scaffold` command that deploys a sample rule for a GET request to `/posts` that returns some sample data. It will scaffold the `index.js` file in `./maestro/mockserver` if there is a `.maestro` directory present in the current path and if not, it will create a directory called `sample-workspace` and use that. It will then do a deployment of the workspace. I also updated the instructions page a bit:

Screenshot 2023-02-07 at 11 18 04

## Testing
Tested locally both with and without a `.maestro` folder present in the current directory.


maestro-cli/src/main/java/maestro/cli/command/mockserver/MockServerScaffoldCommand.kt
```diff
@@ -0,0 +1,72 @@
...
+        return CloudInteractor(
+            client = ApiClient(apiUrl),
+        ).deployMaestroMockServerWorkspace(
+            workspace = dir,
+            apiKey = apiKey,
+        )
+    }
+}
+
+val sampleIndexJs = """
```
dmitry-zaitsev, line 63: minor nit: better put it in a `companion object`, otherwise this field will show up every time you try to use autocomplete in IDE (since it is in a global scope)

maestro-studio/web/src/MockServerInstructions.tsx
```diff
@@ -0,0 +1,35 @@
+import { CodeSnippet } from "./Examples"
+
+const MockServerInstructions = ({ projectId }: { projectId: string }) => (
+  <div className="w-full h-full flex justify-center items-center">
+    <div className="flex flex-col w-3/4 space-y-6 p-8 shadow-xl border-4 border-blue-300 margin-auto">
+      <p className="text-lg font-bold">No events found! Follow the guide below to integrate Maestro SDK.</p>
+      
+      <div>
+        <p className="text-md">First, add the Maestro SDK dependency to your app:</p>
+        <CodeSnippet>{`implementation 'dev.mobile:maestro-sdk-android:1.21.4-SNAPSHOT'`}</CodeSnippet>
```
dmitry-zaitsev, line 10: Replace `1.21.4-SNAPSHOT` with `+` **OR** get a version from `gradle.properties`, whichever you prefer

dmitry-zaitsev (approved):

PR #749, opened 2023-02-08
Feature: global tags

dmitry-zaitsev:
## Proposed Changes

Test tags give you further flexibility to customize which Flows are executed in a test suite. This can also be configured in the config.yaml file:

```
# Include all Flows tagged with onboarding
includeTags:
  - onboarding

# Exclude any Flows that are tagged with subflow
excludeTags:
  - subflow
```


Leland-Takamine (approved):

dmitry-zaitsev:
Let's **not** include it in the 22 yet, this needs some Cloud changes to be made

Issue #732, opened 2023-02-03
[BUG] - Launching app with clearState: true is still crashing the app on every subsequent startup


simon-gilmurray:
Using 

```
 - launchApp:
    clearState: true
```

makes the app crash on every subsequent open

simon-gilmurray:
Slack threads and others experiencing same issue:

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1672762188380479

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1674121973358979

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1674480226578929

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1673881720161289

dmitry-zaitsev:
Could you please give Maestro 1.22.1 a try and see if that is still the case?

simon-gilmurray:
This looks to be fixed on 1.22.1, thanks @berikv and @dmitry-zaitsev

PR #752, opened 2023-02-08
Resolve crashes after clearState due to missing system directories

berikv:
## Proposed Changes

After removing the app's data folder, create all directories that are supplied by the OS by default.

## Testing

Ran flows on various customer apps that used to crash with clearState

## Issues Fixed

https://github.com/mobile-dev-inc/maestro/issues/732


dmitry-zaitsev:
Nice!

PR #755, opened 2023-02-08
Implement `clearAppState` in new `SimctlIOSDevice`

berikv:
## Proposed Changes

Move the clearAppState logic from idb to new SimctliOSDevice.

## Testing
Tested manually on several flows

### DIFF
--- a/maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
@@ -317,49 +317,7 @@ class IdbIOSDevice(
     }
 
     override fun clearAppState(id: String): Result<Unit, Throwable> {
-
-        // Stop the app before clearing the file system
-        // This prevents the app from saving its state after it has been cleared
-        stop(id)
-
-        // Wait for the app to be stopped, unfortunately idb's stop()
-        // does not wait for the process to finish
-        Thread.sleep(1500)
-
-        // deletes app data, including container folder
-        val result = runCatching {
-            blockingStub.rm(rmRequest {
-                container = fileContainer {
-                    kind = Idb.FileContainer.Kind.APPLICATION
-                    bundleId = id
-                }
-                paths.add("/")
-            })
-        }
-
-        // forces app container folder to be re-created
-        val paths = listOf(
-            "Documents",
-            "Library",
-            "Library/Caches",
-            "Library/Preferences",
-            "SystemData",
-            "tmp"
-        )
-
-        runCatching {
-            paths.forEach { path ->
-                blockingStub.mkdir(mkdirRequest {
-                    container = fileContainer {
-                        kind = Idb.FileContainer.Kind.APPLICATION
-                        bundleId = id
-                    }
-                    this.path = path
-                })
-            }
-        }
-
-        return result.map {  }
+        error("Not supported")
     }
 
     override fun clearKeychain(): Result<Unit, Throwable> {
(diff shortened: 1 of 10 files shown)


berikv:
It'd also be possible to let `LocalIOSDevice` call out to `Simctl` directly (not via XCTestDevice).

maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
```diff
@@ -317,49 +317,7 @@ class IdbIOSDevice(
...
-                        kind = Idb.FileContainer.Kind.APPLICATION
-                        bundleId = id
-                    }
-                    this.path = path
-                })
-            }
-        }
-
-        return result.map {  }
+        error("Not supported")
```
dmitry-zaitsev, line 320: Let's keep the IDB implementation, that might be handy if we want to quickly rollback to a previous behaviour in Cloud (or even do a fallback)

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
```diff
@@ -169,7 +175,8 @@ class XCTestIOSDevice(
     }
 
     override fun stop(id: String): Result<Unit, Throwable> {
-        error("Not supported")
+        deviceId?.let { deviceId -> Simctl.terminate(deviceId, id) }
```
dmitry-zaitsev, line 178: I don't think this fits in `XCTestIOSDevice` - there is a dependency on Simctl that we can't satisfy in Cloud. I suggest creating a new `SimctlIOSDevice` that has those methods implemented in there and route to them from `LocalIOSDevice`

dmitry-zaitsev (dismissed):
The way it is implemented now will not work on Cloud - we would need to keep IDB implementation intact until we re-implement it on Cloud

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
berikv: Having many (3) half-implemented classes does not help with programming safety.
I thought of XCTestIOSDevice as the 'idb-less future', so the unimplemented methods make sense because we're in a transition period.
I'm opposed to make a permanent setup where half of the methods are not implemented and the caller needs to understand what implementation it is talking to to safely call a method.

berikv: One way forward is to drop the XCTestIOSDevice: IOSDevice adherence (rename to XCTestIOSHelper?), and create a SimcltIOSHelper which implements the simctl code and also does not adhere to IOSDevice.

berikv:
> The way it is implemented now will not work on Cloud - we would need to keep IDB implementation intact until we re-implement it on Cloud

Cloud implementation will intercept the call before reaching LocalIOSDevice

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
dmitry-zaitsev: That approach did not work in the past as we ended up with a bunch of mix-ins in an IOSDevice implementation, some of which worked in Cloud and some didn’t.

Half-implemented versions of this class allow us to clearly separate at Local/Cloud device level what implementation is expected to work when. I.e. in current setup calling XCTest device is expected to work in cloud for all implemented method except the ones that were just added. Let’s keep it consistent - any given implementation should either fully support cloud or not support it at all.

dmitry-zaitsev:
Let’s keep IDB implementation in place until the new one is ready. Otherwise we won’t be able to upgrade to a new Maestro version until that change is done

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
berikv: I'll go forward with the SimctlIOSDevice approach you suggested for now and see if I can find an architectural improvement separately.

artem888 (approved):

dmitry-zaitsev (approved):

PR #764, opened 2023-02-09
Solve an issue where Element selector crashes the XCTestRunner

berikv:
## Proposed Changes

Remove the XCUIElement based text input variant because it crashes with apps with big view hierarchies (react native).

## Testing

Manual


berikv:
> did it ever happened in prod? asking because 45+ symbols input + RN app with complex hierarchy sounds like edge-case scenario.
Yes I see it with a customers app that I'm testing.

> the downside is the input will be slow and might even timeout
Slow is better than broken. Let's do a real copy/paste implementation instead of falling back to element selectors.

artem888 (approved):

PR #766, opened 2023-02-09
Use XCTest for tapping on iOS

berikv:
## Proposed Changes

Use XCTest for tapping on iOS


dmitry-zaitsev (approved):

berikv:
> Can we use IDB as a fallback just in case?

I'd rather go for no-idb directly

dmitry-zaitsev:
We are risking to break this for every user of Maestro, let's do it safely and add a fallback

amanjeetsingh150:
> Can we use IDB as a fallback just in case?
@dmitry-zaitsev last I remember adding private apis have resolved for long screens/react native as well. Do we still know any failure cases?

berikv:
As discussed offline, will first run an extensive test before merging

PR #767, opened 2023-02-09
Use xcrun for launching app on iOS

berikv:
## Proposed Changes

Use XCTest for launching app on iOS


artem888 (approved):

dmitry-zaitsev (approved):

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -77,7 +77,8 @@ class SimctlIOSDevice(
     }
 
     override fun launch(id: String): Result<Unit, Throwable> {
-        TODO("Not yet implemented")
+        Simctl.launch(deviceId, id)
```
amanjeetsingh150, line 80: Same this can fail as well hence I would say lets do runCatching

PR #769, opened 2023-02-09
Use xcrun for opening URL on iOS

berikv:
## Proposed Changes

Use XCTest for opening URL on iOS

### DIFF
--- a/maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
@@ -90,7 +90,9 @@ class SimctlIOSDevice(
     }
 
     override fun openLink(link: String): Result<Unit, Throwable> {
-        TODO("Not yet implemented")
+        return runCatching {
+            Simctl.openURL(deviceId, link)
+        }
     }
 
     override fun takeScreenshot(out: Sink, compressed: Boolean): Result<Unit, Throwable> {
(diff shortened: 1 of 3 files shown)


dmitry-zaitsev (approved):

artem888 (approved):

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -87,7 +87,8 @@ class SimctlIOSDevice(
     }
 
     override fun openLink(link: String): Result<Unit, Throwable> {
-        TODO("Not yet implemented")
+        Simctl.openURL(deviceId, link)
```
amanjeetsingh150, line 90: Actually this can fail, so I would still recommend returning with runCatching

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -86,7 +86,10 @@ class SimctlIOSDevice(
     }
 
     override fun openLink(link: String): Result<Unit, Throwable> {
-        TODO("Not yet implemented")
+        runCatching {
+            Simctl.openURL(deviceId, link)
+        }
+        return Ok(Unit)
```
amanjeetsingh150, line 92: Same, this should be:

Returning result of runCatching:

runCatching {
    Simctl.openURL(deviceId, link)
}
The result of above is either Ok or Err that helps caller telling if the call actually succeeded or failed

berikv: Aha, I'll update all functions in SimctlIOSDevice that are applicable in a separate PR

PR #772, opened 2023-02-09
Use xcrun for setting location in iOS

berikv:
## Proposed Changes

Use XCTest for setting location in iOS

## Testing

Tested locally

### DIFF
--- a/maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
@@ -10,6 +10,7 @@ import maestro.logger.Logger
 import okio.Sink
 import java.io.File
 import java.io.InputStream
+import com.github.michaelbull.result.runCatching
 
 class SimctlIOSDevice(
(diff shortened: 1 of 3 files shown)


dmitry-zaitsev (approved):
Oh, didn't know it was possible via `simctl` 👍

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -98,7 +98,10 @@ class SimctlIOSDevice(
     }
 
     override fun setLocation(latitude: Double, longitude: Double): Result<Unit, Throwable> {
-        TODO("Not yet implemented")
+        runCatching {
+            Simctl.setLocation(deviceId, latitude, longitude)
+        }
+        return Ok(Unit)
```
amanjeetsingh150, line 104: We should actually return result of runCatching:

```
runCatching {
            Simctl.setLocation(deviceId, latitude, longitude)
}
```

The result of above is either Ok or Err that helps caller telling if the call actually succeeded or failed

PR #774, opened 2023-02-09
Don't call proceed twice in network interceptor

amanjeetsingh150:


maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -19,6 +19,7 @@ import java.util.concurrent.TimeUnit
 class XCTestDriverClient(
     private val host: String = "localhost",
     private val port: Int = 22087,
+    private val isStudio: Boolean,
```
berikv, line 22: Let's not couple XCTestDriverClient to studio (semantically). Can you find a better name for this property?

amanjeetsingh150: Done renamed it! ✅

berikv (approved):

PR #778, opened 2023-02-10
Remove expect and silently fail inputText if focus not found

amanjeetsingh150:
## Proposed Changes

Silently fail input text if the focus is not found by removing expect call which is throwing an exception explicitly.

### Bug
https://user-images.githubusercontent.com/12881364/218088287-2b906532-a9ad-446d-99d2-56507fb36bde.mov

### Fix

https://user-images.githubusercontent.com/12881364/218089063-928424fd-fdee-4114-b5e7-c358d38f37e8.mov

## Issues Fixed


berikv (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -434,7 +434,7 @@ class IOSDriver(
     override fun inputText(text: String) {
         iosDevice.input(
             text = text,
-        ).expect {}
+        )
```
berikv, line 437: Could you add a comment string mentioning the silent fail?

PR #779, opened 2023-02-10
Remove ScreenshotDriver and move logic to device drivers instead

artem888:
## Proposed Changes

Remove excessive ScreenshotDriver classes and moves logic to device drivers.

## Testing
- [x] Food Truck
- [x] Wikipedia (UIKit)
- [x] ReactNative


maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
```diff
@@ -341,6 +338,14 @@ class FakeDriver : Driver {
         }
     }
 
+    override fun waitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy {
+        return ViewHierarchy(TreeNode())
```
dmitry-zaitsev, line 342: This should perform in the same way as Android _or_ iOS would

artem888: what if they perform differently, should I rely on Android implementation or iOS? :)

dmitry-zaitsev: That is a good question. Perhaps make it iOS-like by default and then maybe we can make it configurable later if we want to

dmitry-zaitsev (approved):

PR #783, opened 2023-02-10
Use xcrun for clearKeychain

berikv:
## Testing
Tested using patch in Wikipedia app:

```
    func readKeychain() -> Bool {
        let query = [
            kSecClass: kSecClassGenericPassword,
            kSecReturnData: true
        ] as CFDictionary

        var result: AnyObject?
        SecItemCopyMatching(query, &result)

        return (result as? Data)?.isEmpty == false
    }

    func setKeychain() {
        let query = [
            kSecValueData: "data".data(using: .utf8)!,
            kSecClass: kSecClassGenericPassword
        ] as CFDictionary

        let status = SecItemAdd(query, nil)

        if status != errSecSuccess {
            print("Keychain error: \(status)")
        }
    }

    private func updateUIStrings() {

        let text = readKeychain() ? "Keychain set" : "Keychain cleared"
        setKeychain()
```

* launch app: Keychain cleared
* kill and relaunch app: Keychain set
* Run flow with launchApp with clearKeychain: Keychain cleared

### DIFF
--- a/maestro-ios/src/main/java/ios/simctl/Simctl.kt
+++ b/maestro-ios/src/main/java/ios/simctl/Simctl.kt
@@ -9,6 +9,7 @@ import java.io.File
 object Simctl {
 
     data class SimctlError(override val message: String): Throwable(message)
+    private val homedir = System.getProperty("user.home")
 
     fun list(): SimctlList {
         val command = listOf("xcrun", "simctl", "list", "-j")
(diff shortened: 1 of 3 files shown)


maestro-ios/src/main/java/ios/simctl/Simctl.kt
```diff
@@ -200,4 +201,37 @@ object Simctl {
...
+        )
+
+        CommandLineUtils.runCommand(
+            listOf(
+                "xcrun",
+                "simctl",
+                "spawn",
+                deviceId,
+                "launchctl",
+                "start",
```
artem888, line 232: nice 👏 I guess `start` recreates /Keychains folder as well

artem888 (approved):

dmitry-zaitsev (approved):

PR #785, opened 2023-02-10
Use xcrun for uninstall

berikv:
## Testing
Tested on cli

### DIFF
--- a/maestro-ios/src/main/java/ios/simctl/Simctl.kt
+++ b/maestro-ios/src/main/java/ios/simctl/Simctl.kt
@@ -202,6 +202,18 @@ object Simctl {
         )
     }
 
+    fun uninstall(deviceId: String, bundleId: String) {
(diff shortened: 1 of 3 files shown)


maestro-ios/src/main/java/ios/simctl/Simctl.kt
```diff
@@ -200,4 +200,15 @@ object Simctl {
             )
         )
     }
+
+    fun uninstall(deviceId: String, bundleId: String) {
+        CommandLineUtils.runCommand(
+            listOf(
+                "xcrun",
+                "simctl",
+                deviceId,
```
artem888, line 209: "uninstall" seems to be missing

berikv: Good catch! Thanks!

artem888 (approved):

PR #786, opened 2023-02-10
Fix: Repeat-while-true did not work properly with JS

dmitry-zaitsev:


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -397,6 +397,8 @@ class Maestro(private val driver: Driver,
     }
 
     fun inputText(text: String) {
+        LOGGER.info("Inputting text: $text")
```
axelniklasson, line 400: Is this change intended?

axelniklasson (approved):

maestro-client/src/main/java/maestro/Maestro.kt
dmitry-zaitsev: Yes, for some reason we did not have a log message for that command

artem888 (approved):

PR #787, opened 2023-02-10
Remove hierarchy request for waitForAppToSettle (iOS)

artem888:
## Proposed Changes

If screenshot comparison was successful (static screen is detected), do not do an extra hierarchy request. This helps to optimize flow execution times, especially for RN apps with comlex view hierarchies.

Some benchmarking examples shows how this optimization helps.

SwiftUI:

Screenshot 2023-02-10 at 18 54 18

ReactNative (flow1):

Screenshot 2023-02-10 at 18 54 27

ReactNative (flow2):

Screenshot 2023-02-10 at 18 54 32

## Testing
- [x] RN app
- [x] SwiftUI app
- [x] UIKit app


maestro-client/src/main/java/maestro/Driver.kt
```diff
@@ -84,5 +84,15 @@ interface Driver {
 
     fun isUnicodeInputSupported(): Boolean
 
-    fun isScreenStatic(): Boolean
+    fun waitUntilScreenIsStatic(timeoutMs: Long): Boolean
+
+    fun waitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy?
+
+    fun performTap(
```
dmitry-zaitsev, line 91: Let's not include it into a `Driver`, it should not be concerned with that higher-level logic

dmitry-zaitsev (changes requested):
Architectural changes

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -181,7 +170,7 @@ class Maestro(private val driver: Driver,
     ) {
         LOGGER.info("Tapping on element: $element")
 
-        val hierarchyBeforeTap = waitForAppToSettle(initialHierarchy)
+        val hierarchyBeforeTap = waitForAppToSettle(initialHierarchy) ?: initialHierarchy
```
dmitry-zaitsev, line 173: We already pass `initialHierarchy` to `waitForAppToSettle`, why do we need an `?:` operator here?

maestro-client/src/main/java/maestro/utils/TapUtils.kt
```diff
@@ -0,0 +1,73 @@
+package maestro.utils
+
+import com.github.romankh3.image.comparison.ImageComparison
+import maestro.Driver
+import maestro.Point
+import maestro.ViewHierarchy
+import org.slf4j.LoggerFactory
+import java.awt.image.BufferedImage
+
+class TapUtils {
```
dmitry-zaitsev, line 10: Util classes such as this one (and actually `ScreenshotUtils` too) seem to break encapsulation and spread business logic across the codebase. Ideally, `Maestro` class should serve that role. Can we find a way to contain all that branching logic there instead?

maestro-client/src/main/java/maestro/Maestro.kt
artem888: for iOS driver might return null right now in case of screenshot comparison and function parameter is also optional

dmitry-zaitsev (approved):

PR #788, opened 2023-02-12
Introduce Action Modal

Leland-Takamine:
https://user-images.githubusercontent.com/847683/218292547-4aa5b911-7fc3-493b-81ea-0164f7821d50.mp4


dmitry-zaitsev (approved):

maestro-studio/web/src/ActionModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+        className="outline-none py-5 pl-5 pr-5 flex-1 bg-transparent min-w-[175px]"
+        placeholder="Search commands"
+        autoFocus={true}
+        value={query}
+        onChange={e => onQuery(e.target.value)}
+      />
+      <div className="flex gap-x-3 gap-y-2 pr-5 flex-wrap py-5 justify-end">
+        <Shortcut text={['\u25B2', '\u25BC']} action={"Select"} />
+        <Shortcut text={'\u2318 C'} action={"Copy"} />
+        <Shortcut text={'\u2318 D'} action={"Docs"} />
```
axelniklasson, line 54: Nit: here and in other places we should replace the macOS meta key symbol with Linux and Windows alternatives for users on those platforms.

maestro-studio/web/src/commandExample.ts
```diff
@@ -0,0 +1,125 @@
...
+        definition: {
+          id: uiElement.resourceId,
+        }
+      })
+    }
+  } else {
+    selectors.push({
+      title: 'Resource Id',
+      status: 'unavailable',
+      message: 'This element has no resource id associated with it. Type ‘\u2318 D’ to view platform-specific documentation on how to assign resource ids to ui elements.',
```
axelniklasson, line 61: Nit: can we implement an inline code snippet component instead of using backticks for `\u2318 D`?

maestro-studio/web/src/ActionModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const fuse = useMemo(() => new Fuse(unfilteredExamples, {keys: ['title', 'content']}), [unfilteredExamples])
+  const examples = useMemo(() => query ? fuse.search(query).map(r => r.item): unfilteredExamples, [fuse, unfilteredExamples, query])
+  const { focused, moveFocus, setFocus } = useFocused(examples);
+  useHotkeys('up, down', e => {
+    if (e.code === 'ArrowUp') {
+      moveFocus(-1)
+    } else if (e.code === 'ArrowDown') {
+      moveFocus(1)
+    }
+  }, {preventDefault: true, enableOnFormTags: true})
```
axelniklasson, line 185: Is there a way to add `{preventDefault: true, enableOnFormTags: true}` as options to all `useHotkeys` invocations? Otherwise we could extract it out into an object to avoid having to repeat it for all invocations here.

maestro-studio/web/src/InteractPage.tsx
```diff
@@ -70,28 +37,44 @@ const InteractPage = () => {
...
+          />
         </div>
         <div
           className="flex items-center gap-1 justify-center px-3 bg-slate-50 border-t h-10 text-slate-500 overflow-hidden whitespace-nowrap data-[error]:bg-red-100 data-[error]:text-red-800"
           data-error={footerHint ? null : replError}
         >
           {footerHint || replError}
         </div>
+        {inspectedElement && (
+          <ActionModal
```
axelniklasson, line 60: Something that I noticed when testing this out -- having to swap between mouse and keyboard when building my flow is a little tedious, so I wonder if we should also consider clicking on an element in action upload as the run action? That removes having to hit enter to confirm each command and makes for a more smooth UX in my opinion.

axelniklasson (approved):
Awesome stuff!

maestro-studio/web/src/InteractPage.tsx
Leland-Takamine: I was thinking about this as well but maybe double-click makes sense? Check out Raycast UX for instance

axelniklasson: Yeah I like that, double-click sounds great!

maestro-studio/web/src/ActionModal.tsx
Leland-Takamine: Let's tackle as a follow up if needed

PR #790, opened 2023-02-13
Refactor unsafe subprocess calls to safer ones

berikv:
## Proposed Changes

Refactor all subprocess calls which use string interpolation to pass separate arguments instead.

### DIFF
--- a/maestro-xcuitest-driver/src/main/kotlin/util/XCRunnerSimctl.kt
+++ b/maestro-xcuitest-driver/src/main/kotlin/util/XCRunnerSimctl.kt
@@ -70,7 +70,15 @@ object XCRunnerSimctl {
     }
 
     fun uninstall(bundleId: String) {
-        CommandLineUtils.runCommand("xcrun simctl uninstall booted $bundleId")
+        CommandLineUtils.runCommand(
+            listOf(
+                "xcrun",
+                "simctl",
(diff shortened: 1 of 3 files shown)


maestro-xcuitest-driver/src/main/kotlin/util/XCRunnerSimctl.kt
```diff
@@ -81,9 +89,12 @@ object XCRunnerSimctl {
 
     fun isAppAlive(bundleId: String): Boolean {
         val process = ProcessBuilder(
-            "bash",
-            "-c",
-            "xcrun simctl spawn booted launchctl list | grep $bundleId | awk '/$bundleId/ {print \$3}'"
+            "xcrun",
```
dmitry-zaitsev, line 92: Is it functionally the same as before?

berikv: The grep filters lines that contain $bundleId, awk filters so that each line only contains $bundleId.
Then two lines down, the kotlin code does:

```
val processOutput =  process.inputStream.source().buffer().readUtf8().trim()
return processOutput.contains(bundleId)
```

So yes, removing the filtering using grep and awk should functionally have no impact.

PR #792, opened 2023-02-13
Use XCTest private API for scrolling on iOS

berikv:
## Proposed Changes

Use XCTest Private API for swiping / scrolling.
The private API takes a duration and point coordinates instead of speed (in points per second) and locations delta's (0..1).
This makes much of the code simpler. Though, there is a small change in the scrolling behaviour. The XCUIElement implementation scrolls a tiny bit further than private API. I see two possible causes:

- We used to calculate velocity incorrectly (maybe velocity is in pixels vs points?)
- XCUIElement implementation does something a little different internally which results in faster scrolling

## Testing

- [x] `samples/ios-advanced-flow.yaml`
- [x] `- scroll`
- [x] `- swipe direction: UP|DOWN|LEFT|RIGHT`
- [x] `- swipe start: end:`
- [x] `- swipe from: direction`
- [x] `- swipe duration:`


maestro-driver-ios/maestro-driver-iosUITests/Routes/Models/SwipeRequest.swift
```diff
@@ -1,7 +1,31 @@
...
+
+struct SwipeRequest: Decodable {
+
+    enum CodingKeys: String, CodingKey {
+        case startX, startY, endX, endY, duration
+    }
+
+    let start: CGPoint
+    let end: CGPoint
+    let duration: Double
```
artem888, line 11: may be we could call it durationMs for better readability

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandler.swift
```diff
@@ -5,63 +5,40 @@ import os
...
+            end: requestBody.end,
+            duration: requestBody.duration)
+
+        return HTTPResponse(statusCode: .ok)
     }
-    
+
+    func swipePrivateAPI(start: CGPoint, end: CGPoint, duration: Double) async throws {
+
+        logger.info("Swiping from \(start.debugDescription) to \(end.debugDescription) with \(duration) duration")
```
artem888, line 28: nit : I think .debugDescription is not needed, because Swift will use it by default when interpolating string inside log

maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/PointerEventPath.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+struct PointerEventPath {
```
artem888, line 3: nit : better feels like class, because `let path: NSObject` reference type is used inside

artem888 (approved):
LGTM after comments addressed

maestro-driver-ios/maestro-driver-iosUITests/Routes/Models/SwipeRequest.swift
berikv: These are seconds, the units of iOS, just like Double (CGFloat) for pixel locations

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandler.swift
berikv: It is was needed because there were multiple implementations for string interpolation and the compiler couldn't decide. However, that may have been for CGVector, which I was using to test both old uielement and private api implementation.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Models/SwipeRequest.swift
artem888: may be could use TimeInterval for clarity

maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/PointerEventPath.swift
berikv: I don't agree here: The `XCPointerEventPath` would nowadays have been implemented as a struct. There is no downside of wrapping a reference type in a struct. The struct api is more safe and convenient.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandler.swift
berikv: Screenshot 2023-02-14 at 12 46 52
Meh..

maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/PointerEventPath.swift
artem888: yeah, it is just a nit, I used to prefer clean value types back then when coding in Swift (may be my knowledge is rusty :D), but there is no bad side-effects of using reference type inside, except that copying + mutating it might bring some unexpected mutations for the reference.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandler.swift
artem888: hmm, looks like smth os.log specific, works well with normal strings and print() 🤔

maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/PointerEventPath.swift
berikv: Totally agree on the surprising mutating behaviour. Updated the code to use `mutating func` where the struct is mutating the underlying value type.

dmitry-zaitsev (approved):
Sweet!

PR #797, opened 2023-02-14
Use PointerEventPath with shouldRedact:false when inserting text

berikv:
## Proposed Changes

Use PointerEventPath instead of `_XCT_sendString`.
PointerEventPath allows to set `shouldRedact:false`, which disables autocorrect on the input.
A quick test shows that a frequency of 60 characters / second is no issue.

## Testing

Ran several flows with text input

## Issues Fixed

* Long text inputs time out because the typing frequency is too slow
* autocorrect corrects the text from inputText, causing flows to fail unexpectedly


dmitry-zaitsev (approved):

artem888 (approved):
🔝

berikv:
Ran a test suite with > 10 customer flows, no errors on text input.

PR #806, opened 2023-02-17
Remove SetLocationCommand

axelniklasson:
## Proposed Changes
This removes support for `setLocation` since this command is not implemented on Android/Web and not properly supported on iOS. It does leave the iOS code implementation though so whenever we do decide to tackle this down the road we can reuse the existing implementation. 

## Testing
Tested locally, entering `setLocation` is not considered a valid command.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -567,20 +567,6 @@ data class RunFlowCommand(
 
 }
 
-data class SetLocationCommand(
-    val latitude: Double,
-    val longitude: Double,
-) : Command {
-
-    override fun description(): String {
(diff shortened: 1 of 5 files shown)


dmitry-zaitsev:
Implementation actually functions correctly on iOS (works on my machine), we could consider adding Android support if there is a popular demand for that

wollac11:
> Implementation actually functions correctly on iOS (works on my machine), we could consider adding Android support if there is a popular demand for that

I think it is a useful feature and would be pleased to see support for it on Android.

axelniklasson:
Thanks for the feedback @wollac11! Android support is on our radar and we're looking to add support for it soon and once we are able to release it I'll close this PR. cc @dmitry-zaitsev

axelniklasson:
https://github.com/mobile-dev-inc/maestro/pull/875

PR #807, opened 2023-02-17
Feat/introduce toml usage

nilsjr:
## Proposed Changes
After checking the build setup I identified some differences in versions of used dependencies.

Example:
```
File, Module, Plugin
build.gradle (maestro-android) -> "com.google.protobuf" version "0.8.18"
build.gradle (maestro-client) -> "com.google.protobuf" version "0.9.1"
```
To fix this issue I want to introduce version catalog (libs.versions.toml) so every dependencies + version is in one place. It is now easier to update & maintain all deps used in this repository.

## Testing
Tested it by running **./gradlew test** , installDist, using cli & maestro studio locally.

## Issues Fixed
see above


maestro-android/build.gradle
```diff
@@ -80,10 +83,6 @@ tasks.register("copyMaestroAndroid", Copy) {
     into layout.buildDirectory.file(maestroAndroidApkDest)
 
     it.doLast {
-        if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
```
nilsjr, line 83: Why was this a requirement?

dmitry-zaitsev: Compiling Android modules with versions of Java higher than 8 is causing bytecode incompatibility issues on certain devices. We should preserve that behavior

.github/workflows/publish-cli.yml
```diff
@@ -12,11 +12,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install JDK
+      - name: Install JDK 11
```
dmitry-zaitsev, line 15: We would need to stick with Java 8 due to requirements in our Android modules

dmitry-zaitsev:
TOML usage is a neat and welcome change! However, can we stick with Java 8 (or at least address it in a different PR)?

maestro-android/build.gradle
nilsjr: Which devices? Or do you mean a specific android api level?

.github/workflows/publish-cli.yml
nilsjr: Ok I will revert these changes and will do a follow up PR for the java 11 update. It also is required when Android Gradle Plugin is updated to 7.4.0 or higher.
[Android Gradle Plugin](https://developer.android.com/studio/releases/gradle-plugin#compatibility-7-4-0)

nilsjr:
Java 11 branch in fork created. [update-java-11-agp-7.4.1](https://github.com/nilsjr/maestro/tree/feat/update-java-11-agp-7.4.1) depends on toml change so waiting until this is merged.

dmitry-zaitsev (approved):

dmitry-zaitsev:
Thanks @nilsjr ! Everything works 👍

PR #810, opened 2023-02-17
Use in6addr any instead of loopback xctest driver

artem888:
## Proposed Changes
Changing xctest driver server to use in6addr_any instead of in6addr_loopback. This might fix some connection issues, i.e. `java.net.ConnectException: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:22087`


berikv (approved):

amanjeetsingh150:
@artem888 Connect Exceptions are mostly due to HTTP server not being reachable not sure how switching the port can affect this 🤔 Do we have some reports or issues that can validate loopback address is causing the server not reachable?

PR #815, opened 2023-02-20
Fix uninitialized jsEngine in Orchestra; Respect duration in relative swipe command

axelniklasson:
Fixes the following issues:
- Use specified duration instead of default 400 ms for relative swipe command
- Initialize `jsEngine` in `Orchestra.executeCommand` to avoid an unhandled exception thrown as part of executing repeat commands from Studio:

```
kotlin.UninitializedPropertyAccessException: lateinit property currentScope has not been initialized
```

## Testing
This has been tested locally and I also update test case `078` to have a correct assertion.

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -2143,7 +2143,7 @@ class IntegrationTest {
         val expectedStart = Point(deviceInfo.widthGrid / 2, deviceInfo.heightGrid * 30 / 100)
         val expectedEnd = Point(deviceInfo.widthGrid / 2, deviceInfo.heightGrid * 60 / 100)
         driver.assertHasEvent(
-            Event.Swipe(start = expectedStart, End = expectedEnd, durationMs = 400)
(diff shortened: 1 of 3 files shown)


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -2143,7 +2143,7 @@ class IntegrationTest {
         val expectedStart = Point(deviceInfo.widthGrid / 2, deviceInfo.heightGrid * 30 / 100)
         val expectedEnd = Point(deviceInfo.widthGrid / 2, deviceInfo.heightGrid * 60 / 100)
         driver.assertHasEvent(
-            Event.Swipe(start = expectedStart, End = expectedEnd, durationMs = 400)
+            Event.Swipe(start = expectedStart, End = expectedEnd, durationMs = 3000)
```
amanjeetsingh150, line 2146: What is the intent behind changing this default duration for relative swipe?

amanjeetsingh150 (approved):

amanjeetsingh150:
Can you also mention this PR to the issues it's going to fix?

PR #817, opened 2023-02-20
Private swipe

berikv:
## Proposed Changes

Use XCTest private API to swipe / scroll.

Note: This is a patch-up of [#792](https://github.com/mobile-dev-inc/maestro/pull/792)

## Testing

Ran 14 different flows from various apps.

## Issues Fixed

Apps which have big view hierarchies would cause an error in XCTest.framework / IDB' scroll commands.
By using XCTest.framework's private API we can avoid fetching the apps view hierarchy for scroll commands; avoiding the errors.


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -100,7 +99,6 @@ class IOSDriver(
             .getOrThrow {
                 MaestroException.UnableToLaunchApp("Unable to launch app $appId ${it.message}")
             }
-        waitForAppToSettle(null)
```
dmitry-zaitsev, line 103: I believe this is required

berikv: I've added this line to try and solve the scrolling issue. Now that it's solved a different way, the line can go.
It is more efficient to wait Before running a command when needed instead of waiting after a command because some commands don't need the view to be in quiescence state.

maestro-client/src/test/java/maestro/PointTest.kt
```diff
@@ -22,21 +22,4 @@ internal class PointTest {
         // Then
         assertThat(point).isEqualTo(Point(1, 2))
     }
-
-    @Test
-    internal fun `Deserialize PointF`() {
```
dmitry-zaitsev, line 27: I think it should be fine to remove `PointF`, but heads up that there was some issue in Cloud that made me to add this test. Unfortunately I no longer remember what the issue was :)

dmitry-zaitsev (approved):

maestro-client/src/test/java/maestro/PointTest.kt
berikv: The way I see this is that IOSDriver.kt is responsible for mapping the maestro world into the iOS world.
iOS uses Double for its pixel locations and time intervals, and iOSDriver maps the Int pixel values directly to Double without a Float in between. 

The denormalizedDistance transformation is moved into `width|heightPercentToPoint`. Thus the PointF is now unused:
```
val denormalizedDistance = PointF(start.x * width, start.y * height)
             .distance(PointF(end.x * width, end.y * height))
```

PR #823, opened 2023-02-21
[MOB 1869] Auto verify opening deeplinks on Android

amanjeetsingh150:
## Proposed Changes

1. Add openBrowser command separately for opening the browser for https/http link. Sometimes different apps intercept these links which cause unexpected behavior.
2. Add autoVerify flag on openLink command which will optionally auto verify opening HTTP links to user app and also agree the chrome agreement.

```
appId: com.chatloop.my
---
- openLink: 
    link: https://something.com/linking
    autoVerify: true
```

By default this flag is false.

## Testing
Attaching internally.

## Issues Fixed

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1676539339478699

### DIFF
--- a/maestro-client/build.gradle
+++ b/maestro-client/build.gradle
@@ -71,6 +71,7 @@ dependencies {
     api(libs.jackson.module.kotlin)
     api(libs.jackson.dataformat.yaml)
     api(libs.jackson.dataformat.xml)
+    api(libs.apk.parser)
 
     implementation project(':maestro-ios')
     implementation(libs.google.findbugs)
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -25,8 +25,8 @@ import maestro.Filters.asFilter
 import maestro.UiElement.Companion.toUiElementOrNull
 import maestro.drivers.AndroidDriver
 import maestro.drivers.WebDriver
-import maestro.utils.ScreenshotUtils
 import maestro.utils.MaestroTimer
+import maestro.utils.ScreenshotUtils
 import maestro.utils.SocketUtils
 import okio.Sink
 import okio.buffer
@@ -423,23 +423,11 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         waitForAppToSettle()
     }
 
-    fun openLink(link: String, autoVerify: Boolean) {
-        driver.openLink(link, autoVerify)
+    fun openLink(link: String, appId: String?, autoVerify: Boolean) {
+        LOGGER.info("Opening link $link for app: $appId with autoVerify config as $autoVerify")
+
+        driver.openLink(link, appId, autoVerify)
         waitForAppToSettle()
-    }
-
-    fun autoVerifyApp(appId: String?) {
-        if (driver is AndroidDriver) {
-            driver.autoVerifyApp(appId)
-            waitForAppToSettle()
-        }
-    }
-
-    fun autoVerifyChromeAgreement(verifyGoogleChromeAgreement: () -> Unit) {
-        if (driver is AndroidDriver) {
-            verifyGoogleChromeAgreement()
-            waitForAppToSettle()
-        }
     }
 
     fun openBrowser(link: String) {
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -423,15 +423,10 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         waitForAppToSettle()
     }
 
-    fun openLink(link: String, appId: String?, autoVerify: Boolean) {
+    fun openLink(link: String, appId: String?, autoVerify: Boolean, browser: Boolean) {
         LOGGER.info("Opening link $link for app: $appId with autoVerify config as $autoVerify")
 
-        driver.openLink(link, appId, autoVerify)
-        waitForAppToSettle()
-    }
-
-    fun openBrowser(link: String) {
-        driver.openBrowser(link)
+        driver.openLink(link, appId, autoVerify, browser)
         waitForAppToSettle()
     }
 
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -24,7 +24,6 @@ import dadb.Dadb
 import maestro.Filters.asFilter
 import maestro.UiElement.Companion.toUiElementOrNull
 import maestro.drivers.AndroidDriver
-import maestro.drivers.IOSDriver
 import maestro.drivers.WebDriver
 import maestro.utils.ScreenshotUtils
 import maestro.utils.MaestroTimer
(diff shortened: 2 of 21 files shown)


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -424,8 +424,27 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
 
-    fun openLink(link: String) {
-        driver.openLink(link)
+    fun openLink(link: String, autoVerify: Boolean) {
+        driver.openLink(link, autoVerify)
+        waitForAppToSettle()
+    }
+
+    fun autoVerifyApp(appId: String?, autoVerifyApp: (text: String) -> Unit) {
+        if (driver is AndroidDriver) {
```
dmitry-zaitsev, line 433: Let's try to avoid any OS-specific checks at the top level, otherwise it will get out of hand very quickly. Instead, you can add a no-op implementation of `autoVerifyApp` to other drivers

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -424,8 +424,27 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+            driver.autoVerifyApp(appId, autoVerifyApp)
+            waitForAppToSettle()
+        }
+    }
+
+    fun autoVerifyChromeAgreement(verifyGoogleChromeAgreement: () -> Unit) {
```
dmitry-zaitsev, line 439: This top-level method is:

- Leaking OS-specific details (it is Android-only)
- Is always called in tandem with `autoVerifyApp`

I think it can simply be a part of the `autoVerifyApp` method

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -396,10 +397,38 @@ class AndroidDriver(
...
     }
 
-    override fun openLink(link: String) {
+    override fun openLink(link: String, autoVerify: Boolean) {
         dadb.shell("am start -a android.intent.action.VIEW -d \"$link\"")
     }
 
+    fun autoVerifyApp(appId: String?, autoVerifyApp: (text: String) -> Unit) {
+        if (appId != null) {
+            val apkFile = AndroidAppFiles.getApkFile(dadb, appId)
```
dmitry-zaitsev, line 406: Just to double check - does this work on real devices?

dmitry-zaitsev:
In what cases would a user want to set `autoVerify` to `false`?

dmitry-zaitsev (changes requested):
Comments

maestro-client/src/main/java/maestro/Maestro.kt
amanjeetsingh150: Done 💯  I have done following changes:

1. Change autoVerify method to private 
2. Used autoVerify flag passed in openLink method.

This will avoid adding explicit method for this in driver.

maestro-client/src/main/java/maestro/Maestro.kt
amanjeetsingh150: Done added this in AndroidDriver. This is no more leaking in Maestro or Orchestra class 💯

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
amanjeetsingh150: Yes I checked on my real device and was able to get app name. Should we add a try-catch around it and not fail this if somehow this throws an exception? WDYT?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -518,8 +521,65 @@ class Orchestra(
         return true
     }
 
-    private fun openLinkCommand(command: OpenLinkCommand): Boolean {
-        maestro.openLink(command.link)
+    private fun openLinkCommand(command: OpenLinkCommand, config: MaestroConfig?): Boolean {
+        maestro.openLink(command.link, command.autoVerify ?: false)
+
+        if (command.autoVerify == true) {
+            val autoVerifyApp: (appName: String) -> Unit = {
```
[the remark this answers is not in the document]
amanjeetsingh150: Done moved explicitly in `AndroidDriver`

amanjeetsingh150:
> In what cases would a user want to set autoVerify to false?

1. This flag defaults to false
2. If user wants to add testing between the interaction of the system dialog then it would be false. For example if dialog shows the text in different locale.

For the users who don't want to write the interaction with dialog can add autoVerify as true.


Btw do you think if we fail with ElementNotFound after openLink command we should add this in error message that try using autoVerify flag to automatically verify your app for the link passed?

maestro-client/build.gradle
```diff
@@ -71,6 +71,7 @@ dependencies {
     api(libs.jackson.module.kotlin)
     api(libs.jackson.dataformat.yaml)
     api(libs.jackson.dataformat.xml)
+    api("net.dongliu:apk-parser:2.6.10")
```
amanjeetsingh150, line 74: @dmitry-zaitsev I was not able to reference my dependency here by `api(libs.apkParser)`. Am I missing something?

dmitry-zaitsev: I believe it should be `api(libs.apk.parser)`

dmitry-zaitsev:
> Btw do you think if we fail with ElementNotFound after openLink command we should add this in error message that try using autoVerify flag to automatically verify your app for the link passed?

If I understand your point correctly, I think we should not throw `ElementNotFound` nor fail the test if auto-verify fails to find an element. We don't know the locale of user's device nor can we react to update in OS/Browser fast enough - hence we should do auto-verification as a best-effort attempt

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -424,8 +423,15 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
-        driver.openLink(link)
+    fun openLink(link: String, appId: String?, autoVerify: Boolean) {
+        LOGGER.info("Opening link $link for app: $appId with autoVerify config as $autoVerify")
+
+        driver.openLink(link, appId, autoVerify)
+        waitForAppToSettle()
+    }
+
+    fun openBrowser(link: String) {
+        driver.openBrowser(link)
```
dmitry-zaitsev, line 434: nit: add a log message, similarly to `openLink`

dmitry-zaitsev (approved):
Looks good! Let's just make sure to use a version from toml file

berikv:
I'm a bit late to the discussion here, forgive me if this was already answered.
Wouldn't it be better to specify the 'browserness' of the link opening as an extra parameter to the openLink instruction?
Eg:
```
- openLink: "https://wikipedia.org"
    browser: true # or "com.android.chrome"
```

On iOS there is no difference between openLink and openBrowser. Having a separate command do exactly the same is not great. Both for understanding what happens and for documentation.
Also I believe pattern can be reused for other "open this or that" methods, preventing a whole lot of commands like `- openInAppleMaps`.

amanjeetsingh150:
Hey @berikv thanks for the suggestion, I am open to both. However, if I think as a user I think the more explicit the command the better. openBrowser seems to be explicit as its altogether a different command with a browser true attribute its also explicit but it can confuse me  by the name since "browser" is not a behaviour which needs true or false. I could be biased by the implementation though open to change to anything better.


cc: @dmitry-zaitsev WDYT?

maestro-client/src/main/java/maestro/Maestro.kt
amanjeetsingh150: Updated to add browser attribute

PR #828, opened 2023-02-22
Add quotes around string values; Better dark mode support for ReplView

axelniklasson:
## Proposed Changes
- Add proper dark mode support for play icon in `ReplView`
- All text values will now have double quotes around them to avoid values such as `yes`, `no`, `off`, `on` to be interpreted as booleans as defined in the [YAML 1.1 spec](https://yaml.org/type/bool.html).

**Screenshot**
See the `- tapOn: "25h 22m 22s"` command

| Before  | After |
| ------------- | ------------- |
|  before |  after |

## Testing
Tested locally.

### DIFF
--- a/maestro-studio/web/src/ReplView.tsx
+++ b/maestro-studio/web/src/ReplView.tsx
@@ -386,13 +386,13 @@ const ReplView = ({input, onInput, onError}: {
           }}
         >
           <AutosizingTextArea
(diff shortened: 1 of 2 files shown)


berikv (approved):

axelniklasson:
Will tackle this in another way, closing this PR

PR #835, opened 2023-02-23
WIP: Add custom YAML parser to allow parsing boolean-ish words

axelniklasson:
## Issues Fixed


axelniklasson:
Note: [landed a fix in jackson-dataformats-text](https://github.com/FasterXML/jackson-dataformats-text/pull/389) instead so once that ships I'll update maestro to use that instead of rolling our own custom parser.

PR #838, opened 2023-02-23
Simplify making json calls from XCTestDriverClient

berikv:
## Proposed Changes

Refactor in XCTestDriverClient to simplify json requests and reduce repetition.

### DIFF
--- a/maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -122,19 +122,7 @@ class XCTestDriverClient(
     }
 
     fun runningAppId(appIds: Set<String>): Response {
-        val mediaType = "application/json; charset=utf-8".toMediaType()
-        val appIdsRequest = GetRunningAppRequest(appIds)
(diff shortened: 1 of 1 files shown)


maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -145,71 +133,29 @@ class XCTestDriverClient(
...
             startX = startX,
             startY = startY,
             endX = endX,
             endY = endY,
             duration = duration
-        )
-        val body = mapper.writeValueAsString(request).toRequestBody(mediaType)
-
-        val url = xctestAPIBuilder("swipe")
-            .addQueryParameter("appId", appId)
```
igorsmotto, line 159: Isn't this necessary anymore?

igorsmotto (approved):

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
berikv: Wipe does not need the appId argument indeed. See https://github.com/mobile-dev-inc/maestro/blob/main/maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandler.swift

PR #840, opened 2023-02-24
Use textInput for enter pressKey

berikv:
## Proposed Changes

Use `inputText "\n"` for `keyPress: enter` on iOS

## Issues Fixed

### DIFF
--- a/maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
@@ -91,7 +91,9 @@ class XCTestIOSDevice(
     }
 
     override fun pressKey(code: Int): Result<Unit, Throwable> {
-        error("Not supported")
+        return runCatching {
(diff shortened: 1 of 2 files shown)


maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
```diff
@@ -91,7 +91,9 @@ class XCTestIOSDevice(
     }
 
     override fun pressKey(code: Int): Result<Unit, Throwable> {
-        error("Not supported")
+        return runCatching {
+            client.inputText("\n")
+        }
```
igorsmotto, line 96: Should return an error if code != 40?

igorsmotto (approved):

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
berikv: Thanks for the suggestion, that would be better.

Issue #133, opened 2022-09-06
[Feature] Extend `launchApp` to allow permissions


wadey1:
eg:
```
- launchApp:
      clearState: true
      permissions: { notifications: 'YES', location: 'always' }
```

dmitry-zaitsev:
For now, our recommended approach is to use optional commands to handle that:

```
# Tap on Allow on permission dialogue if one shows up
- tapOn:
     text: Allow
     optional: true
```

berikv:
Hi @wadey1, maestro version 1.24.0 includes the change you requested. Can you give it a try and let us know what you think?

PR #849, opened 2023-03-01
Support for permission parameter on launchApp

berikv:
## Proposed Changes

Allow to set permissions in `launchApp`:
```
- launchApp:
      permissions: { notifications: deny }
```

## Testing

Tested on customer app flows.

## Issues Fixed

* https://github.com/mobile-dev-inc/maestro/issues/133
* There is a difference in permission handling on maestro cloud vs maestro cli on iOS, this PR is a step to close that gap

### DIFF
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -89,5 +89,6 @@ interface Driver {
     fun waitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy?
 
     fun capabilities(): List<Capability>
+
     fun setPermissions(appId: String, permissions: Map<String, String>)
 }
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -89,4 +89,5 @@ interface Driver {
     fun waitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy?
 
     fun capabilities(): List<Capability>
+    fun setPermissions(appId: String, permissions: Map<String, String>)
 }
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -41,8 +41,8 @@ import maestro.ViewHierarchy
 import maestro.android.AndroidAppFiles
 import maestro.android.asManifest
 import maestro.android.resolveLauncherActivity
-import maestro.utils.ScreenshotUtils
 import maestro.utils.MaestroTimer
+import maestro.utils.ScreenshotUtils
 import maestro.utils.StringUtils.toRegexSafe
 import maestro_android.MaestroDriverGrpc
 import maestro_android.deviceInfoRequest
@@ -530,23 +530,23 @@ class AndroidDriver(
     }
 
     private fun setAllPermissions(appId: String, value: String) {
-        val argument = argumentForPermissionValue(value)
-        shell("pm $argument -g $appId")
+        val translated = translatePermissionValue(value)
+        shell("pm $translated -g $appId")
     }
 
     private fun setPermission(appId: String, permission: String, value: String) {
         val name = permission.replace("[^A-Za-z0-9._]+".toRegex(), "")
-        val argument = argumentForPermissionValue(value)
+        val translated = translatePermissionValue(value)
 
-        shell("pm $argument $appId $name")
+        shell("pm $translated $appId $name")
     }
 
-    private fun argumentForPermissionValue(value: String): String {
+    private fun translatePermissionValue(value: String): String {
         return when (value) {
             "allow" -> "grant"
             "deny" -> "revoke"
             "unset" -> "revoke"
-            else -> throw IllegalArgumentException("Permission 'all' can be set to 'allow', 'deny' or 'unset', not '$value'")
+            else -> throw IllegalArgumentException("Permissions can be set to 'allow', 'deny' or 'unset' on Android, not '$value'")
         }
     }
 
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -525,20 +525,77 @@ class AndroidDriver(
         }
 
         mutable.forEach { permission ->
-            setPermission(appId, permission.key, permission.value)
+            val permissionValue = translatePermissionValue(permission.value)
+            translatePermissionName(permission.key).forEach { permissionName ->
+                setPermissionInternal(appId, permissionName, permissionValue)
+            }
         }
     }
 
-    private fun setAllPermissions(appId: String, value: String) {
-        val translated = translatePermissionValue(value)
-        shell("pm $translated -g $appId")
+    private fun setAllPermissions(appId: String, permissionValue: String) {
+        val permissionsResult = runCatching {
+            val apkFile = AndroidAppFiles.getApkFile(dadb, appId)
+            ApkFile(apkFile).apkMeta.usesPermissions
+        }
+        if (permissionsResult.isSuccess) {
+            permissionsResult.getOrNull()?.let {
+                it.forEach { permission ->
+                    setPermissionInternal(appId, permission, translatePermissionValue(permissionValue))
+                }
+            }
+        }
     }
 
-    private fun setPermission(appId: String, permission: String, value: String) {
-        val name = permission.replace("[^A-Za-z0-9._]+".toRegex(), "")
-        val translated = translatePermissionValue(value)
+    private fun setPermissionInternal(appId: String, permission: String, permissionValue: String) {
+        try {
+            dadb.shell("pm $permissionValue $appId $permission")
+        } catch (exception: Exception) {
+            /* no-op */
+        }
+    }
 
-        shell("pm $translated $appId $name")
+    private fun translatePermissionName(name: String): List<String> {
+        return when (name) {
+            "location" -> listOf(
+                "android.permission.ACCESS_FINE_LOCATION",
+                "android.permission.ACCESS_COARSE_LOCATION",
+            )
+            "camera" -> listOf("android.permission.CAMERA")
+            "contacts" -> listOf(
+                "android.permission.READ_CONTACTS",
+                "android.permission.WRITE_CONTACTS"
+            )
+            "phone" -> listOf(
+                "android.permission.CALL_PHONE",
+                "android.permission.ANSWER_PHONE_CALLS",
+            )
+            "microphone" -> listOf(
+                "android.permission.RECORD_AUDIO"
+            )
+            "bluetooth" -> listOf(
+                "android.permission.BLUETOOTH_CONNECT",
+                "android.permission.BLUETOOTH_SCAN",
+            )
+            "storage" -> listOf(
+                "android.permission.WRITE_EXTERNAL_STORAGE",
+                "android.permission.READ_EXTERNAL_STORAGE"
+            )
+            "notifications" -> listOf(
+                "android.permission.POST_NOTIFICATIONS"
+            )
+            "medialibrary" -> listOf(
+                "android.permission.WRITE_EXTERNAL_STORAGE",
+                "android.permission.READ_EXTERNAL_STORAGE",
+                "android.permission.READ_MEDIA_AUDIO",
+                "android.permission.READ_MEDIA_IMAGES",
+                "android.permission.READ_MEDIA_VIDEO"
+            )
+            "calendar" -> listOf(
+                "android.permission.WRITE_CALENDAR",
+                "android.permission.READ_CALENDAR"
+            )
+            else -> listOf(name.replace("[^A-Za-z0-9._]+".toRegex(), ""))
+        }
     }
 
     private fun translatePermissionValue(value: String): String {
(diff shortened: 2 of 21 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -267,6 +267,9 @@ class Orchestra(
 
     private fun clearAppStateCommand(command: ClearStateCommand): Boolean {
         maestro.clearAppState(command.appId)
+        // Android's clear command also resets permissions
+        // Reset all permissions to unset so both platforms behave the same
+        maestro.setPermissions(command.appId, mapOf("all" to "unset"))
```
amanjeetsingh150, line 272: On Android call to clearAppState will bring back the permissions to unset case by default. Isn't that the for iOS?

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
```diff
@@ -358,6 +357,12 @@ class FakeDriver : Driver {
         )
     }
 
+    override fun setPermissions(appId: String, permissions: Map<String, String>) {
+        ensureOpen()
+
+        events.add(Event.SetPermissions(appId, permissions))
```
amanjeetsingh150, line 363: Can we add IntegrationTest for this?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -518,6 +518,38 @@ class AndroidDriver(
...
+    private fun setAllPermissions(appId: String, value: String) {
+        val translated = translatePermissionValue(value)
+        shell("pm $translated -g $appId")
+    }
+
+    private fun setPermission(appId: String, permission: String, value: String) {
+        val name = permission.replace("[^A-Za-z0-9._]+".toRegex(), "")
+        val translated = translatePermissionValue(value)
+
+        shell("pm $translated $appId $name")
```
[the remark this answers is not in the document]
berikv: I'm not sure what you're asking?

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -123,6 +121,11 @@ class SimctlIOSDevice(
         TODO("Not yet implemented")
     }
 
+    override fun setPermissions(id: String, permissions: Map<String, String>) {
+        Simctl.setPermissions(deviceId, id, permissions)
+        xcTestDevice.restartXCTestRunnerService()
```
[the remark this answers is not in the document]
berikv: Because it is killed

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
berikv: Though it is a no-change for Android; it's good to be explicit about what the expected result is by setting it for both platforms instead of relying on platform specific behaviours. If the setPermission call takes a long time on Android, then we better only call it on iOS.

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
berikv: Sure, but lets first make sure that the direction is approved and the code works on both platforms.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -518,6 +518,38 @@ class AndroidDriver(
...
+        }
+
+        mutable.forEach { permission ->
+            setPermission(appId, permission.key, permission.value)
+        }
+    }
+
+    private fun setAllPermissions(appId: String, value: String) {
+        val translated = translatePermissionValue(value)
+        shell("pm $translated -g $appId")
```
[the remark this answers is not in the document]
berikv: Here's someone suggesting this command to work: https://www.reddit.com/r/NetHunter/comments/thcxse/comment/i1h034s/?utm_source=reddit&utm_medium=web2x&context=3

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
amanjeetsingh150: Ohh interesting, yes I remember in cloud we needed to restart idb as well can you check if idb operations work as well after this ?

dmitry-zaitsev (approved):
Generally the change makes sense to me, however I defer to @amanjeetsingh150 for the final approval as he has more knowledge on the topic than I do.

Also, could you please add an integration test? That would be a good demonstration of how API looks like from user's standpoint

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
berikv: Good point, I'll test this

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -133,4 +133,8 @@ class LocalIOSDevice(
     override fun isScreenStatic(): Result<Boolean, Throwable> {
         return xcTestDevice.isScreenStatic()
     }
+
+    override fun setPermissions(id: String, permissions: Map<String, String>) {
+        simctlIOSDevice.setPermissions(id, permissions)
```
amanjeetsingh150, line 138: What will happen if a user gives following input:

```
- launchApp:
      permissions: { notifications: deny , all: allow }
```

maestro-client/src/main/java/maestro/Driver.kt
```diff
@@ -89,4 +89,6 @@ interface Driver {
     fun waitForAppToSettle(initialHierarchy: ViewHierarchy?): ViewHierarchy?
 
     fun capabilities(): List<Capability>
+
+    fun setPermissions(appId: String, permissions: Map<String, String>)
```
amanjeetsingh150, line 93: What are all possible values for permissions, since it could be different for iOS and android?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -518,6 +518,38 @@ class AndroidDriver(
         )
     }
 
+    override fun setPermissions(appId: String, permissions: Map<String, String>) {
+        val mutable = permissions.toMutableMap()
```
amanjeetsingh150, line 522: Launch app fails due to this, creating a fix for this. Let's defer this PR until then

amanjeetsingh150, line 522: Launch app fails due to this, creating a fix for this. Let's defer this PR until then

amanjeetsingh150 (changes requested):

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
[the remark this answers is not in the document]
amanjeetsingh150: What will be the value of `name` here?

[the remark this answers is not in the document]
berikv: Please see the implementation of `translatePermissionValue`. It is either "grant" or "revoke".

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
berikv: It is expected that all permissions are set to allow except for the notification permission.

maestro-client/src/main/java/maestro/Driver.kt
berikv: The values are platform specific. On iOS the list is short and mostly static.

On android the list is quite big, which also makes the list more dynamic. Also I believe that apps on Android can only ask for permissions that they've listed in their manifest, which we can read out with adb. This means we can make "all permissions" mean "all requested permissions" on android, which avoids hard-coding a separate list of "all permissions" that exist on Android.

berikv: Also, I think the current implementation of `adb pm grant -g <appi>` does exactly that.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
berikv: Does it fail on creating a mutable map? That's unexpected..

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
berikv: Tested, idb works fine after `setPermissions`

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
berikv: Validated that the following command on iOS will deny notification permissions, but allow the rest
```
- launchApp:
      permissions: { notifications: deny, all: allow }
```

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
amanjeetsingh150: Nope, sorry pointing at wrong line. While setting permission in the line below, no worries I am going to raise a fix for that very soon against this PR.

maestro-client/src/main/java/maestro/Driver.kt
berikv: Aha, that's unfortunate.

amanjeetsingh150: This PR is going to fix it: https://github.com/mobile-dev-inc/maestro/pull/864

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
berikv: Not every flow has a launchApp command, so that can only be done by first inspecting if the flow starts with a launchApp command. Which is not easy because the first command could be a runFlow or a runScript command, etc.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
berikv: Are there any open action items for me on this PR now that android works?

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
berikv: I'll add the integration test now.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
amanjeetsingh150: Integration tests

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
amanjeetsingh150: This is future improvement but I see it as follows:

1. Whenever `setPermissions` is called in test initially with `launchApp` we run the xctest runner after that
2. Subsequently, after that, all the `setPermissions` calls will not install/run xctest runner again

This will avoid installing xctest runner multiple times, right now its done in open as well as again everytime setPermission is called as part of launchApp.

berikv:
> Should we do setPermission with openLink command since that is also another way of opening any app?

The issue here is that the appId might not be available.
I think we should add a top-level setPermissions command, that could be useful on iOS in very specific cases, and it provides a way to set the permissions when the launch command is done via openLink.
Though, first I'd like to see how this setup will be used. openLink with permissions is already quite an edge case, and there is an easy workaround available: first launchApp with the correct permission settings, then stopApp then openLink.

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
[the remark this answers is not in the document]
berikv: We should definitely look into speeding up the startup steps we do here. I've measured, this change introduces ~1.5 extra time on a flow start with launchApp with clearState. I think that is acceptable. Speed improvements are always welcome..

maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
berikv: Integration tests don't cover the code with logic in this PR. I've added one, but its not really testing anything.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
berikv: Done (but, see other thread)

maestro-test/src/test/resources/088_launchApp_with_all_permissions_to_deny_and_notification_to_allow.yaml
```diff
@@ -0,0 +1,6 @@
+appId: com.example.app
+---
+- launchApp:
+    permissions:
+      all: deny
+      notifications: allow
```
amanjeetsingh150, line 6: Shouldn't this be: `permissions: { all: deny, notifications: allow }`?

amanjeetsingh150 (approved):

maestro-test/src/test/resources/088_launchApp_with_all_permissions_to_deny_and_notification_to_allow.yaml
berikv: You can write it either way..

PR #856, opened 2023-03-03
Experimental feature: Natural language commands

dmitry-zaitsev:
## Proposed Changes

Allowing certain commands to be entered in a natural English language within `yaml` files. For example:

```
appId: com.example.myapp
---
- Launch app
- Tap on Continue
- Tap on Email
- Input user@example.com
- Tap on Password
- Enter 123456
- User should see Home
- Scroll 3 times
- Assert that Checkbox is checked
```

### Supported commands (and some examples)

- `launchApp` 
  - `Launch app`
  - `Open app`
  - `Launch com.example.someBundleId`
- `tapOn` 
  - `Tap on a Button`
  - `Press Button`
  - `Click Button`
- `swipe` (swiping from a view is not yet supported)
  - `Swipe`
  - `Swipe left` (or right, top, bottom)
  - `Swipe from right to left`
  - `Swipe from top to bottom`
- `scroll`
  - `Scroll`
  - `Scroll up` (or down)
  - `Scroll left` (or right)
- `repeat` (only `times` is supported, conditions are not yet supported)
  - `Tap on Button 3 times`
  - `Swipe left 1 time`
- `assertVisible` / `assertNotVisible` (including checks on `checked`, `selected`, `enabled`, `focused` properties)
  - `Assert that Hello World is visible`
  - `User should see a Button`
  - `Ensure My Checkbox is checked`
  - `Button should not be visible`
- `pressKey`
  - `Press Enter`
- `back`
  - `Go back`
- `eraseText`
  - `Clear text`
  - `Erase field`
  - `Delete input`

### Limitations

- JS usage is not supported in natural-language commands
- `scrollUntilVisible` is not yet supported
- `swipe` from an element is not yet supported
- `repeat` until condition is not yet supported
- `runFlow` and `runScript` are not yet supported

Note that the aforementioned unsupported commands can still be mixed in using a conventional syntax. For example:

```
- Tap on a Button
- runFlow: myFlow.yaml
```


marckrenn:
While I think this is a great initiative, these are "just" some synonyms with fixed syntax, right? So I'm not quite sure if it really makes the life of testers so much easier, as they'd still have to look those up in the docs.

So just as a food for thought:

Wouldn't this be a great example for applying a fine-tuned / few-shot GPT3.5(turbo) or custom LM to preprocess any arbitrary command – be it in the style of natural language paragraphs or a lists – into valid maestro yamls?

If you'd go with the OpenAI API, an API key could be provided by the user so there'd be no additional cost on your side. Plus, this feature could be tacked onto Maestro Studio ("Convert text to test"), that would require no additional code changes downstream in the Maestro codebase.

Doing so, it would also have the benefit of having some sort of validation GUI for the testers, to check if the LM didn't misinterpret their prompt.

If you're interested in that approach, I think a prototype can be spooled up rather quickly.

dmitry-zaitsev:
@marckrenn thanks for the detailed write up!

I was considering going Open API route initially, but decided to settle on pattern-matching approach for now just to validate whether the feature will have its use. If the value is proven _and_ pattern-matching would feel limited - we are going to explore ML models to do the job. Though if you have some interesting ideas on what kind of features would be possible with GPT - I would be curious to know!

Note that the primary audience for this kind of feature are not so much testers but rather people who are not always deeply technical (i.e. product managers/owners). This feature offers a way to potentially write tests in natural language _before_ a feature is even implemented. 

With that in mind, I think for engineering and QA teams an existing syntax offers the most flexibility and in combination with Studio is highly unlikely to be replaced by NLP.

amanjeetsingh150:
> validate whether the feature will have its use

Do we plan to add some analytics to validate how the feature is performing in practical use? How much people are using it? Who all use it, so that we can plan to improve this further as you suggested.

I remember we added ability to add analytics in our CLI

dmitry-zaitsev:
We decided to not proceed with this feature in its current form yet

PR #875, opened 2023-03-08
Feature: setLocation on Android

dmitry-zaitsev:
Making `setLocation` work on Android


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -486,7 +487,14 @@ class AndroidDriver(
         .map { parts: Array<String> -> parts[1] }
 
     override fun setLocation(latitude: Double, longitude: Double) {
```
axelniklasson, line 489: Not sure if it is easy to do, but could we also add an integration test for this?

axelniklasson (approved):
Awesome stuff Dima, thanks for handling this!

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
dmitry-zaitsev: There is already one, `051`

maestro-android/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:tools="http://schemas.android.com/tools"
+          xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <uses-permission android:name="android.permission.ACCESS_MOCK_LOCATION"
+                     tools:ignore="MockLocation,ProtectedPermissions"/>
```
amanjeetsingh150, line 6: Does this permission get allowed on installed or its dangerous?

Also from what I know for mocking location we have to switch the mock app in developer options, are we doing that with some command?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -196,6 +206,66 @@ class Service(
...
+                }
+                locMgr.setTestProviderStatus(
+                    LocationManager.GPS_PROVIDER,
+                    LocationProvider.AVAILABLE,
+                    null, System.currentTimeMillis()
+                )
+
+                locMgr.setTestProviderLocation(LocationManager.GPS_PROVIDER, newLocation)
+
+                postLocation()
```
amanjeetsingh150, line 259: This seems like a recursive call, what is the terminal condition of this recursion 🤔 ?

maestro-android/src/androidTest/AndroidManifest.xml
dmitry-zaitsev: > Does this permission get allowed on installed or its dangerous?

It appears to be granted automatically, didn't see any issue with that.

> Also from what I know for mocking location we have to switch the mock app in developer options, are we doing that with some command?

Correct, we do that via automatically via `dadb` in `AndroidDriver`

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
dmitry-zaitsev: The terminal condition is `locationCounter != version`. Note that this is a non-blocking recursion since we are using `Handler` to add message to the end of the message queue

dmitry-zaitsev: Also, worth noting, it is expected that the recursion loop will continue indefinitely to make sure the location remains consistent and is not overwritten by other location providers (as I have observed without the loop)

PR #878, opened 2023-03-08
Feature: Travel command

dmitry-zaitsev:
## Proposed Changes

`travel` command that takes a list of points and speed (in m/s) and then navigates user through the route

https://user-images.githubusercontent.com/2990722/223780598-d7354170-a3ed-4f67-bb2a-6997a14025d8.mov


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -702,3 +702,40 @@ data class MockNetworkCommand(
...
+    val points: List<GeoPoint>,
+    val speedMPS: Double? = null,
+) : Command {
+
+    data class GeoPoint(
+        val latitude: Double,
+        val longitude: Double,
+    ) {
+
+        fun getDistanceInMeters(another: GeoPoint): Double {
```
dmitry-zaitsev, line 716: Fun fact - this function is entirely written by ChatGPT

ArthurSav:
Great addition Dima. I'm a bit sceptical on the name. Perhaps something with `location` would be more suitable

dmitry-zaitsev:
@ArthurSav on the contrary, I think the name is nearly perfect since it captures the whole concept in just 1 word that is also a verb. I struggle to find a better alternative and certainly think that it is more natural-sounding than any variations of `updateLocation`, `traverseLocations`, etc.

ArthurSav (approved):

PR #882, opened 2023-03-09
Auto-restart xctest runner and idb_companion on connection error

berikv:
## Proposed Changes

When a connection error occurs on either xctest runner or on idb_companion, restart that specific service.

## Testing

Early testing looks promising, more testing is needed.
Tested by:

- Adjusting LocalIOSDevice.tap to use IdbIOSDevice.tap
- Running `./maestro studio` in one terminal
- Running `./maestro test samples/ios-flow.yaml` several times in another terminal
- Executing a tapOn command in the still open studio

## Issues Fixed

launchApp would cause idb_companion to shut down due to its internal setPermissions call. idb_companion is needed as a fallback for getViewHierarchy for ReactNative (and other apps with big view hierarchies).


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -309,13 +308,18 @@ object MaestroSessionManager {
...
-            if (SessionStore.activeSessions().isNotEmpty()) {
-                IdbInstaller.setup(device)
-                return@XCTestDriverClient xcTestInstaller.setup()
+        val xcTestDriverClient = XCTestDriverClient(
+            host = defaultHost,
+            port = xcTestPort,
+            installNetworkInterceptor = isStudio,
+            restoreConnection = {
+                if (SessionStore.activeSessions().isNotEmpty()) {
+//                    IdbInstaller.setup(device)
```
amanjeetsingh150, line 317: I think this is needed since the original java process stops which started both xctest runner and idb.

amanjeetsingh150: What is the problem we are facing with idb starting here?

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -177,4 +135,42 @@ class XCTestDriverClient(
...
+                    // to prevent a stack trace in the cli.
+                    if (isShuttingDown) {
+                        Response.Builder()
+                            .request(it.request())
+                            .protocol(Protocol.HTTP_1_1)
+                            .code(200)
+                            .build()
+                    } else {
+                        throw XCTestDriverUnreachable("Failed to reach out XCUITest Server")
+                    }
```
berikv, line 169: I assume the stack trace issue is why this network interceptor exists in the first place. Is that correct @amanjeetsingh150 ?

artem888 (approved):

dmitry-zaitsev (commented):
Let's make sure to somehow inform the user on _why_ their device is being rebooted, otherwise it can get very confusing

berikv:
> Let's make sure to somehow inform the user on _why_ their device is being rebooted, otherwise it can get very confusing

Devices are not rebooted, only idb is restarted

artem888:
> > Let's make sure to somehow inform the user on _why_ their device is being rebooted, otherwise it can get very confusing
> 
> Devices are not rebooted, only idb is restarted

seems that applesimutils causes some kinda device reboot https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1678455454132919?thread_ts=1678452522.490319&cid=C041FU72T54

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: Thats correct

berikv: In that case, could we just add this network interceptor to both studio and non-studio runs? Seems quite harmless to fake the 200 while shutting down in the non-studio runs.

amanjeetsingh150: I didn't saw usecase for shutting downs in non studio runs that's why avoided there

berikv:
> > > Let's make sure to somehow inform the user on _why_ their device is being rebooted, otherwise it can get very confusing
> > 
> > 
> > Devices are not rebooted, only idb is restarted
> 
> seems that applesimutils causes some kinda device reboot, but I don't see an easy workaround there https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1678455454132919?thread_ts=1678452522.490319&cid=C041FU72T54

It looks more like a UI refresh, applesimutils probably kills some simulator os-level gui process to make its changes persistent.

Do we really want to warn the user about that though? And through which channel? We currently don't have a very simple way to reliable warn the user about something. Maestro's ansi UI overwrites any println() that we do. I could trigger a system alert, but I don't think we'd make friends that way..

maestro-cli/src/main/java/maestro/cli/idb/IdbCompanion.kt
```diff
@@ -0,0 +1,137 @@
...
+import maestro.debuglog.DebugLogStore
+import maestro.utils.MaestroTimer
+import java.net.Socket
+import java.util.concurrent.TimeUnit
+import kotlin.concurrent.thread
+
+object IdbCompanion {
+    private val logger = DebugLogStore.loggerFor(IdbCompanion::class.java)
+
+    // TODO: Understand why this is a separate method from strartIdbCompanion
```
berikv, line 24: @amanjeetsingh150 Do you know the details on why this was implemented differently than `startIdbCompanion`?

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
berikv: Alright, I've simplified by always installing the shutdown handler.

PR #892, opened 2023-03-14
Add device API to allow listing and selecting devices

axelniklasson:
## Proposed Changes
- Make it possible to start Studio without a selected device
- Add a device api with three routes:
  - `GET /api/devices` that returns connected devices as well as devices available for launch
  - `GET /api/devices/selected` that returns the currently selected device if any
  - `POST /api/devices/selected` that allows selecting a device via either `instanceId` or `modelId`

## Testing
Tested locally

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
@@ -72,9 +72,3 @@ object DevicesService {
         }
     }
 }
-
-fun main() {
-    val connected = DeviceService.listConnectedDevices()
-    val availableForLaunch = DeviceService.listAvailableForLaunchDevices()
-    println(connected.size)
-}
--- a/maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
+++ b/maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import io.ktor.http.HttpStatusCode
 import io.ktor.server.application.call
 import io.ktor.server.response.respondText
+import io.ktor.server.routing.Route
 import io.ktor.server.routing.Routing
 import io.ktor.server.routing.get
 import io.ktor.server.routing.post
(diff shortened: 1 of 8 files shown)


maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+private data class SelectDeviceRequest(
+    val instanceId: String?,
+    val modelId: String?,
+)
+
+object DevicesService {
+    private var selectedDevice: Device.Connected? = null
+
+    fun routes(routing: Routing) {
```
dmitry-zaitsev, line 23: The choice is entirely up to you, but just FYI one convenient way of declaring such routes is by using Kotlin's extension functions:

```
fun Route.deviceRoutes() {
    devicesRoute()
    getSelectedDeviceRoute()
    selectDeviceRoute()
}

fun Route.devicesRoute() {
    get("/api/devices") {
        // etc
    } 
}

// etc.
```

Then you can mount all the routes like that:

```
routing {
    DeviceService.deviceRoutes()
}
```

maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
```diff
@@ -0,0 +1,74 @@
...
+        routing.get("/api/devices") {
+            val connected = DeviceService.listConnectedDevices()
+            val availableForLaunch = DeviceService.listAvailableForLaunchDevices()
+            val data = Devices(connected, availableForLaunch)
+
+            val response = jacksonObjectMapper()
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                .writerWithDefaultPrettyPrinter()
+                .writeValueAsString(data)
+            call.respondText(response)
```
dmitry-zaitsev, line 33: Ktor can take care of serialization by itself. It does take a bit of setup: https://ktor.io/docs/serialization.html#serialization_dependency

Not a blocker, functionally your implementation would work too, just a cleaner way of doing it

maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
```diff
@@ -0,0 +1,74 @@
...
+            val response = jacksonObjectMapper()
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                .writerWithDefaultPrettyPrinter()
+                .writeValueAsString(selectedDevice)
+            call.respondText(response)
+        }
+
+        routing.post("api/devices/select") {
+            val request = call.parseBody<SelectDeviceRequest>()
+            if (selectedDevice?.instanceId == request.instanceId) throw HttpException(HttpStatusCode.BadRequest, "Device already selected")
```
dmitry-zaitsev, line 46: nit: wouldn't a simple no-op be more robust/friendly here?

dmitry-zaitsev (approved):

maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
axelniklasson: Good point, will update 👍

maestro-cli/src/main/java/maestro/cli/studio/DevicesService.kt
axelniklasson: Ah, neat suggestion! However, won't it be 

```kotlin
routing {
   this.devicesRoutes()
}
```
instead given that it is an extension function for `Route`?

PR #894, opened 2023-03-14
feat: update env var injection and propagate to cloud

axelniklasson:
## Proposed Changes
- Only inject shell env vars prefixed with `MAESTRO_`
- Also inject env vars when running `maestro cloud`, to avoid different results compared to running locally when Flows are depending on shell env vars.

Note that this changes the current behaviour, which is to inject _all_ env vars when running locally and _not_ propagate these env vars to Maestro Cloud.

## Testing
Tested locally.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
@@ -32,15 +32,16 @@ object Env {
         return listOf(MaestroCommand(DefineVariablesCommand(env))) + this
     }
 
-    fun MutableMap<String, String>.withInjectedShellEnvVars(): MutableMap<String, String> {
+    fun Map<String, String>.withInjectedShellEnvVars(): Map<String, String> {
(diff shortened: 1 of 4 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
```diff
@@ -34,7 +34,7 @@ object Env {
 
     fun MutableMap<String, String>.withInjectedShellEnvVars(): MutableMap<String, String> {
```
berikv, line 35: This does not need to be a MutableMap if the implementation makes a mutable copy:
```
val mutable = this.asMutableMap()
...
```

berikv (approved):

berikv:
Thanks, great improvement!

maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
axelniklasson: Good one, updated 👍 thanks Berik!

Issue #876, opened 2023-03-08
tapOn contentDescription is not working when android:hint or android:text is present


jaychang0917test:
Hi, for the`tapOn` command, can we force to find the view by `contentDescription` (in android)? The problem is if the `android:hint` and `android:contentDescription` are both present, maestro will use the hint one to locate the view, but what i expected is find the view by `contentDescription` instead.

For example: the command - tapOn: "content" is not working to locate this EditText
```
<EditText
     android:layout_height="wrap_content"
     android:layout_width="match_parent"
     android:hint="hint"
     android:contentDescription="content"
 />
```
The same issue happens for TextView when both android:text and `android:contentDescription` are present, maestro always use the `android:text`

So, is this behaviour expected? or can we force to find the view by `contentDescription` (in android)? Thanks

amanjeetsingh150:
Hey @jaychang0917test we have added support to select elements through contentDescription as well. You just need to mention the accessibility text in the text attribute:

```
- tapOn: "accessibility-text-here"
```

Let me know if this works. Feel free to reopen issue if something works unexpected.

PR #897, opened 2023-03-15
[MOB 1954] Text matchers extended to match hint text and accessibility text as well

### LINKED ISSUE #850: Maestro still can not recognize `placeholder` in the `UITextField`
My screen hierarchy:
image

In the maestro studio:
image

amanjeetsingh150:
## Proposed Changes

1. Support matching contentDescription in android which is an accessibilityText
2. Support matching through hint text in android and placeHolder text in iOS 
3. Support matching with value by separating the XCUIElement label from the text

Adds two attributes to platform-agnostic hierarchy `TreeNode`: `hintText` and `accessibilityText`

## Studio changes
Add hintText and accessibilityText in studio when available

#### Matching Hint and value 
https://user-images.githubusercontent.com/12881364/225227227-7c759238-03cd-4184-bf52-e6b25a5d52c9.mov

#### Matching accessibility text
https://user-images.githubusercontent.com/12881364/225227263-c33d73d1-6eac-43c7-9091-8bf1b6fef6f8.mov

#### Studio

Screenshot 2023-03-15 at 12 26 54 PM

Screenshot 2023-03-15 at 12 27 09 PM

### iOS

https://user-images.githubusercontent.com/12881364/225235306-c4ca7c21-13a1-4150-84aa-9aac858a6b50.mov

Screenshot 2023-03-15 at 12 46 28 PM

## Issues Fixed
#850 
#876


maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
```diff
@@ -93,6 +93,7 @@ object ViewHierarchy {
             serializer.attribute("", "NAF", java.lang.Boolean.toString(true))
         }
         serializer.attribute("", "index", Integer.toString(index))
+        serializer.attribute("", "hintText", safeCharSeqToString(node.hintText))
```
axelniklasson, line 96: Do we need one of these for `accessibilityText` as well?

axelniklasson (approved):
Awesome 🔝

maestro-client/src/main/java/maestro/Filters.kt
```diff
@@ -54,26 +54,41 @@ object Filters {
...
     }
 
-    fun textMatches(text: String): ElementLookupPredicate {
-        return {
-            it.attributes["text"]?.let { value ->
-                text == value
-                    || text == value.replace('\n', ' ')
-            } ?: false
-        }
+    fun textMatches(regex: Regex): ElementLookupPredicate {
```
dmitry-zaitsev, line 57: We need to match in order of priority:

- Text
- Hint
- Accessibility

Otherwise, this risks being a breaking change. 

For inspiration how to do it, take a look at `idMatches` method that returns `ElementFilter` instead of `ElementLookupPredicate`

amanjeetsingh150: Why would this be a breaking change. If someone uses hint already it should match same for content description.

Am I missing something?

dmitry-zaitsev: Say you have 3 views in your hierarchy that appear in this order:

- `View(hint="Hello World")`
- `View(accessibilityText="Hello World")`
- `View(text="Hello World")`

In previous implementation a third view will be matched. In your implementation a first view will be matched. This is a breaking change. First or second view should only match if there is no third view.

amanjeetsingh150: Done ✅

maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
amanjeetsingh150: So, accessibilityText is a platform agnostic added new attribute. The one attribute which resolves to this is `content-desc` (contentDescription) which is already here.

axelniklasson: Got it, thanks!

PR #898, opened 2023-03-15
Stop XCTest runner before starting

berikv:
## Proposed Changes

Stop a running XCTest runner before starting it.

## Issues Fixed

Exception where the XCTest runner is restarted but was already running: "Failed to reach out XCUITest Server".

### DIFF
--- a/maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
+++ b/maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
@@ -33,16 +33,23 @@ class LocalXCTestInstaller(
 
         stop()
 
-        logger.info("[Start] Uninstalling the XCUITest runner app")
+        logger.info("[Start] Uninstall XCUITest runner")
         XCRunnerSimctl.uninstall(UI_TEST_RUNNER_APP_BUNDLE_ID)
-        logger.info("[Done] Uninstalling the XCUITest runner app")
+        logger.info("[Done] Uninstall XCUITest runner")
     }
 
     private fun stop() {
         if (xcTestProcess?.isAlive == true) {
-            logger.info("[Start] Killing the started XCUITest")
+            logger.info("[Start] Stop XCUITest runner")
             xcTestProcess?.destroy()
-            logger.info("[Done] Killing the started XCUITest")
+            logger.info("[Done] Stop XCUITest runner")
+        }
+
+        val pid = XCRunnerSimctl.pidForApp(UI_TEST_RUNNER_APP_BUNDLE_ID)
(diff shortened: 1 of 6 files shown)


maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
```diff
@@ -26,29 +26,36 @@ class LocalXCTestInstaller(
 
     private var xcTestProcess: Process? = null
 
-    override fun killAndUninstall() {
+    override fun uninstall() {
```
amanjeetsingh150, line 29: This doesn't aligns with what function is doing, I see that its killing a process as well. We might want to call `stop` with `uninstall` wherever its needed or rename this.

maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
```diff
@@ -26,29 +26,36 @@ class LocalXCTestInstaller(
...
-        logger.info("[Done] Uninstalling the XCUITest runner app")
     }
 
-    override fun setup(): Boolean {
+    override fun start(): Boolean {
         if (useXcodeTestRunner) {
             return ensureOpen()
         }
 
+        stop()
```
amanjeetsingh150, line 54: Why not uninstall here and only stop?

berikv: Because it would take time without any benefit.

maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
berikv: uninstalling implies also stopping the process

artem888 (approved):

maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
amanjeetsingh150: Ummm, uninstalling gave me the impression that it's just removing the app. That's why I was being skeptical about the name.

maestro-xcuitest-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
amanjeetsingh150: Got it! Yup if the app is not going to be present here I think we are ok.

PR #905, opened 2023-03-15
feat: add longPress to tap Private API

igorsmotto:
## Proposed Changes
This PR adds support for long press using Private APIs. 
The Private API used is the same one as the tap (`[XCPointerEventPath initForTouchAtPoint:offset:]`) but now with a configurable offset: tap: 0.1s, longPress: 3s

## Testing
Yet to be done


maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/EventRecord.swift
```diff
@@ -20,9 +20,9 @@ struct EventRecord {
             .takeUnretainedValue() as! NSObject
     }
 
-    mutating func addPointerTouchEvent(at point: CGPoint, touchUpAfter: TimeInterval = defaultTapDuration) {
+    mutating func addPointerTouchEvent(at point: CGPoint, touchUpAfter: TimeInterval?) {
         var path = PointerEventPath.pathForTouch(at: point)
-        path.offset += touchUpAfter
+        path.offset +=  touchUpAfter ?? EventRecord.defaultTapDuration
```
berikv, line 25: Nit: extra space

berikv: Also, I'm not sure why you've moved the default argument to an optional. Default argument value is preferred, if the compiler allows it.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Models/TouchRequest.swift
```diff
@@ -1,4 +1,5 @@
 struct TouchRequest : Codable {
     let x: Float
     let y: Float
+    let durationInMs: Int?
```
berikv, line 4: As a datatype, please use TimeInterval which is a Double value, in seconds.
For the name, I think `pressDuration` would be clearer.

berikv:
Amazing contribution! Congrats to your first swift code change @igorsmotto!

maestro-driver-ios/maestro-driver-iosUITests/Routes/XCTest/EventRecord.swift
berikv: Oh I see, the value you're passing is an optional. In that case, great!

maestro-driver-ios/maestro-driver-iosUITests/Routes/Models/TouchRequest.swift
igorsmotto: Agreed! 
https://github.com/mobile-dev-inc/maestro/pull/905/commits/9ad0a4760d0ef183931b7ef98ae3f3e12d96bc26: I decided to go only for `duration` because the `swipe` request uses `duration` in seconds as well.

berikv (approved):

PR #909, opened 2023-03-16
Pass XCTestInstaller to XCTestDriverClient instead of XCTestIOSDevice

berikv:
## Proposed Changes

This simplifies the setup:
* DriverClient already needs Installer to restart the service on error
* Device already needs DriverClient to perform the API calls
* With this change Device no longer needs to know about the Installer

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -31,7 +31,6 @@ import maestro.Maestro
 import maestro.cli.device.Device
 import maestro.cli.device.PickDeviceInteractor
 import maestro.cli.device.Platform
-import maestro.cli.idb.IdbCompanion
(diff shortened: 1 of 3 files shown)


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -164,18 +163,13 @@ object MaestroSessionManager {
                         val xcTestDriverClient = XCTestDriverClient(
                             host = defaultHost,
                             port = defaultXcTestPort,
-                            restoreConnection = {
-                            if (SessionStore.activeSessions().isNotEmpty()) {
```
berikv, line 168: Note that the condition for restarting `if (SessionStore.activeSessions().isNotEmpty()) { <restart> }` is removed, and restarts now happen unconditionally.

@amanjeetsingh150 Do you happen to know why this condition is put in place? Why stop a xctest runner restart if there are no active sessions?

artem888 (approved):

PR #912, opened 2023-03-16
[iOS] Configure maestro app language

artem888:
## Proposed Changes

Adding additional `language` parameter for LaunchApp command, this way users can test different locales

## Testing
- [x] UIKit app
- [ ] ReactNative app
- [x] Flutter app

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -361,7 +361,9 @@ class IntegrationTest {
         driver.assertEvents(
             listOf(
                 Event.StopApp("com.example.app"),
-                Event.LaunchApp("com.example.app", null)
+                Event.LaunchApp(
+                    appId = "com.example.app",
+                )
             )
         )
     }
@@ -592,7 +594,11 @@ class IntegrationTest {
         // Then
         // No test failure
         driver.assertHasEvent(Event.ClearState("com.example.app"))
-        driver.assertHasEvent(Event.LaunchApp("com.example.app", null))
+        driver.assertHasEvent(
+            Event.LaunchApp(
+                appId = "com.example.app",
+            )
+        )
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -361,7 +361,7 @@ class IntegrationTest {
         driver.assertEvents(
             listOf(
                 Event.StopApp("com.example.app"),
-                Event.LaunchApp("com.example.app")
+                Event.LaunchApp("com.example.app", null)
```
berikv, line 364: Naming the arguments will help understanding what null means here.

berikv (approved):

maestro-ios/src/main/java/ios/simctl/Simctl.kt
```diff
@@ -181,15 +189,26 @@ object Simctl {
...
+        var command = mutableListOf(
+            "xcrun",
+            "simctl",
+            "launch",
+            deviceId,
+            bundleId,
+        )
+
+        language?.let {
+            if (supportedLanguages.contains(it)) {
```
amanjeetsingh150, line 202: Are we throwing exceptions if someone makes an invalid selection?

maestro-ios/src/main/java/ios/simctl/Simctl.kt
```diff
@@ -10,6 +10,14 @@ object Simctl {
 
     data class SimctlError(override val message: String): Throwable(message)
     private val homedir = System.getProperty("user.home")
+    private val supportedLanguages = listOf(
```
dmitry-zaitsev, line 13: This list seems rather short. Are those really _all_ locales that we support?

maestro-ios/src/main/java/ios/simctl/Simctl.kt
dmitry-zaitsev: We should certainly inform user that their language selection was invalid

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -2496,6 +2496,30 @@ class IntegrationTest {
         )
     }
 
+    @Test
+    fun `Case 091 - Launch app with language set`() {
```
dmitry-zaitsev, line 2500: Please add a test for invalid language codes as well

Issue #287, opened 2022-10-18
Hide Keyboard fails intermittently.


imanishverma:
Hide Keyboard is still not 100% reliable. I found following flow where it fails:

```
# flow_contacts_android.yaml
appId: com.android.contacts
---
- launchApp
- tapOn: "Create new contact"
- tapOn: "First name"
- inputText: "FName01"
- tapOn: "Last name"
- inputText: "LName01"
- tapOn: "Phone"
- inputText: "80011085400"
- hideKeyboard
- tapOn: "Email"
- inputText: "lname01.fname01@gmail.com"
- tapOn: "Save"
```

Device: Pixel 6 Emulator

raulriera:
It also seems that this command is introducing a line break (Nexus S, SDK 32) https://github.com/mobile-dev-inc/maestro/blob/5f866556debc28a8157ab5949fb1f40ed24268a3/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt#L351

nhaarman:
KeyEvent `4` seems to work ('back', which dismisses the keyboard) http://www.temblast.com/ref/akeyscode.htm

PR #917, opened 2023-03-17
Use the 'back' key event to dismiss keyboard

nhaarman:
'Enter' does never dismiss the keyboard when there is a 'next' TextField to focus on.

Fixes https://github.com/mobile-dev-inc/maestro/issues/287


dmitry-zaitsev:
Thank you for the contribution!

ice-cap0:
this triggers back actions
`hideKeyboard` is probably not the right name for this if it's also pressing the back button on Android

PR #918, opened 2023-03-17
Adds proper text description for long press

ArthurSav:
## Testing
- Manual

## Issues Fixed
- longPressOn would display as "Tap" as command description. Now it correctly shows "Long press"


igorsmotto (approved):

Dambakk (commented):
🙌

PR #924, opened 2023-03-20
Use the XCTest runner version of keyPress() buttonPress() and eraseText()

berikv:
## Proposed Changes

Use the XCTest runner version of keyPress() buttonPress() and eraseText()

## Testing

Ran customer flows and crafter flows which execute these commands.

## Issues Fixed

IDB crashes when commands are executed on an app with a big view hierarchy. The XCTest runner implementation of these commands should also work on apps with big view hierarchies.


artem888 (approved):
🔝

PR #926, opened 2023-03-20
Feature: React Native SDK

dmitry-zaitsev:
Sorry for the large PR, but there are tons of configs that have to go together for it to work


remotenode:
@dmitry-zaitsev can you please describe in a few words the purpose of adding ReactNative SDK?

dmitry-zaitsev:
@remotenode absolutely. 

In the short-term:

- SDK is going to make RN apps compatible with our [Mock Server](https://maestro.mobile.dev/advanced/experimental/maestro-mock-server) feature (that requires you to obtain a mock server URL at runtime).
- We are planning to add a much requested method to check whether app is running under a Maestro test at any given moment (something like `isMaestroRunning()`)

Longer term, having an SDK would allow us to expose even more features for interacting with the application and further enrich Maestro Studio experience.

It is important to note that our philosophy remains that Maestro should function without any SDKs and with as little setup as possible. What that means for SDK is that it will likely remain an optional component targeted at users who want to get this extra functionality.

ArthurSav (approved):

PR #927, opened 2023-03-20
Use XCTest runner for deviceInfo()

berikv:
## Proposed Changes

Use XCTest runner for deviceInfo()

## Testing

Tested with customer flows and by running:
```
        let request = HTTPRequest(method: .POST, version: .http11, path: "", query: [], headers: [:], body: Data())
        let response = try await DeviceInfoHandler().handleRequest(request)
        print(String(data: response.body!, encoding: .utf8)!)
```

## Issues Fixed


maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
```diff
@@ -40,7 +41,31 @@ class XCTestIOSDevice(
...
+                    logger.info("Device info failed: $message")
+                    throw UnknownFailure(message)
+                }
+
+                body ?: throw UnknownFailure("Error: response body missing")
+
+                val responseBody = mapper.readValue(body, DeviceInfoResponse::class.java)
+                logger.info("Device info $responseBody")
+
+                // Warning: assuming pixels == points
```
artem888, line 59: technically we can get them as well by multiplying on `UIScreen.main.scale`

artem888 (approved):

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
artem888: hmm, just tested , worked for me

berikv: You are correct sir! Thanks for testing! I'll update the PR accordingly

PR #951, opened 2023-03-27
Retry starting XCTest runner multiple times

berikv:
## Proposed Changes

Retry the starting the XCTest runner on connection error.

## Testing

todo

## Issues Fixed

During an launchApp command, xctest may not be able to run for a while because the simulator is not fully initialised (due to permissions being set). This causes errors in the CLI to be reported about XCTest runner. This fix will retry starting XCTest runner multiple times, making it far more likely that the simulator is fully initialised before an error is emitted.

### DIFF
--- a/maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -178,15 +178,15 @@ class XCTestDriverClient(
             try {
                 return it.proceed(request)
             } catch (connectException: IOException) {
-                if (retryAttempt < retryCount) {
-                    retryAttempt += 1
-                    if (!installer.start()) {
-                        throw XCTestDriverUnreachable("Failed to start XCUITest Server in RetryOnError")
-                    }
-                    continue
-                } else {
+                if (retryAttempt >= retryCount) {
                     throw XCTestDriverUnreachable("Failed to reach out XCUITest Server in RetryOnError")
                 }
+
+                if (!installer.start()) {
+                    throw XCTestDriverUnreachable("Failed to start XCUITest Server in RetryOnError")
+                }
+
+                retryAttempt += 1
             }
         }
     }


maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -167,17 +167,29 @@ class XCTestDriverClient(
...
-        } catch (connectException: IOException) {
-            if (installer.start()) {
-                it.proceed(request)
-            } else {
-                throw XCTestDriverUnreachable("Failed to reach out XCUITest Server in RetryOnError")
+        while (true) {
+            try {
+                return it.proceed(request)
+            } catch (connectException: IOException) {
+                if (retryAttempt >= retryCount) {
```
igorsmotto, line 181: Instead of having a `while(true)`. What do you think about having something like `while(retryAttempt <= retryCount)`?
This way we **guarantee** that we will never softlock maestro

igorsmotto (approved):

maestro-xcuitest-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
berikv: It would not execute the first proceed if retryCount == 0

berikv: The retry logic should really be in the catch block

igorsmotto: My concern is only that for some unrelated reason the `proceed` fails with a different exception than IOException, which will not be captured by this catch block, resulting in an infinite loop.

berikv: Ah, yes that case needs handling

berikv: Actually, unhandled exceptions can never be trapped in the while loop. They would exit the loop immediately

igorsmotto: Oh yeah. You're right. It would throw the exception.

axelniklasson (approved):

axelniklasson:
@berikv is this a change we should go ahead with or is it not needed anymore?

PR #952, opened 2023-03-27
feat: exit process on SIGTSTP

igorsmotto:
## Proposed Changes
Add a signal handler that reacts to SIGTSTP (when ctrl+z is used). 
This signal handler will exit the process with 0 as the return code. 
This will trigger runtime shutdown hooks.

## Testing
- ✅ Opened `./maestro studio`
- ✅ Ctrl + Z
- ✅ Opened `./maestro studio`

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -34,6 +34,8 @@ import maestro.cli.device.Platform
 import maestro.debuglog.IOSDriverLogger
 import maestro.drivers.IOSDriver
 import org.slf4j.LoggerFactory
+import sun.misc.Signal
+import sun.misc.SignalHandler
 import util.XCRunnerSimctl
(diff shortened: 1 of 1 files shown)


felipevolpone (approved):

axelniklasson (approved):

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -336,4 +339,15 @@ object MaestroSessionManager {
 
     }
 
+    private class CustomSignalHandler() : SignalHandler {
+        override fun handle(signal: Signal) {
+            when (signal) {
+                suspendSignal -> exitProcess(0)
```
dmitry-zaitsev, line 345: Will this trigger a shutdown hook on line 86?

igorsmotto: Yes. But since ctrl+z suspend the main process, so it will run in the background

PR #955, opened 2023-03-28
Restart iOS simulator after permissions reset

artem888:
## Proposed Changes
This helps to mitigate some xctest runner / idb connection issues after applesimutils permissions reset

## Testing
- [x] tested all existing flows with maestro-flow-tester

### DIFF
--- a/maestro-ios/src/main/java/ios/simctl/Simctl.kt
+++ b/maestro-ios/src/main/java/ios/simctl/Simctl.kt
@@ -110,6 +110,41 @@ object Simctl {
         awaitLaunch(deviceId)
     }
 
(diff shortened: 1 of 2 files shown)


artem888:
https://user-images.githubusercontent.com/961175/228314994-491b3800-11da-4d6c-9a77-b9e0edaf24da.mov

@dmitry-zaitsev not perfect, but it is planned to be a temporary fix until we implement .hierarchy command with the new APIs

berikv (approved):

dmitry-zaitsev:
@artem888 this is an unacceptable delay, please revert this change. Can we reduce the impact of this somehow?

PR #959, opened 2023-03-29
Use return instead of enter for `pressKey: ENTER` on iOS

berikv:
## Proposed Changes

Some apps behave differently to return vs enter. In general the behaviour for return is more holistic and expected when running `pressKey: ENTER`.
Enter can be still be entered by running `inputText: "\n"`


artem888 (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -147,7 +147,7 @@ class IOSDriver(
     override fun pressKey(code: KeyCode) {
         val keyCodeNameMap = mapOf(
             KeyCode.BACKSPACE to "delete",
-            KeyCode.ENTER to "enter",
+            KeyCode.ENTER to "return",
             // Supported by iOS but not yet by maestro:
 //        KeyCode.RETURN to "return",
```
axelniklasson, line 152: Should we map both `ENTER` and `RETURN` to `return` or remove `RETURN` altogether and guide users to use `ENTER` instead?

berikv: maestro does not support `pressKey: RETURN`.
We can either
* switch the implementation of `pressKey: ENTER` to use "return" 
* add support for `pressKey: RETURN` and document the subtle difference and ask users to change their flows

axelniklasson: Yep, but keeping this in the code looks odd to me:

```
            KeyCode.ENTER to "return",
            // Supported by iOS but not yet by maestro:
//        KeyCode.RETURN to "return",
```

which is why I raised the question. No need to decide on the direction here right now though we can work on that later.

berikv: I don't follow your suggestion. We can't map RETURN to anything, because it is currently not supported. Only if we start supporting it, we can map it to "return". But if we go that route, it would make sense to map ENTER to "enter". Which is fine by me. But I think it'd be confusing for users, since even Apple didn't get this right in every app they made. If "return" works, I suggest we just map ENTER to "return".

berikv: (In new keyboards you can press return by fn+enter)

berikv: I'll clean up the comment to make it make more sense.

PR #966, opened 2023-03-31
Implement notifications permission setting logic on xctest driver side

artem888:
## Proposed Changes
This PR:
- removes "notifications" and "health" permissions from `applesimutils` since they were restarting Springboard and causing some idb / xctest connection issues
- moves notification permission setting under XCTest driver with a logical change that alert would be automatically dismissed depending on "notifications" permission setting

## Testing
- [x] maestro-flow-tester

## Issues Fixed
Majority of "Failed to reach out XCUITest Server" issues


maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
```diff
@@ -12,27 +12,30 @@ final class SubTreeRouteHandler: HTTPHandler {
...
             let springboardBundleId = "com.apple.springboard"
             logger.info("Trying to capture hierarchy snapshot for \(appId)")
             let start = NSDate().timeIntervalSince1970 * 1000
             let xcuiApplication = XCUIApplication(bundleIdentifier: appId)
             let springboardApplication = XCUIApplication(bundleIdentifier: springboardBundleId)
-
+            
+            tapOnSystemPermissionAlertIfNeeded(springboardApplication: springboardApplication,
+                                               appName: xcuiApplication.label)
+
```
berikv, line 25: Instead of mixing this into the implementation of SubTree, can the permission code go into its own file. Then each handler could first call into the permission handler and then continue with its work.
Later we could then refactor it as a pre-action for every handler so that the handlers are free from permission logic.

artem888: do you think it makes sense to do it on every Handler or just subTree?

berikv:
Yes I think so, but lets first have the implementation that you’ve tested to work.
My point is to contain the permission code as much as possible, so that calling it from other Handlers would be easy.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SetPermissionsHandler.swift
```diff
@@ -0,0 +1,36 @@
...
+    func handleRequest(_ request: HTTPRequest) async throws -> HTTPResponse {
+        let decoder = JSONDecoder()
+        
+        guard let requestBody = try? decoder.decode(SetPermissionsRequest.self, from: request.body) else {
+            let errorData = handleError(message: "incorrect request body provided")
+            return HTTPResponse(statusCode: HTTPStatusCode.badRequest, body: errorData)
+        }
+        
+        if let encoded = try? JSONEncoder().encode(requestBody.permissions) {
+            UserDefaults.standard.set(encoded, forKey: "permissions")
```
amanjeetsingh150, line 21: This can't fail?

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
```diff
@@ -53,9 +56,38 @@ final class SubTreeRouteHandler: HTTPHandler {
...
+            return "unknown-snapshot-failure"
+        }
+    }
+    
+    private func tapOnSystemPermissionAlertIfNeeded(springboardApplication: XCUIApplication, appName: String) {
+        let predicate = NSPredicate(format: "label CONTAINS[c] %@", appName)
+        
+        guard let data = UserDefaults.standard.object(forKey: "permissions") as? Data,
+              let permissions = try? JSONDecoder().decode([String : PermissionValue].self, from: data),
+              let notificationsPermission = permissions.first(where: { $0.key == "notifications" }) else {
```
amanjeetsingh150, line 70: Can this work with other permissions as well?

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
```diff
@@ -53,9 +56,38 @@ final class SubTreeRouteHandler: HTTPHandler {
...
+            return
```
amanjeetsingh150, line 71: Another question would be: "notifications" would work in case of localization?

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SetPermissionsHandler.swift
berikv: UserDefaults is very reliable

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
berikv: Lets for now only use this on tho notification permission. We can check other permissions later

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
berikv: This string comes from the flow definition. Flow definitions currently do not support localisation.

maestro-driver-ios/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
artem888: resolved

berikv (approved):

PR #967, opened 2023-03-31
Rename ios modules and Simctl

berikv:
## Proposed Changes

Rename:
* Module maestro-xcuitest-driver to maestro-ios-driver (contains the code to manage iOS Simulator and talk to the XCTest runner)
* Module maestro-driver-ios to maestro-ios-xctest-runner (contains the xctest runner code)
* Class maestro-xcuitest-driver/XCRunnerSimctl to maestro-ios-driver/XCRunnerCLIUtils (contains cli utils for managing the XCTest runner)
* Class maestro-ios/Simctl to maestro-ios-driver/LocalSimulatorUtils (contains code to control the simulator via cli / simctl)

These renames make it easier to navigate the code:
* ios code is grouped by the `maestro-ios` prefix
* Clearly different spelling between XCRunnerCLIUtils and LocalSimulatorUtils  

## Testing

No functional change. Just renames.
Tested running `./maestro-ios-xctest-runner/build-maestro-ios-runner.sh`


.github/workflows/publish-release.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.SIGNING_PASSWORD }}
 
       - name: Upload Maestro XCUiTest Driver
-        run: ./gradlew clean :maestro-xcuitest-driver:publish --no-daemon --no-parallel
+        run: ./gradlew clean :maestro-ios-driver:publish --no-daemon --no-parallel
```
amanjeetsingh150, line 98: Wouldn't this be confused with the other directory where we have server code?

There would be maestro-ios-driver and maestro-driver-ios (swift one) in codebase, which is confusing for me. Sounds like same.

berikv: From the description:

> Rename:
>
> Module maestro-xcuitest-driver to maestro-ios-driver (contains the code to manage iOS Simulator and talk to the XCTest runner)
> Module maestro-driver-ios to maestro-ios-xctest-runner (contains the xctest runner code)

axelniklasson (approved):

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -143,9 +143,8 @@ class LocalIOSDevice(
         return xcTestDevice.isScreenStatic()
     }
 
-    override fun setPermissions(id: String, permissions: Map<String, String>) {
-        simctlIOSDevice.setPermissions(id, permissions)
-        xcTestDevice.setPermissions(id, permissions)
+    override fun setPermissions(id: String, permissions: Map<String, String>): Result<Unit, Throwable> {
+        return simctlIOSDevice.setPermissions(id, permissions)
```
amanjeetsingh150, line 147: This is a rebase problem

amanjeetsingh150, line 147: This is breaking the notification change, and regresses maestro flow tester and the runs we do staging. I propose we take these rename changes in next release if these are not dependent on something important. Let me know your thoughts.

amanjeetsingh150 (changes requested):
This is breaking the notification change, and regresses maestro flow tester and the runs we do staging. I propose we take these rename changes in next release if these are not dependent on something important. Let me know your thoughts.

maestro-ios/src/main/proto/idb.proto
```diff
@@ -870,4 +870,4 @@ message DapResponse {
 
 message SimulateMemoryWarningRequest {}
 
-message SimulateMemoryWarningResponse {}
\ No newline at end of file
+message SimulateMemoryWarningResponse {}
```
amanjeetsingh150, line 873: Is something extra changed here?

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
berikv: Thanks for finding this! I have updated the cod to call both xctest and simctl in case of setPermissions

maestro-ios/src/main/proto/idb.proto
berikv: A newline was added at the end of the file, my IntellIJ does that when I open the file

.github/workflows/publish-release.yml
berikv: Guys, this is both in the description and in the changes made. If there are any doubts, use `git switch rename_simctl` and see for yourself.

amanjeetsingh150 (approved):

PR #969, opened 2023-04-03
Hide keyboard on iOS

amanjeetsingh150:
## Proposed Changes

There are no straightforward APIs to hide the keyboard on iOS. This PR attempts to hide the keyboard as a user through maestro. Three attempts at how maestro tries hiding the keyboard:

1. Scroll with a small delta on y axis
2. Scroll with a small delta on x axis
3. Clicking on static texts XCUIElement

## Testing
Tested on multiple client apps.

## Issues Fixed

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -31,6 +31,7 @@ import ios.IOSDevice
 import maestro.Capability
 import maestro.DeviceInfo
 import maestro.Driver
+import maestro.Filters
 import maestro.KeyCode
 import maestro.MaestroException
 import maestro.Platform
@@ -38,6 +39,8 @@ import maestro.Point
 import maestro.ScreenRecording
 import maestro.SwipeDirection
 import maestro.TreeNode
+import maestro.UiElement.Companion.toUiElement
+import maestro.UiElement.Companion.toUiElementOrNull
 import maestro.ViewHierarchy
 import maestro.utils.FileUtils
 import maestro.utils.MaestroTimer
@@ -224,6 +227,7 @@ class IOSDriver(
         val text = xcUiElement.title?.ifEmpty {
             xcUiElement.value
         }
+        val clickable = if (xcUiElement.elementType == ELEMENT_TYPE_STATIC_TEXT) false else null
         attributes["accessibilityText"] = xcUiElement.label
         attributes["text"] = text ?: ""
         attributes["hintText"] = xcUiElement.placeholderValue ?: ""
@@ -248,6 +252,7 @@ class IOSDriver(
 
         return TreeNode(
             attributes = attributes,
+            clickable = clickable,
             children = children,
             enabled = xcUiElement.enabled,
             focused = xcUiElement.hasFocus,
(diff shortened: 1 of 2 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -373,7 +378,48 @@ class IOSDriver(
...
+                .any {
+                    tap(it.bounds.center())
+                    isKeyboardHidden()
+                }
+            return@retryUntilTrue result
+        }
+    }
+
+    private fun isKeyboardHidden(): Boolean {
+        val filter = Filters.textMatches("[rR]".toRegex())
```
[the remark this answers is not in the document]
amanjeetsingh150: What about I do this with keys that are not locale based for example delete key that has id linked with it?

Screenshot 2023-04-03 at 1 19 33 PM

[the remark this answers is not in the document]
axelniklasson: That would be great -- I assume most (if not all) keyboards have a delete button? I think adding this to our docs for the `hideKeyboard` command would be good too, so users are aware that the `delete` key needs to be there for this command to work on iOS.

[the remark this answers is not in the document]
amanjeetsingh150: Done, sure will be updating the docs accordingly

axelniklasson (approved):
LGTM but let's also get eyes from @berikv or @artem888 on this one

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -373,7 +378,48 @@ class IOSDriver(
...
+
+        swipe(
+            start = Point(widthPercentToPoint(0.5), heightPercentToPoint(0.5)),
+            end = Point(widthPercentToPoint(0.47), heightPercentToPoint(0.5)),
+            durationMs = 50,
+        )
+
+        if (isKeyboardHidden()) {
+            return
+        }
```
berikv, line 399: I imagined this to be implemented in the XCTest runner.
The runner can create a single EventRecord with both the swipe down and up in there. Reducing the change that the `hidekeyboard` command changes anything in the app: Separate swipe events are more likely to trigger some app logic than a single swipe down then up event.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -373,7 +378,48 @@ class IOSDriver(
...
+
+        tapOnNonClickableText()
```
berikv, line 401: Is tapping still needed if we implement the swipe solution?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -373,7 +378,48 @@ class IOSDriver(
...
+    }
+
+    private fun tapOnNonClickableText() {
```
berikv, line 404: Lets not mix tap and click in the names: Mouses click and touch panels tap.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
berikv: Although, on second thought.
For buttons to not be tapped, the touch point needs to be swiped outside of the tappable area. If the user swipes back to the tappable area, the press will not be cancelled.
Did you test with swipe down event right on a button?
A quick check with the cancel button in Setting Search app, the swipe needs to be roughly 1/5th the screen height before it is considered cancelled.

berikv (changes requested):
Can you verify that the swipe action does not accidentally presses a button if it happens to be exactly on the swipe starting point?

artem888 (approved):
LGTM once comments from @berikv addressed

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Yes, it's just there as a fallback if the swipe fails to hide the keyboard.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Done ✅

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: This is the change I did:

```
func addHideKeyboardSwipeEvent(start: CGPoint, end: CGPoint, duration: TimeInterval) -> Self {
        var path = PointerEventPath.pathForTouch(at: start)
        path.offset += Self.defaultTapDuration
        path.moveTo(point: end)
        path.moveTo(point: start)
        path.moveTo(point: end)
        path.offset += duration
        path.liftUp()
        return add(path)
}
```


## EventRecord Changes

https://user-images.githubusercontent.com/12881364/229527384-cd15d780-e5f0-4f40-9828-f8eb93a270f1.mov

## Current Implementation

https://user-images.githubusercontent.com/12881364/229528802-47718aac-e06a-4ca6-ad59-c8b9bb847f98.mov


Seems like same, unless I am missing something in that eventRecord code.

berikv (approved):

PR #970, opened 2023-04-03
Feature: new assertOutgoingRequests command

felipevolpone:
## Proposed Changes
This PR implements a new command to assert outgoing requests from the app. This will be used to check if the app does some network calls for certain APIs/backend.
The `mock-server` will be used to fetch and match the events stored in Maestro Cloud.

This is the command definition:
```yaml
- assertOutgoingRequest:
    url: ".*api.company.com\/[^\/]+\/login"
    headersPresent:
      - "cache-control"
    headersAndValues:
      - "Content-Type": "*application/json*"
      - "Authorization": "Bearer .*"
    httpMethodIs: "GET"
    requestBodyContains: '"foo": "bar"'
```

Important to say that this command will only work if the user have set the Maestro SDK properly (https://maestro.mobile.dev/advanced/experimental/maestro-mock-server/getting-started#setup)

The implementation for Android and iOS relies on getting the sessionId (and UUID). For Android, the maestro-sdk generates the random UUID; for iOS the driver generates and the SDK fetches from there. With the sessionId, maestro can call the Maestro Cloud backend and filter the events and make the assert logic.

I know this is a very long PR and I'm sorry for that, but if you follow the commit timeline it will make it easier to review it.


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -287,6 +301,29 @@ class Service(
...
+    override fun fetchSession(
+        request: MaestroAndroid.SessionRequest,
+        responseObserver: StreamObserver<MaestroAndroid.SessionResponse>
+    ) {
+        val sessionReq = Request.Builder()
+            .get()
+            .url("http://localhost:7008/session")
+            .build()
+
+        val sessionInfo = httpClient
```
dmitry-zaitsev, line 313: This call might throw an error, you need to propagate it properly to `responseObserver`

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -287,6 +301,29 @@ class Service(
         responseObserver.onCompleted()
     }
 
+    override fun fetchSession(
+        request: MaestroAndroid.SessionRequest,
+        responseObserver: StreamObserver<MaestroAndroid.SessionResponse>
+    ) {
+        val sessionReq = Request.Builder()
+            .get()
+            .url("http://localhost:7008/session")
```
dmitry-zaitsev, line 310: This service is running in the same process, what is the point of using HTTP to communicate with it? You can just access memory directly

dmitry-zaitsev: Specifically, you are looking for `SdkService` class

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -287,6 +287,14 @@ class Service(
         responseObserver.onCompleted()
     }
 
+    override fun fetchSession(
+        request: MaestroAndroid.SessionRequest,
+        responseObserver: StreamObserver<MaestroAndroid.SessionResponse>
+    ) {
+        responseObserver.onNext(sessionResponse { id = SdkService.sessionInfo.sessionId })
```
amanjeetsingh150, line 294: Can session id be null?

maestro-client/src/main/java/maestro/OutgoingRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        val eventsFilteredByHeadersAndValues = rules.headersAndValues.entries.fold(eventsFilteredByHeader) { eventsList, (header, value) ->
+            eventsList.filter { e -> e.headers?.get(header.lowercase()) == value }
+        }
+
+        val eventsMatching = rules.requestBodyContains?.let { requestBody ->
+            eventsFilteredByHeadersAndValues.filter { e -> e.bodyAsString?.contains(requestBody) == true }
+        } ?: eventsFilteredByHeadersAndValues
+
+        println("from ${events.size} events, ${eventsMatching.size} match the url ${rules.url}")
```
amanjeetsingh150, line 39: nit: Capitalization `F`. If this to be shown to user

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -494,6 +495,10 @@ class IOSDriver(
         iosDevice.setPermissions(appId, permissions)
     }
 
+    override fun fetchSessionId(): UUID {
+        return iosDevice.fetchSessionInfo().expect {  }
```
amanjeetsingh150, line 499: This will fail if fetchSessionInfo errors. Is this intended?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -748,6 +748,22 @@ data class TravelCommand(
...
+    val headersAndValues: Map<String, String> = emptyMap(),
+    val httpMethodIs: String? = null,
+    val requestBodyContains: String? = null,
+) : Command {
+    override fun description(): String {
+        return "Assert outgoing requests to $url"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
```
amanjeetsingh150, line 763: Assertions can have JS resolutions from runScript output I would recommend using evaluate script in case of string.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import FlyingFox
+import os
+import XCTest
+import Network
+
+@MainActor
+struct SessionInfoHandler: HTTPHandler {
+
```
amanjeetsingh150, line 9: I see changes in xctest but the updated binaries are not here?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import FlyingFox
+import os
+import XCTest
+import Network
+
+@MainActor
```
amanjeetsingh150, line 7: Do we need to do this on main thread?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -287,6 +287,14 @@ class Service(
...
+        responseObserver.onCompleted()
+    }
```
amanjeetsingh150, line 296: I see changes here but not updated maestro server and maestro-android APKs

amanjeetsingh150:
Can you add integration tests for this? I am finding hard to understand how and when users will write it in their flow file?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
felipevolpone: Yes, all other handlers also contain the `@MainActor`

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
felipevolpone: How can I do that?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
felipevolpone: Yes

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
felipevolpone: No, the sessionId is the first thing created when the users uses the SDK

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
felipevolpone: How can I do that?

felipevolpone:
@amanjeetsingh150 I added an integration test, but I don't think it is that useful for this feature. I updated the PR description to explain more about this feature, let me know if it's not enough

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
felipevolpone: Done

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -158,4 +159,8 @@ class LocalIOSDevice(
     override fun eraseText(charactersToErase: Int) {
         xcTestDevice.eraseText(charactersToErase)
     }
+
+    override fun fetchSessionInfo(): Result<UUID, Throwable> {
+        return simctlIOSDevice.fetchSessionInfo()
```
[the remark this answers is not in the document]
felipevolpone: Fixed 👍

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
berikv: Thats not a great reason. The other handlers need @MainActor because they interact with XCTest.frawemork, which requires to be called on the main thread. I don't see code in the SessionInfoHandler that requires the main thread.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
berikv: (cd maestro; ./maestro-ios-xctest-runner/build-maestro-ios-runner.sh)

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
```diff
@@ -0,0 +1,33 @@
...
+struct SessionInfoHandler: HTTPHandler {
+
+    let logger = Logger(
+        subsystem: Bundle.main.bundleIdentifier!,
+        category: String(describing: Self.self)
+    )
+    
+    private let sessionId = UUID()
```
berikv, line 15: Why is the XCTest runner involved to create a UUID? Seems like this can be done much faster and simpler in the worker..?

berikv (changes requested):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
felipevolpone: When the user uses the maestro SDK to use the Mock Server feature (https://maestro.mobile.dev/advanced/experimental/maestro-mock-server/getting-started#setup), it will proxy the app requests through our backend. The backend stores the events and then the maestro client needs to check if the data used in the `assertOutgoingRequest` matches the info stored.

The sessionId is related to a specific maestro session, meaning every time it runs `maestro test <flow>`. Because of that, the driver should create a new UUID every time and then, the client will get the sessionId and fetch the backend to get the events. That's why we can't use worker for that.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
felipevolpone: @berikv this command fails for me with long output. I'll send to you over Slack

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
berikv: Sounds like XCTest runner is really not the right place for that. It can be restarted throughout the flow and should not be used for storing UUID's. Can't maestro store the session UUID?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
berikv: You need to set up Xcode and install the 15.4 simulator runtime to use this command. Let me know if that resolves the error.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
berikv: How about the SDK doing a call-out to our backend with the sessionId to match on as a HTTP call?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
amanjeetsingh150: 👍

amanjeetsingh150:
> @amanjeetsingh150 I added an integration test, but I don't think it is that useful for this feature. I updated the PR description to explain more about this feature, let me know if it's not enough

Thanks @felipevolpone now I understand a bit more. Few questions:

1. Does these assertions can only be done with network mocking, right? If yes how do we plan to give a feedback to user if they try to use with live network?
2. I find it hard to understand where should I assert requests. Generally the requests are on user actions, and since requests can be in parallel I wouldn't which one to assert first?
3. Do we plan to include this assertion in flow file only?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SessionInfoHandler.swift
axelniklasson: @berikv the request to our backend with the sessionId to match on HTTP call happens from maestro and not from the SDK, as maestro needs the response in order to validate the assertions. Is there any way we can generate a session id from within the iOS SDK and keep that session id throughout the flow execution?

berikv: What is a session exactly? How long does it live? Where does it depend on?

* If it is a flow run, then maestro should be the source of truth, eg Orchestra or SessionManager could keep this state.
* If a session is longer than a single flow run, then the source of truth needs to be outside of maestro (and not in the XCTest runner, which is part of maestro).
 * If a session is bound by the SDK, then let the SDK be the source of truth. If maestro needs to know the session ID, it can fetch it using HTTP calls from the backend or directly to the app, whatever works.
* If it is just a part of a flow run, the source of truth can still be maestro. A session object that's closed and reopened when needed. It could live alongside Orchestra.

PR #972, opened 2023-04-03
Feature: launch arguments for Android

dmitry-zaitsev:
Kudos to @artem888 and @amanjeetsingh150 for implementation idea


maestro-android/src/androidTest/AndroidManifest.xml
```diff
@@ -4,5 +4,8 @@
 
     <uses-permission android:name="android.permission.ACCESS_MOCK_LOCATION"
                      tools:ignore="MockLocation,ProtectedPermissions"/>
+    <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"
```
artem888, line 7: from my memory it was only needed for the target app

dmitry-zaitsev: I think so too, but I am putting it in both on the off chance that some Android version might interpret it differently. It doesn't hurt

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -108,6 +110,35 @@ class Service(
...
+                val keyValue = arg.split("=")
+
+                if (keyValue.size == 2) {
+                    intent.putExtra(keyValue[0], keyValue[1])
+                } else {
+                    intent.putExtra(arg, true)
+                }
+            }
+
+        context.startActivity(intent)
```
amanjeetsingh150, line 136: Did we also checked with apps that have multiple launchers? For example the ones with leak canary.

maestro-test/src/test/resources/095_launch_arguments.yaml
```diff
@@ -0,0 +1,11 @@
...
+env:
+  ARGUMENT_B: argumentB
+  INCLUDE_ARGUMENT_C: true
+---
+- launchApp:
+    arguments:
+      - argumentA
+      - ${ARGUMENT_B}
+      - ${INCLUDE_ARGUMENT_C && 'argumentC'}
+      - ${INCLUDE_ARGUMENT_D && 'argumentD'}
```
amanjeetsingh150, line 11: Slightly confused with the arguments here. Can it be an independent variable or key-value pairs only?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
dmitry-zaitsev: Hm, isn't `getLaunchIntentForPackage` solving that already? I am not 100% certain

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -168,24 +165,20 @@ class AndroidDriver(
...
-            shell("monkey --pct-syskeys 0 -p $appId 1")
-        } catch (saxException: SAXException) {
-            shell("monkey --pct-syskeys 0 -p $appId 1")
-        }
+        blockingStub.launchApp(
+            launchAppRequest {
+                this.packageName = appId
+                this.arguments.addAll(launchArguments)
+            }
+        )
```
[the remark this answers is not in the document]
dmitry-zaitsev: That will indeed be the only way. I was considering leaving the previous option as a fallback, but after working with it - it becomes very confusing on what is going on and why arguments are not being respected

maestro-test/src/test/resources/095_launch_arguments.yaml
dmitry-zaitsev: It can be independent variable (that is what iOS typically expects). In Android terms that could be either a value (in which case it will be passed as a Boolean flag) or key-value (in which case it will be passed as a String)

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
amanjeetsingh150: I'm not sure too, let's verify maybe verify from the sample app we use.

maestro-test/src/test/resources/095_launch_arguments.yaml
amanjeetsingh150: How would android behave if an argument only is supplied will it ignore?

amanjeetsingh150: Curious how iOS reads a single argument. @artem888 any idea how would single argument (argumentA) be read in iOS apps?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
amanjeetsingh150: Verified should be handling ✅

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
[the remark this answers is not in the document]
amanjeetsingh150: Sounds good!

amanjeetsingh150 (approved):

PR #973, opened 2023-04-04
[iOS] fix permission reset crashing xctest server

artem888:
## Proposed Changes

`XCUIApplication.label` causes a crash for some apps with big view hierarchy

## Testing

tested locally with some RN apps


maestro-driver-ios/maestro-driver-iosUITests/Routes/Helpers/SystemPermissionHelper.swift
```diff
@@ -1,8 +1,10 @@
 import XCTest
 
 final class SystemPermissionHelper {
-    static func handleSystemPermissionAlertIfNeeded(springboardApplication: XCUIApplication, appName: String) {
-        let predicate = NSPredicate(format: "label CONTAINS[c] %@", appName)
+    private static let notificationsPermissionLabel = "Would Like to Send You Notifications"
+    
+    static func handleSystemPermissionAlertIfNeeded(springboardApplication: XCUIApplication) {
+        let predicate = NSPredicate(format: "label CONTAINS[c] %@", "")
```
amanjeetsingh150, line 7: What would be difference in filtering if we don't have appName in this implementation?

artem888: this would only dismiss Notification Permission alerts and wouldn't work if app is localized to non-english language, but those are the drawback we're currently aware of.

berikv (approved):

PR #986, opened 2023-04-12
fix: Error shows on flow runs "Failed to reach out XCUITest Server"

artem888:
## Proposed Changes

Throwing a custom exception inside network interceptor makes OkHttpClient library logging it. To fix it we need to return an error response instead.


berikv (approved):
👍

PR #988, opened 2023-04-13
Enhanced debug output

ArthurSav:
## Proposed Changes

This PR enhances debug output. 

More specifically, it adds the following:
- Flow failure: Failure reason, command failure screenshot, maestro logs ✅
- Test suite (folder) failure: Failure reason(s), command failure screenshot(s), maestro logs ✅

### Before
(single flow)
simple-flow

(multiple flows)
testsuite

### After
(single flow)
simple-flow-updated

(multiple flows)
testsuite-updated

**Note: `maestro cloud` CLI output is still the same. Remains to be implemented with failure reason.**

### Debug Output folder
For each test run, a folder will be generated with the following:
- Failure screenshot, captured after a command fails (if any)
- Maestro logs (Useful to identify unknown or obscure issues)
- Serialized commands for each flow, with metadata. Can be useful for debugging purposes.

Example output for `/Users/yourname/.maestro/tests`
Screenshot 2023-04-13 at 18 14 04

## Testing
Local / wikipedia flows


simon-gilmurray:
This is super useful, thanks for adding! One thing that would also greatly help if possible is to include line number somewhere on the error.

venkat2789:
Super cool enhancement. Wonder if this will make any changes to the junit report.

ArthurSav:
Thanks for the feedback folks.

@simon-gilmurray Unfortunately it was not straight forward to add the line number. Will have to keep it in mind for the next improvement.

@venkat2789  Added some JUNIT improvements. More specifically, Maestro Cloud junit output will now include failure reason as well.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -400,12 +402,14 @@ class ApiClient(
     }
 }
 
+@JsonIgnoreProperties(ignoreUnknown = true)
```
amanjeetsingh150, line 405: Why this option?

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -0,0 +1,102 @@
...
+    init {
+
+        // folder
+        val dateFormatter = DateTimeFormatter.ofPattern(dateFormat)
+        val folderName = dateFormatter.format(LocalDateTime.now())
+        path = Paths.get(System.getProperty("user.home"), ".maestro", "tests", folderName)
+
+        Files.createDirectories(path)
+
+        // gson
```
amanjeetsingh150, line 35: very small nit: This is actually not gson, but jackson

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -0,0 +1,102 @@
...
+            val name = "screenshot-$status-${it.timestamp}-(${flowName}).png"
+            val file = File(path.absolutePathString(), name)
+
+            it.screenshot.copyTo(file)
+        }
+    }
+
+    fun saveLogs() {
+        // maestro logs
+        DebugLogStore.copyTo(File(path.absolutePathString(), "maestro.log"))
```
amanjeetsingh150, line 67: There is a bug in the `maestro bugreport` flow. When we do a maestro bug report it creates an empty log file can you verify if it's not interfering with this functionality?

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -74,16 +108,35 @@ object MaestroCommandRunner {
...
                 refreshUi()
             },
             onCommandFailed = { _, command, e ->
+                debugCommands[command]?.let {
+                    it.status = CommandStatus.FAILED
+                    it.calculateDuration()
+                    it.error = e
+                }
+
+                takeDebugScreenshot(CommandStatus.FAILED)
```
amanjeetsingh150, line 134: Do we take screenshot only when command fails?

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -43,25 +47,55 @@ object MaestroCommandRunner {
...
 
         val commandStatuses = IdentityHashMap<MaestroCommand, CommandStatus>()
         val commandMetadata = IdentityHashMap<MaestroCommand, Orchestra.CommandMetadata>()
 
+        // debug
+        val debugCommands = debug.commands
+        val debugScreenshots = debug.screenshots
+
+        fun takeDebugScreenshot(status: CommandStatus): File? {
+            val containsFailed = debugScreenshots.any { it.status == CommandStatus.FAILED }
```
amanjeetsingh150, line 62: Is this function intended to be called when flow fails?

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -124,14 +129,77 @@ class TestSuiteInteractor(
         var flowStatus: FlowStatus
         var errorMessage: String? = null
 
+        // debug
+        val debug = FlowDebugMetadata()
+        val debugCommands = debug.commands
+        val debugScreenshots = debug.screenshots
+
+        fun takeDebugScreenshot(status: CommandStatus): File? {
+            val containsFailed = debugScreenshots.any { it.status == CommandStatus.FAILED }
```
amanjeetsingh150, line 138: This method seems to be a duplicate of another`takeDebugScreenshot` in MaestroCommandRunner. Can we common this implementation?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
ArthurSav: Currently any api changes will break the client. This option will future proof such cases

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
ArthurSav: This is actually a blocker that i'm investigating. I have encountered the issue myself.

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
ArthurSav: yes

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
ArthurSav: It's a convenience method for local references, doesn't make much sense to abstract it. The actual issue imo is that we have 2 difference entry points for running commands - but that's a problem for another day.

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
ArthurSav: Yes. Initially, it was for more steps but it created too many screenshots. For now it's for failure only.

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
ArthurSav: Tested and it seems to be working

maestro-cli/src/main/java/maestro/cli/App.kt
```diff
@@ -112,7 +115,7 @@ fun main(args: Array<String>) {
                 ex.stackTraceToString()
             }
 
-            logger.info(message)
+            logger.error(message)
```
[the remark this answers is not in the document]
ArthurSav: Because it's an error

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -0,0 +1,133 @@
...
+        // todo - device logs
+    }
```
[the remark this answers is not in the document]
ArthurSav: See the PRs description images

ArthurSav:
Hey everyone, will try to wrap up this week and ship it with the next update. Need to tackle a blocker with logging before merging.

davydspotnana:
Any updates?

maestro-cli/build.gradle
```diff
@@ -45,9 +45,6 @@ dependencies {
     implementation(libs.ktor.client.cio)
     implementation(libs.jarchivelib)
 
-    // Silence logs from WireMock
-    implementation(libs.slf4j.nop)
```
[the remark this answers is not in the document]
ArthurSav: Didn't see any while running a flow.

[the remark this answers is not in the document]
axelniklasson: Checking the code base it seems to be used primarily for the old version of experimental network mocking -- if we don't need to drop the dependency for now I think we should keep it in until we decide to deprecate it fully. Another option is of course to run a Flow using the network mocking feature and check whether logs are not silenced correctly.

maestro-cli/src/main/java/maestro/cli/report/JUnitTestSuiteReporter.kt
```diff
@@ -40,9 +40,9 @@ class JUnitTestSuiteReporter(
                                 testCases = suite.flows
                                     .map { flow ->
                                         TestCase(
-                                            id = flow.fileName ?: flow.name,
+                                            id = flow.name,
                                             name = flow.name,
-                                            classname = flow.fileName ?: flow.name,
+                                            classname = flow.name,
```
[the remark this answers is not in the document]
ArthurSav: Cloud runs and local runs produce different names. This is done for consistency of output.
This will indeed change naming behaviour, depending on user configuration.

maestro-cli/build.gradle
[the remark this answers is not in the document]
ArthurSav: >    implementation(libs.slf4j.nop)
This interferes with logging on CLI so it can't stay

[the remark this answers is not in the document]
axelniklasson: Could you then please check that this does not cause any unwanted side effects? Thanks!

axelniklasson (approved):
LGTM with one pending comment, approving assuming that will be addressed. Let's also ensure that this is good to go by @amanjeetsingh150 before merging!

maestro-cli/build.gradle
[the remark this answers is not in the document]
ArthurSav: Already tested multiple times.

The reason `implementation(libs.slf4j.nop)` was added was to silence logs in CLI. However, in this case we actually want to log things from the CLI.

If you look at the implementation you'll notice that for CLI logs will be written to file only and not console. This is done programmatically upon CLI init.

Can't think of a side effect.

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -31,37 +33,69 @@ import maestro.orchestra.MaestroCommand
...
+
+            // Avoids duplicate failed images from parent commands
+            if (containsFailed && status == CommandStatus.FAILED) {
+                return null
+            }
```
amanjeetsingh150, line 67: Will this coincide with what Felipe worked on recently surfacing screenshots?

amanjeetsingh150 (approved):

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
ArthurSav: This will work only on CLI. Maestro Cloud uses Orchestra directly to run commands which bypasses MaestroCommandRunner.

PR #989, opened 2023-04-13
Missing message response

amanjeetsingh150:
## Proposed Changes

* Add message and response body in case of shutting down studio as well failures while executing XCTest requests
* Set a timeout for hideKeyboard if the swipe or other calls with hierarchy, activeApp underneath fails for some reason

## Testing
Tested locally with known apps.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -50,12 +50,16 @@ import org.slf4j.LoggerFactory
 import util.XCRunnerCLIUtils
 import java.io.File
 import java.nio.file.Files
+import java.util.concurrent.Executors
+import java.util.concurrent.Future
+import java.util.concurrent.TimeUnit
 import kotlin.collections.set
 
 class IOSDriver(
     private val iosDevice: IOSDevice,
 ) : Driver {
 
+    private val executor by lazy { Executors.newSingleThreadExecutor() }
     private val deviceInfo by lazy {
         iosDevice.deviceInfo().expect {}
     }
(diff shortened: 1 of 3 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -379,24 +383,35 @@ class IOSDriver(
...
+                end = Point(widthPercentToPoint(0.5), heightPercentToPoint(0.47)),
+                durationMs = 50,
+            )
 
-        swipe(
-            start = Point(widthPercentToPoint(0.5), heightPercentToPoint(0.5)),
-            end = Point(widthPercentToPoint(0.47), heightPercentToPoint(0.5)),
-            durationMs = 50,
-        )
+            if (isKeyboardHidden()) {
```
felipevolpone, line 397: super nit: to keep consistency, let's make it the same as in line 389, as a one-line if like `if (isKeyboardHidden()) return@submit`

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -379,24 +383,35 @@ class IOSDriver(
...
+                start = Point(widthPercentToPoint(0.5), heightPercentToPoint(0.5)),
+                end = Point(widthPercentToPoint(0.47), heightPercentToPoint(0.5)),
+                durationMs = 50,
+            )
+
+            waitForAppToSettle(null)
+        }
+        try {
+            future.get(5, TimeUnit.SECONDS)
+        } catch (exception: Exception) {
```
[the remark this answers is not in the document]
amanjeetsingh150: Done ✅

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Done ✅

PR #992, opened 2023-04-13
feat: new assertOutgoingRequest

felipevolpone:
This PR implements a new command to assert outgoing requests from the app. This will be used to check if the app does some network calls for certain APIs/backend.
The mock-server will be used to fetch and match the events stored in Maestro Cloud.

This is the command definition:

```
- assertOutgoingRequest:
    path: "/login"
    headersPresent:
      - "cache-control"
    headersAndValues:
      - "Content-Type": "*application/json*"
      - "Authorization": "Bearer .*"
    httpMethodIs: "GET"
    requestBodyContains: '"foo": "bar"'
```

Important to say that this command will only work if the user have set the Maestro SDK properly (https://maestro.mobile.dev/advanced/experimental/maestro-mock-server/getting-started#setup)

How it works:
- Maestro creates a sessionId (UUID)
- Maestro injects the sessionId in the device in launchApp, setting it as env var
  - I know that a user could re-launch the app in the flow, making the uuid be recreated, but that's ok for now as we want to launch the feature soon and get feedback from the users
-  SDK read the env var and initializes the SDK


maestro-client/src/main/java/maestro/OutgoingRequest.kt
```diff
@@ -0,0 +1,42 @@
...
+    val headersAndValues: Map<String, String> = emptyMap(),
+    val httpMethodIs: String? = null,
+    val requestBodyContains: String? = null,
+)
+
+object AssertOutgoingRequestService {
+
+    private val REGEX_OPTIONS = setOf(RegexOption.IGNORE_CASE, RegexOption.DOT_MATCHES_ALL, RegexOption.MULTILINE)
+
+    fun assert(events: List<MockEvent>, rules: OutgoingRequestRules): List<MockEvent> {
```
axelniklasson, line 18: Big nit: shall we use a different name since `assert` is a reserved word in Java?

axelniklasson (approved):
LGTM!

- Did you verify that the mock server UI is still working as intended with these changes?
- Are there changes required on Maestro Cloud's side before merging this?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -180,6 +181,8 @@ class AndroidDriver(
             val apkFile = AndroidAppFiles.getApkFile(dadb, appId)
             val manifest = apkFile.asManifest()
             runCatching {
+                val sessionUUID = sessionId ?: UUID.randomUUID()
+                dadb.shell("setprop debug.maestro.sessionId $sessionUUID")
```
amanjeetsingh150, line 185: Do network mocking and all related features currently support emulator or can be physical device as well?

maestro-ios-driver/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -71,7 +69,7 @@
 			<key>DYLD_FRAMEWORK_PATH</key>
 			<string>__TESTROOT__/Debug-iphonesimulator:__TESTROOT__/Debug-iphonesimulator/PackageFrameworks:__PLATFORMS__/iPhoneSimulator.platform/Developer/Library/Frameworks</string>
 			<key>DYLD_INSERT_LIBRARIES</key>
-			<string>__SIMRUNTIMEROOT__/usr/lib/libMainThreadChecker.dylib</string>
+			<string>/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 15.4.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libMainThreadChecker.dylib</string>
```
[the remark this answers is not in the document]
amanjeetsingh150: Yup let's ignore these changes, this seems to be modified sometimes building iOS driver.

berikv (approved):

maestro-client/src/main/java/maestro/mockserver/MockInteractor.kt
```diff
@@ -46,6 +46,7 @@ class MockInteractor(
         .build()
 
     fun getCachedAuthToken(): String? {
+        if (!System.getenv("MAESTRO_CLOUD_API_KEY").isNullOrEmpty()) return System.getenv("MAESTRO_CLOUD_API_KEY")
```
axelniklasson, line 49: Nit: you can use `EnvUtils.maestroCloudApiKey()` here

felipevolpone:
@axelniklasson thanks for reviewing this PR

> Did you verify that the mock server UI in Studio is still working as intended with these changes?

Yes, it works 👍 

> Are there changes required on Maestro Cloud's side before merging this?

Yes, there are. I'm opening a PR for that on Maestro Cloud's repo

maestro-client/src/main/java/maestro/mockserver/MockInteractor.kt
felipevolpone: EnvUtils is in another package and it would need to add it as a dependency, that's why didn't do it.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
felipevolpone: I'm not sure if physical devices will work as well. Didn't tested on them

felipevolpone:
@axelniklasson @berikv I've done some changes to make it work for iOS in Maestro Cloud. Could you please review this PR again?

maestro-ios/src/main/java/ios/idb/IdbIOSDevice.kt
```diff
@@ -394,6 +396,7 @@ class IdbIOSDevice(
                         if (launchArguments.isNotEmpty()) {
                             appArgs.addAll(launchArguments)
                         }
+                        sessionId = maestroSessionId?.toString() ?: UUID.randomUUID().toString()
```
berikv, line 399: IDB does not support sessionId. idb.proto is not for IDB, which makes everything confusing. We'll improve that soon. For now lets not pass it here, because IDB will break if we do (this code is not unused anyway).

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -190,7 +190,21 @@ object LocalSimulatorUtils {
         deviceId: String,
         bundleId: String,
         launchArguments: List<String> = emptyList(),
+        sessionId: String,
     ) {
+        runCommand(
```
berikv, line 195: The runCommand can be conditional If the sessionId is null.

maestro-sdk/ios/Maestro SDK/Maestro SDK/MaestroMockServerSdk.swift
```diff
@@ -7,7 +7,14 @@ public class MaestroMockServerSdk {
...
         
-        let sessionId = NSUUID().uuidString
+        let sessionIdFromEnvVar = ProcessInfo.processInfo.environment["MAESTRO_SESSION_ID"]
+
+        var sessionId = NSUUID().uuidString
+        if (sessionIdFromEnvVar != nil) {
+            if let uuid = UUID(uuidString: sessionIdFromEnvVar!) {
+                sessionId = uuid.uuidString
+            }
+        }
```
berikv, line 17: This can be written more swift-y:
```
let sessionId: String
if let sessionIdFromEnvVar = ProcessInfo.processInfo.environment["MAESTRO_SESSION_ID"],
    let uuid = UUID(uuidString: sessionIdFromEnvVar) {
    sessionId = uuid.uuidString
} else {
    sessionId = UUID().uuidString
}
```

berikv (approved):
Overall looks good!
I left some suggestions for improving further.

PR #996, opened 2023-04-14
Changed upgrade instructions to include command instead of url

bobpozun:
## Proposed Changes
A simple quality of life change to update the upgrade instructions to contain the command for upgrading instead of a url to a page that has the command to upgrade, streamlining that process a bit.

## Testing
This is a simple text change

## Issues Fixed
None


axelniklasson (approved):
Thanks for the contribution @bobpozun ✅

PR #1008, opened 2023-04-20
[iOS] Resolve swipe flakiness

artem888:
## Proposed Changes

Often when there is a running animation XCTest framework gets into the idle state. as a result it either crashes runner app or not being able to resolve it with provided app id. This becomes especially visible for ReactNative apps with complex view hierarchy in combination with scroll commands.

As a fix swizzling `waitForQuiescenceIncludingAnimationsIdle` method to an empty implementation is done.

## Testing

- [x] test all the flows with maestro-flow-tester


maestro-ios-xctest-runner/maestro-driver-iosUITests/maestro_driver_iosUITests.swift
```diff
@@ -2,20 +2,29 @@ import XCTest
...
         // In UI tests it is usually best to stop immediately when a failure occurs.
         continueAfterFailure = false
 
         // In UI tests it’s important to set the initial state - such as interface orientation - required for your tests before they run. The setUp method is a good place to do this.
     }
-
-    override func tearDownWithError() throws {
-        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    
+    @objc func replace() {
```
berikv, line 24: Lets call this `replace_waitForQuiescenceIncludingAnimationsIdle`, to make the swizzle a bit less confusing.

berikv (commented):
Swift change looks good to me.
Why are there so many changes to the .app?

maestro-ios-xctest-runner/maestro-driver-iosUITests/maestro_driver_iosUITests.swift
berikv: And inside a comment:

```
    // Disable waiting for quiescence
```

berikv (approved):

PR #1015, opened 2023-04-24
feat: add reverse indexing

igorsmotto:
## Proposed Changes
- Add reverse indexing. Example:
```yaml
# Tapping the second to last element with the same text
- tapOn:
   text: "text"
   index: -2 
```

## Testing
- Updated Wikipedia Advanced Android Sample to use the feature. 

## Limitation
- Both for Android and iOS, this would only work for **visible** elements. More specifically, **rendered** elements.


ArthurSav:
@igorsmotto Shall we close this one?

pravz71:
Hi @igorsmotto @ArthurSav why did we close this?
if there is an alternative solution for this, could you please let me know.

PR #1020, opened 2023-04-26
[iOS] Notification permissions do not work

artem888:
## Proposed Changes
Map unknown permission values to .unknown instead of failing the setPermissions xctest request

## Testing
tested locally:
- [x] default enable all
- [x] single notifications permission enabled/disabled
- [x] mixed permissions


maestro-ios-driver/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -69,7 +71,7 @@
 			<key>DYLD_FRAMEWORK_PATH</key>
 			<string>__TESTROOT__/Debug-iphonesimulator:__TESTROOT__/Debug-iphonesimulator/PackageFrameworks:__PLATFORMS__/iPhoneSimulator.platform/Developer/Library/Frameworks</string>
 			<key>DYLD_INSERT_LIBRARIES</key>
-			<string>/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 15.4.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libMainThreadChecker.dylib</string>
+			<string>__SIMRUNTIMEROOT__/usr/lib/libMainThreadChecker.dylib</string>
```
amanjeetsingh150, line 74: Same! is this intentional?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/SetPermissionsRequest.swift
```diff
@@ -4,6 +4,11 @@ enum PermissionValue: String, Codable {
     case allow
     case deny
     case unset
+    case unknown
+        
+    init(from decoder: Decoder) throws {
+        self = try PermissionValue(rawValue: decoder.singleValueContainer().decode(RawValue.self)) ?? .unknown
+    }
```
amanjeetsingh150, line 11: Just to understand the case, was permission handler not handling the non-notification permissions and was failing for other shown permissions?

artem888: It was failing when the value was `never` when decoding. and since we only have `allow` `deny` `unset` supported for permissions it makes sense not to fail if some other value is recieved for other permissions

maestro-ios-driver/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -35,6 +35,8 @@
 			<string>__TESTROOT__/Debug-iphonesimulator/maestro-driver-iosUITests-Runner.app</string>
 			<string>__TESTROOT__/Debug-iphonesimulator/maestro-driver-iosUITests-Runner.app/PlugIns/maestro-driver-iosUITests.xctest</string>
 		</array>
+		<key>DiagnosticCollectionPolicy</key>
+		<integer>1</integer>
```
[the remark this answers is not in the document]
amanjeetsingh150: I would say let's take whatever there is on production to be safe.

PR #1025, opened 2023-04-27
[iOS] runner app uninstall fails when multiple iOS simulators are launched

artem888:
## Proposed Changes

when multiple simulators are launched, simctl command with booted parameter often fail

## Testing
- [x] maestro flow tester flows locally


berikv (approved):
Great work! 🎉

PR #1040, opened 2023-05-03
Adds long value to enable timestamp like values

amanjeetsingh150:
## Proposed Changes

Adds long value to enable passing timestamp like values on Android

## Testing
- [x] Local
- [x] Staging


maestro-client/src/test/java/maestro/android/AndroidLaunchArgumentsTest.kt
```diff
@@ -14,7 +14,8 @@ class AndroidLaunchArgumentsTest {
             "isMaestro" to true,
             "cartValue" to 4,
             "cartValueDouble" to 4.4,
-            "cartColor" to "Hello this is cart value which is orange"
+            "cartColor" to "Hello this is cart value which is orange",
+            "cartTimeStamp" to 1683113805263
```
axelniklasson, line 18: Could we add another test for `0` to ensure that gets parsed correctly?

axelniklasson (approved):

maestro-client/src/test/java/maestro/android/AndroidLaunchArgumentsTest.kt
amanjeetsingh150: Done ✅

Issue #1044, opened 2023-05-05
Feature: `defineSelectors` command to support Screen/Page Object design pattern


larryng:
Additional context in [Slack](https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1683300725606219)

I have working locally a command `defineSelectors` intended to help implement the Screen/Page Object pattern. It enables defining custom selectors that you can put in a subflow for loading in test flows. For example:
```yaml
# screens/login.yaml
...
- defineSelectors:
    Login.ForgotPassword: "Forgot password"
    Login.SignIn:
      id: id_sign_in
```

```yaml
# flow.yaml
...
- runFlow: screens/login.yaml  # Load the selectors
- assertVisible:
    selector: Login.ForgotPassword
- tapOn:
    selector: Login.SignIn
```

I'm aware that currently there is a JS approach for doing this, something like:
```js
// screens/login.js
output.Login = {
  ForgotPassword: "Forgot Password",
  SignIn: "id_sign_in"
}
```

```yaml
# flow.yaml
...
- runScript: screens/login.yaml  # Load the selectors
- assertVisible: ${output.Login.ForgotPassword}
- tapOn:
    id: ${output.Login.SignIn}
```

Advantages of `defineSelectors` would be:
 * It's more concise. No need for `output` or `${}`.
 * You can compose multiple selectors.
 * Less leaky. JS approach requires you to know what selector a variable maps to (text, ID, etc.)
 * Better readability.

I'll leave the Disadvantages as an exercise for the reader. 😉 

I can push a PR but would love to gather comments first.

simon-gilmurray:
@larryng Thanks for suggesting this, it looks pretty cool! I was looking for a Page Object Model solution for a while before Maestro added JS integration which made it possible. 

It does read cleaner than having to prefix every element with `${output.` so from that perspective it is an improvement.

**Questions:**

1. Is there any way to improve organisation/readability with your approach? Currently it reads as if it will just be a long list. One current benefit of the JS way is you can group via arrays or nesting e.g. I can reference `${output.cards.allCards.createNewBtn}` using:

```
output.cards = {
    allCardsBtn: 'all_cards_button'

    // All Cards Screen
    allCards: {
        createNewBtn: 'btn_create'
    }
```


2. One thing that would be extremely useful if possible is to provide an id and text to a selector as a combination. I've encountered some screens that have multiple of the same IDs but the text differs. It would be great if we could have something like:

```
- defineSelectors:
[...]

PR #1055, opened 2023-05-09
[WIP] Add `defineSelectors` command

larryng:
WIP. Working proof of concept, but API is subject to change and it's insufficiently tested.

### <samp>🤖 Generated by Copilot at 82fe314</samp>

This pull request adds a new feature to the `maestro` framework that allows defining and using custom element selectors in YAML and Kotlin commands. Custom selectors can reference other selectors and filter elements based on nested criteria. The feature is implemented by a new `DefineSelectorsCommand` class and a corresponding `YamlDefineSelectors` class. The feature is tested by a new integration test case and a YAML file.

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/1044

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -650,6 +650,25 @@ data class DefineVariablesCommand(
 
 }
 
+data class DefineSelectorsCommand(
+    val selectors: Map<String, ElementSelector>,
+) : Command {
+
+    override fun description(): String {
+        return "Define selectors"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): DefineSelectorsCommand {
+        return copy(
+            selectors = selectors.mapValues { (_, value) ->
+                value.evaluateScripts(jsEngine)
+            }
+        )
+    }
+
+    override fun visible(): Boolean = false
+}
+
 data class RunScriptCommand(
     val script: String,
     val env: Map<String, String> = emptyMap(),
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/ElementSelector.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/ElementSelector.kt
@@ -23,6 +23,7 @@ import maestro.js.JsEngine
 import maestro.orchestra.util.Env.evaluateScripts
 
 data class ElementSelector(
+    val selector: String? = null,
     val textRegex: String? = null,
     val idRegex: String? = null,
     val size: SizeSelector? = null,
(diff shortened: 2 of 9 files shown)


dmitry-zaitsev:
That is a neat API, I personally like it! @axelniklasson and the team will be making the final call on the proposal

One high-level comment - we need to be clear about scoping. For example, if a user will run `defineSelectors` in sub-flow (via `runFlow`), is it expected that it will override a top-level definition? I guess this is acceptable, but it could be tricky to debug if such problem occurs. One way to solve it is to move custom selectors out of the flow itself, for example in `config.yaml`. This way we guarantee that selectors are only defined once and also make it easier to use as you don't need to re-define them with each flow.

ArthurSav:
Also like the API and I think we can proceed after a few changes

> One way to solve it is to move custom selectors out of the flow itself

This will have a double benefit of keeping flows clear of clutter and make it easier to read. With that said, I don't think it should be located in config.yaml.

I propose we allow imports via an import command, allow definitions inside a flow as well and **make it clear via docs that all imports live under one scope**. If too much confusion arrises from duplicate imports and different scopes, we can throw a warning via maestro.


flow.yaml
```yaml
appId: com.example.app
---
- includeSelectors: "selectors.yaml"
- defineSelectors:
    SomeSelector:
      id: some_selector
- assertVisible:
    selector: LoginScreen.EmailText
- assertNotVisible:
    selector: PasswordText
    text: Password
- tapOn:
    selector: LoginButton
```

selectors.yaml
```yaml
- defineSelectors:
    LoginScreen.EmailText:
      id: id_email_text
      text: Email
    LoginButton: Login
- defineSelectors:
    PasswordText:
      id: id_password_text
```
@larryng What do you think?

larryng:
@ArthurSav Sorry for the delayed response. I leave it to y'all to decide on the best path forward -- please don't let me be a blocker for this.

My thoughts:
 * nit: A separate `includeSelectors` command doesn't seem necessary as we could overload `defineSelectors` to accept a filepath, e.g. `defineSelectors: "selectors.yaml"`
 * It's not clear to me what would happen if a command used an undefined selector. Would it error when configuring the flow, or when the command runs? If it's the former, then we shouldn't be defining selectors via commands. If it's the latter, then there's no difference to what I already have.
 * IMO what I already have is consistent with how JS and `output` works (I think) and consistency is good, since if you have an understanding of how the JS scoping works, then you understand how custom selectors work as well. You'd also understand the downsides as well: you can erroneously reference an undefined `output` property just as you can an undefined custom selector. Ditto for overwriting accidentally, etc.

axelniklasson:
Thanks for taking the time to raise this PR @larryng -- we really value the time that went into this. We won't be moving forward with this right now due to resourcing constraints, but we appreciate your effort and the proposal.

PR #1060, opened 2023-05-11
Add deprecation notices

axelniklasson:
## Proposed Changes

copilot:summary

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/mockserver/MockServerDeployCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/mockserver/MockServerDeployCommand.kt
@@ -2,6 +2,7 @@ package maestro.cli.command.mockserver
 
 import maestro.cli.api.ApiClient
 import maestro.cli.cloud.CloudInteractor
+import maestro.cli.view.red
 import picocli.CommandLine
 import picocli.CommandLine.Command
 import picocli.CommandLine.Option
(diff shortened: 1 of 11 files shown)


ArthurSav (approved):

erawhctim:
Is there more detail/documentation anywhere on why this is deprecated/being removed, and if there is any future plans to support network mocking in a different way?

axelniklasson:
@erawhctim I shared the following note with the community [in our public Slack here](https://mobile-dev-inc.slack.com/archives/C01A2QJL51T/p1683811831135009).

> At [mobile.dev](http://mobile.dev/) we are continuously innovating and investing into various Maestro features, some of which become available as experimental features to be tested along with this community. The plan with this is to get feedback from the community and understand if we are solving the right problems or if we should focus our efforts elsewhere. After careful consideration we have decided to deprecate the following experimental products from our offering:
> - Network Mocking
> - Maestro Mock Server
> - Maestro SDK

> These products will be phased out over the coming maestro versions and no further development will be made. We know that some of may have been using these features and we apologize for any inconvenience this may cause. However, we need to focus our resources on other things that bring more value to the community, like physical iOS device support.

> We want to thank everyone who has provided feedback and support for these products - your input has been very valuable and we we will continue to listen to this amazing community to guide the future of Maestro’s product suite.

For now there is no plan to focus on network mocking in the immediate future as the team is heads down focusing on other features that are deemed more important, but it is not impossible that we get back to network mocking in the future when bandwidth allows.

erawhctim:
Exactly what I was looking for, thank you for re-posting!

Issue #1061, opened 2023-05-11
[v1.27.0] `inputText` failing on `secureTextEntry`


KrisLau:
**Describe the bug**
`inputText` failing on `secureTextEntry`. The Input is showing a  "Strong Password" message and text entry just becomes disabled

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Start the simulator
2. Run `maestro test` with `- inputText: 'whatever"` on a secure text entry Input
3. The Flow fails at the text input

**Expected behavior**
Enter the password

**Screenshots**
![image](https://github.com/mobile-dev-inc/maestro/assets/46275463/e86f57f7-d2a8-4f97-b609-2c52a4eae5f8)

**Environment information (please complete the following information):**

- Maestro version: v1.27.0
- Platform: iOS 16.2
- Device model and OS version: iPhone 11
- Simulator
- Host: Mac Mini

**Workspace (if applicable)**
N/A

**Bug report output (if applicable)**
N/A

**Additional context**
N/A

filipgerat:
Facing the same issue. Android does not suffer from this. It also doesn't help to set the secureTextEntry property to false

andrejguran:
I have the same issue and can't get through login/signup step

heba-mekawi:
Same here, any recommendation to fix it please ?

orbiteleven:
It's not great, but as a workaround I'm setting those fields to be normal text fields when I detect I'm in testing mode.

thalmann-christof:
As a workaround, you could disable auto-filling in your simulator.
_Inside your simulator: Settings → Passwords → Type in a random password → Password options → Disable toggle_

In Maestro Cloud, it seems to be disabled by default, so there are no issues there.

KrisLau:
@axelniklasson

solidcell:
For anyone else running into this, I have two solutions:

Maestro
```
- swipe:
    direction: RIGHT
- tapOn: "Settings"
# ^^ that could probably be done via some command to open Settings.app's app ID
- tapOn: "Passwords"
- inputText: "a" # anything works here as the password
- pressKey: Enter
- tapOn: "Password Options"
- runFlow:
    when:
      visible: "1" # Only toggle when currently ON. This isn't great, but I don't know of a way to be more explicit about which toggle it is. Luckily, there's just the one (for now...)
    commands:
      - tapOn: "AutoFill Passwords"
```

That's obviously not great.

[...]

PR #1094, opened 2023-05-19
[iOS] Resolve inputText flakiness

artem888:
## Proposed Changes

This PR implements a fix for inputText flakiness on iOS. The main reason for the flakiness is that in case of typingSpeed being too fast characters are being skipped. Can be reliably reproduced when typingSpeed is set to 30 chars per second.

New logic has several improvements:
- taps automatically on the cell, so additional tapOn command is not needed anymore. supported params (`id`, `accessibilityText`, `point`)
- in case of text length equal or less 50 characters inputText would be used as normally with reduced typingSpeed (10 characters per second)
- in case of text length exceeding 50 characters new logic of copy-paste will be used

before:
```
- tapOn:
    id: "login-email-input"
- inputText: "user@example.com"
```

after:
```
- inputText:
    text: "user@example.com"
    id: "login-email-input"
```

short text (<50 chars) input video:

https://github.com/mobile-dev-inc/maestro/assets/961175/3e778ff5-8d5d-414a-9f8b-e7de6cb51a7a

long text (>50 chars) input video:

https://github.com/mobile-dev-inc/maestro/assets/961175/78fc8dbf-55c8-424a-bb8c-e11f0cecda25

## Testing
maestro-flow-tester tests:
- [x] UIKit app
- [x] ReactNative app
- [x] Flutter app

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/395


igorsmotto:
Hey! Great to see this PR, I have some general questions:
1. Is this API change required?
2. How it works for other platforms (i.e.: Android)?
3. Does this solution works for secureTextEntry? Related: https://github.com/mobile-dev-inc/maestro/issues/1061?

artem888:
> Hey! Great to see this PR, I have some general questions:
> 
> 1. Is this API change required?
> 2. How it works for other platforms (i.e.: Android)?
> 3. Does this solution works for secureTextEntry? Related: [[v1.27.0] `inputText` failing on `secureTextEntry` #1061](https://github.com/mobile-dev-inc/maestro/issues/1061)?

1. yes, it is required, because we need a textfield element to doubleTap on it on iOS (for copy-paste logic)
2. on Android there will be no change, besides when the new API is used we do the tap on the cell automatically
3. I've tested with secure text-fields on iOS, since this change is iOS only and it works fine there

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -327,21 +327,43 @@ data class AssertConditionCommand(
...
 
+data class InputTextCommandV2(
+    val text: String,
+    val accessibilityText: String? = null,
+    val id: String? = null,
+    val point: String? = null,
+    val pasteTitle: String? = null
+) : Command {
+    override fun description(): String {
+        return "Input text v2 $text"
```
axelniklasson, line 353: This is being presented to users, so maybe we should remove the `v2` part? We could also update it to add something indicating that it is inputting text to an element directly, thus indicating it is the v2 version.

```suggestion
        return "Input text $text"
```

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -39,6 +40,7 @@ import maestro.orchestra.yaml.YamlCommandReader
 import maestro.toSwipeDirection
 import maestro.utils.MaestroTimer
 import maestro.utils.StringUtils.toRegexSafe
+import java.awt.SystemColor.text
```
axelniklasson, line 43: Is this import required?

artem888:
> Awesome work on this Artem!
> 
> Some follow up questions:
> 
> * Are there changes required on Maestro Cloud?
> * What tests have been made to ensure the changes in the tap implementation does not introduce any regressions?
> 
> Let's also prepare a change request with documentation changes for this command so it is ready for the next release.

- for maestro cloud I need to check, but probably yes
- testing is still WIP, plan is to rerun all maestro-flow-tester tests + I'm doing couple of extra new ones for RN, Flutter and SwiftUI

Leland-Takamine:
Is it possible to query the currently active element to past into instead of changing the inputText API https://github.com/appium/appium/issues/16419

I know when @dmitry-zaitsev tried this he ran into some issues where reactnative apps didn't report the selected element correctly, but curious if you explored this route @artem888 ?

artem888:
> Is it possible to query the currently active element to past into instead of changing the inputText API [appium/appium#16419](https://github.com/appium/appium/issues/16419)
> 
> I know when @dmitry-zaitsev tried this he ran into some issues where reactnative apps didn't report the selected element correctly, but curious if you explored this route @artem888 ?

yes, I did consider that as a first option. there is a parameter `hasKeyboardFocus` in the public API, but I didn't manage to make it work with private APIs + you still need to query the element somehow + it is not even supported by public API queries

Leland-Takamine:
Got it thanks for the context Artem!

Leland-Takamine:
@artem888 @axelniklasson Shouldn't the API allow for any element selector as opposed to just finding the element by id?

artem888:
> @artem888 @axelniklasson Shouldn't the API allow for any element selector as opposed to just finding the element by id?

@axelniklasson @Leland-Takamine that was also a deliberate decision, since the textfield position can be changed after it is tapped on. with default element selector many options wouldn't work in that case. text fields are normally well defined elements that have either accessibility_id or some text content on them, so it is fine to use only those two.

Leland-Takamine:
Ah yeah that makes sense Artem. Ok let's see if disabling autocorrect is a better alternative and make a decision from there

dmitry-zaitsev:
> I know when @dmitry-zaitsev tried this he ran into some issues where reactnative apps didn't report the selected element correctly, but curious if you explored this route @artem888 ?

Just to confirm - yes, that was the case. Querying a currently focused view did not work on React Native.

On selector for `inputText`, I understand that this is not a breaking change but just wanted to share a use-case we have where it is _preferrable_ that `inputText` does not have selector - we want to verify that a view has auto-focus and so we are starting to do `inputText` without doing any tapping prior to that.

artem888:
closing this PR, since another one with better solution is coming.

from this PR we can reuse the logic in the future to implement:
- `pasteTextFrom` command
- `doubleTap` command

ArthurSav:
https://github.com/mobile-dev-inc/maestro/pull/1119

PR #1109, opened 2023-05-26
Add screen recording commands

### LINKED ISSUE #336: Screen recording
Is there any thoughts on recording the screen as a video? Is there a way to trigger a screen recording?

tokou:
## Proposed Changes

- This PR adds the `startRecording: <recording_path>` and `stopRecording` commands as suggested in #336 
- The recording logic was already implemented so it was quite easy to add

## Testing

- Wrote an integration test
- Built maestro and ran it locally both on an Android emulator and an iOS simulator

## Issues Fixed

Fixes #336

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -26,6 +26,7 @@ import maestro.Filters.asFilter
 import maestro.FindElementResult
 import maestro.Maestro
 import maestro.MaestroException
+import maestro.ScreenRecording
 import maestro.js.Js
 import maestro.js.JsEngine
 import maestro.networkproxy.NetworkProxy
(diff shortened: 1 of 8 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -557,6 +564,20 @@ class Orchestra(
         return false
     }
 
+    private fun startRecordingCommand(command: StartRecordingCommand): Boolean {
+        val pathStr = command.path + ".mp4"
```
berikv, line 568: Lets use the exact path as specified, not appending the extension.
Otherwise `screenRecording: login.mp4` would be recording `login.mp4.mp4`.

berikv: I think changing the behaviour of screenshot over this does not bring enough value to be worth it.
Lets keep the ".mp4" addition extension as in the PR now to be in line with screenshot.

berikv (approved):

tokou:
@Angelk90 I don't see a reason why it wouldn't.

tokou:
@Angelk90 You can use the local script `maestro` which builds and installs the distribution and then uses it instead of the default binary.

https://github.com/mobile-dev-inc/maestro/blob/cfcebe186c8f5eee24f095188b4eea9e58c529eb/maestro#L5

PR #1119, opened 2023-05-30
[iOS] Resolve inputText flakiness

artem888:
**Problem Statement:**

During our recent investigations, we've identified an issue with the textInput character skipping event. We discovered that characters after the 1st one are consistently skipped, and this behavior seems to be linked to keyboard listener events that occur during these keystrokes. Toggling the ConnectHardwareKeyboard and KeyboardAutocorrection settings alleviates the issue to some extent, but it doesn't completely resolve it. Even with all of these settings disabled, higher typing speeds still result in these characters being skipped.

**Proposed Solution:**

This pull request introduces a two-step approach to mitigate the issue:

- The first character will be input at a slower typing speed, which should prevent the following characters from being skipped.
- The remaining text will be input at the default typing speed.

## Testing
- [x] maestro flow tester all textinput test cases

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/395


ArthurSav (approved):

axelniklasson (approved):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
```diff
@@ -21,11 +24,25 @@ struct InputTextRouteHandler : HTTPHandler {
         do {
             let start = Date()
 
+            // due to different keyboard input listener events (i.e. autocorrection or hardware keyboard connection)
+            // 2nd and 3rd characters are often skipped, so we'll input them with lower typing frequency
```
axelniklasson, line 28: Let's update this comment to reflect the new changes

PR #1146, opened 2023-06-12
Change wait for idle time to min value

amanjeetsingh150:
Modify the wait time to lower value


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -197,7 +197,7 @@ class Service(
      */
     private fun refreshAccessibilityCache() {
         try {
-            uiDevice.waitForIdle(2000)
+            uiDevice.waitForIdle(500)
```
berikv, line 200: Is half a second fast enough?
Adding half a second to each view hierarchy request seems to be excessive. What is the impact on customer flow runtime?

amanjeetsingh150: Yes i ran customer flow with this change and without change. It took almost same time. But i can try with completely removing the wait time as well.

berikv (approved):

Issue #1081, opened 2023-05-17
[Feature Request] Upgrade Javascript Rhino Engine to something more recent 


bernardotwistag:
**Is your feature request related to a problem? Please describe.**
Yes, the problem arises when there is a need to execute more complex external JavaScript, such as tasks that involve reading an email inbox to retrieve a verification code, resetting URLs, or other related tasks. These tasks are necessary to ensure E2E testing flows operate as expected for our end-users. Currently, the 'runScript' function that uses the Rhino JavaScript engine is inadequate, as its limited capabilities do not fully support these tasks.

**Describe the solution you'd like**
To resolve this issue, we propose the incorporation of a more recent and robust JavaScript engine supporting ES6. This engine should be capable of handling complex tasks and offer a wider range of capabilities, surpassing the limitations encountered with the Rhino JavaScript engine. In essence, the new engine should be able to better support the execution of intricate JavaScript required for our E2E testing.

**Describe alternatives you've considered**
No alternatives found yet.

**Additional context**
The implementation of a more sophisticated JavaScript engine is crucial to enhance the efficiency and effectiveness of our E2E testing processes. Not only will it simplify the execution of complex tasks, but it will also ensure a better user experience by more reliably verifying that all systems are functioning as intended.

orbiteleven:
Did you check out https://maestro.mobile.dev/advanced/javascript/make-http-s-requests ? I had the same question, but this seems to do the trick for me.

bsarmento:
I haven't, but I will take a look if it meets my needs -- thanks!

bartekpacia:
Does [GraalJS](https://maestro.mobile.dev/advanced/javascript/graaljs-support) fulfill your needs?

PR #1147, opened 2023-06-13
Implement Experimental GraalJsEngine

Leland-Takamine:
# Proposed Changes

Enables opting in to use GraalJS instead of Rhino JS for javascript evaluation. GraalJS is fully ECMAScript 2022 compliant while Rhino JS only supports ES5.

This feature is experimental for now, but we intend to make GraalJS the default if all goes well.

### Enable GraalJS via Flow config

> **Note**
> This only has an effect if present at the top-level flow. It is ignored when included in any subflows.

```yaml
appId: com.example.app
jsEngine: graaljs
---
# Note: The ?? operator is an example of an ES2020 feature and is not supported by Rhino JS
- inputText: ${null ?? 'foo'}
```

### Enable GraalJS via environment variable

> **Warning**
> This env var will have no effect when running on Maestro Cloud. Use the Flow config above instead to opt into GraalJS on Maestro Cloud.

```shell
export MAESTRO_USE_GRAALJS=true
maestro test my-flow.yaml
```

# GraalJS behavior differences

There are some differences between the new `GraalJsEngine` and the current `RhinoJsEngine` implementation that are worth noting. All of the differences below and some others are documented and tested by the `GraalJsEngineTest` and `RhinoJsEngineTest` tests.

**tldr; The variable scoping with the GraalJS implementation is more consistent and understandable**

<table>
<tr><th>Rhino JS</th><th>GraalJS</th></tr>
<tr>
<td>
<li><b>Can not</b> reuse variable names across scripts within the same Flow</li>
<li><b>Can</b> reuse variable names if redeclaration lives in a subflow</li>
<li><b>Can</b> access variables declared earlier in the Flow in a separate script</li>
<li><code>output</code> variable is shared across all scripts</li>
</td>
<td>
<li>Variables are local to each script</li>
<li><code>output</code> variable is shared across all scripts</li>
</td>
</tr>
</table>

**Some examples**

<table>
<tr><th></th><th>Example</th><th>Rhino JS</th><th>GraalJS</th></tr>

<tr>
  <td><b>Redeclaring variables across scripts</b></td>
  <td>
  <pre>
appId: com.example
---
- evalScript: ${const foo = null}
- evalScript: ${const foo = null}
</pre>
  </td>
  <td>❌ Variable redeclarations throw an error</td>
  <td>✅ Variable names can be reused across scripts</td>
</tr>

<tr>
  <td><b>Accessing variables across scripts</b></td>
  <td>
  <pre>
appId: com.example
---
- evalScript: ${const name = 'joe'}
- assertTrue: ${name === 'joe'}
</pre>
  </td>
  <td>✅ Variables are accessible across scripts</td>
  <td>❌ Variables can't be accessed across scripts</td>
</tr>

<tr>
  <td><b>Special character handling</b></td>
  <td>
  <pre>
appId: com.example.app
env:
  FOO: \
---
- inputText: ${FOO}
</pre>
  </td>
  <td>❌ Single backslash causes an exception</td>
  <td>✅ Single backslash and all other special chars are handled correctly</td>
</tr>

</table>

# Testing

Ran integration tests with `MAESTRO_USE_GRAALJS=true`. 2 tests failed due to known differences between GraalJS and RhinoJS described above. All other tests passed.

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
@@ -1,0 +1,115 @@
+package maestro.js
+
+import okhttp3.OkHttpClient
+import okhttp3.Protocol
+import org.graalvm.polyglot.Context
+import org.graalvm.polyglot.Source
+import org.graalvm.polyglot.Value
+import org.graalvm.polyglot.proxy.ProxyObject
+import java.io.ByteArrayOutputStream
+import java.util.concurrent.TimeUnit
+import java.util.logging.Handler
+import java.util.logging.LogRecord
+
+private val NULL_HANDLER = object : Handler() {
+    override fun publish(record: LogRecord?) {}
+
+    override fun flush() {}
+
+    override fun close() {}
+}
+
+class GraalJsEngine(
+    httpClient: OkHttpClient = OkHttpClient.Builder()
+        .readTimeout(5, TimeUnit.MINUTES)
+        .writeTimeout(5, TimeUnit.MINUTES)
+        .protocols(listOf(Protocol.HTTP_1_1))
+        .build(),
+) : JsEngine {
+
+    private val openContexts = HashSet<Context>()
+
+    private val httpBinding = GraalJsHttp(httpClient)
+    private val outputBinding = HashMap<String, Any>()
+    private val maestroBinding = HashMap<String, Any?>()
+    private val envBinding = HashMap<String, String>()
+
+    private var onLogMessage: (String) -> Unit = {}
+
+    override fun close() {
+        openContexts.forEach { it.close() }
+    }
+
+    override fun onLogMessage(callback: (String) -> Unit) {
+        onLogMessage = callback
+    }
+
+    override fun enterScope() {}
+
+    override fun leaveScope() {}
+
+    override fun putEnv(key: String, value: String) {
+        this.envBinding[key] = value
+    }
+
+    override fun setCopiedText(text: String?) {
+        this.maestroBinding["copiedText"] = text
+    }
+
+    override fun evaluateScript(
+        script: String,
+        env: Map<String, String>,
+        sourceName: String,
+        runInSubScope: Boolean,
+    ): Value {
+        envBinding.putAll(env)
+        val source = Source.newBuilder("js", script, sourceName).build()
+        return createContext().eval(source)
+    }
+
+    private fun createContext(): Context {
+        val outputStream = object : ByteArrayOutputStream() {
+            override fun flush() {
+                super.flush()
+                val log = toByteArray().decodeToString().removeSuffix("\n")
+                onLogMessage(log)
+                reset()
+            }
+        }
+
+        val context = Context.newBuilder("js")
+            .option("js.strict", "true")
+            .logHandler(NULL_HANDLER)
+            .out(outputStream)
+            .build()
+
+        openContexts.add(context)
+
+        envBinding.forEach { (key, value) -> context.getBindings("js").putMember(key, value) }
+
+        context.getBindings("js").putMember("http", httpBinding)
+        context.getBindings("js").putMember("output", ProxyObject.fromMap(outputBinding))
+        context.getBindings("js").putMember("maestro", ProxyObject.fromMap(maestroBinding))
+
+        context.eval("js", """
+            // Prevent a reference error on referencing undeclared variables. Enables patterns like {MY_ENV_VAR || 'default-value'}.
+            // Instead of throwing an error, undeclared variables will evaluate to undefined.
+            Object.setPrototypeOf(globalThis, new Proxy(Object.prototype, {
+                has(target, key) {
+                    return true;
+                }
+            }))
+            function json(text) {
+                return JSON.parse(text)
+            }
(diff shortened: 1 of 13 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -165,6 +168,15 @@ class Orchestra(
         return true
     }
 
+    @Synchronized
+    private fun initJsEngine(config: MaestroConfig?) {
+        if (this::jsEngine.isInitialized) {
+            jsEngine.close()
+        }
+        val shouldUseGraalJs = config?.ext?.get("jsEngine") == "graaljs" || System.getenv("MAESTRO_USE_GRAALJS") == "true"
+        jsEngine = if (shouldUseGraalJs) GraalJsEngine() else RhinoJsEngine()
```
Leland-Takamine, line 177: Note, we are now creating a new instance of the JsEngine on every flow execution. Before we were caching the JsEngine instance. This shouldn't have any significant impact on performance.

Leland-Takamine:
Addresses #1081

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -0,0 +1,115 @@
...
+        env: Map<String, String>,
+        sourceName: String,
+        runInSubScope: Boolean,
+    ): Value {
+        envBinding.putAll(env)
+        val source = Source.newBuilder("js", script, sourceName).build()
+        return createContext().eval(source)
+    }
+
+    private fun createContext(): Context {
```
Leland-Takamine, line 70: Creates a new context for each script evaluation but this is a fast operation based on my adhoc testing

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -0,0 +1,115 @@
...
+            .logHandler(NULL_HANDLER)
+            .out(outputStream)
+            .build()
+
+        openContexts.add(context)
+
+        envBinding.forEach { (key, value) -> context.getBindings("js").putMember(key, value) }
+
+        context.getBindings("js").putMember("http", httpBinding)
+        context.getBindings("js").putMember("output", ProxyObject.fromMap(outputBinding))
```
Leland-Takamine, line 91: `output` and `maestro` are the two variables (aka "bindings") shared across scripts

maestro-test/src/test/kotlin/maestro/test/GraalJsEngineTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.google.common.truth.Truth
+import com.google.common.truth.Truth.assertThat
+import maestro.js.GraalJsEngine
+import org.graalvm.polyglot.PolyglotException
+import org.graalvm.polyglot.Value
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class GraalJsEngineTest : JsEngineTest() {
```
Leland-Takamine, line 12: These tests document known behaviors that are different from Rhino JS. Common tests live in the base JsEngineTest class.

maestro-test/src/test/kotlin/maestro/test/RhinoJsEngineTest.kt
```diff
@@ -0,0 +1,58 @@
+package maestro.test
+
+import com.google.common.truth.Truth.assertThat
+import maestro.js.RhinoJsEngine
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.mozilla.javascript.RhinoException
+
+class RhinoJsEngineTest : JsEngineTest() {
```
Leland-Takamine, line 10: These tests document known behaviors that are different from GraalJS. Common tests live in the base JsEngineTest class.

axelniklasson (approved):
Nice, works like a charm!

igorsmotto (approved):

igorsmotto:
Should the env var `MAESTRO_USE_GRAALJS` even exists? Given that it only works on Maestro (not on Cloud) and there is already a place to set the JsEngine (MaestroConfig)?

Leland-Takamine:
Hm I was was thinking the env var would be useful for testing an entire test suite with graaljs without editing the flow file. This may come in handy when testing against existing test suites (which I think is useful for both Maestro users and for us). I agree that it has the potential to cause confusion but on the other hand it will only exist while graaljs is experimental. But I'm not married to it so we can remove if it seems too problematic

Leland-Takamine:
@igorsmotto I was thinking about that but I wanted to also support this case: `maestro test myflow.yaml` which doesn't use a config file

Leland-Takamine:
@igorsmotto I couldn't find a better alternative that doesn't require a lot more rearchitecting, and given that this is a temporary config while graaljs is still experimental, I've landed on the following compromise:

* Both the env var and flow are available to enable graaljs
* The env var only works locally
* The flow config only has an effect if present on the top-level flow
* **The flow config does not have an effect if present in a subflow**

PR #1148, opened 2023-06-13
Add support for sequential execution API

axelniklasson:
### <samp>🤖 Generated by Copilot at abfcba9</samp>

This pull request adds a new feature to maestro-cli and maestro-orchestra to allow users to customize the execution order and failure handling of flows in a test suite. It also updates the models and the workspace execution planner to support the new feature. It deprecates the old class that handled deterministic order.

Some notes:
- Adds a deprecation warning when using `local.deterministicOrder`
- If `continueOnFailure` is set to `false` (default is `true`) and a Flow in the sequence fails, the following Flwos in the sequence will not be executed
    - Other Flows will be executed in normal fashion
- Executes the Flows defined in `executionOrder.flowsOrder` first before all the other Flows

## Testing
- [x] Tested locally
- [x] Unit tests

The following tests uses the following config for sequential execution:

```yaml
# config.yaml
executionOrder:
  flowsOrder:
    - ios-flow
    - ios-flow-3
    - ios-flow-2
```

with the following `.maestro` directory structure:
```
.maestro/
   config.yaml
   ios-advanced-flow.yaml
   ios-flow-2-custom-name.yaml
   ios-flow-3.yaml
   ios-flow.yaml
```

where `ios-flow-2-custom-name.yaml` has `name: ios-flow-2` configured in the config section, to showcase that this takes [custom names](https://maestro.mobile.dev/api-reference/configuration/flow-configuration) into consideration.

| Test case  | Output |
| ------------- | ------------- |
| Normal test suite  | random_order  |
| Deterministic ordering  | deterministic_order |
| Sequential execution with failing Flow and default `continueOnFailure` (`true`)  | sequence_continue_failure_true  |
| Sequential execution with failing Flow and `continueOnFailure` set to `false`  | sequence_continue_failure_false  |
| Sequential execution all Flows passing  | Screenshot 2023-06-13 at 13 28 18  |

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
@@ -43,7 +43,7 @@ class TestSuiteInteractor(
     ): TestExecutionSummary {
         return if (input.isFile) {
             runTestSuite(
-                listOf(input),
+                WorkspaceExecutionPlanner.ExecutionPlan(flowsToRun = listOf(input.toPath())),
                 reportOut,
                 env,
             )
@@ -61,19 +61,17 @@ class TestSuiteInteractor(
             }
 
             runTestSuite(
-                flowFiles.map { it.toFile() },
+                plan,
                 reportOut,
                 env,
-                plan.stopOnFailure
             )
         }
     }
 
     private fun runTestSuite(
-        flows: List<File>,
+        executionPlan: WorkspaceExecutionPlanner.ExecutionPlan,
         reportOut: Sink?,
         env: Map<String, String>,
-        stopOnFailure: Boolean? = false,
     ): TestExecutionSummary {
         val flowResults = mutableListOf<TestExecutionSummary.FlowResult>()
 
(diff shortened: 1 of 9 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
```diff
@@ -8,17 +8,24 @@ data class WorkspaceConfig(
...
     val local: Local? = null,
+    val executionOrder: ExecutionOrder? = null
 ) {
 
     @JsonAnySetter
     fun setOtherField(key: String, other: Any?) {
         // Do nothing
     }
 
+    // deprecated in favour of ExecutionOrder
```
felipevolpone, line 19: Let's add the `@Deprecated` annotation as well so it's easier to map and find in the code

felipevolpone (approved):
well done 👏 thanks for doing this

maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -81,8 +96,37 @@ object WorkspaceExecutionPlanner {
                 .takeIf { it.exists() }
     }
 
+    private fun parseFileName(file: Path): String {
+        return file.fileName.toString().substringBeforeLast(".")
+    }
+
+    private fun getFlowsToRunInSequence(
```
felipevolpone, line 103: In cloud, we also check if the list has duplicated items, but I think it's fine to not have that locally

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -79,19 +79,34 @@ class TestSuiteInteractor(
...
+        // first run sequence of flows if present
+        val flowSequence = executionPlan.sequence
+        for (flow in flowSequence?.flows ?: emptyList()) {
+            val result = runFlow(flow.toFile(), env, maestro)
+            flowResults.add(result)
 
             if (result.status == FlowStatus.ERROR) {
                 passed = false
+                if (executionPlan.sequence?.continueOnFailure != true) {
+                    PrintUtils.message("Flow ${result.name} failed and continueOnFailure is set to false, aborting running sequential Flows")
```
felipevolpone, line 92: Shouldn't we use the `error` method?

maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
axelniklasson: https://github.com/mobile-dev-inc/maestro/pull/1148/commits/275e0b987cc8d0deece3bb2dbecb1752015c5ec8

Leland-Takamine (approved):

PR #1149, opened 2023-06-13
Adds multiple tap option in tapOn (double tap)

ArthurSav:
### <samp>🤖 Generated by Copilot at dbd1a2c</samp>

This pull request adds a new feature to the `Maestro` and `Orchestra` modules that allows the user to repeat and delay the tap action on elements, points, and relative points on the screen. The feature is implemented by introducing a new data class `TapRepeat` and adding a `tapRepeat` parameter to the relevant functions and data classes. The feature is also supported in the YAML commands, by adding `repeat` and `delay` properties to the element selectors. The purpose of this feature is to enhance the functionality and flexibility of the `Maestro` and `Orchestra` modules.

Introduces multi tap
```yaml
appId: com.other.app
---
- tapOn:
    text: Button
    repeat: 3
```

Possible to customise delay
```yaml
appId: com.other.app
---
- tapOn:
    text: Button
    repeat: 10
    delay: 500
```

For convenience, a wrapper command was added for double tap
```yaml
appId: com.other.app
---
- doubleTapOn:
    text: Button
```

## Testing
- Integration test
- iOS: double tap on Apple Maps
- Android: double tap on Google Maps

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -175,7 +175,8 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         retryIfNoChange: Boolean = true,
         waitUntilVisible: Boolean = false,
         longPress: Boolean = false,
-        appId: String? = null
+        appId: String? = null,
+        tapRepeat: TapRepeat? = null
     ) {
         LOGGER.info("Tapping on element: $element")
 
@@ -190,11 +191,12 @@ class Maestro(private val driver: Driver) : AutoCloseable {
             ).bounds
             .center()
         performTap(
-            center.x,
-            center.y,
-            retryIfNoChange,
-            longPress,
-            hierarchyBeforeTap,
+            x = center.x,
+            y = center.y,
+            retryIfNoChange = retryIfNoChange,
+            longPress = longPress,
+            initialHierarchy = hierarchyBeforeTap,
+            tapRepeat = tapRepeat
         )
 
         if (waitUntilVisible) {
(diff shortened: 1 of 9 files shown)


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -278,9 +292,12 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
                 driver.longPress(Point(x, y))
-            } else {
-                driver.tap(Point(x, y))
-            }
+            } else if (tapRepeat != null) {
+                for (i in 0 until tapRepeat.repeat) {
+                    driver.tap(Point(x, y))
+                    if (tapRepeat.repeat > 1) Thread.sleep(tapRepeat.delay) // do not wait for single taps
+                }
+            } else driver.tap(Point(x, y))
```
berikv, line 300: When tapRepeat is passed to the driver, we can better fit the specified delay.
In the current implementation, the roundtrip from driver to device is added to the delay for each tap.

Note that XCTest has api to send a sequence of taps with delays to the device, which allows for much-much more precise timing.
(see `XCPointerEventPath` in the xcuitest driver code)

berikv: There's also `XCPointerEventPath.pressDownAtOffset(TimeInterval)` available in private API

berikv (approved):

maestro-client/src/main/java/maestro/Maestro.kt
ArthurSav: > In the current implementation, the roundtrip from driver to device is added to the delay for each tap.

I can subtract the execution delay from the overall delay. That would also be consistent with both platforms.
Tbh I don't think we need to complicate this any further by using other apis.

PR #1160, opened 2023-06-16
feat: create XCTestClient

igorsmotto:
### Proposed Changes

This PR refactors the port number for `LocalXCTestInstaller` to allow a random port in case of not using a default port (22087). This PR still keeps the default port to avoid any compatibility issues.
`LocalXCTestInstaller` now returns the connection in terms of `XCTestClient`, to be used by the connection above. This keeps the connection contained and can be improved later on.

### Testing
✅ Manual tests on ios samples
✅ Integrated tests ran

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
+++ b/maestro-ios-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
@@ -138,26 +138,25 @@ class LocalXCTestInstaller(
             .inputStream.source().buffer().readUtf8()
             .trim()
 
-        if (processOutput.contains(UI_TEST_RUNNER_APP_BUNDLE_ID)) {
-            logger.info("Not able to find port running - Uninstalling UI test runner")
-            uninstall()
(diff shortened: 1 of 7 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
```diff
@@ -132,33 +138,34 @@ class LocalXCTestInstaller(
...
-            logger.info("[Start] Running XcUITest with xcode build command")
-            xcTestProcess = XCRunnerCLIUtils.runXcTestWithoutBuild(
-                deviceId,
-                xctestRunFile.absolutePath,
-                port
-            )
-            logger.info("[Done] Running XcUITest with xcode build command")
-        } else {
-            logger.info("UI test runner already installed and running")
+        if (processOutput.contains(UI_TEST_RUNNER_APP_BUNDLE_ID)) {
```
berikv, line 141: This change I don't understand.
Can't we keep the already running xcuitest runner as we did before?

PR #1194, opened 2023-06-26
chore: access modifier of jsengine

igorsmotto:
## Proposed Changes

This change should do nothing, and it only makes jsEngine public.


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -60,7 +60,7 @@ class Orchestra(
     private val onCommandMetadataUpdate: (MaestroCommand, CommandMetadata) -> Unit = { _, _ -> },
 ) {
 
-    private lateinit var jsEngine: JsEngine
+    lateinit var jsEngine: JsEngine
```
amanjeetsingh150, line 63: I am assuming this is done public for a reason. Do you want to use it in some other class?

igorsmotto:
Closed because this won't work.

PR #1206, opened 2023-06-28
setup / teardown hooks

artem888:
## Proposed Changes

This PR implements two hooks that can be executed before and after each maestro flow, i.e. to execute some setup / teardown actions.

Example of how the API will look like:

```
# myFlow.yaml
appId: myId
onFlowStart:
          - runFlow: setup.yaml
	  - runScript: setup.js
	  - <any other command>
onFlowComplete:
	  - runFlow: teardown.yaml
	  - runScript: teardown.js
	  - <any other command>
---
- launchApp
```

## Testing
- [x] tested with successful case (both `onFlowStart` and `onFlowComplete` executed)
- [x] tested that `onFlowComplete` is executed if main flow fails
- [x] tested with multiple commands under `onFlowStart` and `onFlowComplete`

successful case:

https://github.com/mobile-dev-inc/maestro/assets/961175/49aa1ed7-2c75-4658-8c18-f00081be556c

failed case:

https://github.com/mobile-dev-inc/maestro/assets/961175/fd58cb97-9efb-4f46-b356-973bb0008ffb

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -50,6 +50,7 @@ object MaestroCommandRunner {
         debug: FlowDebugMetadata
     ): Result {
         val initFlow = YamlCommandReader.getConfig(commands)?.initFlow
+        val onFlowComplete = YamlCommandReader.getConfig(commands)?.onFlowComplete
 
         val commandStatuses = IdentityHashMap<MaestroCommand, CommandStatus>()
         val commandMetadata = IdentityHashMap<MaestroCommand, Orchestra.CommandMetadata>()
@@ -89,6 +90,11 @@ object MaestroCommandRunner {
                     device = device,
                     initCommands = toCommandStates(
                         initFlow?.commands ?: emptyList(),
+                        commandStatuses,
+                        commandMetadata
+                    ),
+                    onFlowCompleteCommands = toCommandStates(
+                        onFlowComplete?.commands ?: emptyList(),
                         commandStatuses,
                         commandMetadata
                     ),
@@ -177,6 +183,11 @@ object MaestroCommandRunner {
         }
 
         val flowSuccess = orchestra.runFlow(commands, cachedState)
+
+        if (onFlowComplete != null) {
+            orchestra.runFlow(onFlowComplete.commands)
+        }
+
         return Result(flowSuccess = flowSuccess, cachedAppState = cachedState)
     }
 
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -190,15 +190,7 @@ object MaestroCommandRunner {
             cachedAppState
         }
 
-        if (onFlowStart != null) {
-            orchestra.runFlow(onFlowStart.commands)
-        }
-
         val flowSuccess = orchestra.runFlow(commands, cachedState)
-
-        if (onFlowComplete != null) {
-            orchestra.runFlow(onFlowComplete.commands)
-        }
 
         return Result(flowSuccess = flowSuccess, cachedAppState = cachedState)
     }
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -19,6 +19,7 @@
 
 package maestro.cli.runner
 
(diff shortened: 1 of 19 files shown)


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -2795,6 +2798,48 @@ class IntegrationTest {
...
+                    }
+                }
+            }
+        }
+
+        // Then
+        assertThat(receivedLogs).containsExactly(
+            "setup",
+            "teardown",
+        ).inOrder()
```
axelniklasson, line 2840: Let's update this test to ensure that the `tapOn` command is being executed in between the setup / teardown scripts

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -2795,6 +2798,48 @@ class IntegrationTest {
         )
     }
 
+    @Test
```
axelniklasson, line 2801: Can we add another test with a few commands that fails after the first one which then asserts that the `onFlowComplete` hook triggers correctly?

axelniklasson:
@artem888 nice work on this 🙌 

- Can you prepare a documentation PR for the maestro docs?
- Are there any changes required on Maestro Cloud for this to be supported?

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -176,7 +190,16 @@ object MaestroCommandRunner {
             cachedAppState
         }
 
+        if (onFlowStart != null) {
+            orchestra.runFlow(onFlowStart.commands)
```
amanjeetsingh150, line 194: In case if someone is using both init flow and onFlowStart what happens?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
```diff
@@ -16,6 +18,8 @@ data class YamlConfig(
     val initFlow: YamlInitFlowUnion?,
     val tags: List<String>? = emptyList(),
     val env: Map<String, String> = emptyMap(),
+    val onFlowStart: YamlOnFlowStart?,
+    val onFlowComplete: YamlOnFlowComplete?,
```
amanjeetsingh150, line 22: What was the point to prefer this over before/after again? I remember some point over extensibility but just want to ensure I understand that correct

maestro-orchestra/src/test/resources/YamlCommandReaderTest/022_on_flow_start_complete.yaml
```diff
@@ -0,0 +1,7 @@
+appId: com.example.app
+onFlowStart:
+  - back
+onFlowComplete:
+  - scroll
```
amanjeetsingh150, line 5: These are actually a set of Maestro commands only right?

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
artem888: this order logic is done in MaestroCommandRunner, so it is impossible to test this scenario with IntegrationTest as for now.

we could add a tech debt ticket to make MaestroCommandRunner testable and have IntegrationTest scenarios wrapped around it.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
artem888: ditto

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
artem888: in this case both will be executed, but I doubt someone will try to do that, plus the plan is to deprecate initFlow asap

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
artem888: for the case of extensibility, i.e. if someone will need onFlowFailed / onFlowSucceeded hooks in the future

maestro-orchestra/src/test/resources/YamlCommandReaderTest/022_on_flow_start_complete.yaml
artem888: yes, correct, the subtype is `List<MaestroCommand>`

axelniklasson (approved):

axelniklasson (approved):

PR #1209, opened 2023-06-29
New /viewHierarchy call on xctest-runner to improve RN support

berikv:
## Proposed Changes

Add /viewHierarchy as a more capable replacement of the /subTree call:
* Upon error (due to an ReactNative vs XCTest issue) retries getting the view hierarchy from the first child of the app node
* Recursively retries to get the hierarchy from child nodes until either the child node contains multiple children, or it succeeds
* If the hierarchy returned from .snapshot() has a depth == maxdepth, we can't be sure that we got everything. In that case, the algorithm tries to get the .snapshot() of the children one level down.

## Testing

Tested on staging with multiple customer flows

## Issues Fixed

Maestro currently falls back to IDB for React Native apps if an error occurs with XCTest. IDB's view hierarchy is a different structure and misses some key elements (like TabBar items).
By making the XCTest view hierarchy retrieval more capable, the IDB fallback can be removed, making maestro work more predictable with React Native apps.


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -0,0 +1,182 @@
...
+
+        do {
+            let viewHierarchy = try logger.measure(message: "View hierarchy snapshot for \(appId)") {
+                try getViewHierarchy(appId: appId)
+            }
+
+            let body = try JSONEncoder().encode(viewHierarchy)
+
+            // TODO: Remove debug print here
+            print(String(data: body, encoding: .utf8)!)
```
artem888, line 37: should we remove or PR still WIP?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -0,0 +1,182 @@
+import FlyingFox
+import XCTest
+import os
+
+extension NSException: Error {}
+
+@MainActor
+struct ViewHierarchyHandler: HTTPHandler {
+
+    let maxDepth = 60
```
artem888, line 10: some clarification on why value 60 was picked would be nice in here

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -0,0 +1,182 @@
...
+        do {
+            var hierarchy = try elementHierarchy(xcuiElement: element)
+            let hierarchyDepth = hierarchy.depth()
+
+            if hierarchyDepth < maxDepth {
+                logger.info("Hierarchy dept below maxdepth \(hierarchyDepth)")
+                return hierarchy
+            }
+
+            // When the hierarchy depth is equal to maxDepth, it is unknown if
```
artem888, line 87: did you manage to reproduce this case? my expectation was that `xcuiElement.snapshot().dictionaryRepresentation` always gives us complete hierarchy and crashes with `kAXErrorIllegalArgument` exception when its depth limits are reached

also it should be `is greater or equal to maxDepth` here looking at the code

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -187,6 +189,41 @@ class IOSDriver(
         }
     }
 
+    fun viewHierarchy(): TreeNode {
+        val hierarchy = iosDevice.viewHierarchy().get() ?: return TreeNode()
+        return mapViewHierarchy(hierarchy)
+    }
+
+    private fun mapViewHierarchy(element: AXElement): TreeNode {
```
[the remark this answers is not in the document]
amanjeetsingh150: CMIIW driver is responsible to convert the native Hierarchy to cross-platform `TreeNode` are you suggesting we should have a separate abstraction that should take responsibility for that conversion for both Android and iOS drivers? Heads up, this was already there for `subTree` call so we might need to also consider refactoring that

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -182,10 +189,11 @@ class XCTestDriverClient(
     }
 
     fun setPermissions(permissions: Map<String, String>) {
-        executeJsonRequest("setPermissions", SetPermissionsRequest(permissions))
+        executeJsonRequestUNCHECKED("setPermissions", SetPermissionsRequest(permissions))
     }
 
-    private fun executeJsonRequest(pathSegment: String, body: Any): Response {
+
+    private fun executeJsonRequestUNCHECKED(pathSegment: String, body: Any): Response {
```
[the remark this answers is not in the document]
amanjeetsingh150: This is actually unchecked because we are not closing the response body in these calls, which was actually case for every call to iOS XCUITest driver. We have added a new method `executeJsonRequest` to ensure we use that for any new call.

So, there is no such exception handled here. Let me know if that clears the naming?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -0,0 +1,137 @@
+import FlyingFox
+import XCTest
+import os
+
+@MainActor
+struct ViewHierarchyHandler: HTTPHandler {
+
+    let maxDepth = 60
```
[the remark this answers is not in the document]
amanjeetsingh150: Done

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
amanjeetsingh150: Yes I have shifted to draft. I'll be adding more improvements and addressing comments in upcoming PR

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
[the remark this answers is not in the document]
artem888: for me drivers already have too many responsibilities (i.e. IOSDriver is >500 lines already), so it is better to abstract some of this logic. sounds like it is out of scope from this PR though.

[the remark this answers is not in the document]
berikv: I doubted to put this logic in AXElement, but AXElement is in maestro-ios, and therefore can't access TreeNode which is in maestro-client. Then I considered adding this method in TreeNode, which would be messy (why should TreeNode know about AXElement). 

I suggest to keep the logic where it is now.

Lets look at this later, but fwiw, I think we can extract TreeNode into a helper module and have maestro-ios use it directly. That way we don't have to do all the translations between the different models. It would also be easier to add eg. the frame bounds and centerpoint as calculated model values.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -178,8 +183,8 @@ class IOSDriver(
     }
 
     override fun contentDescriptor(): TreeNode {
-        return when (val contentDescriptorResult = iosDevice.contentDescriptor()) {
-            is Ok -> mapHierarchy(contentDescriptorResult.value)
+        return when (val viewHierarchy = iosDevice.viewHierarchy()) {
```
amanjeetsingh150, line 186: Wait why are we using this here don't we need to keep the original implementation intact?

berikv: You are correct! Will revert this once testing is done

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -58,30 +58,23 @@ struct ViewHierarchyHandler: HTTPHandler {
...
-    func getRunningAppId(_ runningAppIds: [String]) -> String {
-        return runningAppIds.first { appId in
-            let app = XCUIApplication(bundleIdentifier: appId)
-            
-            return app.state == .runningForeground
-        } ?? ViewHierarchyHandler.springboardBundleId
+    func getRunningApp(_ runningAppIds: [String]) -> XCUIApplication? {
+        runningAppIds
+            .map { XCUIApplication(bundleIdentifier: $0) }
+            .first { app in app.state == .runningForeground }
```
amanjeetsingh150: Why this function doesn't return the value?

berikv: It does?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
berikv: Turns out we don't need to set the maxDepth anyway

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
berikv: This code has been removed since we don't need to set maxDepth

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
berikv: Removed

amanjeetsingh150:
I see that after this commit https://github.com/mobile-dev-inc/maestro/pull/1209/commits/a2132cda37a0e76a27e576f4c1c3522a7da80f79 the artifacts were not updated. Can you update the archives and also rebase with main.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/AXClientSwizzler.swift
```diff
@@ -0,0 +1,53 @@
...
+    // Make this type not-initializable
+    private init() {}
+
+    static var overwriteDefaultParameters: [String: Int] {
+        get { _overwriteDefaultParameters }
+        set { setup; _overwriteDefaultParameters = newValue }
+    }
+
+    static let setup: Void = {
+        print("Swizzle XCAXClient_iOS.defaultParameters")
```
artem888, line 20: better use os log to be consistent

artem888 (approved):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/AXClientSwizzler.swift
amanjeetsingh150: Done ✅

PR #1213, opened 2023-07-03
chore: add stacktrace and message to response

igorsmotto:
## Proposed Changes

This PR adds the exception message and the stacktrace to the Fake Ok response when receiving an error from XCTest Server.
This should facilitate debug.

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -233,7 +233,7 @@ class XCTestDriverClient(
             } else {
                 val message = "Failed request for XCTest server"
                 val responseBody = """
-                    { "message" : "$message" }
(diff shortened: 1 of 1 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -233,7 +233,7 @@ class XCTestDriverClient(
             } else {
                 val message = "Failed request for XCTest server"
                 val responseBody = """
-                    { "message" : "$message" }
+                    { "exceptionMessage": "${connectException.localizedMessage}, "stackTrace": "${connectException.stackTraceToString()} }
```
amanjeetsingh150, line 236: This wouldn't be that useful if I am not wrong. Reason because IOException is going to be thrown mostly in cases:

1. Trying to reach the XCTest server when it's not yet started
2. Server crashed
3. Test got canceled
In most cases, we'll have connected exceptions extending the IOException we are catching here. AFAICT, more useful would be looking at benchmarks that are affected by this and validating if either of these three cases are true. Let me know your thoughts!

PR #1216, opened 2023-07-04
Improving the new `viewHierarchy` call

amanjeetsingh150:
## Proposed Changes

1. Merging activeAppId call in new `ViewHierarchyHandler`
2. Check existence of the keyboard and alerts before taking a snapshot
3. Remove swizzling of max depth since skipping first few views is enough
4. Corrected falling back on idb for `contentDescriptor` call
5. Return `appHierarchy` nil in case of unknown exceptions


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
         }
 
-        return AXElement(children: alertHierarchies)
+        return try elementHierarchy(xcuiElement: alert)
     }
 
     let useFirstParentWithMultipleChildren = false
-    func findRecoveryElement(_ element: XCUIElement) -> XCUIElement {
+    private func findRecoveryElement(_ element: XCUIElement) -> XCUIElement {
         if !useFirstParentWithMultipleChildren {
```
amanjeetsingh150, line 161: @berikv not sure but when does `useFirstParentWithMultipleChildren` this flag is true? I just see constant false value

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
+            element.alerts.firstMatch
         }
 
-        return AXElement(children: alertHierarchies)
+        return try elementHierarchy(xcuiElement: alert)
     }
 
     let useFirstParentWithMultipleChildren = false
-    func findRecoveryElement(_ element: XCUIElement) -> XCUIElement {
+    private func findRecoveryElement(_ element: XCUIElement) -> XCUIElement {
```
amanjeetsingh150, line 160: Right now we just find the next Any element to find the recoveryElement. Should we make this skipping condition stricter? For example: Checking frame size to see we are not skipping some important XCUIElement

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -175,8 +177,12 @@ struct ViewHierarchyHandler: HTTPHandler {
         }
     }
 
-    func elementHierarchy(xcuiElement: XCUIElement) throws -> AXElement {
+    private func elementHierarchy(xcuiElement: XCUIElement) throws -> AXElement {
         let snapshotDictionary = try xcuiElement.snapshot().dictionaryRepresentation
         return AXElement(snapshotDictionary)
     }
+    
+    enum ViewHierarchyError: Error {
```
berikv, line 185: Better use struct for Error types, even better to put that struct in the modules/ folder and call it:
```
enum AppErrorType { case client, server }
struct AppError: Error, Codable {
    let message: String
    let errorType: AppErrorType
}
```

Then we can (in a separate PR) refactor all places from:
```
            let errorJson = """
                 { "errorMessage" : "Snapshot failure while getting view hierarchy", "errorCode": "\(errorCode)" }
                """
            return HTTPResponse(statusCode: .badRequest, body:  Data(errorJson.utf8))
```
to:
```
 throw AppError(message: "foo", errorType: .server)
```

And add a server-wide exception handler that does something like:
```
catch {
 if error is AppError {
  let body = JSONEncoder().encode(error)
  return HTTPResponse(statusCode: error.errorTypeAsHTTPCode, body:  body)
 } else {
  let appError = AppError(message: error.localizedDescription, errorType: .server)
  return HTTPResponse(statusCode: appError.errorTypeAsHTTPCode, body:  JSONEncoder().encode(appError))
 }
```

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -181,7 +178,10 @@ class IOSDriver(
     }
 
     override fun contentDescriptor(): TreeNode {
-        return viewHierarchy()
+        return when (val contentDescriptorResult = iosDevice.contentDescriptor()) {
+            is Ok -> mapHierarchy(contentDescriptorResult.value)
+            is Err -> TreeNode()
+        }
```
berikv, line 184: Why are we using Result here, instead of handling exceptions?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
amanjeetsingh150: Ahh got it 👍

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
amanjeetsingh150: The only risk I see in finding `recoveryElement` is that with:

```
element
    .children(matching: .any)
    .firstMatch
```
we might risk skipping important elements. But that's a tradeoff also I think it's just not going to be possible and for React native apps that's mostly the view containers for increasing padding

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -33,17 +34,13 @@ struct ViewHierarchyHandler: HTTPHandler {
...
-
             return HTTPResponse(statusCode: .ok, body: body)
         } catch {
-            print(error)
             let errorInfo = [
                 "errorMessage": "Snapshot failure while getting view hierarchy. Error: \(error)",
             ]
             let body = try? JSONEncoder().encode(errorInfo)
-            return HTTPResponse(statusCode: .internalServerError, body: body ?? Data())
+            return HTTPResponse(statusCode: .badRequest, body: body ?? Data())
```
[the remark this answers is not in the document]
amanjeetsingh150: I included this to stay in parity with the previous code. But open to change that.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
+        }
 
         return AXElement(children: [
             springboardHierarchy,
             appHierarchy,
         ].compactMap { $0 })
     }
+    
+    func getRunningAppId(_ requestBody: ViewHierarchyRequest) -> String {
+        return requestBody.appIds.first { appId in
```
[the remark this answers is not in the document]
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -24,7 +25,7 @@ struct ViewHierarchyHandler: HTTPHandler {
             return HTTPResponse(statusCode: .badRequest, body: body ?? Data())
         }
 
-        let appId = requestBody.appId
+        let appId = getRunningAppId(requestBody)
```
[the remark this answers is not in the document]
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -7,8 +7,9 @@ extension NSException: Error {}
 @MainActor
 struct ViewHierarchyHandler: HTTPHandler {
 
-    let maxDepth = 60
-    let springboardApplication = XCUIApplication(bundleIdentifier: "com.apple.springboard")
+    private let maxDepth = 60
+    private let springboardApplication = XCUIApplication(bundleIdentifier: "com.apple.springboard")
+    private static let springboardBundleId = "com.apple.springboard"
```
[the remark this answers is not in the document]
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
+        return requestBody.appIds.first { appId in
+            let app = XCUIApplication(bundleIdentifier: appId)
+            
+            return app.state == .runningForeground
+        } ?? ViewHierarchyHandler.springboardBundleId
+    }
 
-    func appHierarchy(_ xcuiApplication: XCUIApplication) throws -> AXElement {
-        AXClientSwizzler.overwriteDefaultParameters["maxDepth"] = maxDepth
+    func appHierarchy(_ xcuiApplication: XCUIApplication) throws -> AXElement? {
```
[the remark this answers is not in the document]
amanjeetsingh150: Changed the return type back to AxElement and now I am throwing an explicit error where we are unable to get view hierarchy.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
-
-                return AXElement(children: [
-                    hierarchy,
-                    keyboard,
-                    alerts
-                ].compactMap { $0 })
+                return hierarchy
+            } else {
+                logger.error("Snapshot failure, cannot return view hierarchy due to \(message)")
+                return nil
```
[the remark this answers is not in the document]
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -57,105 +54,110 @@ struct ViewHierarchyHandler: HTTPHandler {
...
-
-            return hierarchy
+        }
+    }
+    
+    private func getErrorCode(message: String) -> String {
+        if message.contains("Error kAXErrorIllegalArgument getting snapshot for element") {
+            return "illegal-argument-snapshot-failure"
+        } else {
+            return "unknown-snapshot-failure"
```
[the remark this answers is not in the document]
amanjeetsingh150: Done replaced with enums ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
amanjeetsingh150: Makes sense, I have changed accordingly and added `ViewHierarchyError`. Let me know if that looks good!

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: In case of unknown errors we just return empty hierarchy. This is in parity with production

berikv: I mean, why do we use Result in IOSDevice for each methods return type, instead of using exception throwing?
Inside IOSDriver all Result values are then `.expect {}`'ed anyway..

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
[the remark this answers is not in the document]
berikv: Yeah, we should return 5xx for errors that happen during execution and 4xx for errors related to invalid requests (eg request model not valid). I changed the code in my branch specifically to be in line with the http standard, and would like to keep it that way.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
berikv: If the view frame is checked, then what do we do when the frame size is unexpected? And why would that be better than continuing the algorithm as normal?
Why not check other things, like the accessibilityLabel?
I still don't understand what is special about the frame size that makes it something that should be checked in the view hierarchy recovery algorithm.

berikv (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
berikv: No action required here though, we can have a look at this later.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
amanjeetsingh150: > If the view frame is checked, then what do we do when the frame size is unexpected? And why would that be better than continuing the algorithm as normal?

If we check the frame size and it's equal to screen size we know that we can safely skip these elements. If let's say it goes unexpected and we need to still skip it for successful operation we do it intentionally. Its better because we would know where our algorithm could go wrong.

Issue #1124, opened 2023-06-01
Implement retry strategy


testableapple:
**Is your feature request related to a problem? Please describe.**
Most of the frameworks allow us to retry failed tests. Maestro currently does not. It would be great to have such an option.

**Describe the solution you'd like**
- Maestro users can set up a `retry-tests-count` in the config file globally or in some flow in particular (similar to `-retry-tests-on-failure` flag in XCTest).
- By default, this count should be equal to `0` (if flow fails, then it fails straight away)
- If a user sets it up (e.g.: `3`) and the flow fails on the first run, then Maestro automatically starts it from scratch, and so on
- These repetitions are clearly displayed in the logs

**Describe alternatives you've considered**
N/A

**Additional context**
Reference: https://twitter.com/testableapple/status/1662915551927193609

ubuntudroid:
This is an essential feature. As the [Maestro Cloud docs](https://cloud.mobile.dev/reference/retries) correctly state:

> The behavior of an application can be complex and challenging to verify, leading to unreliable test results in some instances. This can be caused by various factors such as animations, API calls, backend availability, network interruptions, and temporary outages.

We could for sure just restart the whole test suite ourselves if it fails, but that's far less efficient than just rerunning the failing tests. So a built-in mechanism similar to Maestro Cloud would be much appreciated.

abhishekbedi1432:
We too are this basic feature. Currently we rely on scripts to re-run the test based on its failure status iteratively

Fishbowler:
Inherently, we dislike global retries, the same way we dislike hard waits.
We appreciate that there's a middle-ground though, for stuff outside the app's control, like flaky server infrastructure.
A couple of months back, we added this: https://docs.maestro.dev/api-reference/commands/retry

Fishbowler:
Of course, if you really want to, you can do something horrid like:

```
- retry:
    maxRetries: 3
    commands:
      - runFlow: myTest.yaml
```

But hopefully, it'll be something much nicer like:

```
- retry:
    maxRetries: 3
    commands:
      - runScript: createUserViaAPI.js
      - tapOn: 'login'
      - tapOn: 'username'
      - inputText: 'test_user'
      - tapOn: 'password'
      - inputText: 'test_password'
      - tapOn: 'Log In'
      - assertVisible: 'Your Account'
- tapOn: 'Product List'
- tapOn; 'Add to Cart'
- tapOn: 'Checkout'
```

[...]

PR #1228, opened 2023-07-08
Retry flows on failure

carlosmuvi:
## Proposed Changes

- Include `retryTestsCount` config param to auto-retry a flow instead of failing when a command execution does not succeed.

## Testing
- Create a flow with failing commands that includes `retryTestsCount` in its config. Tests should auto retry as many times as specified.
- No retries should happen if this config field is not provided.

## Issues Fixed
- https://github.com/mobile-dev-inc/maestro/issues/1124


igorsmotto:
Hi @carlosmuvi,

While I appreciate the effort and thought put into this feature, I have decided to close the pull request as it doesn't align well with our current requirements.

The main reasons for closing it are:

- Achieving the same functionality without this feature is possible by creating a script that reads the response of Maestro tests and reruns Maestro as needed. We believe that implementing a separate configuration parameter for retries could be redundant in this context.
- The introduction of this feature could potentially hide flakiness in the local execution of Maestro, which could make flow code from users less robust. While we understand that flakiness can have various causes, we believe that it's important to address and minimize them directly rather than relying heavily on retries.

I appreciate your efforts in contributing to the project, and I encourage you to continue exploring other opportunities to enhance Maestro. If you have any questions or need further clarification regarding the decision to close the PR, please feel free to reach out to me.

Thank you for your understanding.

omegasoft7:
@igorsmotto 

That is actually a good feature to have in Maestro.
Regarding the flakiness, lets the users decide what they need to have
and regarding implementing it via a script, imagine that I have a flow that contains 10 flows inside it. If I want to manage it with scripts, then in case of failures I have to run everything from the beginning, but if Maestro can handle it, then for instance if flow number 6 failed, I can only retry that flow once more, instead of everything again.
It would be great if you can consider having these functionalities at least for runFlow

Thanks

freak4pc:
Sorry for the bump but I highly disagree this isn't needed - The point isn't retrying the entire flow which can be mostly simple in shell, but it's also to retry the specific tests which failed, which means parsing the results of the test report and then having the script retry only those failures. Seems like something that could be really nice to have in the tool instead of everyone writing their own.

saihemanth121:
@igorsmotto Maestro Cloud supports test [retries](https://cloud.mobile.dev/reference/retries), doesn't it? Is it a lot different from the retry strategy being added here?

PR #1239, opened 2023-07-12
feat: add a more generic retry interceptor to XCTestDriverClient

igorsmotto:
## Proposed Changes

This PR adds a more generic retry interceptor, trying more aggressively to retry if XCTestServer fails to respond.

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -47,7 +47,7 @@ class XCTestDriverClient(
     private val okHttpClient = OkHttpClient.Builder()
         .connectTimeout(10, TimeUnit.SECONDS)
         .readTimeout(30, TimeUnit.SECONDS)
-        .addRetryOnErrorInterceptor()
+        .addRetryInterceptor()
         .addReturnOkOnShutdownInterceptor()
         .build()
 
@@ -197,16 +197,25 @@ class XCTestDriverClient(
         return okHttpClient.newCall(requestBuilder.build()).execute()
     }
 
-    private fun OkHttpClient.Builder.addRetryOnErrorInterceptor() = addInterceptor(Interceptor {
-        val request = it.request()
-        try {
-            it.proceed(request)
-        } catch (connectException: IOException) {
-            installer.start()?.let { newClient ->
-                client = newClient
-                it.proceed(request)
-            } ?: throw XCTestDriverUnreachable("Failed to reach out XCUITest Server in RetryOnError")
+    private fun OkHttpClient.Builder.addRetryInterceptor() = addInterceptor(Interceptor { chain ->
+        var exception: IOException? = null
+        repeat(3) {
+            try {
+                val response = chain.proceed(chain.request())
+                if (response.isSuccessful) {
+                    return@Interceptor response
+                }
+            } catch (e: IOException) {
+                installer.start()?.let {
+                    client = it
+                }
+                exception = e
+            }
+            Thread.sleep(200L)
         }
+
+        exception?.let { throw it }
+            ?: throw XCTestDriverUnreachable("Failed to reach out XCUITest Server after 3 attempts")
     })
 
     private fun OkHttpClient.Builder.addReturnOkOnShutdownInterceptor() = addNetworkInterceptor(Interceptor {


felipevolpone (approved):

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -197,16 +197,22 @@ class XCTestDriverClient(
...
+        repeat(3) {
+            try {
+                val response = chain.proceed(chain.request())
+                if (response.isSuccessful) {
+                    return@Interceptor response
+                }
+            } catch (e: IOException) {
+                exception = e
+            }
+            Thread.sleep(200L)
```
amanjeetsingh150, line 211: Confused about this sleep 🤔 ?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -197,16 +197,22 @@ class XCTestDriverClient(
...
-        try {
-            it.proceed(request)
-        } catch (connectException: IOException) {
-            installer.start()?.let { newClient ->
-                client = newClient
-                it.proceed(request)
-            } ?: throw XCTestDriverUnreachable("Failed to reach out XCUITest Server in RetryOnError")
+    private fun OkHttpClient.Builder.addRetryInterceptor() = addInterceptor(Interceptor { chain ->
+        var exception: IOException? = null
+        repeat(3) {
```
amanjeetsingh150, line 202: Didn't we a have a retrying mechanism already in place? How this is different?

amanjeetsingh150: Also there is https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/retry-on-connection-failure/
API if we want to retry on connection failures.

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
igorsmotto: Just to wait 200ms between attempts. Instead of retrying right away. Avoiding overloading the XCTest Server.

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
igorsmotto: The current retrying mechanism does not retry based on the status code of the response. So it only retries on connection failure (through an exception). Now this retry is more generic will retry based if the status code is not successful (200)

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -197,16 +197,22 @@ class XCTestDriverClient(
         return okHttpClient.newCall(requestBuilder.build()).execute()
     }
 
-    private fun OkHttpClient.Builder.addRetryOnErrorInterceptor() = addInterceptor(Interceptor {
-        val request = it.request()
-        try {
-            it.proceed(request)
-        } catch (connectException: IOException) {
-            installer.start()?.let { newClient ->
```
berikv, line 205: Why is XCTest-runner not re-installed anymore?

igorsmotto: Sorry, forgot about that

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: Got it and when do we see this retry happening I see its done in IOExceptions are there special cases we want to handle. I'm asking this because retryConnectionOnFailure(true) do cover most of it. So, either we have some specific case which is not covered by retryConnectionOnFailure 🤔

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: > Now this retry is more generic will retry based if the status code is not successful (200)

But why do we need that having can 4xx response be flaky from server?

PR #1255, opened 2023-07-17
Scroll improvements

amanjeetsingh150:
## Proposed Changes

1. Keeping the XCUIApplication instance as root fails in big nested screens due to which the app doesn't scroll
2. Changes root to first any child for scroll which avoids resolving of XCUIApplication on scroll


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
```diff
@@ -26,7 +26,8 @@ struct SwipeRouteHandlerV2: HTTPHandler {
         let description = "Swipe from \(request.start) to \(request.end) with \(request.duration) duration"
         logger.info("\(description)")
 
-        let eventTarget = EventTarget(bundleId: request.appId)
+        let runningAppId = RunningApp.getForegroundAppId(request.appIds ?? [""])
```
berikv, line 29: Is the empty string needed here?
I'd expect an empty array: `request.appIds ?? []`

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/RunningApp.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  RunningApp.swift
+//  maestro-driver-iosUITests
+//
+//  Created by Amanjeet Singh on 17/07/23.
+//
+
```
berikv, line 7: Nit: iOS engineers tend to remove these bogus headers on file creation

berikv (approved):

artem888 (approved):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/RunningApp.swift
```diff
@@ -0,0 +1,25 @@
...
+import Foundation
+import XCTest
+
+struct RunningApp {
+    
+    private static let springboardBundleId = "com.apple.springboard"
+    private init() {}
```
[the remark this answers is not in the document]
berikv: I'm not super in favour of misusing the enum type for that purpose. A `private init` more clearly communicates what is going on, and for the compiler there's no difference.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/RunningApp.swift
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
amanjeetsingh150: Done ✅

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/SwipeRequest.swift
```diff
@@ -30,5 +32,6 @@ struct SwipeRequest: Decodable {
             y: try container.decode(Double.self, forKey: .endY)
         )
         duration = try container.decode(Double.self, forKey: .duration)
+        appIds = try container.decodeIfPresent([String].self, forKey: .appIds)
```
berikv, line 35: For safety and reduction of optionals in models, it may be better to create a `swipeV3()` and `SwipeRequestV3`.
WDYT?

amanjeetsingh150: Hmmm, open to do that but do we know when we should completely delete swipe v1 and v2 🤔

berikv: When both cli and cloud are using V3, the others can be removed

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/XCTest/RunningApp.swift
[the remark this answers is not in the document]
amanjeetsingh150: Agreed with Berik here! Lets avoid a Enum here

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/SwipeRequest.swift
amanjeetsingh150: Umm, I was looking at the scope and it's indeed just a couple of line changes do you still think v3 makes sense for that?

PR #1270, opened 2023-07-24
Test port number for android

kshitij6325:
### <samp>🤖 Generated by Copilot at 58e447b</samp>

Added a debug statement to `maestro-cli` to print the port number of the `maestro-server` endpoint. This helps diagnose the connection issue between the two components.

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - spike/kshitij/test-build-with-logs
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


felipevolpone:
Hey @kshitij6325 thanks for the PR, but this is not the way to go. Adding a `println` there isn't the best option. Instead, we could have a new parameter (like `--verbose` or `--emulator-verbose`) in the `test` command that would print the port and other information about the emulator.

felipevolpone (changes requested):
Added a comment on how to proceed with this PR

felipevolpone:
Closing the PR for inactivity. @kshitij6325 please feel free to leave a comment if you will still work on this PR

PR #1273, opened 2023-07-24
Add insights from the max depth on the screens

amanjeetsingh150:
## Proposed Changes

Add insights at shutdown hook for the max depth on iOS screens

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -31,6 +31,7 @@ import maestro.Maestro
 import maestro.cli.device.Device
 import maestro.cli.device.PickDeviceInteractor
 import maestro.cli.device.Platform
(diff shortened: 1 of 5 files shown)


igorsmotto (approved):

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -89,6 +90,7 @@ object MaestroSessionManager {
             SessionStore.withExclusiveLock {
                 heartbeatFuture.cancel(true)
                 SessionStore.delete(sessionId, selectedDevice.platform)
+                ScreenReporter.reportMaxDepth()
```
igorsmotto, line 93: Should we place this call under `runCatching { }`?  Just to avoid spilling any errors during the shutdown process

amanjeetsingh150: Makes sense done 💯

PR #1274, opened 2023-07-24
[MOB 2206] Raising insights on view hierarchy

amanjeetsingh150:
## Proposed Changes
Raise insights when the view hierarchy of screen is too deep

1. Raise the insight when it start taking lot of time 
2. Raise insight on migration of new architecture and the current obtained depth

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
@@ -25,7 +25,11 @@ import maestro.cli.App
 import maestro.cli.DisableAnsiMixin
 import maestro.cli.session.MaestroSessionManager
 import maestro.cli.view.green
+import maestro.cli.view.yellow
+import maestro.utils.Insights
+import maestro.utils.chunkStringByWordCount
 import picocli.CommandLine
+import java.lang.StringBuilder
 
 @CommandLine.Command(
     name = "hierarchy",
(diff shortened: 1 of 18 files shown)


maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -44,6 +48,14 @@ class PrintHierarchyCommand : Runnable {
 
     override fun run() {
         MaestroSessionManager.newSession(parent?.host, parent?.port, parent?.deviceId) { session ->
+            Insights.onInsightsUpdated {
+                val message = StringBuilder()
+                message.append("Warning!".yellow() + ": ")
+                it.first().message.chunkStringByWordCount(12).forEach { chunkedMessage ->
```
Leland-Takamine, line 54: This will only ever show the first insight even if there are multiple

maestro-utils/src/main/kotlin/Insight.kt
```diff
@@ -0,0 +1,31 @@
+package maestro.utils
+
+object Insights {
+
+    private val insights = mutableSetOf<Insight>()
+    private var onInsightsUpdated: (List<Insight>) -> Unit = {}
+
+    fun report(insight: Insight) {
+        insights.add(insight)
```
Leland-Takamine, line 9: doesn't seem necessary to store the insights since the consumers are responsible for reacting to any of these events and they don't rely on the history of insights.

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -20,6 +24,9 @@ class LocalIOSDevice(
     private val simctlIOSDevice: SimctlIOSDevice,
 ) : IOSDevice {
 
+    private var isViewHierarchyInProgress = false
```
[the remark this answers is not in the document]
amanjeetsingh150: Done changed that ✅

maestro-utils/src/main/kotlin/Insight.kt
amanjeetsingh150: Done changed that ✅

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
amanjeetsingh150: Done, fixed that by having single insight

Leland-Takamine (approved):

PR #1282, opened 2023-07-25
[fix] inpuText first character is being skipped

artem888:
## Proposed Changes

This PR implements logic to check whether keyboard is physically presented before start sending text input events to XCTest. This should fix the issues when first character is skipped or first characters are appended to the text buffer while typing.

## Testing
- [x] local test
- [x] staging test


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
```diff
@@ -61,4 +63,18 @@ struct InputTextRouteHandler : HTTPHandler {
         let errorData = Data(jsonString.utf8)
         return HTTPResponse(statusCode: HTTPStatusCode.badRequest, body: errorData)
     }
+    
+    private func waitUntilKeyboardIsPresented(appId: String?) async {
+        let timeout = Date().addingTimeInterval(2) // Set the timeout for 2 second from now
```
berikv, line 68: Can you extract the timeout logic here?
Something like:
```
func repeatUntil(timeout: TimeInterval, delta: TimeInterval, block: () -> Bool) {
     let endDate = ...
     while Date() < endDate {
        Task.sleep(delta)
        if block() { return }
    }
}
```

berikv, line 68: I'm a bit worried that the 2 seconds delay will have an impact on various flow runs. Do you know how often the the keyboard is not found and maestro waits for 2 seconds?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -152,8 +152,9 @@ class XCTestDriverClient(
 
     fun inputText(
         text: String,
+        appId: String?,
```
amanjeetsingh150, line 155: Let's not put a separate api call here for running app id. We have RunningApp struct which can help reducing these HTTP calls to 1.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/InputTextRouteHandler.swift
artem888: In all of my test cases it was available after the first iteration (meaning only 200ms delay), I think it is safe to decrease the timeout to 1s

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
artem888: the thing is we also need to call getInstalledApps on simctl before doing this call, so it'd not be possible to do that completely on xctest driver side.

we could pass appIds list to the request, but that'd also look ugly imo.

amanjeetsingh150: What weirdness you are referring to? We did that in viewHierarchy maybe be we can address there as well

artem888: ok, I see. I don't see a big issue with extra network request here anyway since it is done on localhost and we can reuse this logic. 

but if we choose a pattern of resolving runningAppId on the xctest driver side I'm fine with doing this change.

I'd also deprecate runningAppId endpoint then

amanjeetsingh150: >  I don't see a big issue with extra network request

Well, it's about okhttp. Doing too many calls does affect the connection pools and network latency. But since we haven't seen such reports I would say for me apart from this it also makes less complex. But open to discuss this and we can definitely deprecate runningAppId endpoint.


@berikv would appreciate your opinion about this as well.

berikv: >  it is done on localhost and we can reuse this logic.
That's not always true.
If it is easy to switch to a similar setup as /viewHierarchy, I suggest we implement that. If otoh it is much more difficult, then lets keep it as in the PR now.

artem888: thanks for the arguments guys! I think it makes sense to have it on the xctest driver side now

artem888: fixed ✅

PR #1285, opened 2023-07-26
[MOB-2209] Fail the flow if one of onStart / Complete hooks fails

artem888:
## Proposed Changes

This PR implements failure handling in onStart / onComplete hooks. The behaviour will follow the logic in the following table:

| Test case                                              | JUnit behavior        | XCTest behavior      |
| ------------------------------------------------------ | --------------------- | -------------------- |
| when before function fails, does that fail the whole test? | test marked :red_circle: | test marked :red_circle: |
| when before function fails, does that skip the main body of the test execution? | test execution skipped | test execution skipped |
| when before function fails, does the after function still run? | after function is still called | after function is still called |
| when the after function fails, does the test fail?   | test marked :red_circle: | test marked :red_circle:   |

## Testing
- [x] local test
- [ ] staging test

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -582,16 +582,21 @@ class Orchestra(
         // filter out DefineVariablesCommand to not execute it twice
         val filteredCommands = commands.filter { it.asCommand() !is DefineVariablesCommand }
 
-        subflowConfig?.onFlowStart?.commands?.let {
-            executeSubflowCommands(it, config)
-        }
-
+        var flowSuccess = false
         try {
-            return executeSubflowCommands(filteredCommands, config)
-        } finally {
-            subflowConfig?.onFlowComplete?.commands?.let {
+            val onStartSuccess = subflowConfig?.onFlowStart?.commands?.let {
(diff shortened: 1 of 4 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -579,16 +585,26 @@ class Orchestra(
...
+        } catch (e: Throwable) {
+            exception = e
+        } finally {
+            val onCompleteSuccess = subflowConfig?.onFlowComplete?.commands?.let {
+                executeSubflowCommands(it, config)
+            } ?: true
+
+            exception?.let { throw it }
+
+            return onCompleteSuccess && flowSuccess
```
berikv, line 607: This bit confuses me, why not:
```
   return onComplete && success
} finally {
  onFlowComplete.execute()
}
```

artem888: the thing here is we also need to know `onFlowComplete.execute()` result before returning.

in case `onFlowComplete.execute()` returns false we mark flow as failed as documented. 

also `onFlowComplete.execute()` needs to be executed nevermind the result of the main flow execution

berikv (approved):

PR #1286, opened 2023-07-27
Removing future call from hideKeyboard to avoid leaks 

amanjeetsingh150:
## Proposed Changes

The future is not needed any more since we have fixed swipe and scroll and also the animation problem we had with wait for app being idle.

## Testing
Locally ✅

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -61,7 +61,6 @@ class IOSDriver(
     private val iosDevice: IOSDevice,
 ) : Driver {
 
-    private val executor by lazy { Executors.newSingleThreadExecutor() }
     private val deviceInfo by lazy {
         iosDevice.deviceInfo().expect {}
     }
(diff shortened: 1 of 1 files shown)


berikv:
The future was used to guard the hideKeyboard operation to take too long. Can we put back a way to set a timeout, but do close the response when the timeout fires?
For example by using a try/catch hook in the cloud/swipeV2 method, and closing the response body in the catch block, then rethrowing the exception.

amanjeetsingh150:
> The future was used to guard the hideKeyboard operation to take too long. Can we put back a way to set a timeout, but do close the response when the timeout fires?
> For example by using a try/catch hook in the cloud/swipeV2 method, and closing the response body in the catch block, then rethrowing the exception.

We actually added because of 2 issues. One being scroll not working and other one due to xcuitest waiting on app being idle. You can see the context over added this future here in this PR: https://github.com/mobile-dev-inc/maestro/pull/989. The whole operation only took long whenever we had these errors. Now since both of them are addressed I don't think we need that future.

berikv (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -423,33 +422,25 @@ class IOSDriver(
     override fun backPress() {}
 
     override fun hideKeyboard() {
-        val future = executor.submit {
```
igorsmotto, line 426: I'm curious, do you happen to know why it used a different thread to run the hideKeyboard?

igorsmotto (approved):

berikv (approved):

PR #1293, opened 2023-07-28
Adds an API to raise insights on studio

amanjeetsingh150:
## Proposed Changes

1. Adds `InsightsService` api
2. Checks for 3 seconds for insights and responds with either empty string or valid insight

## Testing
- [ ] Needs testing with frontend integration.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -194,7 +194,7 @@ class IOSDriver(
                     "is ${hierarchyResult.depth}. If you are using React native, consider migrating to the new " +
                     "architecture where view flattening is available. For more information on the " +
                     "migration process, please visit: https://reactnative.dev/docs/new-architecture-intro"
-            Insights.report(Insight(message, Insight.Level.INFO))
+            Insights.report(Insight(message, Insight.Level.INFO, Insight.Visibility.VISIBLE))
         }
         val hierarchy = hierarchyResult?.axElement ?: return TreeNode()
         return mapViewHierarchy(hierarchy)
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -195,6 +195,8 @@ class IOSDriver(
                     "architecture where view flattening is available. For more information on the " +
                     "migration process, please visit: https://reactnative.dev/docs/new-architecture-intro"
             Insights.report(Insight(message, Insight.Level.INFO))
+        } else {
+            Insights.report(Insight("", Insight.Level.NONE))
         }
         val hierarchy = hierarchyResult?.axElement ?: return TreeNode()
         return mapViewHierarchy(hierarchy)
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -194,7 +194,7 @@ class IOSDriver(
                     "is ${hierarchyResult.depth}. If you are using React native, consider migrating to the new " +
                     "architecture where view flattening is available. For more information on the " +
                     "migration process, please visit: https://reactnative.dev/docs/new-architecture-intro"
-            Insights.report(Insight(message, Insight.Level.INFO, Insight.Visibility.VISIBLE))
+            Insights.report(Insight(message, Insight.Level.INFO))
         }
         val hierarchy = hierarchyResult?.axElement ?: return TreeNode()
         return mapViewHierarchy(hierarchy)
--- a/maestro-ios/src/main/java/ios/LocalIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/LocalIOSDevice.kt
@@ -54,7 +54,8 @@ class LocalIOSDevice(
                         Insight(
                             message = "Retrieving the hierarchy is taking longer than usual. This might be due to a " +
                                     "deep hierarchy in the current view. Please wait a bit more to complete the operation.",
-                            level = Insight.Level.WARNING
+                            level = Insight.Level.WARNING,
+                            visibility = Insight.Visibility.VISIBLE
                         )
                     )
                 }
--- a/maestro-ios/src/main/java/ios/LocalIOSDevice.kt
+++ b/maestro-ios/src/main/java/ios/LocalIOSDevice.kt
@@ -47,7 +47,7 @@ class LocalIOSDevice(
 
     override fun viewHierarchy(): Result<ViewHierarchy, Throwable> {
         var isViewHierarchyInProgress = true
-        executor.schedule(
+        val future = executor.schedule(
             {
                 if (isViewHierarchyInProgress) {
                     Insights.report(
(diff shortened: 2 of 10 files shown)


Leland-Takamine:
* Do we want to surface all insights in the banner (including future ones) or just the hierarchy warning?
* The implementation seems convoluted. Why not just something like this pseudocode:

```kotlin
onHierarchyWarning { message ->
  BannerService.message = message
}

// BannerService.kt
var message: String? = null
Routing.get("/api/insights") {
  call.respond(message)
}
```

amanjeetsingh150:
> Do we want to surface all insights in the banner (including future ones) or just the hierarchy warning?

Hmm, good question I am planning to change the implementation to only propagate those insights that studio or terminal needs to show. So, there would be separate StudioInsight and CliInsight.

> The implementation seems convoluted. Why not just something like this pseudocode:
Not sure if I understand 🤔  As implemented in this PR: https://github.com/mobile-dev-inc/maestro/pull/1274/files. The insights are not synchronous in nature we need to register callbacks for this. Are you suggesting this as part of already present APIs like `/api/device-screen`? 🤔  The idea implementing behind this was originally because of async nature of insights

Leland-Takamine:
> The insights are not synchronous in nature we need to register callbacks for this

Sure but in that callback can't we just store the latest message in some field and clear the field once the message is no longer applicable?

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -56,7 +56,8 @@ class LocalIOSDevice(
                         Insight(
                             message = "Retrieving the hierarchy is taking longer than usual. This might be due to a " +
                                     "deep hierarchy in the current view. Please wait a bit more to complete the operation.",
-                            level = Insight.Level.WARNING
+                            level = Insight.Level.WARNING,
+                            visibility = Insight.Visibility.VISIBLE
```
Leland-Takamine, line 60: This is a bit confusing to me since visibility seems to only applies to studio and not the terminal output

maestro-studio/server/src/main/java/maestro/studio/InsightService.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.coroutines.suspendCoroutine
+
+object InsightService {
+
+    private var insight: Insight? = null
+
+    fun routes(routing: Route) {
+        routing.get("/api/banner-message") {
+            Insights.onInsightsUpdated {
+                insight = it.first { insight -> insight.visibility == Insight.Visibility.VISIBLE }
```
Leland-Takamine, line 21: A few issues I see with this:

* This will only ever show the first insight that was reported. Subsequent insights will not be shown since you're using `first`
* There's no way to clear an insight once you're displayed it. In our specific case, we want to clear the hierarchy message once the hierarchy call is no longer taking too long.
* Insights are all stored in memory for the entire session. (This might be ok if we don't expect to be storing many insights, but that might not be a safe assumption)

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
amanjeetsingh150: It is reflecting there its on diffferent PR. Request you to review that as well: https://github.com/mobile-dev-inc/maestro/pull/1274/files

Leland-Takamine: What I meant is that the visibility field only seems to control whether this is displayed in the studio banner. This is surprising since I would expect a field named `visibility` to apply in all cases.

amanjeetsingh150: Ahhh right thats true. I'll do that

maestro-studio/server/src/main/java/maestro/studio/InsightService.kt
amanjeetsingh150: Ok updated:

1. Removed the first call, instead we return each and every insight we get from callback also mapping to different class, BannerMessage.
2. For clearing insight I've also added updated callback to clear function of Insights
3. For this one I'll not defer any kind of storage optimization just yet, since insights are less. But agreed if this comes to more insights we can back this with a storage.

maestro-studio/server/src/main/java/maestro/studio/InsightService.kt
```diff
@@ -0,0 +1,58 @@
...
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+object InsightService {
+
+    private var currentInsights: List<Insight>? = null
+
+    fun routes(routing: Route) {
+        routing.get("/api/banner-message") {
+            Insights.onInsightsUpdated {
```
Leland-Takamine, line 21: Move this outside of API callback handler

maestro-utils/src/main/kotlin/Insight.kt
```diff
@@ -16,16 +16,20 @@ object Insights {
 
     fun clear() {
         insights.clear()
+        onInsightsUpdated(insights.toList())
     }
 
     fun onInsightsUpdated(callback: (List<Insight>) -> Unit) {
```
Leland-Takamine, line 22: This only allows a single listener at a time

maestro-studio/server/src/main/java/maestro/studio/InsightService.kt
```diff
@@ -0,0 +1,58 @@
...
+import kotlinx.coroutines.*
+import maestro.studio.BannerMessage.*
+import maestro.utils.Insight
+import maestro.utils.Insights
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+object InsightService {
+
+    private var currentInsights: List<Insight>? = null
```
Leland-Takamine, line 17: Should this actually be a list?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -195,6 +195,8 @@ class IOSDriver(
                     "architecture where view flattening is available. For more information on the " +
                     "migration process, please visit: https://reactnative.dev/docs/new-architecture-intro"
             Insights.report(Insight(message, Insight.Level.INFO))
+        } else {
+            Insights.report(Insight("", Insight.Level.NONE))
```
Leland-Takamine, line 199: This API still feels weird to me since we're emitting an empty insight on every hierarchy call. I understand that this allows us to clear the banner message, but it doesn't make sense in the context of the terminal, and also doesn't makes sense if we ever introduce any other insights. I'm fine shipping with this as a initial implementation but I think we're going to have to rethink how this is architected as soon as we want to do anything else with the Insights class.

maestro-utils/src/main/kotlin/Insight.kt
amanjeetsingh150: Updated to list of listeners ✅

maestro-studio/server/src/main/java/maestro/studio/InsightService.kt
amanjeetsingh150: Done ✅

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Why does this doesn't fit on terminal. We are filtering on terminal side as well on the basis of level of insight in the same way how we do it for studio 🤔 

Do you mean we need more intuitive API to clear the current shown insight irrespective its studio or CLI?

Leland-Takamine: I mean in the terminal does `Insights.report(Insight("", Insight.Level.NONE))` do anything or is it a no-op?

Leland-Takamine (approved):

PR #1318, opened 2023-08-04
Removing implementation for initFlow

amanjeetsingh150:
## Proposed Changes

Removing IDB depends on removing pushAppState and pullAppState which are part of deprecated initFlow. This PR aims to remove initFlow implementation defering cleaning up for later

## Testing
- [x] Local
- [x] Staging


artem888 (approved):

maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlCommandReaderTest.kt
```diff
@@ -82,29 +83,6 @@ internal class YamlCommandReaderTest {
         assertThat(e.message).contains("Flow files must contain a config section and a commands section")
     }
 
-    @Test
-    fun initFlow(
-        @YamlFile("007_initFlow.yaml") commands: List<Command>,
```
artem888, line 87: lets not completely remove those tests , rather test that they throw an exception, you can copy from my PR

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -612,75 +602,6 @@ class IntegrationTest {
         }
     }
 
-    @Test
-    fun `Case 023 - runFlow with initFlow`() {
```
artem888, line 616: ditto, keeping one test at least which asserts the correct exception would be nice until we fully deprecate it

maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlCommandReaderTest.kt
amanjeetsingh150: Done ✅

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
amanjeetsingh150: Done ✅

PR #1326, opened 2023-08-09
Studio: Fixed device overflow on window resize

proksh:
**_Issue:_** The device's size expanded disproportionately when the window was resized.

**_Solution:_** Adjusted the styling by setting `flex-grow` to `flex-1` and positioning inner elements as `absolute`.


Leland-Takamine:
Nice! Tested locally and it works great!

(For future reference: a screen recording would've been helpful here to show the fixed behavior)

Leland-Takamine (approved):

PR #1328, opened 2023-08-09
Studio: Added play button with commands

proksh:
Added back a visible play button with the commands.


Leland-Takamine:
For future PRs, a screenshot would be helpful to see the changes visually

Issue #805, opened 2023-02-16
App (React Native) seems to become unresponsive when running with Maestro


matt-dalton:
I'm finding that with certain test runs, the test will get to a certain point, then the app will immediately freeze (i.e. it won't crash, but animations will stop and it will become unresponsive). Eventually maestro will timeout & fail the test. 

What's weird is this exact test suite was working reliably yesterday, and a different one was failing repeatedly.

I'm guessing it's when I get to something consuming more memory/CPU:
- Yesterday it happened when opening up a deep link from app close & navigating to a screen => I guess lots things happening at once
- Today it's freezing on a loading placeholder screen while navigating => Quite a bit of data fetched in the background, but not a lot on the screen

If I follow the exact same steps manually on the simulator (without maestro running) everything works reliably & smoothly every time.

There's no visible error, but if I rebuild Xcode after the crash I get a debug report:
```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Incident Identifier: 3A1127B9-D600-41D4-80BA-3BAA935E475F
CrashReporter Key:   5CDB127C-9FD0-635F-E716-77C74B52A34C
Hardware Model:      MacBookPro18,3
Process:             my_app [3211]
Path:                /Users/USER/Library/Developer/CoreSimulator/Devices/71BBA01C-2A5D-42CF-BC1C-1D14D1C3C031/data/Containers/Bundle/Application/15AC1364-CC10-43D5-B01C-D39BB5E504BB/my_app.app/my_app
Identifier:          com.my_app
Version:             0.0.0 (1)
Code Type:           X86-64 (Native)
Role:                Foreground
Parent Process:      launchd_sim [97381]
Coalition:           com.apple.CoreSimulator.SimDevice.71BBA01C-2A5D-42CF-BC1C-1D14D1C3C031 [30920]
Responsible Process: SimulatorTrampoline [37118]

Date/Time:           2023-02-16 19:05:34.3661 +0000
Launch Time:         2023-02-16 18:56:56.0366 +0000
OS Version:          macOS 12.5.1 (21G83)
Release Type:        User
Report Version:      104

Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: FRONTBOARD 2343432205 
[...]

PR #1337, opened 2023-08-11
Fix performance issues with maestro studio device refresh

artem888:
## Proposed Changes

This PR:
- adds compression for maestro studio screen refresh device screenshots to improve refresh performance

## Testing
- [x] tested locally

## Issues Fixed
Some of the problems related to the issues mentioned [here](https://github.com/mobile-dev-inc/maestro/issues/805) might be resolved

### DIFF
--- a/maestro-studio/web/src/pages/InteractPage.tsx
+++ b/maestro-studio/web/src/pages/InteractPage.tsx
@@ -49,6 +49,7 @@ const InteractPage = () => {
         try {
           const deviceScreen = await API.getDeviceScreen();
           setDeviceScreen(deviceScreen);
+          await wait(250);
         } catch (e) {
           console.error(e);
           await wait(1000);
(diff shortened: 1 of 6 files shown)


maestro-studio/web/src/pages/InteractPage.tsx
```diff
@@ -49,6 +49,7 @@ const InteractPage = () => {
         try {
           const deviceScreen = await API.getDeviceScreen();
           setDeviceScreen(deviceScreen);
+          await wait(250);
```
Leland-Takamine, line 52: Would it make more sense for this to be a minimum duration per iteration instead of a flat 250ms additional to every iteration? I guess it depends on what the hypothesis is here - can you explain the reasoning behind this?

artem888: I have researched a bit more about adding this delay, I think it doesn't make sense here actually, since ```const deviceScreen = await API.getDeviceScreen();
          setDeviceScreen(deviceScreen);``` takes about 600ms at best to complete and for more complex screens it is >1s.

axelniklasson (approved):

PR #1346, opened 2023-08-14
fix: cleaning up logs from TestDebugReporter

amanjeetsingh150:
## Proposed Changes

Cleaning up logs where TestDebugReporter is installed

## Testing
- [x] Locally


axelniklasson (approved):

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -68,12 +67,12 @@ object TestDebugReporter {
...
+    fun deleteOldFiles(days: Long = 14) {
         try {
             val currentTime = Instant.now()
             val daysLimit = currentTime.minus(Duration.of(days, ChronoUnit.DAYS))
 
-            Files.walk(path)
+            Files.walk(getDebugOutputPath())
                 .filter {
                     val fileTime = Files.getAttribute(it, "basic:lastModifiedTime") as FileTime
                     val isOlderThanLimit = fileTime.toInstant().isBefore(daysLimit)
```
ArthurSav, line 78: This is dangerous. 

Previously the test directory was defined by us so it was safe to delete but now users get to specify a custom directory. We'll be automatically deleting all files from that directory.

I propose we only delete if it's not a custom dir and let the user cleanup logs in their custom dir.

ArthurSav (changes requested):

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
amanjeetsingh150: Good point! Right now we delete debugOutputPathAsString which will have tests folder and all. Should we select that particular for cleanup or remove cleanup altogether?

ArthurSav (approved):
Ok, this is solved already

PR #1349, opened 2023-08-14
Add primary support for snapshot testing

denil-ct:
### <samp>🤖 Generated by Copilot at 99cbba9</samp>

This pull request adds a new feature to the maestro-orchestra module that allows users to compare screenshots with reference images using a new command type, `AssertSnapshotCommand`. It also adds support for this command type in the `MaestroCommand` sealed class, the `Orchestra` class, and the YAML syntax. It modifies the `maestro-orchestra-models` and `maestro-orchestra` modules to handle and execute this new command type.

The proposed YAML syntax and the comparison algorithm (right now its a simple pixel by pixel comparison) are open to suggestions. I haven't written tests, as the API or the implementation is not finalized yet.

You pass in a screenshot file path that you want the current screen to be compared against. You can also specify a threshold value, for example 98% meaning the images should have the same pixel across 98% of the image and only a 2% variation is allowed. This property is optional and if not declared, will be assumed to be a perfect match.

I'm new to Kotlin, so feel free to point out any silly mistakes I might have made.

## Testing
Manually tested changes with different combinations of yaml

```yaml
appId: com.apple.Preferences
---
- launchApp
- assertVisible: "Screen Time"
- assertVisible: "General"
- assertSnapshot:
    path: "test.png"
    threshold: "98%" 
```

```yaml
appId: com.apple.Preferences
---
- launchApp
- assertVisible: "Screen Time"
- assertVisible: "General"
- assertSnapshot:
    path: "test.png"
```

```yaml
appId: com.apple.Preferences
---
- launchApp
- assertVisible: "Screen Time"
- assertVisible: "General"
- assertSnapshot: "test.png"
```

## Issues Fixed

Added primary support for snapshot testing

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -476,6 +476,22 @@ data class TakeScreenshotCommand(
     }
 }
 
+data class AssertSnapshotCommand(
+    val path: String,
+    val threshold: String? = null,
+) : Command {
+
+    override fun description(): String {
+        return "Assert that current screen matches $path"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): AssertSnapshotCommand {
+        return copy(
+            path = path.evaluateScripts(jsEngine)
+        )
+    }
+}
+
 data class StopAppCommand(
     val appId: String,
 ) : Command {
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
@@ -61,6 +61,7 @@ data class MaestroCommand(
     val travelCommand: TravelCommand? = null,
     val startRecordingCommand: StartRecordingCommand? = null,
     val stopRecordingCommand: StopRecordingCommand? = null,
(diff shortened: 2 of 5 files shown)


axelniklasson:
Hi there @denil-ct. First, I want to take the time to thank you for writing up this PR. Screenshot testing is something that we consider to be out of scope for Maestro CLI and something that we plan on tackling holistically in the future. For the more immediate term: Maestro does output screenshots via the [takeScreenshot command](https://maestro.mobile.dev/api-reference/commands/takescreenshot) (as I'm sure you know), so it should be relatively straightforward to wire that into other screenshot comparison tooling to support your needs.

Again, I recognize the work that went into this - thanks again for your time!

orelviz:
@axelniklasson I'm not sure why you think this is not important. For some apps the ability to test snapshots are very important, not for design purposes, but for app functionality.
For example, an app that generate photos, or an app that displays them. 

 do understand how to implement that locally using libraries like pixelmatch and the likes, but how would you do that on maestro cloud?

axelniklasson:
@orelviz I'm not saying that it is not important, I believe screenshot testing is something that can provide great value and something I have used successfully in several settings before 🙂 That said, we consider it out of scope for Maestro CLI for now (which is why I closed this PR) and something we plan on integrating into Maestro Cloud in the future.

PR #1350, opened 2023-08-15
Improve CLI exit messages for Maestro Cloud

ArthurSav:
### <samp>🤖 Generated by Copilot at bba624d</samp>

This pull request enhances the maestro CLI tool to support better cloud upload functionality and error reporting. It adds new options for `CloudCommand` and `CloudInteractor` to control the timeout behavior and exit code. It also improves the `TestSuiteStatusView` to display more information about the test suite results and any issues with the cloud upload.

## Proposed Changes
These changes are mainly for CI users using the `maestro cloud` command

### Flow processing timeout
- New default timeout added (30min -> 60min)
- Changed default to exit with 1 (Failure) on timeout. Previously it would exit with 0 (Success)
- Added new CLI option to change timeout on demand `maestro cloud --result-wait-timeout=<timeout in minutes> ...`
- Added helpful message to clarify that is a waiting timeout and how to change the limit
- Added helpful message to clarify how to change exit code on timeout `maestro cloud --fail-on-timeout=<true|false>`
- Added helpful message that show exit code

**Previous**
x-timeout-skipped

**Updated**
timeout-with-success

### Flow Cancellation status & exit code
- If there's at least one FAILED flow, process will exit as failure.
   -  Previously, if there was at least on CANCELLED flow, the process would exit as SUCCESS. Even if there was a FAILED flow.
- Added helpful message to clarify what CANCELLED means
- Added helpful message to clarify how to change exit code on CANCELLATION `maestro cloud --fail-on-cancellation=<true|false>`
- Added helpful message that show exit code

**Previous**
x-complete-all-cancelled

x-complete-2cancelled-1success

**Updated**
complete-all-cancelled

complete-2cancelled-1success

## Testing
Manual & Local testing

## Issues Fixed
- User confusion around processing wait timeout.
- User confusion around exit codes when Cancellations occurred.
- Proper exit code when at least one flow failure exists.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/view/TestSuiteStatusView.kt
+++ b/maestro-cli/src/main/java/maestro/cli/view/TestSuiteStatusView.kt
@@ -53,7 +53,7 @@ object TestSuiteStatusView {
             )
 
             if (canceledFlows.isNotEmpty()) {
-                PrintUtils.warn("${canceledFlows.size} ${flowWord(canceledFlows.size)} Canceled (Internal Maestro Cloud Error)")
+                PrintUtils.warn("${canceledFlows.size} ${flowWord(canceledFlows.size)} Canceled")
             }
 
         } else {
@@ -69,11 +69,11 @@ object TestSuiteStatusView {
                 )
 
                 if (canceledFlows.isNotEmpty()) {
-                    PrintUtils.warn("${canceledFlows.size} ${flowWord(canceledFlows.size)} Canceled (Internal Maestro Cloud Error)")
+                    PrintUtils.warn("${canceledFlows.size} ${flowWord(canceledFlows.size)} Canceled")
                 }
             } else {
                 println()
-                PrintUtils.err("Upload was canceled (Internal Maestro Cloud Error)")
+                PrintUtils.err("All flows were canceled")
             }
         }
         println()


maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
```diff
@@ -139,9 +140,15 @@ class CloudCommand : Callable<Int> {
     @Option(hidden = true, names = ["--fail-on-cancellation"], description = ["Fail the command if the upload is marked as cancelled"])
     private var failOnCancellation: Boolean = false
 
+    @Option(hidden = true, names = ["--fail-on-timeout"], description = ["Fail the command if the upload times outs"])
+    private var failOnTimeout: Boolean = true
+
     @Option(hidden = true, names = ["--disable-notifications"], description = ["Do not send the notifications configured in config.yaml"])
     private var disableNotifications = false
 
+    @Option(hidden = true, names = ["--result-wait-timeout"], description = ["Minutes to wait until all flows complete"])
```
axelniklasson, line 149:
```suggestion
    @Option(hidden = true, names = ["--timeout"], description = ["Minutes to wait until all flows complete"])
```

What about just `--timeout` as arg name?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -257,10 +278,19 @@ class CloudInteractor(
...
+        return if (passed) {
+            PrintUtils.message("Process will exit with code 0 (SUCCESS)")
+            if (isCancelled) {
+                PrintUtils.message("* To change exit code on Cancellation, run maestro with this option: `maestro cloud --fail-on-cancellation=<true|false>`")
+            }
             0
+        } else {
+            PrintUtils.message("Process will exit with code 1 (FAIL)")
+            if (isCancelled && !containsFailure) {
+                PrintUtils.message("* To change exit code on Cancellation, run maestro with this option: `maestro cloud --fail-on-cancellation=<true|false>`")
```
axelniklasson, line 291: Huge nit: I don't think we should capitalize cancellation. We generally only capitalize `Flows` and `Uploads`.
```suggestion
                PrintUtils.message("* To change exit code on cancellation, run maestro with this option: `maestro cloud --fail-on-cancellation=<true|false>`")
```

maestro-cli/src/main/java/maestro/cli/view/TestSuiteStatusView.kt
```diff
@@ -63,11 +69,11 @@ object TestSuiteStatusView {
                 )
 
                 if (canceledFlows.isNotEmpty()) {
-                    println("(${canceledFlows.size} ${flowWord(canceledFlows.size)} Canceled)")
+                    PrintUtils.warn("${canceledFlows.size} ${flowWord(canceledFlows.size)} Cancelled (Internal Maestro Cloud Error)")
```
axelniklasson, line 72: Let's be consistent with the spelling of `canceled` in terms of how many `l` we use. In the console we use one l and say `canceled` so let's use the same here.

image

axelniklasson, line 72: What happens in the event of timeouts or skipped Flows due to dependency Flows fail? Will it say `Internal Maestro Cloud Error` then even though this is not the case?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -242,9 +260,12 @@ class CloudInteractor(
...
 
-        val passed = if (upload.status == UploadStatus.Status.CANCELED && failOnCancellation) {
-            false
-        } else upload.status != UploadStatus.Status.ERROR
+        val isCancelled = upload.status == UploadStatus.Status.CANCELED
+        val isFailure = upload.status == UploadStatus.Status.ERROR
+        val containsFailure =
+            upload.flows.find { it.status == UploadStatus.Status.ERROR } != null // status can be cancelled but also contain flow with failure
+
+        val passed = !isFailure && !containsFailure && !(isCancelled && failOnCancellation)
```
axelniklasson, line 268: Could we invert this logic to avoid all the negations? It is pretty hard to read 😅

maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
ArthurSav: I thought it would be too generic but then again, there aren't other cases for timeout for this command.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
ArthurSav: Simplified it a bit for easier readability

maestro-cli/src/main/java/maestro/cli/view/TestSuiteStatusView.kt
ArthurSav: > What happens in the event of timeouts or skipped Flows due to dependency Flows fail

Yes, there's no way to know the difference client side atm

ArthurSav: @axelniklasson Removed the Internal Error message as it might cause more confusion.

axelniklasson (approved):

ferdy-roz:
does this change also return a non-zero exit code in case of `SKIPPED`? or whatever happened here:

```
Upload did not complete in time due to an issue on mobile.dev side. Follow the results of your upload below:
https://console.mobile.dev/uploads/f4d270cc-932f-47f5-894e-8a59fbfde799?teamId=12ffe931-ef42-4f64-8b41-43cff78f82cc&appId=efaf4b2b-73f0-43df-ba49-0b1e8d39beca
SKIPPED
```

the above happened in our CI pipeline and it returned an exit code of 0. I only noticed it because I happened to check the logs.

ArthurSav:
@ferdy-roz That is correct. The error you're seeing is due to CLI Timeout which was set to 30min.
With the new changes you will be able to change the Timeout and control exit code

PR #1354, opened 2023-08-16
Fixing the leaking response bodies and improving XCTestDriverClient code

amanjeetsingh150:
## Proposed Changes

Untitled

1. Fixing the retry of calls after reinstalling the XcTest driver
2. Handling IOExceptions by mapping them to network exceptions
3. Showing meaningful messages when retrying to install the driver and failing
4. Fix open response bodies because of returning fake responses

## Testing
- [ ] Local
- [ ] Staging
- [ ] Change the executing of request to fail in case of 502 for each request
 
## Issues Fixed
1. Response bodies being closed
2. Improves error messages for XCTestUnreachable exception
3. Improves debugging for XCTestUnreachable exception


maestro-ios-driver/src/main/kotlin/xcuitest/api/NetworkErrorHandler.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    fun retryConnection(call: Call, response: Response): Response {
+        return if (retry < 3) {
+            xcTestInstaller.start()?.let {
+                client = it
+            }
+            response.close()
+            retry++
+            PrintUtils.log("ℹ️ Retrying connection to the XCUITest server for ${retry}...")
```
amanjeetsingh150, line 64: What should be alternative to the emoji in case of no-ansi?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -225,34 +237,31 @@ class XCTestDriverClient(
...
-                    client = it
-                }
-                exception = e
-            }
-            Thread.sleep(300)
+        val response = try {
+             chain.proceed(chain.request())
+        } catch (ioException: IOException) {
+            val networkException = mapNetworkException(ioException)
+            networkErrorHandler.reinitializedXCTestClient()?.let { client = it }
```
amanjeetsingh150, line 244: Might need to do this after installation

maestro-ios-driver/src/main/kotlin/xcuitest/api/NetworkErrorHandler.kt
axelniklasson: I think you can skip an alternative and just go with the text for no-ansi

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
axelniklasson: Hm, not sure I follow - could you elaborate on this? What do we need to do after installation? Installation of the XCTest runner app?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -168,8 +165,8 @@ class XCTestDriverClient(
         return executeJsonRequestUNCHECKED("eraseText", EraseTextRequest(charactersToErase, appIds))
     }
 
-    fun deviceInfo(): Response {
-        return executeJsonRequestUNCHECKED("deviceInfo", Unit)
+    fun deviceInfo(httpUrl: HttpUrl = client.xctestAPIBuilder("deviceInfo").build()): Response {
```
felipevolpone, line 168: why this changed?

felipevolpone (approved):
Great work Amanjeet. I don't have lot of context about this part, but it looks good to me

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: To make this function testable, so that we can configure HTTP URL and test with mockwebserver

ArthurSav (approved):

PR #1361, opened 2023-08-17
Use Server-Sent Events instead of custom long-polling in Maestro Studio

Leland-Takamine:
## Impact
* Maestro Studio can now run in multiple tabs simultaneously
* Improves Maestro Studio performance

## Background
* Custom long-polling was causing simultaneous API calls to hang (mostly in Chrome)
* Switching to Server-Sent Events instead
* In order to mock SSEs we need to upgrade MSW to the fetch release candidate version. Context: https://github.com/mswjs/msw/blob/feat/standard-api/MIGRATING.md#support-readablestream-mocked-responses

### DIFF
--- a/maestro-studio/web/src/helpers/models.ts
+++ b/maestro-studio/web/src/helpers/models.ts
@@ -38,7 +38,6 @@ export type ReplCommand = {
 }
 
 export type Repl = {
-  version: number
   commands: ReplCommand[]
 }
 
--- a/maestro-studio/web/storybook-assets/mockServiceWorker.js
+++ b/maestro-studio/web/storybook-assets/mockServiceWorker.js
@@ -2,13 +2,14 @@
 /* tslint:disable */
 
 /**
- * Mock Service Worker (0.49.2).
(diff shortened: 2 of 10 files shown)


maestro-studio/web/storybook-assets/mockServiceWorker.js
```diff
@@ -2,13 +2,14 @@
 /* tslint:disable */
 
 /**
- * Mock Service Worker (0.49.2).
+ * Mock Service Worker (0.0.0-fetch.rc-16).
```
Leland-Takamine, line 5: These changes are all generate by MSW

maestro-studio/web/src/helpers/models.ts
```diff
@@ -38,7 +38,6 @@ export type ReplCommand = {
 }
 
 export type Repl = {
-  version: number
```
Leland-Takamine, line 41: This version was only used for long polling. No longer necessary now that we migrated to SSE

proksh:
Hey @Leland-Takamine Some integration test are failing with the last commit.

proksh (approved):

PR #1367, opened 2023-08-21
Added device context to make the code clean

proksh:
## Proposed Changes

Rather than sending deviceScreen, inspectedElement and hover elements deep inside, created context so that it can be shared everywhere without need of props.


maestro-studio/web/src/components/device-and-device-elements/ElementsPanel.tsx
```diff
@@ -13,22 +14,29 @@ const compare = (a: string | undefined, b: string | undefined) => {
...
 
 export default function ElementsPanel({
-  deviceScreen,
-  onElementSelected,
-  hoveredElement,
-  setHoveredElement,
+  // deviceScreen,
+  // onElementSelected,
+  // hoveredElement,
+  // setHoveredElement,
```
axelniklasson, line 29: Should the commented properties be deleted?

axelniklasson (approved):

PR #1368, opened 2023-08-21
Fix: use correct tmp dir in LocalXCTestInstaller

axelniklasson:
### <samp>🤖 Generated by Copilot at 0fcc35a</samp>

Fixed a bug in `LocalXCTestInstaller.kt` that caused test installation to fail on some machines. Changed `tempDir` to use the system property `java.io.tmpdir` instead of the environment variable `TMPDIR`.

## Testing
Tested locally through running iOS sample Flow.


artem888 (commented):
could you please explain the fix a little more, from my understanding `TMPDIR` should work just fine on macOS and `java.io.tmpdir` would be needed for cross-platform tmp dir locations, which we don't need here

PR #1370, opened 2023-08-21
Fix creating connections

amanjeetsingh150:
## Proposed Changes

1. Catching IOException in The `XCTestClient#ensureOpen` call which only caught connectException but driver is really going to be ready when it doesn't have any IOException on a call
2. Increase the retries to 5 instead, since I saw 4 is enough
3. Change call to expect instead of get() which gives null

## Testing
- [x] Local
- [ ] Staging

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -157,9 +157,9 @@ class IOSDriver(
 
     fun viewHierarchy(): TreeNode {
         LOGGER.info("Requesting view hierarchy of the screen")
-        val hierarchyResult = iosDevice.viewHierarchy().get()
-        LOGGER.info("Depth of the screen is ${hierarchyResult?.depth ?: 0}")
-        if (hierarchyResult?.depth != null && hierarchyResult.depth > WARNING_MAX_DEPTH) {
+        val hierarchyResult = iosDevice.viewHierarchy().expect {  }
+        LOGGER.info("Depth of the screen is ${hierarchyResult.depth}")
+        if (hierarchyResult.depth > WARNING_MAX_DEPTH) {
             val message = "The view hierarchy has been calculated. The current depth of the hierarchy " +
                     "is ${hierarchyResult.depth}. This might affect the execution time of your test. " +
                     "If you are using React native, consider migrating to the new " +
@@ -169,7 +169,7 @@ class IOSDriver(
         } else {
             Insights.report(Insight("", Insight.Level.NONE))
         }
-        val hierarchy = hierarchyResult?.axElement ?: return TreeNode()
+        val hierarchy = hierarchyResult.axElement
         return mapViewHierarchy(hierarchy)
     }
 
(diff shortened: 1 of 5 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/api/NetworkErrorHandler.kt
```diff
@@ -21,7 +21,7 @@ class NetworkErrorHandler(
     companion object {
         const val RETRY_RESPONSE_CODE = 503
         private const val NO_RETRY_RESPONSE_CODE = 502
-        private const val MAX_RETRY = 3
+        private const val MAX_RETRY = 5
```
amanjeetsingh150, line 24: I'm holding off making this configurable right away, since I noticed that 4 is enough

maestro-ios-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
```diff
@@ -93,21 +93,21 @@ class LocalXCTestInstaller(
...
     }
 
     private fun ensureOpen(): Boolean {
         XCRunnerCLIUtils.ensureAppAlive(UI_TEST_RUNNER_APP_BUNDLE_ID, deviceId)
         return MaestroTimer.retryUntilTrue(10_000, 100) {
             try {
-                subTreeOfRunnerApp().use { it.isSuccessful }
-            } catch (ignore: ConnectException) {
+                xcTestDriverStatusCheck().use { it.isSuccessful }
+            } catch (ignore: IOException) {
```
amanjeetsingh150, line 104: Because connect exception is just one case there could be many and we should only consider driver ready when it has no IOException

maestro-ios-driver/src/main/kotlin/xcuitest/api/NetworkErrorHandler.kt
artem888: just out of curiousity, how were you able to reproduce this, also using MockWebServer?

amanjeetsingh150: This I noticed in production that 4th call was successful hence increasing retries

felipevolpone (approved):

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -157,9 +157,9 @@ class IOSDriver(
 
     fun viewHierarchy(): TreeNode {
         LOGGER.info("Requesting view hierarchy of the screen")
-        val hierarchyResult = iosDevice.viewHierarchy().get()
-        LOGGER.info("Depth of the screen is ${hierarchyResult?.depth ?: 0}")
-        if (hierarchyResult?.depth != null && hierarchyResult.depth > WARNING_MAX_DEPTH) {
+        val hierarchyResult = iosDevice.viewHierarchy().expect {  }
```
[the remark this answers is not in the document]
amanjeetsingh150: Actually that is the problem, right now the cases it errors specifically where the XCUITest server returns 5xx or 4xx it basically gives null, we should be failing this hard since we don't expect any known errors to happen here. Let me know if you think of any error that should be handled here?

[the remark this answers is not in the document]
felipevolpone: Gotcha, no worries then. Thanks for the context Amanjeet

PR #1371, opened 2023-08-21
Added element id and copy option for it

proksh:
## Changes

- Updated Action Modal to have element id (if element has id)
- Copy command wont work if any text is selected. That text will be copied instead

Screenshot 2023-08-21 at 5 36 55 PM


proksh:
Changed ui to include text as well -
Screenshot 2023-08-22 at 12 35 25 PM

maestro-studio/web/src/components/device-and-device-elements/ActionModal.tsx
```diff
@@ -263,8 +263,9 @@ export default function ActionModal({
           break;
         case "KeyC":
           if (
-            (isMac && e.metaKey) ||
-            (!isMac && e.ctrlKey && !e.altKey && !e.shiftKey)
+            ((isMac && e.metaKey) ||
+              (!isMac && e.ctrlKey && !e.altKey && !e.shiftKey)) &&
+            (!window || (window && window.getSelection()?.toString() === "")) // No text is selected
```
axelniklasson, line 268: Nit: This is a bit challenging to read -- could we simplify this logic somehow or add comments for each condition?

axelniklasson (approved):
🔝

Leland-Takamine (approved):

Leland-Takamine:
I noticed in the second screenshot that the id is ellipsized. I believe one of the users mentioned that they would like to be able to copy part of the id in some cases. I'm not certain, but might be work checking out that

proksh:
Got it. Removed the truncation change ui slightly @Leland-Takamine  -
Screenshot 2023-08-23 at 11 09 36 AM

maestro-studio/web/src/components/device-and-device-elements/ActionModal.tsx
proksh: Can you check now @axelniklasson ? Does this looks better?

```
case "KeyC":
          if (
            (isMac && e.metaKey) || // If mac - Command is pressed
            (!isMac && e.ctrlKey && !e.altKey && !e.shiftKey) // Or If not mac - Only control key is pressed
          ) {
            // If no text is selected
            if (window && window.getSelection()?.toString() === "") {
              e.preventDefault();
              if (typeof selectedCommand?.content === "string") {
                copyCommand(selectedCommand.content);
              }
            }
          }
          break;
```

Leland-Takamine:
Looks good!

simon-gilmurray:
Thanks for adding this - looks great! 
I've noticed on all the commands when running on Android, Studio prefixes the ID with the long package name - this feels kind of unnecessary as all Maestro flows just pass it in as an `appID` and then never use it again, which leads to having to delete it from every command when copying and pasting. Appreciate it is probably nice to have in one place somewhere on the UI though?
Also iOS doesn't seem to have this issue and just uses the ID without the `appID` prefix.

PR #1372, opened 2023-08-22
Error message improvements

Leland-Takamine:
### Overview

* Proper error messages for empty workspaces
* Surface workspace validation errors before trying to connect to device when running `maestro test`
* Lay test fixture groundwork for future error improvements

### Behavior Change

This PR also changes the behavior when includeTags are specified in both config.yaml and as a command line option. Previously, in order for a flow to be included, it would need to match one of the config.yaml includeTags AND one of the command line includeTags. Even though this might be useful for some specific scenarios, this behavior is confusing and hard to explain. Now, in order for a flow to be included, it needs to match only one includeTag regardless of where the includeTag was specified.

#### Flow path doesn't exist

```
Flow path does not exist: {PROJECT_DIR}/src/test/resources/workspaces/e000_flow_path_does_not_exist/workspace
```

#### Flow directory doesn't contain any Flow files (recursively)

```
Flow directory does not contain any Flow files: {PROJECT_DIR}/src/test/resources/workspaces/e001_directory_does_not_contain_flow_files/workspace
```

#### Top-level flow directory doesn't contain any files (but subdirectories do)

```
Top-level directory does not contain any Flows: {PROJECT_DIR}/src/test/resources/workspaces/e002_top_level_directory_does_not_contain_flow_files/workspace
To configure Maestro to run Flows in subdirectories, check out the following resources:
  * https://maestro.mobile.dev/cli/test-suites-and-reports#inclusion-patterns
  * https://blog.mobile.dev/maestro-best-practices-structuring-your-test-suite-54ec390c5c82
```

#### Inclusion patterns (ie: flows config in config.yaml) don't match any Flows

```
Flow inclusion pattern(s) did not match any Flow files:
- FlowA.yaml
- FlowB.yaml
```

#### Include tags / exclude tags don't match any Flow files:

```
Include / Exclude tags did not match any Flows:

Include Tags:
- parameterInclude
- configInclude

Exclude Tags:
- parameterExclude
- configExclude
```

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/App.kt
+++ b/maestro-cli/src/main/java/maestro/cli/App.kt
@@ -19,29 +19,16 @@
 
 package maestro.cli
 
-import maestro.cli.command.BugReportCommand
-import maestro.cli.command.CloudCommand
-import maestro.cli.command.DownloadSamplesCommand
-import maestro.cli.command.LoginCommand
-import maestro.cli.command.LogoutCommand
-import maestro.cli.command.PrintHierarchyCommand
-import maestro.cli.command.QueryCommand
-import maestro.cli.command.RecordCommand
-import maestro.cli.command.StudioCommand
-import maestro.cli.command.TestCommand
-import maestro.cli.command.UploadCommand
+import maestro.cli.command.*
 import maestro.cli.report.TestDebugReporter
 import maestro.cli.update.Updates
 import maestro.cli.util.ErrorReporter
 import maestro.cli.view.box
 import maestro.debuglog.DebugLogStore
-import maestro.debuglog.LogConfig
-import org.slf4j.LoggerFactory
 import picocli.CommandLine
 import picocli.CommandLine.Command
 import picocli.CommandLine.Option
-import java.util.Properties
-import kotlin.io.path.absolutePathString
+import java.util.*
 import kotlin.system.exitProcess
 
 @Command(
@@ -94,9 +81,6 @@ fun main(args: Array<String>) {
     // https://stackoverflow.com/a/17544259
     System.setProperty("apple.awt.UIElement", "true")
 
-    // logs & debug output
-    val logger = LoggerFactory.getLogger(App::class.java)
-
     Dependencies.install()
     Updates.fetchUpdatesAsync()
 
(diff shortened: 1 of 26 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -119,6 +146,18 @@ object WorkspaceExecutionPlanner {
         return null
     }
 
+    private fun isFlowFile(path: Path): Boolean {
+        if (!path.isRegularFile()) return false // Not a file
+        val extension = path.extension
+        if (extension != "yaml" && extension != "yml") return false // Not YAML
+        if (path.nameWithoutExtension == "config") return false // Config file
+        return true
```
igorsmotto, line 154: Can we have different error messages for the wrong extension or missing extension as well?
If the flow isn't found because of a type in the extension it can be hard to find the reason, especially on Windows where extensions are usually not shown to the user.


You can ignore this comment if you think this is out of the scope of this PR.

Leland-Takamine: Hmm the workspace may have other types of files so we can't know whether a non-yaml extension is a mistake or not

maestro-cli/src/main/java/maestro/cli/App.kt
```diff
@@ -19,29 +19,16 @@
...
-import maestro.cli.command.DownloadSamplesCommand
-import maestro.cli.command.LoginCommand
-import maestro.cli.command.LogoutCommand
-import maestro.cli.command.PrintHierarchyCommand
-import maestro.cli.command.QueryCommand
-import maestro.cli.command.RecordCommand
-import maestro.cli.command.StudioCommand
-import maestro.cli.command.TestCommand
-import maestro.cli.command.UploadCommand
+import maestro.cli.command.*
```
igorsmotto, line 22: I personally prefer having all the imports explicitly in the file since IntelliJ IDEA does a good job of hiding them from me and we avoid using the wrong dependency.

Do we have any style rules to this? If not, feel free to ignore my comment.

igorsmotto (approved):
Interesting work with WorkspaceExecutionPlannerErrorsTest!

maestro-cli/src/main/java/maestro/cli/App.kt
Leland-Takamine: Agree - Updated my IDE settings to not use star imports and updated the PR

PR #1377, opened 2023-08-23
Start device command

ArthurSav:
## Proposed Changes

- Adds new command that starts or creates a new device (Android emulator, iOS simulator)
`maestro start-device --platform <platform> --os-version<version>`

- Modifies the `maestro test` command to either start or create a device, when no devices are running.

### maestro start-device
This new command will create a Maestro recommended device (similar specs as Maestro Cloud) for either Android or iOS.
- If no similar device exists, Maestro will attempt to create it.
- If required dependencies are missing, Maestro will attempt to install them or throw a message.
- If successful, it will launch the device.

Menu & options
Screenshot 2023-08-24 at 18 31 19

Android - Missing system image
Screenshot 2023-08-24 at 18 34 43

iOS - Missing runtime version
Screenshot 2023-08-24 at 18 40 39

### maestro test
Running `maestro test` with **no devices running** will:
- Prompt the user with device config details
- If device doesn't exist, it will be created (same as with `start-device` command)
- If successful, it will launch the device.

**Note: No changes in behaviour, if one or multiple devices are already running.**

New: Starts a Maestro recommended device
Screenshot 2023-08-24 at 19 16 19

Previous: It would list all devices on the computer
Screenshot 2023-08-24 at 18 53 25

## Testing
- Local / Manually


maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -0,0 +1,145 @@
...
+        return when (platform) {
+            Platform.ANDROID -> {
+                getOrCreateAndroidDevice(osVersion, forceCreate)
+            }
+
+            Platform.IOS -> {
+                getOrCreateIosDevice(osVersion, forceCreate)
+            }
+
+            else -> throw CliError("Unsupported platform $platform. Please one of: android, ios")
```
axelniklasson, line 20:
```suggestion
            else -> throw CliError("Unsupported platform $platform. Please specify one of: android, ios")
```

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -0,0 +1,145 @@
...
+            existingDeviceId ?: DeviceService.createIosDevice(deviceName, device, runtime).toString()
+        } catch (e: IllegalStateException) {
+            val error = e.message ?: ""
+            if (error.contains("Invalid runtime")) {
+                val msg = "Required runtime to create the simulator is not installed: $runtime\n\n" +
+                        "To install additional iOS runtimes checkout this guide:\n" +
+                        "* https://developer.apple.com/documentation/xcode/installing-additional-simulator-runtimes"
+                throw CliError(msg)
+            } else if (error.contains("Invalid device type")) {
+                throw CliError("Device type $device is either not supported or not found.")
```
axelniklasson, line 64: Is there a way we can guide users to recover from this error?

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -0,0 +1,145 @@
...
+            } else {
+                throw CliError(error)
+            }
+        }
+
+        if (existingDeviceId == null) PrintUtils.message("Created simulator $deviceName ($deviceUUID)")
```
axelniklasson, line 70:
```suggestion
        if (existingDeviceId == null) PrintUtils.message("Created simulator with name $deviceName and udid $deviceUUID)")
```

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -0,0 +1,145 @@
...
+        // dependencies
+        if (existingDevice == null && !DeviceService.isAndroidSystemImageInstalled(systemImage)) {
+            PrintUtils.err("The required system image $systemImage is not installed.")
+
+            PrintUtils.message("Would you like to install it? y/n")
+            val r = readlnOrNull()?.lowercase()
+            if (r == "y" || r == "yes") {
+                PrintUtils.message("Attempting to install $systemImage via Android SDK Manager...\n")
+                if (!DeviceService.installAndroidSystemImage(systemImage)) {
+                    throw CliError("Was unable to install required dependencies.")
```
axelniklasson, line 108: Can we provide a more detailed error message here and/or guide users to self-recover like we do below?

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -0,0 +1,145 @@
...
+                systemImage = systemImage,
+                tag = DeviceConfigAndroid.tag,
+                abi = DeviceConfigAndroid.abi,
+                force = forceCreate
+            )
+        } catch (e: IllegalStateException) {
+            throw CliError("${e.message}")
+        }
+
+        if (existingDevice == null) PrintUtils.message("Created Android emulator: $name ($systemImage)")
```
axelniklasson, line 137: Is `$systemImage` the same as `deviceId` that we reference in maestro CLI and that users can use to specify which device to run on? 

```
      --udid, --device=<deviceId>
                  (Optional) Select a device to run on explicitly
```

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -168,6 +177,217 @@ object DeviceService {
         }
     }
 
+    /**
```
axelniklasson, line 180: For all the device creation and env related functions when it comes to Android, do we know whether that works on Windows / Linux too?

maestro-cli/src/main/java/maestro/cli/device/PickDeviceInteractor.kt
```diff
@@ -17,6 +18,12 @@ object PickDeviceInteractor {
                 var result: Device = pickedDevice
 
                 if (result is Device.AvailableForLaunch) {
+                    when (result.platform) {
+                        Platform.ANDROID -> PrintUtils.message("Launching Android device...")
```
axelniklasson, line 22:
```suggestion
                        Platform.ANDROID -> PrintUtils.message("Launching Android emulator...")
```

maestro-cli/src/main/java/maestro/cli/device/PickDeviceInteractor.kt
```diff
@@ -47,13 +54,22 @@ object PickDeviceInteractor {
     }
 
     private fun startDevice(): Device {
-        val availableDevices = DeviceService.listAvailableForLaunchDevices()
+        PrintUtils.message("No devices found. Would you like to start a new one? y/n")
+        val proceed = readlnOrNull()?.lowercase()?.trim()
+        if (proceed == "no" || proceed == "n") {
+            throw CliError("Please either start a device manually or via Maestro to proceed running your flows")
```
axelniklasson, line 60: Does starting via Maestro refer to running `maestro start-device`? If so, let's update the error message to be explicit here

```suggestion
            throw CliError("Please either start a device manually or via running maestro start-device to proceed running your flows")
```

maestro-cli/src/main/java/maestro/cli/util/DeviceConfig.kt
```diff
@@ -0,0 +1,32 @@
+package maestro.cli.util
+
+internal object DeviceConfigIos {
+
+    val device: String = "iPhone-11"
+    val versions = listOf(15, 16)
+    val runtimes = mapOf(
+        15 to "iOS-15-5",
+        16 to "iOS-16-2"
+    )
```
axelniklasson, line 10: Should we fetch this from Cloud instead? If not old CLI clients will be using the wrong minor version if we decide to move Cloud to for example iOS 16.4.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -0,0 +1,70 @@
...
+                Platform.ANDROID -> DeviceConfigAndroid.defaultVersion.toString()
+                else -> ""
+            }
+        }
+        val o = osVersion.toIntOrNull()
+
+        DeviceCreateUtil.getOrCreateDevice(p, o, forceCreate).let {
+            PrintUtils.message(if (p == Platform.IOS) "Launching simulator..." else "Launching emulator...")
+            DeviceService.startDevice(it)
+        }
```
axelniklasson, line 64: What is the error message here when trying to start an iOS device from a Windows/Linux machine?

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
ArthurSav: Ideally, this should never happen. From what I know, Xcode has all the devices available without having to install additional components.

This is to cover edge cases.

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
ArthurSav: Added a better message with self recovery instructions

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
ArthurSav: It should work, as we use standard tools from the Android SDK i.e  `cmdline-tools/latest/bin/avdmanager`

With that said, I have not tested in these environments. I can reach out to Manish to run a few command to validate

ArthurSav: Reached out to Manish to assist with testing

maestro-cli/src/main/java/maestro/cli/util/DeviceConfig.kt
ArthurSav: I'm ok with it being hardocded for the time being. Minor versions usually don't introduce major OS changes and if it is an issue for a user then we can push them to update (provided we support it).

Open to discuss introducing remote config though.

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
ArthurSav: According to Manish device listing never worked and by extension these updates will also not work. This is a [limitation](https://maestro.mobile.dev/getting-started/installing-maestro/windows#2884) with WSL + ANDROID_HOME.

What I propose is to detect if it's running on Windows and show a proper error, until we decide to properly support it.

I will also see if there's a quick solution for referencing ANDROID_HOME from WSL

ArthurSav: We tried to install sdk tools in the WSL environment but it still didn't work. 
Will check if we can show a more informative error but Windows will not work with this command.

ArthurSav:
Added better handling for missing `cmdline-tools` + Windows WSL

PR #1382, opened 2023-08-24
feat: add HTTPInterceptor and add more log messages to installer

igorsmotto:
### <samp>🤖 Generated by Copilot at 99e4080</samp>

This pull request enhances the logging functionality of the maestro-ios-driver module, which enables automated testing of iOS apps. It adds a new library for logging HTTP traffic, `square-okhttp-logs`, and updates the gradle configuration and the `libs.versions.toml` file accordingly. It also refactors and improves the log messages in the `LocalXCTestInstaller` and `XCTestDriverClient` classes, which handle the installation and execution of XCUITest runner apps.

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -23,6 +23,7 @@ import kotlin.reflect.KClass
 class XCTestDriverClient(
     private val installer: XCTestInstaller,
     private val logger: Logger,
+    private val httpInterceptor: HttpLoggingInterceptor? = null
 ) {
     private lateinit var client: XCTestClient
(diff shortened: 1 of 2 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -31,24 +32,29 @@ class XCTestDriverClient(
     private var isShuttingDown = false
 
     private val networkErrorHandler by lazy { NetworkErrorHandler(installer, logger) }
+    private val httpInterceptor = HttpLoggingInterceptor()
```
amanjeetsingh150, line 35: I think we somehow need these logs to be streamed to LOGGER right? Where does it output logs by default. Having LOGGER will ensure its streamed on gcp as well

amanjeetsingh150 (approved):

PR #1405, opened 2023-08-30
decouple simctl boot and launch

artem888:
## Proposed Changes

Decouples boot simulator out of launch function (change needed for improving inputText stability on cloud)


igorsmotto:
Heads-up: You also need to update CompanionService which uses the LocalSimulatorUtils on Cloud

igorsmotto (approved):

PR #1415, opened 2023-09-05
Move REPL logic to the frontend

Leland-Takamine:


maestro-studio/web/src/context/ReplContext.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export const ReplProvider = ({ children }: {
+  children: ReactNode
+}) => {
+  const [repl, setRepl] = useState<Repl>(() => restoreRepl());
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+  useEffect(() => {
+    localStorage.setItem('repl', JSON.stringify(repl));
```
Leland-Takamine, line 34: @proksh Thoughts on how to best deal with malformatted data in localstorage? Is that necessary?

proksh (approved):

maestro-studio/web/src/context/ReplContext.tsx
Leland-Takamine: Let's leave as is for now since it's not clear if versioning will be needed long term. We can just remember to add versioning whenever we make a breaking change (and we can treat the absence of a version as version 0)

PR #1417, opened 2023-09-05
Unify logic for eraseText and inputText commands on iOS

artem888:
## Proposed Changes

This PR unifies logic for inputText and eraseText commands on iOS, because they're using the same XCTest API under the hood. It will also reuse various delay that we've implemented for inputText to improve the command stability

## Testing

- [x] local test
- [ ] staging test


ArthurSav (approved):

axelniklasson (approved):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/EraseTextHandler.swift
```diff
@@ -6,38 +6,44 @@ import Network
...
+            
+            let duration = Date().timeIntervalSince(start)
+            logger.info("Erase text duration took \(duration)")
+            return HTTPResponse(statusCode: .ok)
+        } catch {
+            logger.error("Error erasing text of \(requestBody.charactersToErase) characters: \(error)")
+            return errorResponse(message: "internal error")
+        }
 
         return HTTPResponse(statusCode: .ok)
```
axelniklasson, line 37: This code will never be executed so we can delete it

PR #1419, opened 2023-09-05
[MOB 2118] Add Media command for android and iOS

amanjeetsingh150:
## Proposed Changes

1. Add media command to support JPG, PNG, GIF, and MP4 for iOS and Android.
2. Deletion of media at the end
3. Validation:
- maestro test command and input is a single flow file it only validates if the path of media is valid or not
- maestro test command and input is workspace, along with checking valid media path it also check if media exist in workspace or not
- maestro cloud command and input is workspace, along with checking valid media path it also check if media exist in workspace or not

## Testing
- [ ] Local
- [ ] Staging


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -396,7 +414,14 @@ class Service(
...
-        private const val LENGTH_KEY_VALUE_PAIR = 2
+    internal fun Throwable.internalError() = Status.INTERNAL.withDescription(message).asException()
+
+    enum class FileType(val ext: String, val mimeType: String) {
+        JPG("jpg", "image/jpg"),
+        JPEG("jpeg", "image/jpeg"),
+        PNG("png", "image/png"),
+        GIF("gif", "image/gif"),
+        MP4("mp4", "video/mp4"),
+        MP3("mp3", "audio/mp3")
```
amanjeetsingh150, line 425: Going to remove audio for this first release since it needs a little investigation for iOS

maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -25,6 +28,7 @@ object WorkspaceExecutionPlanner {
 
         if (input.isRegularFile()) {
             validateFlowFile(input)
+            MediaCommandsValidator(input).validate(input)
```
amanjeetsingh150, line 31: Might need to handle one more case here, which is `maestro cloud flow.yaml` this would be cloud command and single flow file. In this case both validations would be needed

maestro-android/src/androidTest/java/dev/mobile/maestro/Media.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun getOutputStream(mediaName: String, mediaExt: String): OutputStream? {
+        val uri = when (mediaExt) {
+            Service.FileType.JPG.ext,
+            Service.FileType.PNG.ext,
+            Service.FileType.GIF.ext,
+            Service.FileType.JPEG.ext -> MediaStore.Images.Media.EXTERNAL_CONTENT_URI
+            Service.FileType.MP4.ext -> MediaStore.Video.Media.EXTERNAL_CONTENT_URI
+            Service.FileType.MP3.ext -> MediaStore.Audio.Media.EXTERNAL_CONTENT_URI
+            else -> throw IllegalStateException("mime type not yet supported")
```
artem888, line 18: nit : could also output `mediaExt` here

`"mime type $mediaExt not yet supported"`

maestro-client/src/main/java/maestro/Media.kt
```diff
@@ -0,0 +1,24 @@
...
+}
+
+fun String.isMediaFile(): Boolean {
+    return this == MediaExt.MP3.extName ||
+            this == MediaExt.PNG.extName ||
+            this == MediaExt.JPEG.extName ||
+            this == MediaExt.JPG.extName ||
+            this == MediaExt.GIF.extName ||
+            this == MediaExt.MP4.extName ||
+            this == MediaExt.MP3.extName
```
artem888, line 23: MP3 is duplicated here

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
```diff
@@ -142,6 +142,24 @@ class SimctlIOSDevice(
             .also { screenRecording = null }
     }
 
+    override fun addMedia(path: String) {
+        backupSimulatorMedia()
+        LocalSimulatorUtils.addMedia(deviceId, path)
+    }
+
+    private fun backupSimulatorMedia() {
+        if (!LocalSimulatorUtils.photoDataBackupExists()) {
```
artem888, line 151: does it make sense to make a backup on every new addMedia call, not only when lazy var is not initialized?

maestro-orchestra/src/main/java/maestro/orchestra/workspace/MediaCommandsValidator.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import maestro.orchestra.error.MediaFileNotFound
+import maestro.orchestra.yaml.YamlCommandReader
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import kotlin.io.path.isRegularFile
+
+class MediaCommandsValidator(private val input: Path) {
```
artem888, line 11: nit : a separate set of tests would be nice for this file

artem888 (approved):
nice work 👏 LGTM after comments addressed

maestro-orchestra/src/main/java/maestro/orchestra/workspace/MediaCommandsValidator.kt
```diff
@@ -0,0 +1,53 @@
...
+import okio.Path.Companion.toPath
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import kotlin.io.path.absolutePathString
+import kotlin.io.path.isRegularFile
+import kotlin.io.path.notExists
+
+class MediaCommandsValidator(private val input: Path) {
```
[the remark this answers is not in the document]
amanjeetsingh150: Yes, I realized that while testing. I have changed that accordingly.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
amanjeetsingh150: Done ✅

maestro-client/src/main/java/maestro/Media.kt
amanjeetsingh150: Removed ✅

maestro-android/src/androidTest/java/dev/mobile/maestro/Media.kt
amanjeetsingh150: Done ✅

maestro-ios/src/main/java/ios/simctl/SimctlIOSDevice.kt
amanjeetsingh150: I am doing it once before we add media because we want the data to come back to the original state where it was before adding media, for the simulator. 

If we do it for each call then it would keep the recent photos in case of addition of multiple media files.

maestro-orchestra/src/main/java/maestro/orchestra/workspace/MediaCommandsValidator.kt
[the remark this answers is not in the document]
Leland-Takamine: Shouldn't that also be the same logic for all path references in the flow, not just media files?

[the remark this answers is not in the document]
amanjeetsingh150: Yes unifying this logic and removing this means the same nature as run flow:

1. Running locally  (both `maestro test single_flow.yaml` and `maestro test ./workspace`) would fail in one case when the path doesn't exist
2. With maestro cloud command it fail in both cases if its missing in workspace or not existing.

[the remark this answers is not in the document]
Leland-Takamine: I meant that this should also be the same logic for subflow paths:

> Yes, I realized that while testing. I have changed that accordingly. Now this file only checks if all the media files are in workspace in case of input being workspace directory.

So in that case, I'm curious why this validation logic is written in a class specific to media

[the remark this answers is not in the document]
Leland-Takamine: * If path doesn't exist: Should fail in all cases
* If path is outside the workspace: Should fail for `maestro test workspace/` and `maestro cloud workspace/`
* `maestro test single_flow.yaml` doesn't have a concept of a workspace yet so it should only check for existence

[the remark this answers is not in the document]
amanjeetsingh150: > If path is outside the workspace: Should fail for maestro test workspace/ and maestro cloud workspace/

For this case I tried with runFlow command and it doesn't fail in case of `maestro test workspace` command. Is that intended?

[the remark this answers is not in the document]
Leland-Takamine: I don't think that was intentional. To me it would make sense for this case to be an error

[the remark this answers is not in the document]
amanjeetsingh150: Hey @Leland-Takamine I fixed this by creating a common implementation which I am planning to raise as a separate PR. Can you have look at the latest commit if the abstraction looks good and then I'll merge this and raise the runFlow command PR. It would basically be 

```
allFlows.forEach {
            val commands = YamlCommandReader.readCommands(it).mapNotNull {  it.runFlowCommand }
            val flowPaths = commands.flatMap { it.sourceDescription }
            YamlCommandsPathValidator.validatePathsExistInWorkspace(input, it, flowPaths)
}
```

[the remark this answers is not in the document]
Leland-Takamine: Sounds good thanks!

Issue #1299, opened 2023-07-31
[1.30.4] `xcuitest.XCTestDriverClient$XCTestDriverUnreachable`


denil-ct:
**Describe the bug**
When running maestro test, I'm getting an exception
```
xcuitest.XCTestDriverClient$XCTestDriverUnreachable: Failed to reach XCUITest Server in restart
	at xcuitest.XCTestDriverClient.restartXCTestRunnerService(XCTestDriverClient.kt:52)
	at ios.xctest.XCTestIOSDevice.open(XCTestIOSDevice.kt:33)
	at ios.LocalIOSDevice.open(LocalIOSDevice.kt:25)
	at maestro.drivers.IOSDriver.open(IOSDriver.kt:84)
	at maestro.Maestro$Companion.ios(Maestro.kt:579)
	at maestro.cli.session.MaestroSessionManager.createIOS(MaestroSessionManager.kt:329)
	at maestro.cli.session.MaestroSessionManager.createMaestro(MaestroSessionManager.kt:157)
	at maestro.cli.session.MaestroSessionManager.access$createMaestro(MaestroSessionManager.kt:50)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:82)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:81)
	at maestro.cli.db.KeyValueStore.withExclusiveLock(KeyValueStore.kt:37)
	at maestro.cli.session.SessionStore.withExclusiveLock(SessionStore.kt:74)
	at maestro.cli.session.MaestroSessionManager.newSession(MaestroSessionManager.kt:81)
	at maestro.cli.session.MaestroSessionManager.newSession$default(MaestroSessionManager.kt:58)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:126)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:44)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
	at picocli.CommandLine.access$1200(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
[...]

Issue #1257, opened 2023-07-17
[v1.30.2] Failed to reach XCUITest Server in restart


tqirazevedo:
**Describe the bug**
When I run the maestro test command, the following error occurs:

maestro test ios-advanced-flow.yaml

Running on iPhone 14 Pro Max - iOS 16.2 - 308A7B68-C45F-4D9B-AE1E-BD952577DC9A

xcuitest.XCTestDriverClient$XCTestDriverUnreachable: Failed to reach XCUITest Server in restart
	at xcuitest.XCTestDriverClient.restartXCTestRunnerService(XCTestDriverClient.kt:47)
	at ios.xctest.XCTestIOSDevice.open(XCTestIOSDevice.kt:31)
	at ios.LocalIOSDevice.open(LocalIOSDevice.kt:27)
	at maestro.drivers.IOSDriver.open(IOSDriver.kt:82)
	at maestro.Maestro$Companion.ios(Maestro.kt:579)
	at maestro.cli.session.MaestroSessionManager.createIOS(MaestroSessionManager.kt:329)
	at maestro.cli.session.MaestroSessionManager.createMaestro(MaestroSessionManager.kt:157)
	at maestro.cli.session.MaestroSessionManager.access$createMaestro(MaestroSessionManager.kt:50)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:82)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:81)
	at maestro.cli.db.KeyValueStore.withExclusiveLock(KeyValueStore.kt:37)
	at maestro.cli.session.SessionStore.withExclusiveLock(SessionStore.kt:74)
	at maestro.cli.session.MaestroSessionManager.newSession(MaestroSessionManager.kt:81)
	at maestro.cli.session.MaestroSessionManager.newSession$default(MaestroSessionManager.kt:58)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:126)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:44)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
	at picocli.CommandLine.access$1200(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
[...]

PR #1439, opened 2023-09-12
[fix] xcuitest.XCTestDriverClient$XCTestDriverUnreachable issues

artem888:
## Proposed Changes

[FlyingFox](https://github.com/swhitty/FlyingFox) (the library that we use for XCTest HTTP server) implements loopback address using Darwin POSIX address:

```
Darwin.sockaddr_in6(
            sin6_len: UInt8(MemoryLayout<sockaddr_in6>.stride),
            sin6_family: sa_family_t(AF_INET6),
            sin6_port: port.bigEndian,
            sin6_flowinfo: 0,
            sin6_addr: in6addr_loopback,
            sin6_scope_id: 0
)
```

Server is listening on IPv6 all the time and is able to resolve IPv4 connections correctly if `::1 localhost` is present inside `etc/hosts`, however fails if it doesn't.

[SwiftNIO](https://github.com/apple/swift-nio) has its own [convenience method](https://swiftinit.org/docs/swift-nio/niocore/socketaddress.makeaddressresolvinghost(_:port:)) for this case. We could consider in the future to migrate to [SwiftNIO](https://github.com/apple/swift-nio) based server for this and many other reasons.

For now the best solution would be to send requests to IPv6 host directly instead of using localhost as a host.

## Testing
- [x] local test
- [x] cloud test 

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/1299
https://github.com/mobile-dev-inc/maestro/issues/1257

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -48,6 +48,7 @@ import kotlin.system.exitProcess
 
 object MaestroSessionManager {
     private const val defaultHost = "localhost"
+    private const val defaultXctestHost = "[::1]"
     private const val defaultIdbPort = 10882
     private const val defaultXcTestPort = 22087
 
@@ -276,14 +277,14 @@ object MaestroSessionManager {
         val xcTestInstaller = LocalXCTestInstaller(
             logger = IOSDriverLogger(LocalXCTestInstaller::class.java),
             deviceId = deviceId,
-            host = defaultHost,
+            host = defaultXctestHost,
             defaultPort = defaultXcTestPort
         )
 
         val xcTestDriverClient = XCTestDriverClient(
             installer = xcTestInstaller,
             logger = IOSDriverLogger(XCTestDriverClient::class.java),
-            client = XCTestClient(defaultHost, defaultXcTestPort)
+            client = XCTestClient(defaultXctestHost, defaultXcTestPort)
         )
 
         val xcTestDevice = XCTestIOSDevice(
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -276,14 +277,14 @@ object MaestroSessionManager {
         val xcTestInstaller = LocalXCTestInstaller(
             logger = IOSDriverLogger(LocalXCTestInstaller::class.java),
             deviceId = deviceId,
-            host = defaultHost,
+            host = defaultXctestHost,
```
amanjeetsingh150, line 280: How are ipv4 and ipv6 mapped to a host in /etc/hosts?

Can a host with ipv4 communicate to ipv6 as well and vice versa?

amanjeetsingh150 (approved):

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
artem888: this is basically the default mapping:

```
127.0.0.1       localhost
::1             localhost
```

XCTest server only listens on IPv6

> Can a host with ipv4 communicate to ipv6 as well and vice versa?

with correct networking setup yes, with the library we're using (FlyingFox) it is only able to listen on ::1 for loopback addresses

abhishekbedi1432: Is it specific to Xcode 16.2/4 or Xocde 26? or mac osx 15.X ?

PR #1440, opened 2023-09-12
Set device locale as a part of `start-device` command

artem888:
## Proposed Changes

This PR:
- implements logic to set a device locale via `start-device` command
- in case of missing `--device-locale` parameter a default `en_US` locale will be used
- on Android the locale is set via a BroadcastReceiver and `updateConfiguration` private API, on iOS it is set via modifying simulator settings with simctl.

Example API usage (device language will be set to Spanish in this case):
```
maestro start-device --platform android --os-version 31 --device-locale es_ES
maestro start-device --platform ios --os-version 15 --device-locale es_ES
```

feature recording iOS:

https://github.com/mobile-dev-inc/maestro/assets/961175/f725e457-9ac2-4ff5-8a3b-58282efc4e07

feature recording Android:

https://github.com/mobile-dev-inc/maestro/assets/961175/a725ce5b-0c07-4630-ae97-293164007c2c

## Testing
- [x] local test
- [x] staging test


maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.util.*
+
+class LocaleSettingHandler(context: Context) : AbstractSettingHandler(context, listOf(CHANGE_CONFIGURATION)) {
+    fun setLocale(locale: Locale) {
+        try {
+            if (hasPermissions()) {
+                setLocaleWith(locale)
+            }
+        } catch (e: Exception) {
+            Log.e(TAG, "Failed to set locale", e)
```
axelniklasson, line 18: Will this be properly propagated in Maestro Cloud?

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+            Locale.Builder().setLocale(locale).setScript(script).build().also { locale = it }
+        }
+
+        if (!LocaleUtils.isAvailableLocale(locale)) {
+            val approximateMatchesLc = matchLocales(language, country)
+
+            if (approximateMatchesLc.isNotEmpty() && script.isNullOrBlank()) {
+                Log.i(
+                    TAG,
+                    "The locale $locale is not known. Selecting the closest known one ${approximateMatchesLc[0]} instead"
```
axelniklasson, line 40: How will we surface this to users so they know that something other than what they specified was used?

maestro-android/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,29 @@
...
                      tools:ignore="QueryAllPackagesPermission"/>
+    <uses-permission android:name="android.permission.WRITE_SETTINGS"
+                     tools:ignore="ProtectedPermissions" />
+    <uses-permission android:name="android.permission.CHANGE_CONFIGURATION"
+                     tools:ignore="ProtectedPermissions" />
 
-    <application android:usesCleartextTraffic="true" />
+    <application
+            android:usesCleartextTraffic="true">
+            <activity
```
axelniklasson, line 17: Now that we're adding a `MainActivity` to the runner app, might this have any other side effects?

axelniklasson (commented):
Nice! Let's verify that this works on staging

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
artem888: should be part of `adb logcat` @amanjeetsingh150 cmiiw

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: looking at Appium they do some soft handling here, by picking up the closest locale possible if Locale() object is failed to be initialized, I think I'm also fine with this approach, in the end we don't have a control of typos / not supported locales for specific Android versions.

maestro-android/src/main/AndroidManifest.xml
artem888: generally every Android app has a main activity, so I don't see any bad side-effects happening by adding it to maestro driver app. cc @amanjeetsingh150

maestro-android/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,29 @@
...
+                android:name=".MainActivity"
```
amanjeetsingh150, line 18: I don't see a need of MainActivity though why its added ? We don't need a UI

maestro-android/src/main/java/dev/mobile/maestro/MainActivity.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.main)
+
+        Log.d(TAG, "Entering the app")
+
+        registerSettingsReceivers(listOf(
+            LocaleSettingReceiver::class.java
+        ))
```
amanjeetsingh150, line 21: We wouldn't need to register receiver here. Mentioning in manifest should suffice you can issue adb broadcast to cross check.

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
```diff
@@ -0,0 +1,42 @@
+package dev.mobile.maestro.handlers
+
+import android.content.Context
+import android.content.pm.PackageManager
+import android.util.Log
+
+abstract class AbstractSettingHandler(private val context: Context, private val permissions: List<String>) {
+    fun enable(): Boolean {
```
amanjeetsingh150, line 8: Lets include this code in LocationSettingsHandler itself nobody else is using it.

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
```diff
@@ -0,0 +1,42 @@
+package dev.mobile.maestro.handlers
+
+import android.content.Context
+import android.content.pm.PackageManager
+import android.util.Log
+
+abstract class AbstractSettingHandler(private val context: Context, private val permissions: List<String>) {
```
amanjeetsingh150, line 7: Let's include this code to LocationSettingHandler nobody else is using it.

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    fun disable(): Boolean {
+        Log.d(TAG, "Disabling $settingDescription")
+        return if (!hasPermissions()) {
+            false
+        } else setState(false)
+    }
+
+    protected fun hasPermissions(): Boolean {
```
amanjeetsingh150, line 22: Is this really important ? I think permissions required would be auto granted while installation because we include -g. You can check though

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
amanjeetsingh150: Yes should be in logcat

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.content.Context
+import android.content.Intent
+import android.util.Log
+import dev.mobile.maestro.handlers.LocaleSettingHandler
+import org.apache.commons.lang3.LocaleUtils
+import java.util.*
+
+class LocaleSettingReceiver : BroadcastReceiver(), HasAction {
+    override fun onReceive(context: Context?, intent: Intent?) {
+        var language = intent?.getStringExtra(LANG)
```
amanjeetsingh150, line 14: This can be val

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+                                "The following locales are available altogether: ${LocaleUtils.availableLocaleList()}"
+                    )
+                }
+                resultCode = Activity.RESULT_CANCELED
+                resultData = "The locale $locale is not known"
+                return
+            }
+        }
+
+        context?.let {
```
amanjeetsingh150, line 64: Why context is nullable here?

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.util.Log
+import dev.mobile.maestro.handlers.LocaleSettingHandler
+import org.apache.commons.lang3.LocaleUtils
+import java.util.*
+
+class LocaleSettingReceiver : BroadcastReceiver(), HasAction {
+    override fun onReceive(context: Context?, intent: Intent?) {
```
amanjeetsingh150, line 13: The nullable `?` can be removed

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+                matches.add(locale)
+            }
+        }
+        return matches
+    }
+
+    private fun matchLocales(language: String, country: String): List<Locale> {
+        val matches = ArrayList<Locale>()
+        for (locale in LocaleUtils.availableLocaleList()) {
+            if (locale.language.equals(language, ignoreCase = true) &&
```
amanjeetsingh150, line 85: This can be locale.language == language and similarly for country

maestro-android/src/main/res/layout/main.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
amanjeetsingh150, line 2: This should not be needed

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.app.Activity
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.util.Log
+import dev.mobile.maestro.handlers.LocaleSettingHandler
+import org.apache.commons.lang3.LocaleUtils
+import java.util.*
+
+class LocaleSettingReceiver : BroadcastReceiver(), HasAction {
```
amanjeetsingh150, line 12: I am a little confused about how are we planning to ship this:

* This PR doesn't invoke this broadcast receiver
* Where are we planning to send the broadcast from
* How is this going to wire with Maestro Client?

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,105 @@
...
+            LocaleSettingHandler(it).setLocale(locale)
+            Log.i(TAG, "Set locale: $locale")
+            resultCode = Activity.RESULT_OK
+            resultData = locale.toString()
+        } ?: Log.e(TAG, "Failed to set device locale setting (context is null)")
+    }
+
+    private fun matchLocales(language: String): List<Locale> {
+        val matches = ArrayList<Locale>()
+        for (locale in LocaleUtils.availableLocaleList()) {
```
amanjeetsingh150, line 74: What happens when locale is not available?

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
axelniklasson: Sorry if this is a stupid question, but does it being in logcat mean that we can correctly surface client errors in the event of misconfigured locale and infra errors in the event of problems on our side?

maestro-android/src/main/AndroidManifest.xml
axelniklasson: Yep unsure here as well, but given that we're changing internals of how maestro works I want us to be extra careful with testing so we don't break anything. Could you share more on what the plan is to verify that this is a safe change?

artem888: checked, it works without an Activity, thanks for finding 👁️

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
artem888: adb logs are served on MaestroCloud, so later clients can use them to debug the issue

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
artem888: I thought it makes sense to keep it in the separate file, if we'll have more setting handlers in the future, it'd be easy to reuse this logic

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
artem888: in our case yes, from the appium code I see that they added it specifically for location and animation settings.

do you think it is an excessive check?

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: removed nullability

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: I'm resetting it later in l21

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: it sets `resultCode` to `Activity.RESULT_CANCELED` so we can also handle it on maestro side. there is also some logic to find the closest possible locale if one is not available.

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: > Where are we planning to send the broadcast from
coming in a separate PR , this is independent logic, so can be merged separately. small PRs are easier to review imo

> How is this going to wire with Maestro Client?
adb call will be part of DeviceService, integration PR is coming soon

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
amanjeetsingh150: I think this could be implicitly granted. I would suggest experimenting it and then decide to include/not include.

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
```diff
@@ -0,0 +1,68 @@
...
+    @SuppressLint("PrivateApi")
+    @Throws(
+        ClassNotFoundException::class,
+        NoSuchMethodException::class,
+        InvocationTargetException::class,
+        IllegalAccessException::class,
+        NoSuchFieldException::class
+    )
+    private fun setLocaleWith(locale: Locale) {
+        var activityManagerNativeClass = Class.forName("android.app.ActivityManagerNative")
```
amanjeetsingh150, line 31: What happens if something here errors out, maybe reflection failed. How would that behaviour look like in terms of error propagation?

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
amanjeetsingh150: I'm not sure if I follow this why do we have Activity result codes in headless environment. I think we need to return some other result codes

maestro-client/src/main/resources/maestro-app.apk
amanjeetsingh150, line 1: We have not built the server APK which would also be needed. I suggest it would be good idea to include the code there as well to see how are we planning the integration.

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
igorsmotto: >coming in a separate PR , this is independent logic, so can be merged separately. small PRs are easier to review imo

I agree. But how is this going to work end to end? Is there a doc somewhere?

artem888: we decided not to write an RFC for this feature, since it looked rather small in the beginning, but looking retrospectively now it made sense to write an RFC

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
artem888: I'd keep it still, since it is a minor thing and this implementation was inspired by appium, pretty sure they had it there for a reason

maestro-android/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,16 @@
...
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"
                      tools:ignore="QueryAllPackagesPermission"/>
+    <uses-permission android:name="android.permission.WRITE_SETTINGS"
+                     tools:ignore="ProtectedPermissions" />
+    <uses-permission android:name="android.permission.CHANGE_CONFIGURATION"
+                     tools:ignore="ProtectedPermissions" />
 
-    <application android:usesCleartextTraffic="true" />
+    <application
+            android:usesCleartextTraffic="true">
```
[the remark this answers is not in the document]
artem888: no idea, it was there before

maestro-client/src/main/resources/maestro-app.apk
artem888: done

maestro-android/build.gradle
```diff
@@ -144,6 +144,9 @@ dependencies {
     implementation(libs.ktor.server.content.negotiation)
     implementation(libs.ktor.serial.gson)
 
+    implementation('org.apache.commons:commons-lang3:3.13.0')
+    implementation('org.lsposed.hiddenapibypass:hiddenapibypass:4.3')
```
amanjeetsingh150, line 148: What does this library do?

maestro-android/src/main/java/dev/mobile/maestro/handlers/AbstractSettingHandler.kt
amanjeetsingh150, line 8: This seems unused. disable as well

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+import android.content.Context
+import android.content.res.Configuration
+import android.os.Build
+import org.lsposed.hiddenapibypass.HiddenApiBypass
+import java.lang.reflect.InvocationTargetException
+import java.util.*
+
+class LocaleSettingHandler(context: Context) : AbstractSettingHandler(context, listOf(CHANGE_CONFIGURATION)) {
+    fun setLocale(locale: Locale) {
+        if (hasPermissions()) {
```
amanjeetsingh150, line 13: Are we guaranteed permission always?

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+        methodGetConfiguration.isAccessible = true
+        val config = methodGetConfiguration.invoke(amn) as Configuration
+        val configClass: Class<*> = config.javaClass
+        val f = configClass.getField("userSetLocale")
+        f.setBoolean(config, true)
+        config.locale = locale
+        config.setLayoutDirection(locale)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            HiddenApiBypass.invoke(
```
amanjeetsingh150, line 47: How exactly this works?

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    override fun setState(state: Boolean): Boolean {
+        return false
+    }
+
+    override val settingDescription: String = "locale"
+
+    companion object {
+        private const val TAG = "Maestro"
```
amanjeetsingh150, line 70: This seems unused

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,113 @@
...
+import android.content.Context
+import android.content.Intent
+import android.util.Log
+import dev.mobile.maestro.handlers.LocaleSettingHandler
+import org.apache.commons.lang3.LocaleUtils
+import java.util.*
+
+class LocaleSettingReceiver : BroadcastReceiver(), HasAction {
+    override fun onReceive(context: Context, intent: Intent) {
+        var language = intent.getStringExtra(LANG)
```
amanjeetsingh150, line 14: nit: can be val

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,113 @@
+package dev.mobile.maestro.receivers
+
+import android.app.Activity
```
amanjeetsingh150, line 3: nit: Unused import

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        try {
+            LocaleSettingHandler(context).setLocale(locale)
+            Log.i(TAG, "Set locale: $locale")
+            resultCode = RESULT_SUCCESS
+            resultData = locale.toString()
+        } catch (e: Exception) {
+            Log.e(TAG, "Failed to set locale", e)
```
amanjeetsingh150, line 70: Curious, Do we know of any known/unknown exceptions that can happen here?

amanjeetsingh150 (approved):

maestro-android/build.gradle
artem888: since API level 30 `updateConfiguration` API was hidden and is not accessible anymore via `Class.getMethod` API. This library allows to access hidden APIs. Under the hood the library uses Java `Unsafe` API, which can read and write memory in pure Java.

more info [here](https://lovesykun.cn/archives/android-hidden-api-bypass.html)

maestro-android/src/main/java/dev/mobile/maestro/handlers/LocaleSettingHandler.kt
artem888: it should be there always from what I've tested, for now it just checks that `CHANGE_CONFIGURATION` is there

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: it is reassigned later in the code

maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
artem888: since we're using a fixed list of supported locales , the probability that it gets here is close to 0. but it is still worth to propagate this error, since it is hard to test that every language + region combination would work in all the API levels we support.

if exception is related to reflection APIs those might be thrown:

ClassNotFoundException,
NoSuchMethodException,
InvocationTargetException,
IllegalAccessException,
NoSuchFieldException

PR #1447, opened 2023-09-14
Fixes error handling for XCUITestServer communication

amanjeetsingh150:
## Proposed Changes

* Missing try/catch in request handlers
* Handle errors during execution of API call to XCUITestDriver and mapping to XCUITestServerError
* Mapping XCUITestServerError to IOSDeviceErrors
* Properly propagating app crash errors in case of the following messages: `Application com.app.id is not running` and `Lost connection to the application (pid 19985).`
* Adds execute method and runDeviceCall in XCTestDriverClient and IOSDriver

## Testing
- [ ] Local
- [ ] Staging


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -106,45 +109,42 @@ class IOSDriver(
     }
 
     override fun tap(point: Point) {
-        iosDevice.tap(point.x, point.y).expect {}
+        runDeviceCall { iosDevice.tap(point.x, point.y) }
     }
 
     override fun longPress(point: Point) {
-        iosDevice.longPress(point.x, point.y, 3000).expect {}
+        runDeviceCall { iosDevice.longPress(point.x, point.y, 3000) }
```
amanjeetsingh150, line 116: runDeviceCall is method which is going to be responsible to correctly handle exception which can propagate from `SimctlIOSDevice` and `XCTestIOSDevice`.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -481,6 +483,14 @@ class IOSDriver(
         throw TimeoutException("Maestro iOS driver did not start up in time")
     }
 
+    private fun <T> runDeviceCall(call: () -> T): T {
+        return try {
+            call()
+        } catch (appCrashException: IOSDeviceErrors.AppCrash) {
+            throw MaestroException.AppCrash(appCrashException.errorMessage)
+        }
```
amanjeetsingh150, line 491: Right now for ease I have added one of the errors which can propagate from xctest layer but this should also be extended for the errors that can be thrown from SimctlIOSDevice.

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
```diff
@@ -218,95 +182,57 @@ class XCTestIOSDevice(
...
-    class IllegalArgumentSnapshotFailure : Throwable("Failed to capture view hierarchy due to kAXErrorIllegalArgument")
-    class InputFieldNotFound : Throwable("Unable to find focused input field")
-    class UnknownFailure(errorResponse: String) : Throwable(errorResponse)
+    private fun <T> execute(call: () -> T): T {
+        return try {
+            call()
+        } catch (appCrashException: XCUITestServerError.AppCrash) {
+            throw IOSDeviceErrors.AppCrash(
+                "App crashed, please check diagnostic logs: " +
+                        "/Library/Logs/DiagnosticReports directory"
```
amanjeetsingh150, line 230: @artem888 this is a message which users will see on CLI in case their app crash in between flow. Any suggestions what we can show here or this looks good?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
```diff
@@ -1,65 +0,0 @@
-import FlyingFox
-import XCTest
-import os
-
-@MainActor
```
amanjeetsingh150, line 5: @artem888 I hope this can be deleted safely right?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
axelniklasson: What happens in the event of another exception being thrown?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -68,40 +69,32 @@ class XCTestDriverClient(
...
 
     fun viewHierarchy(installedApps: Set<String>): ViewHierarchy {
-        return executeJsonRequest(
+        val responseString = executeJsonRequest(
             "viewHierarchy",
-            ViewHierarchyRequest(installedApps),
-            ViewHierarchy::class
+            ViewHierarchyRequest(installedApps)
         )
+        return mapper.readValue(responseString, ViewHierarchy::class.java)
```
axelniklasson, line 76: Should this be wrapped in a try/catch in the event of the parsing failing?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -185,62 +181,119 @@ class XCTestDriverClient(
         installer.close()
     }
 
-    fun setPermissions(permissions: Map<String, String>): Response {
-        val response = executeJsonRequestUNCHECKED("setPermissions", SetPermissionsRequest(permissions))
-        return response.use { it }
+    fun setPermissions(permissions: Map<String, String>) {
```
axelniklasson, line 184: No return type here?

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -185,62 +181,119 @@ class XCTestDriverClient(
...
+            .url(httpUrl)
             .post(bodyData)
 
         return okHttpClient
             .newCall(requestBuilder.build())
-            .execute()
+            .execute().use { processResponse(it, httpUrl.toString()) }
     }
 
-    private fun executeJsonRequestUNCHECKED(httpUrl: HttpUrl, body: Any): Response {
```
axelniklasson, line 207: Nice to see this being removed 🎉

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -185,62 +181,119 @@ class XCTestDriverClient(
...
+        val error = mapper.readValue(responseBodyAsString, Error::class.java)
+        when {
+            code in 400..499 -> {
+                logger.error("Request for $pathString failed with bad request ${code}, body: $responseBodyAsString")
+                throw XCUITestServerError.BadRequest(
+                    "Request for $pathString failed with bad request ${code}, body: $responseBodyAsString",
+                    responseBodyAsString
+                )
+            }
+            code == 502 -> {
```
axelniklasson, line 270: Do we want to treat 500 -> 599 (except for 502) as network errors too?

artem888 (approved):
Great work, LGTM ! 👍

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: If its not converted to a known exception then it will lead to cancellation.

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: Yes its just a unit operation which just set permissions

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -185,62 +181,119 @@ class XCTestDriverClient(
...
+                logger.error("Request for $pathString failed, because of XCUITest server got crashed/exit, body: $responseBodyAsString")
```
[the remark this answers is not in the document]
amanjeetsingh150: Yes this is because on network layer, 502 is surfaced when we try to reestablish connection. So, its a special case. For that context I'll change 502 to use same constant variable.

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: Apart from 502 (case of fake response code), everything is an unknown exception. As an example 500 is the response code we return from the server in case any handler fails.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/DeviceInfoHandler.swift
```diff
@@ -12,19 +12,23 @@ struct DeviceInfoHandler: HTTPHandler {
...
+                widthPixels: Int(screenSize.width * UIScreen.main.scale),
+                heightPixels: Int(screenSize.height * UIScreen.main.scale)
+            )
 
-        let responseBody = try! JSONEncoder().encode(deviceInfo)
-        return HTTPResponse(statusCode: .ok, body: responseBody)
+            let responseBody = try JSONEncoder().encode(deviceInfo)
+            return HTTPResponse(statusCode: .ok, body: responseBody)
+        } catch let error {
+            return AppError(message: "Getting device info call failed. Error \(error.localizedDescription)").httpResponse
```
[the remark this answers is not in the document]
amanjeetsingh150: AppError is unified object to catch any unknown errors which returns 500 code and a failure reason in message.

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: If this fails we would know it because of the type of exception thrown and would lead to cancellation. Are you recommending to map error to something else

maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt
amanjeetsingh150: Fixed

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SubTreeRouteHandler.swift
amanjeetsingh150: Sounds good!

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
amanjeetsingh150: 🙌

axelniklasson (approved):
🚢

PR #1462, opened 2023-09-19
Revert label addition addition on maestro commands

amanjeetsingh150:


axelniklasson (approved):
Let's also make sure to surface this to the user so they know why this won't make it into this release -- thanks Amanjeet!

szobov:
Hey folks,
Could you please elaborate on why it is reverted and what can we do to get it back?
Thanks a lot!

PR #1471, opened 2023-09-21
initial paste with enter

proksh:
## Changed the input to textarea so that Initial paste of command will now include enter keys.

Screenshot 2023-09-21 at 10 38 30 PM


axelniklasson (approved):
🔝

PR #1474, opened 2023-09-21
Add support for multipart form data file upload

maciejkrolik:
## Proposed Changes

This PR adds support for multipart form data file upload.

The feature can be used to prepare data on the backend before running e2e tests on mobile.

Usage is simple. Instead of a `body` object, we can define `multipartForm` with as many fields as we want. We can also upload multiple files in one request by using many objects with `filePath` properties. `filePath` is required for the files. `mediaType` is optional. If `multipartForm` is available in a request then `body` will be ignored.

PR with the updated docs can be found here: https://github.com/mobile-dev-inc/maestro-docs/pull/28

### Example script
``` js
function uploadFile() {
  const filePath = ``;
  const url = ``;

  const response = http.post(url, {
    multipartForm: {
      "uploadType": "test",
      "data": {
        "filePath": filePath,
        "mediaType": "text/csv"
      }
    },
    headers: {
      'Authorization': `Bearer ${token}`
    },
  });
}
```

## Testing
 - Integration + unit tests
 - Manual testing

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsHttp.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsHttp.kt
@@ -1,11 +1,11 @@
 package maestro.js
 
+import maestro.utils.HttpUtils.toMultipartBody
 import okhttp3.Headers
 import okhttp3.OkHttpClient
 import okhttp3.Request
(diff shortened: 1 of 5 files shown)


axelniklasson (approved):
Nice, thanks for the contribution @maciejkrolik 🙌

Issue #1463, opened 2023-09-19
[v1.32.0] iOS: Location permissions don't work on iOS 17 simulators


danielbyondd:
**Describe the bug**
When specifying the `location: always` permission, the system location permissions prompt still appears on iOS 17 simulators (but not on 16.4)

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Write a flow file that uses the `location: always` permission for the `launchApp` command
2. Run the flow

**Expected behavior**
Location prompt should not appear when the app requests user location

**Screenshots**
Please DM for screenshot

**Environment information (please complete the following information):**

- Maestro version [e.g. v1.2.3]: 1.32.0
- Platform: [e.g. iOS 16.4.1, Android 11.0, Chrome]: iOS 17.0
- Framework: [Native iOS/Android, Flutter, React Native, etc]: Native iOS
- Device model and OS version: [e.g. iPhone6]: iPhone 15 Pro Max
- Simulator, emulator or physical device: Simulator
- Host [e.g. Mac M1, Windows, Linux, etc]: Mac M1

**Workspace (if applicable)**
Please share your workspace with Flow files and application binary if you are able to do so as that greatly helps in reproduction efforts.

**Bugreport output (if applicable)**
The zip files created when `maestro bugreport` can be uploaded here as well if applicable.

**Additional context**

bartekpacia:
cross-referencing probably related issue #1782

elizabethfalkner-jpg:
Seeing the same in iOS 18 and 26; for other permissions as well.

Fishbowler:
@elizabethfalkner-jpg Your issue sounds unrelated. Maybe you should come chat about it on Slack - https://slack.maestro.dev

elizabethfalkner-jpg:
Working with my org do get that set up / enabled. In the mean time, I can create a separate issue if that's more helpful :)

Fishbowler:
Sure! A new issue would be fine!

Fishbowler:
I can't reproduce this. Looks like it was fixed in #1476!

Fishbowler:
I can't reproduce this. Looks like it was fixed in #1476!

PR #1476, opened 2023-09-21
[fix] set location permission is broken on iOS

artem888:
## Proposed Changes

[AppleSimulatorUtils](https://github.com/wix/AppleSimulatorUtils) has some issues with setting permissions on iOS17 and newest Xcode 15 SDK. This PR migrates setting location permission logic (which was the main source of `Unable to clear state for app xxx.xxx` errors) to simctl.

## Testing
- [x] local test
- [x] staging test

## Issues Fixed
https://github.com/mobile-dev-inc/maestro/issues/1463

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
@@ -460,6 +460,20 @@ object LocalSimulatorUtils {
                 )
             }
 
+            "unset" -> {
+                runCommand(
+                    listOf(
+                        "xcrun",
+                        "simctl",
+                        "privacy",
+                        deviceId,
+                        "reset",
+                        "location-always",
+                        bundleId
+                    )
+                )
+            }
+
             else -> throw IllegalArgumentException("wrong argument value '$value' was provided for 'location' permission")
         }
     }
--- a/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
@@ -22,6 +22,26 @@ object LocalSimulatorUtils {
 
     private val homedir = System.getProperty("user.home")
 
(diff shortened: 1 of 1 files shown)


maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -334,49 +354,111 @@ object LocalSimulatorUtils {
...
-            runCommand(
-                listOf(
-                    "$homedir/.maestro/deps/applesimutils",
-                    "--byId",
-                    deviceId,
-                    "--bundle",
-                    bundleId,
-                    "--setPermissions",
-                    argument
+        if (argument.isNotEmpty()) {
```
amanjeetsingh150, line 357: nit: Can you rename the variable argument here? Its difficult to understand what's happening here.

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -334,49 +354,111 @@ object LocalSimulatorUtils {
...
-        "microphone",
-        "motion",
-        "photos",
-        "reminders",
-        "siri",
-        "speech",
-        "userTracking",
-    )
+    private fun setSimctlPermissions(deviceId: String, bundleId: String, permissions: Map<String, String>) {
+        val mutable = permissions.toMutableMap()
```
amanjeetsingh150, line 389: nit: let's rename this `mutable` to something which explains structure of map.

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -334,49 +354,111 @@ object LocalSimulatorUtils {
...
+        if (mutable.containsKey("all")) {
```
amanjeetsingh150, line 390: Can we instead use filter or map operation instead of this if condition?

amanjeetsingh150 (approved):

PR #1498, opened 2023-10-02
fix: the timeout should be equal to readTimeout

amanjeetsingh150:
Give enough time to complete the status call


maestro-ios-driver/src/main/kotlin/xcuitest/installer/LocalXCTestInstaller.kt
```diff
@@ -98,7 +98,7 @@ class LocalXCTestInstaller(
 
     private fun ensureOpen(): Boolean {
         XCRunnerCLIUtils.ensureAppAlive(UI_TEST_RUNNER_APP_BUNDLE_ID, deviceId)
-        return MaestroTimer.retryUntilTrue(10_000, 100) {
+        return MaestroTimer.retryUntilTrue(100_000, 100) {
```
axelniklasson, line 101: Is `readTimeout` defined somewhere? If yes, can we use that value instead of using a seemingly magic number here?

ArthurSav (approved):

amanjeetsingh150:
Closing this since was done in another PR.

PR #1516, opened 2023-10-09
Improve scrollUntilVisible to center element

ArthurSav:
## Proposed Changes
This change will attempt to center (on screen) the element we scroll to.
- If the element is the last element on the list, we will attempt to center a few times and then exit with success (if the element is indeed visible)
- This is an opt-in feature

### <samp>🤖 Generated by Copilot at ef41ce2</samp>

This pull request adds a new feature to the `maestro` framework that allows scrolling until an element is centered on the screen. It modifies the `ScrollUntilVisibleCommand` class, the `Orchestra` class, the `UiElement` class, and the YAML parsing and command execution logic to support the new feature. It also adds a new test case and a YAML file to verify the functionality.

## Testing
- Locally
- Test case

**PS: This is a non breaking change. It's opt-in**

## Issues Fixed
- Sometimes elements would not fully show but it would be considered as visible
- Different UI Frameworks return inconsistent element bounds. Centering on screen should make this command more reliable.

Usage:

```yaml
- scrollUntilVisible:
    centerElement: true
    element:
      text: "Item 6"
```

**Before**

https://github.com/mobile-dev-inc/maestro/assets/2115575/f9d65a3b-dd53-478b-b4cb-4a769b182b67

**After**

https://github.com/mobile-dev-inc/maestro/assets/2115575/593c27ef-ad85-4f2b-84cc-a82b5ddc2eea

### DIFF
--- a/maestro-client/src/main/java/maestro/UiElement.kt
+++ b/maestro-client/src/main/java/maestro/UiElement.kt
@@ -58,23 +58,12 @@ data class UiElement(
             SwipeDirection.LEFT, SwipeDirection.RIGHT -> screenWidth / 5
         }
 
+        // return true when the element center is within the <direction> half of the screen bounds plus margin
         return when(direction) {
-            SwipeDirection.RIGHT -> {
-                // return true when the element center is within the right half of the screen bounds plus margin
-                elementCenterX > centerX - margin
-            }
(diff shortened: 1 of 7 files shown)


artem888 (approved):
nice one 👏

PR #1520, opened 2023-10-10
Feature - Add Status and Time to report.xml

xo0ps:
## Proposed Changes

Currently, `report.xml` file contains the duration of each flow only when it is run locally. But it is also super beneficial to have this data when running the flows on the cloud, enabling us to visualise them on dashboards and create alerts based on them.
The exact duration should be provided by Maestro Cloud's api, but as we don't have it, this PR calculates it using time difference between when the flow's status changes from RUNNING to one of the completed ones.

This PR adds the `status` field for each test flow when running locally.
It also adds both `status` and `time` fields for each test flow when running on the cloud.

The screenshot below demonstrates the fields when using cloud command.

Here's the link to [Slack message](https://mobile-dev-inc.slack.com/archives/C05EV48H5FB/p1689762819324689) asking for this feature.

## Testing
1. Using `maestro test`, the report file contains the status of each test flow (time is already there).
2. Using `maestro cloud`, the report file contains both status and time fields for each test flow.


xo0ps:
@felipevolpone @axelniklasson would it be possible to review this PR as early as possible? We have pending tasks based on this feature to execute. Thank you

felipevolpone:
@xo0ps Hey, sorry for the delay. I'm heading to PTO for the end of the year and early next year I'll get back to this and review it.

Again, thank you for your contribution.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -328,42 +348,34 @@ class CloudInteractor(
...
     }
 
-    companion object {
-
-        private val COMPLETED_STATUSES = setOf(
-            UploadStatus.Status.CANCELED,
-            UploadStatus.Status.WARNING,
-            UploadStatus.Status.SUCCESS,
-            UploadStatus.Status.ERROR,
+    private fun convert(passed: Boolean, upload: UploadStatus, runningFlows: RunningFlows): TestExecutionSummary.SuiteResult {
```
felipevolpone, line 365: let's be more specific about what this method converts by giving it a better name

maestro-cli/src/main/java/maestro/cli/util/TimeUtils.kt
```diff
@@ -0,0 +1,17 @@
+package maestro.cli.util
+
+import kotlin.math.roundToLong
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+object TimeUtils {
+
+    fun durationInSeconds(startTime: Long?, endTime: Long?): Duration {
```
felipevolpone, line 9: let's also be specific that the start and end time should be in milliseconds.
```
durationInSeconds(startTimeInMillis: Long?, endTimeInMillis: Long?)
```

felipevolpone:
Overall the PR looks good. Left some comments to make the code more obvious

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
xo0ps: applied

maestro-cli/src/main/java/maestro/cli/util/TimeUtils.kt
xo0ps: applied

felipevolpone (approved):

felipevolpone:
@xo0ps thank your for your contribution and let's merge this 🙌

PR #1521, opened 2023-10-10
Exception handling for set device locale feature

artem888:
## Proposed Changes

This PR adds better exception handling for setting a device locale feature:
- improves cli error messages and specifications about `--device-locale` in case of exceptions
- adds a list of supported languages and countries by Maestro
- in case of Android we can catch some exceptions happened in the driver via `resultCode` and `resultData` received from BroadcastReceiver

## Testing
- [x] local test
- [ ] staging test


maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -87,13 +87,99 @@ class StartDeviceCommand : Callable<Int> {
...
+
+    private fun validateLocale(language: String, country: String) {
+        if (!SUPPORTED_LANGUAGES.map { it.first }.contains(language)) {
+            val languages = SUPPORTED_LANGUAGES.joinToString("\n")
+            throw CliError("$language language is currently not supported by Maestro, please check that it is a valid ISO-639-1 code. Here is a full list of supported languages:\n" +
+                    "\n" +
+                    languages
+            )
+        }
+        if (!SUPPORTED_COUNTRIES.map { it.first }.contains(country)) {
```
amanjeetsingh150, line 111: How are we getting supported countries and languages? I see this hardcoded.

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -76,10 +76,13 @@ object DeviceService {
                     PrintUtils.message("Setting the device locale to ${device.language}_${device.country}...")
                     val driver = AndroidDriver(dadb)
                     driver.open()
-                    driver.setDeviceLocale(
+                    val result = driver.setDeviceLocale(
                         country = device.country,
                         language = device.language
                     )
+                    if (result != -1) {
+                        throw IllegalStateException("Android failed to update a device locale")
```
amanjeetsingh150, line 84: This is not debuggable really I think we should have failure reasons here and throw proper errors with message

amanjeetsingh150: Also how the failure in setting locale works on iOS?

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -87,13 +87,99 @@ class StartDeviceCommand : Callable<Int> {
...
+            throw CliError("$country country is currently not supported by Maestro, please check that it is a valid ISO-3166-1 code. Here is a full list of supported countries:\n" +
+                    "\n" +
+                    countries
+            )
+        }
+    }
+
+    companion object {
+        // ISO-639-1
+        private val SUPPORTED_LANGUAGES = listOf(
```
axelniklasson, line 122: Did we verify that this list is the same for both Android and iOS?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -568,9 +568,16 @@ class AndroidDriver(
         LOGGER.info("[Done] Adding media files")
     }
 
-    fun setDeviceLocale(country: String, language: String) {
+    fun setDeviceLocale(country: String, language: String): Int {
         dadb.shell("pm grant dev.mobile.maestro android.permission.CHANGE_CONFIGURATION")
-        dadb.shell("am broadcast -a dev.mobile.maestro.locale -n dev.mobile.maestro/.receivers.LocaleSettingReceiver --es lang $language --es country $country")
+        val response = dadb.shell("am broadcast -a dev.mobile.maestro.locale -n dev.mobile.maestro/.receivers.LocaleSettingReceiver --es lang $language --es country $country")
```
axelniklasson, line 573: How will we surface any error thrown by this call?

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
axelniklasson: +1 on what Amanjeet said, we should have as descriptive error messages as possible and avoid these catch all messages if we can. Do we know what can go wrong here? If yes, can we categorize it? If no, let's maybe display it to the user?

artem888: added more helpful failure reasons for Android and made exception message more descriptive

artem888: > Also how the failure in setting locale works on iOS?

There is sadly no way to verify that `xcrun simctl spawn defaults write` will always be successful. I think good documentation and exception messages are enough here for the users to be able to debug the reason

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
artem888: decided to remove hardcoded values here, explained the reasoning offline

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
artem888: answered offline

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
artem888: this call is always successful , exceptions happen on android app level, handled them by propagating certain error codes and data messages for now

amanjeetsingh150 (approved):

PR #1523, opened 2023-10-11
feat: non-generic error messages for the Android Driver

igorsmotto:
## Proposed Changes

Maestro Driver Service now try/catches all exceptions from all operations and transforms them to an INTERNAL Status, this will bring the description of the exception to the GRPC caller.
AndroidDriver now uses a blockingStub with a Deadline of 500ms, if a call takes more than that it will throw a DEADLINE_EXCEEDED - Consequently, I've removed the manual timeout procedure of the contentDescriptor function

Finally, I've created a new wrapper for all calls that make sure that we throw a custom exception for this deadline exceeded.  

This pull request also fixes spelling errors in the `CONTRIBUTING.md` file, and corrects the spelling of the `recipes` folder and its subfolders.


ArthurSav (approved):
Looks good. I would try to test this on Cloud with apps known to crash emulators.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -55,6 +57,7 @@ class AndroidDriver(
         .usePlaintext()
         .build()
     private val blockingStub = MaestroDriverGrpc.newBlockingStub(channel)
+    private val blockingStubWithTimeout get() = blockingStub.withDeadlineAfter(40, TimeUnit.SECONDS)
```
amanjeetsingh150, line 60: Why 40 seconds? 

Do we want to use blockingStubWithTimeout for every call now?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -99,86 +99,98 @@ class Service(
...
+            val intent = context.packageManager.getLaunchIntentForPackage(request.packageName)
 
-        if (intent == null) {
-            Log.e("Maestro", "No launcher intent found for package ${request.packageName}")
-            responseObserver.onError(RuntimeException("No launcher intent found for package ${request.packageName}"))
-            return
-        }
+            if (intent == null) {
+                Log.e("Maestro", "No launcher intent found for package ${request.packageName}")
+                responseObserver.onError(RuntimeException("No launcher intent found for package ${request.packageName}"))
```
amanjeetsingh150, line 109: You need to build the APKs I see the APKs are not changed

igorsmotto: How do I do this? 🤔 Can you show me?

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
igorsmotto: > Why 40 seconds?

I'll make more tests on this number, but don't expect a great ordeal out if this. I think we need just a high number to avoid the client waiting indefinitely.

> Do we want to use blockingStubWithTimeout for every call now?

Either this or use the blockingStub with a specific Deadline. The important here is that using a blockingStub without a Deadline means that the call will wait forever for a response - which makes no sense.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
amanjeetsingh150: Sure, ensure that you are using java 8 in your cli from root of maestro. Then run following commands:

```
./gradlew maestro-android:assembleAndroidTest
./gradlew maestro-android:assemble
```

This should be generating 2 APKs

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
amanjeetsingh150: > Either this or use the blockingStub with a specific Deadline. The important here is that using a blockingStub without a Deadline means that the call will wait forever for a response - which makes no sense.

Yup sounds good

> I'll make more tests on this number, but don't expect a great ordeal out if this. I think we need just a high number to avoid the client waiting indefinitely.

No issues, aligned on this now! nit: if its possible to log if when time is expired on the maestro side that will be useful.

igorsmotto: It will throw a very specific exception, so this will lead to a problem if any

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -254,15 +254,25 @@ class AndroidDriver(
...
-            future.get(5, TimeUnit.SECONDS)
-        } catch (e: TimeoutException) {
-            LOGGER.error("Timeout while fetching view hierarchy", e)
-            null
+            blockingStubWithTimeout.viewHierarchy(viewHierarchyRequest {})
+        } catch (throwable: StatusRuntimeException) {
+            val status = Status.fromThrowable(throwable)
+            if (status.code == Status.Code.DEADLINE_EXCEEDED) {
+                LOGGER.error("Timeout while fetching view hierarchy")
+                throw MaestroException.DriverTimeout("Emulator's driver unreachable")
```
ArthurSav, line 264: Nit: Let's omit the word "Emulator" since it can run on physical devices

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -880,6 +922,20 @@ class AndroidDriver(
...
     }
 
+    private fun <T> runDeviceCall(call: () -> T): T {
+        return try {
+            call()
+        } catch (throwable: StatusRuntimeException) {
+            val status = Status.fromThrowable(throwable)
+            if (status.code == Status.Code.DEADLINE_EXCEEDED) {
+                closed = true
+                throw MaestroException.DriverTimeout("Emulator's driver unreachable")
```
ArthurSav, line 932: Nit: Let's omit the word "Emulator" since it can run on physical devices

ArthurSav (approved):

maestro-client/src/main/java/maestro/Errors.kt
```diff
@@ -31,6 +31,8 @@ sealed class MaestroException(override val message: String) : RuntimeException(m
 
     class AppCrash(message: String): MaestroException(message)
 
+    class DriverTimeout(message: String): MaestroException(message)
```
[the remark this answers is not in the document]
igorsmotto: I see your point, however, I don't see how these relate. We can map a specific Maestro Exception to not be client-error.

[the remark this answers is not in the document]
amanjeetsingh150: There is a risk then right to confidently say it's a client exception. If we really see this as client exception what are those cases and what is client expected to do in that case to fix it?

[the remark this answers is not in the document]
igorsmotto: Sorry if I was not clear.
- this will be an infra error and also a MestroException. 
- These 2 things are not tied - They can be today, but I can untie that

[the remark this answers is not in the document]
amanjeetsingh150: Ohh now I understood! But this will always be mapped as infra error on cloud right?

PR #1531, opened 2023-10-17
Improve command speed for animation heavy screens

ArthurSav:
## Proposed Changes
This PR introduces a `waitForAppToSettle` timeout for tapOn commands that will limit the time we wait for app to settle before proceeding to the next command.

## Issues Fixed
Flows run on animation heavy apps or apps that have moving elements as part of their UI took longer than expected to complete. This addition will allow such flows to limit the waiting time.

Notes:
- This is a best effort timeout. There are future plans to refactor and improve Driver speed / logic.
- This is an opt-in feature and will not break previous behaviour

Usage
```yaml
appId: com.example.app
---
- launchApp
- tapOn:
    text: "Button"
    waitToSettleTimeoutMs: 1500
```

### <samp>🤖 Generated by Copilot at 1f0f32e</samp>

This pull request adds a new feature to the maestro framework that allows customizing the timeout for waiting for the app to settle after tapping an element or a point on the screen. This feature can improve the reliability and performance of the maestro orchestration and testing. The feature is implemented by adding a `waitToSettleTimeoutMs` parameter to the `Driver`, `Maestro`, `Commands`, and `YamlFluentCommand` classes, and modifying the `waitForAppToSettle` functions in the `AndroidDriver`, `IOSDriver`, `WebDriver`, `ScreenshotUtils`, and `FakeDriver` classes. The feature is also tested by adding a new test case `Case 113 - Tap on element - with app settle timeout` in the `IntegrationTest` class and the `113_tap_on_element_settle_timeout.yaml` file.

## Testing
- Local
- Cloud
- Unit

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -240,6 +240,7 @@ data class TapOnElementCommand(
 
     companion object {
         const val DEFAULT_REPEAT_DELAY = 100L
+        const val MAX_TIMEOUT_WAIT_TO_SETTLE_MS = 30000
     }
 }
 
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -224,7 +224,8 @@ data class TapOnElementCommand(
     val retryIfNoChange: Boolean? = null,
     val waitUntilVisible: Boolean? = null,
     val longPress: Boolean? = null,
-    val repeat: TapRepeat? = null
+    val repeat: TapRepeat? = null,
+    val waitToSettleTimeoutMs: Int? = null
 ) : Command {
 
     override fun description(): String {
@@ -265,7 +266,8 @@ data class TapOnPointV2Command(
     val point: String,
     val retryIfNoChange: Boolean? = null,
     val longPress: Boolean? = null,
-    val repeat: TapRepeat? = null
+    val repeat: TapRepeat? = null,
+    val waitToSettleTimeoutMs: Int? = null
 ) : Command {
 
     override fun description(): String {
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -222,7 +222,8 @@ class Orchestra(
         if (this::jsEngine.isInitialized) {
             jsEngine.close()
         }
-        val shouldUseGraalJs = config?.ext?.get("jsEngine") == "graaljs" || System.getenv("MAESTRO_USE_GRAALJS") == "true"
+        val shouldUseGraalJs =
+            config?.ext?.get("jsEngine") == "graaljs" || System.getenv("MAESTRO_USE_GRAALJS") == "true"
         jsEngine = if (shouldUseGraalJs) {
             httpClient?.let { GraalJsEngine(it) } ?: GraalJsEngine()
         } else {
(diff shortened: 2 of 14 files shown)


arildojr7:
would be great to have this option on the global `config.yaml` as well. My tests are spending too much time on each interaction. I manually changed the waitToSettle timeout on my machine and worked fine and faster

ArthurSav:
@arildojr7 This is something we discussed and decided to not implement for the time being.

Main reason is, it would be inconsistent when running 1 flow vs a test suite. Currently, when running a single flow `config.yaml` is not applied.

arildojr7:
@ArthurSav I see, thanks for the reply

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -401,16 +403,18 @@ class Orchestra(
...
             } catch (ignored: MaestroException.ElementNotFound) {
             }
             maestro.swipeFromCenter(direction, durationMs = command.scrollDuration)
         } while (System.currentTimeMillis() < endTime)
 
-        throw MaestroException.ElementNotFound("No visible element found: ${command.selector.description()}", maestro.viewHierarchy().root)
+        throw MaestroException.ElementNotFound(
+            "No visible element found: ${command.selector.description()}",
+            maestro.viewHierarchy().root
+        )
```
amanjeetsingh150, line 417: I see no changes here is this because of some indentation changes?

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -3023,6 +3023,28 @@ class IntegrationTest {
...
+        }
+
+        // When
+        Maestro(driver).use {
+            orchestra(it).runFlow(commands)
+        }
+
+        // Then
+        // No test failure
+        driver.assertEventCount(Event.Tap(Point(50, 50)), expectedCount = 2)
```
amanjeetsingh150, line 3045: This doesn't assert the timeout though?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -265,7 +267,8 @@ data class TapOnPointV2Command(
     val point: String,
     val retryIfNoChange: Boolean? = null,
     val longPress: Boolean? = null,
-    val repeat: TapRepeat? = null
+    val repeat: TapRepeat? = null,
+    val waitToSettleTimeoutMs: Int? = null
```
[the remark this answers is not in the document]
ArthurSav: Maybe. Although it seems we're kinda arbitrarily apply.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ArthurSav: Intellij formatting

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
ArthurSav: This one doesn't check for timeout - just makes sure it runs as expected with a timeout applied. Will add another case.

ArthurSav: Updated the test with command timeout check

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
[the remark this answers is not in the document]
ArthurSav: Kinda torn on this:
1. Most command params don't eval JS, unless users ask for it
2. IMO JS eval should not happen here as it will force all models to use string params (only)

Let's leave it as it is and reconsider if there's demand for it.

amanjeetsingh150 (approved):

PR #1535, opened 2023-10-18
Move device locale validation logic to maestro-client

artem888:
## Proposed Changes

This PR moves logic related to device locale validation to the `maestro-client` module to be able to reuse in in Maestro Cloud

## Testing
- [x] local test

### DIFF
--- a/maestro-client/src/main/java/maestro/utils/LocaleUtils.kt
+++ b/maestro-client/src/main/java/maestro/utils/LocaleUtils.kt
@@ -106,11 +106,11 @@ object LocaleUtils {
 
     val IOS_SUPPORTED_LOCALES = listOf(
         "en_AU" to "Australia",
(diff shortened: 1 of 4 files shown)


maestro-client/src/main/java/maestro/utils/LocaleUtils.kt
```diff
@@ -156,4 +164,50 @@ object LocaleConstants {
 
         return null
     }
+
+    fun parseLocaleParams(deviceLocale: String, platform: Platform): Pair<String, String> {
```
axelniklasson, line 168: Let's cover this with unit tests

artem888: added ✅

axelniklasson (approved):

PR #1538, opened 2023-10-19
fix: raise a invalid url exception when open link url is not valid

amanjeetsingh150:
## Proposed Changes

1. Escape `?` 
2. Raise invalid URL exception exception to avoid throwing illegal state exceptions

## Testing
- [ ] Local
- [ ] Staging

### DIFF
--- a/maestro-client/src/main/java/maestro/utils/StringUtils.kt
+++ b/maestro-client/src/main/java/maestro/utils/StringUtils.kt
@@ -12,4 +12,7 @@ object StringUtils {
         }
     }
 
+    fun String.toSanitizedIOSLink(): String {
+        return this.replace("?", "\\?")
+    }
 }
--- a/maestro-ios-driver/src/main/kotlin/util/CommandLineUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/CommandLineUtils.kt
@@ -24,7 +24,7 @@ object CommandLineUtils {
         } else {
             ProcessBuilder(*parts.toTypedArray())
                 .redirectOutput(nullFile)
-                .redirectError(nullFile)
+                .redirectError(ProcessBuilder.Redirect.PIPE)
         }
 
         processBuilder.environment().putAll(params)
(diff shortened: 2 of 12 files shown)


maestro-client/src/main/java/maestro/utils/StringUtils.kt
```diff
@@ -12,4 +12,7 @@ object StringUtils {
         }
     }
 
+    fun String.toSanitizedIOSLink(): String {
+        return this.replace("?", "\\?")
```
amanjeetsingh150, line 16: @artem888 is this correct we need to escape ? when using the xcrun simctl open URL command?

I was getting: 

```
zsh: no matches found: testapp://testpage?testparam={TEST:test}
```

Also do we need to escape something else as well?

maestro-ios-driver/src/main/kotlin/util/CommandLineUtils.kt
```diff
@@ -24,7 +24,7 @@ object CommandLineUtils {
         } else {
             ProcessBuilder(*parts.toTypedArray())
                 .redirectOutput(nullFile)
-                .redirectError(nullFile)
+                .redirectError(ProcessBuilder.Redirect.PIPE)
```
amanjeetsingh150, line 27: The errors were not even redirecting to the process due to which it just throws illegal state exception without any meaningful message

maestro-client/src/main/java/maestro/utils/StringUtils.kt
artem888: I tested locally , looks like any special character needs to be sanitized, not only `?`.

smth like that should work:

```
fun sanitizeURL(url: String): String {
    val sanitizedURL = StringBuilder()
    for (ch in url) {
        if (ch.isLetterOrDigit() || ch == '.' || ch == '-' || ch == '_' || ch == '*' || ch == '/') {
            sanitizedURL.append(ch)
        } else {
            sanitizedURL.append(URLEncoder.encode(ch.toString(), "UTF-8"))
        }
    }
    return sanitizedURL.toString()
}
```

artem888 (approved):
LGTM after sanitization logic is modified

PR #1527, opened 2023-10-14
FIX: Allow Maestro to work on Android API levels 25 and lower (Android 7.1.1 and lower)

testifyqa:
## Proposed Changes
- Refactor  `AndroidDriver#startInstrumentationSession` so that if the Android API level is 25 or lower, we omit the `-m` flag from the ADB command that is run.

## Testing
I built and ran locally and this change of mine allowed the sample Wikiepedia android flow to run and pass successfully on both an android device  > 26 and another Android 7.1.1 (API 25) device I have.

## Issues Fixed
This fix allows Maestro to work on Android API level lower than 26 :+1:

PR #1548, opened 2023-10-24
fix gradle task dependency issues

NyCodeGHG:
## Proposed Changes

This PR fixes some gradle task dependency issues experienced in #1527.
The maestro-{app,server}.apk files now also get correctly rebuilt when the source has been changed.

Without this PR these warnings get printed, which can lead to wrong results of the build:
```
> Task :maestro-client:processResources
Execution optimizations have been disabled for task ':maestro-client:processResources' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/home/marie/work/maestro/maestro-client/src/main/resources'. Reason: Task ':maestro-client:processResources' uses this output of task ':maestro-android:copyMaestroAndroid' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6/userguide/validation_problems.html#implicit_dependency for more details about this problem.

> Task :maestro-studio:server:processResources
Execution optimizations have been disabled for task ':maestro-studio:server:processResources' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/home/marie/work/maestro/maestro-studio/server/src/main/resources'. Reason: Task ':maestro-studio:server:processResources' uses this output of task ':maestro-studio:server:copyWebFiles' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

## Testing

I ran the android advanced flow sample using the `maestro` script in the project root, as well as running the build multiple times.

## Issues Fixed
None

### DIFF
--- a/maestro-client/build.gradle
+++ b/maestro-client/build.gradle
@@ -91,8 +91,9 @@ dependencies {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
+    }
 }
 
 plugins.withId("com.vanniktech.maven.publish") {
--- a/maestro-client/build.gradle
+++ b/maestro-client/build.gradle
@@ -104,3 +104,9 @@ plugins.withId("com.vanniktech.maven.publish") {
 test {
     useJUnitPlatform()
 }
+
+processResources {
+    def maestroAndroid = project(":maestro-android")
+    dependsOn(maestroAndroid.tasks.copyMaestroAndroid, maestroAndroid.tasks.copyMaestroServer)
+}
+
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,8 +8,9 @@ plugins {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
+    }
 }
 
 detekt {
--- a/maestro-android/build.gradle
+++ b/maestro-android/build.gradle
@@ -75,49 +75,29 @@ android {
 }
 
 tasks.register("copyMaestroAndroid", Copy) {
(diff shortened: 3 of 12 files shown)


maestro-client/build.gradle
```diff
@@ -104,3 +104,8 @@ plugins.withId("com.vanniktech.maven.publish") {
 test {
     useJUnitPlatform()
 }
+
+processResources {
+    def maestroAndroid = project(":maestro-android")
+    dependsOn(maestroAndroid.tasks.copyMaestroAndroid, maestroAndroid.tasks.copyMaestroServer)
+}
```
testifyqa, line 111: nit: EOF

testifyqa (approved):

maestro-client/build.gradle
NyCodeGHG: fixed it!

igorsmotto:
Hi @NyCodeGHG, thanks for this contribution!

Could we also add back the verification of the java version? To make sure we're using the correct one (Java 8)

NyCodeGHG:
@igorsmotto We could do that, but I think it would be better to use [Gradle Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) instead, so it doesn't matter which java version you run gradle with. What do you think?

igorsmotto:
That's an interesting suggestion @NyCodeGHG.

I don't know much about the Toolchain, but if we can make sure that java, kotlin and android compilation tasks will use the same JDK (8) that will definitely works. 

From their docs, it also seems that Toolchain has the option to download the java version if its not autodetected in your local machine, but I believe this can be skipped for now if you want to.

NyCodeGHG:
I changed the project to use toolchains now, maybe it would also be a good idea in the future to reduce code duplication in all of the build.gradle{,.kts} files

axelniklasson:
@NyCodeGHG I'll go ahead and close this PR out as we won't be able to prioritize resources to testing and ensuring that this doesn't break any internal build pipelines that rely on this. Thanks for taking the time to raise this PR!

PR #1556, opened 2023-10-26
Make simulator reboot logic more robust

artem888:
## Proposed Changes

This PR makes awaitLaunch more robust - no timeout needed anymore, bootstatus is better in this case, since it waits until Springboard is fully launched, while just tracking the Booted state from `simctl list` doesn't

## Testing
- [x] tested locally

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
@@ -3,22 +3,17 @@ package util
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import maestro.utils.MaestroTimer
-import org.apache.commons.io.FileUtils
 import org.rauschig.jarchivelib.ArchiveFormat
 import org.rauschig.jarchivelib.ArchiverFactory
 import util.CommandLineUtils.runCommand
 import java.io.File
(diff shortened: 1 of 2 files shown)


maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -51,20 +46,13 @@ object LocalSimulatorUtils {
...
-            ) true else null
-        } ?: throw SimctlError("Device $deviceId did not boot in time")
+    fun awaitLaunch(deviceId: String) {
+        val command = listOf("xcrun", "simctl", "bootstatus", deviceId)
+        runCommand(command, waitForCompletion = true)
     }
 
-    fun awaitShutdown(deviceId: String) {
-        MaestroTimer.withTimeout(30000) {
+    fun awaitShutdown(deviceId: String, timeoutMs: Long = 60000) {
```
amanjeetsingh150, line 54: Are we not going to increase this timeout?

amanjeetsingh150 (approved):

amanjeetsingh150:
Looks good after tested!

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
artem888: I increased it just in case, from what I've tested has no functional impact on local cli logic, but might be safer in the cloud

PR #1564, opened 2023-10-31
conform to XDG base dir spec on linux

NyCodeGHG:
## Proposed Changes
Conform to the [XDG base dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on linux to not spam the users home directory.
This change is backwards compatible and still uses the old directory if it exists.
It also asks the user on startup if they want to migrate to the new directory.

## Testing
Ran tests and the advanced flow on an Android AVD.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/App.kt
+++ b/maestro-cli/src/main/java/maestro/cli/App.kt
@@ -26,6 +26,7 @@ import maestro.cli.update.Updates
 import maestro.cli.util.ErrorReporter
 import maestro.cli.view.box
 import maestro.debuglog.DebugLogStore
+import maestro.utils.MaestroDirectory
 import picocli.CommandLine
 import picocli.CommandLine.Command
 import picocli.CommandLine.Option
@@ -82,6 +83,8 @@ fun main(args: Array<String>) {
     // Disable icon in Mac dock
     // https://stackoverflow.com/a/17544259
     System.setProperty("apple.awt.UIElement", "true")
+
+    MaestroDirectory.migrate()
 
     Dependencies.install()
     Updates.fetchUpdatesAsync()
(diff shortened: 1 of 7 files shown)


axelniklasson:
Hi @NyCodeGHG, I'll take a look at this next week

axelniklasson:
@NyCodeGHG could you provide some more context on what "spamming the users home directory" means?

NyCodeGHG:
@axelniklasson 
If i look in my home directory, there are a lot of directories starting with a dot. (currently 56)
That's really annoying and a lot of clutter.
Thats why the [XDG Basedir Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) exists.
It's comparable to the similar guidelines which exist for macOS and Windows as well.
This spec gives guidelines where applications should store their data.

axelniklasson:
> @axelniklasson If i look in my home directory, there are a lot of directories starting with a dot. (currently 56) That's really annoying and a lot of clutter. Thats why the [XDG Basedir Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) exists. It's comparable to the similar guidelines which exist for macOS and Windows as well. This spec gives guidelines where applications should store their data.

@NyCodeGHG  Are there multiple `.maestro` directories created by accident or are there one `.maestro` and 55 other directories that start with a dot?

NyCodeGHG:
@axelniklasson one `.maestro` and 55 other directories created by other applications. It wouldn't really be a problem for me if only maestro would do that, but way too many applications just dump their data in the home directory where it doesn't really belong.

If the migration thing is too complicated I would also suggest that we could leave the migration thing out and only use the new directory if the old one doesn't exist, which would probably be much simpler to maintain.

axelniklasson:
@NyCodeGHG Thanks for the context. While I see that this is a bit annoying for you and I appreciate that you took the time to create a PR, we won't be merging this PR. Thanks for your time and willingness to contribute to maestro!

PR #999, opened 2023-04-19
Add `containsDescendants` selector

larryng:
## Proposed Changes
Adds selector `containsDescendants` which allows selecting a view that has a selection of descendant elements. The motivation for this was to provide an easier way to verify the contents of individual list items.

## Testing
* Added integration tests
* Using in internal Maestro flows
* Tested in Wikipedia app
    Screenshot 2023-04-18 at 8 55 03 PM

PR #1586, opened 2023-11-11
childOf selector

sandeepjoshi19:
## Proposed Changes
When we are reusing a component/ui inside different different UI. Then to correctly identify which component to select we are adding this `childOf` selector

eg we have a button element and its id is `proceed`. Now this button is used by two UI component in the same page. Now we have two view components id: add_to_cart and buy_now but the inner button have same id `proceed`. So if placement of  these two UI element is dynamic it becomes very difficult to correctly select which one represents add_to_cart and which represents buy_now to tap.

`childOf` selector into rescue. `childOf` selector tells maestro to select id or text which is child of provided selector.

eg if we want to tap on buy now button we can easily write.
```
tapOn:
   text: "proceed"
    childOf:
        - id: "buy-now"
```

this will first search for Element with id  "buy-now" then inside it's child tree it will try to find  element with id "proceed"

View
image

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -457,7 +457,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         return if (element == null) {
             null
         } else {
-            return FindElementResult(element, viewHierarchy)
+            return FindElementResult(element, ViewHierarchy(element.treeNode))
         }
     }
 
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -401,7 +401,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
     fun findElementByRegexp(regex: Regex, timeoutMs: Long): UiElement {
         LOGGER.info("Looking for element by regex: ${regex.pattern} (timeout $timeoutMs)")
 
-        return findElementWithTimeout(timeoutMs, Filters.textMatches(regex), null)?.element
+        return findElementWithTimeout(timeoutMs, Filters.textMatches(regex))?.element
             ?: throw MaestroException.ElementNotFound(
                 "No element that matches regex: $regex",
                 viewHierarchy().root
@@ -415,7 +415,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
     fun findElementByIdRegex(regex: Regex, timeoutMs: Long): UiElement {
         LOGGER.info("Looking for element by id regex: ${regex.pattern} (timeout $timeoutMs)")
 
-        return findElementWithTimeout(timeoutMs, Filters.idMatches(regex), null)?.element
+        return findElementWithTimeout(timeoutMs, Filters.idMatches(regex))?.element
             ?: throw MaestroException.ElementNotFound(
                 "No element has id that matches regex $regex",
                 viewHierarchy().root
@@ -427,15 +427,14 @@ class Maestro(private val driver: Driver) : AutoCloseable {
 
         return findElementWithTimeout(
             timeoutMs,
-            Filters.sizeMatches(width, height, tolerance).asFilter(),
-            null
+            Filters.sizeMatches(width, height, tolerance).asFilter()
         )?.element
     }
 
     fun findElementWithTimeout(
         timeoutMs: Long,
         filter: ElementFilter,
-        viewHierarchy: ViewHierarchy?
+        viewHierarchy: ViewHierarchy? = null
     ): FindElementResult? {
         var hierarchy = viewHierarchy ?: ViewHierarchy(TreeNode())
         val element = MaestroTimer.withTimeout(timeoutMs) {
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -401,7 +401,7 @@ class Maestro(private val driver: Driver) : AutoCloseable {
     fun findElementByRegexp(regex: Regex, timeoutMs: Long): UiElement {
         LOGGER.info("Looking for element by regex: ${regex.pattern} (timeout $timeoutMs)")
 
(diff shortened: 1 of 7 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/ElementSelector.kt
```diff
@@ -116,6 +118,10 @@ data class ElementSelector(
             descriptions.add("Index: ${it.toDoubleOrNull()?.toInt() ?: it}")
         }
 
+        childOf?.let {
+            descriptions.add("Child Of: ${it.description()}")
```
axelniklasson, line 122:
```suggestion
            descriptions.add("Child of: ${it.description()}")
```

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -445,6 +445,22 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         }
     }
 
+    fun findElementFromViewHierarchyWithTimeout(
```
axelniklasson, line 448: Please update `findElementWithTimeout` and make the hierarchy argument optional and when passed use that hierarchy instead of looking it up on the fly

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -855,6 +868,26 @@ class Orchestra(
...
     }
 
+    private fun findElementViewHierarchy(
+        selector: ElementSelector?,
+        timeout: Long
+    ):ViewHierarchy {
+        if(selector == null ){
+            return maestro.viewHierarchy()
+        }
+        val parentViewHierarchy = findElementViewHierarchy(selector.childOf,timeout);
```
axelniklasson, line 878:
```suggestion
        val parentViewHierarchy = findElementViewHierarchy(selector.childOf, timeout)
```

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -855,6 +868,26 @@ class Orchestra(
         )
     }
 
+    private fun findElementViewHierarchy(
+        selector: ElementSelector?,
+        timeout: Long
+    ):ViewHierarchy {
```
axelniklasson, line 874:
```suggestion
    ): ViewHierarchy {
```

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -855,6 +868,26 @@ class Orchestra(
...
+        if(selector == null ){
```
axelniklasson, line 875:
```suggestion
        if (selector == null) {
```

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -855,6 +868,26 @@ class Orchestra(
...
+        val (description, filterFunc) = buildFilter(
+            selector,
+            deviceInfo(),
+        )
+        return maestro.findElementFromViewHierarchyWithTimeout(timeout,
+            filterFunc,
+            parentViewHierarchy)?.hierarchy?: throw MaestroException.ElementNotFound(
+            "Element not found: $description",
+            parentViewHierarchy.root,
+        )
```
axelniklasson, line 888:
```suggestion
        return maestro.findElementFromViewHierarchyWithTimeout(
            timeout,
            filterFunc,
            parentViewHierarchy
        )?.hierarchy ?: throw MaestroException.ElementNotFound(
            "Element not found: $description",
            parentViewHierarchy.root,
        )
```

axelniklasson (changes requested):
There's no automatic linting in place for the maestro repo, so there are quite a few nit comments here -- apologies in advance 🙂 

Let's also add test cases in `IntegrationTest` for this functionality -- you can check https://github.com/mobile-dev-inc/maestro/pull/999 for some inspiration on how to add this

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -425,22 +425,30 @@ class Maestro(private val driver: Driver) : AutoCloseable {
...
+            timeoutMs,
+            Filters.sizeMatches(width, height, tolerance).asFilter(),
+            null
+        )?.element
     }
 
     fun findElementWithTimeout(
         timeoutMs: Long,
         filter: ElementFilter,
+        viewHierarchy: ViewHierarchy?
```
axelniklasson, line 438:
```suggestion
        viewHierarchy: ViewHierarchy? = null
```

axelniklasson: You can set `null` as a default value here to avoid having to pass `null` at all existing callsites

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -840,21 +840,58 @@ class Orchestra(
...
+                selector.childOf,
+                timeout
+            )
+            return maestro.findElementWithTimeout(
+                timeout,
+                filterFunc,
+                parentViewHierarchy
+            ) ?: throw MaestroException.ElementNotFound(
+                "Element not found: $description",
+                maestro.viewHierarchy().root,
```
axelniklasson, line 858:
```suggestion
                parentViewHierarchy.root
```
Probably should be the parent view hierarchy attached to the error here instead of the root as we can't find the child in the parent hierarchy

axelniklasson:
@sandeepjoshi19 please also raise a PR with documentation updates in https://github.com/mobile-dev-inc/maestro-docs and tag me as a reviewer :)

maestro-client/src/main/java/maestro/Maestro.kt
sandeepjoshi19: Yes missed this. Thanks

sandeepjoshi19:
@axelniklasson Raised the PR but I am not able to add reviewers.
PR: https://github.com/mobile-dev-inc/maestro-docs/pull/31

PR #1588, opened 2023-11-14
Improve device specs message for maestro cloud command to include device locale

artem888:
## Proposed Changes

Improves device specs message to include --device-locale parameter. Example UI:

Screenshot 2023-11-15 at 17 30 04

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -172,7 +172,7 @@ class CloudInteractor(
 
         val platform = Platform.fromString(deviceInfo.platform)
 
-        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo}"
+        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo} - locale ${deviceInfo.deviceLocale}"
         val line2 = "To change OS version use this option: ${if (platform == Platform.IOS) "--ios-version=<version>" else "--android-api-level=<version>"}"
 
         val version = when(platform) {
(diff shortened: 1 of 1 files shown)


axelniklasson (approved):

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -172,7 +172,7 @@ class CloudInteractor(
 
         val platform = Platform.fromString(deviceInfo.platform)
 
-        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo}"
+        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo} - locale ${deviceInfo.deviceLocale}"
         val line2 = "To change OS version use this option: ${if (platform == Platform.IOS) "--ios-version=<version>" else "--android-api-level=<version>"}"
```
axelniklasson, line 176: Can we also add a message with instructions on how to change locale like how we do for OS version?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -172,7 +172,7 @@ class CloudInteractor(
 
         val platform = Platform.fromString(deviceInfo.platform)
 
-        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo}"
+        val line1 = "Maestro Cloud device specs:\n* ${deviceInfo.displayInfo} - locale ${deviceInfo.deviceLocale}"
```
ArthurSav, line 175: nit: Since we don't label things, maybe we can skip the "locale" part.

ArthurSav (approved):

PR #1596, opened 2023-11-21
feat: adds scroll attribute

amanjeetsingh150:


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -784,6 +784,10 @@ class AndroidDriver(
                 attributesBuilder["checked"] = node.getAttribute("checked")
             }
 
+            if (node.hasAttribute("scrollable")) {
+                attributesBuilder["scrollable"] = node.getAttribute("scrollable")
+            }
```
axelniklasson, line 789: Is there a similar property that we should surface on iOS as well?

axelniklasson (approved):

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
amanjeetsingh150: Unfortunately no! I'll double check the private apis if we can still get a hand around them but I don't think it exist in public api

PR #1621, opened 2023-12-18
Allow maestro to work on Android API 21-23

NyCodeGHG:
## Proposed Changes

This PR allows maestro to work on Android API 21-23 devices.
It depends on a change in dadb and currently uses a git dependency, which I'll change once https://github.com/mobile-dev-inc/dadb/pull/67 is merged.

## Testing

I ran the advanced flow successfully on emulators with API 21,22,23 and 33.


axelniklasson:
Hey @NyCodeGHG thanks for taking the time to create this PR! Unfortunately we have decided internally that we won't be merging this change as we don't want to add complexity to the code base to support Android API < 24.

PR #1638, opened 2024-01-02
feat: adds `filterOutKeyboardElements` to make backward compatible

amanjeetsingh150:
## Proposed Changes

Adds filterOutKeyboardElements option to filter out elements in keyboard bounds. This will be backward compatible since this only works when this option is true


maestro-client/src/main/java/maestro/Driver.kt
```diff
@@ -51,7 +51,7 @@ interface Driver {
 
     fun pressKey(code: KeyCode)
 
-    fun contentDescriptor(): TreeNode
+    fun contentDescriptor(filterOutKeyboardElements: Boolean = false): TreeNode
```
axelniklasson, line 54:
```suggestion
    fun contentDescriptor(excludeKeyboardElements: Boolean = false): TreeNode
```

axelniklasson (approved):

maestro-client/src/main/java/maestro/Driver.kt
amanjeetsingh150: Done ✅

PR #1671, opened 2024-02-05
Fix "Protocol family unavailable" on rerun after failure

sjchmiela:
## Proposed Changes

I think `xcTestDriverStatusCheck` may return a response object with an error response or throw. In `isChannelAlive`, where we care about _whether_ channel is alive we should treat an exception as an `isChannelAlive = false`.

## Testing

I verified this is valid Kotlin code and looks similar to `ensureOpen`.

## Issues Fixed

I've run into the "Protocol family unavailable" after a CI-run test job fails to connect to XCUITest server. I believe the `xcTestDriverStatusCheck` should be wrapped with `try-catch`, as it is in `ensureOpen()`.


amanjeetsingh150 (approved):

amanjeetsingh150:
Thanks @sjchmiela for the contribution. This should be out next release!

PR #1672, opened 2024-02-06
feat: allow airplane mode control

NyCodeGHG:
## Proposed Changes

add a new command for controlling the airplane mode of android devices.

## Testing

ran the following flow on android 27 and 28 emulators:
```yaml
appId: org.wikipedia
---
- launchApp
- setAirplaneMode: disable
- setAirplaneMode: enable
```

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -889,6 +889,16 @@ data class SetAirplaneModeCommand(
     }
 }
 
+object ToggleAirplaneModeCommand : Command {
+    override fun description(): String {
+        return "Toggle airplane mode"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+}
+
 internal fun tapOnDescription(isLongPress: Boolean?, repeat: TapRepeat?): String {
     return if (isLongPress == true) "Long press"
     else if (repeat != null) {
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -19,6 +19,7 @@
 
 package maestro.orchestra
 
+import com.fasterxml.jackson.annotation.JsonProperty
 import maestro.KeyCode
 import maestro.Point
 import maestro.ScrollDirection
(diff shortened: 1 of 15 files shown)


axelniklasson:
@NyCodeGHG Nice 🎉 Let's update it to be a single command `- toggleAirplaneMode` instead and then we'll move forward with a formal review to get it merged. Thanks!

axelniklasson (changes requested):
https://github.com/mobile-dev-inc/maestro/pull/1672#issuecomment-2125030154

NyCodeGHG:
@axelniklasson hi, thank you for the review.
When I implemented this feature I had another option to toggle the airplane mode, but I removed it as it doesn't explicitly define the desired behavior. 
I can add another option to toggle if you want, but I'd prefer to keep the enable and disable options for those who want to be explicit.

axelniklasson:
> @axelniklasson hi, thank you for the review. When I implemented this feature I had another option to toggle the airplane mode, but I removed it as it doesn't explicitly define the desired behavior. I can add another option to toggle if you want, but I'd prefer to keep the enable and disable options for those who want to be explicit.

Hey, that's fair and a good use case to me. Let's keep the API as it is but also add `- toggleAirplaneMode` as a separate command

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -866,6 +867,28 @@ data class StopRecordingCommand(
     }
 }
 
+enum class AirplaneValue {
+    @JsonProperty("enable")
```
axelniklasson, line 871:
```suggestion
    @JsonProperty("enabled")
```

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -866,6 +867,28 @@ data class StopRecordingCommand(
...
+    Enable,
+    @JsonProperty("disable")
```
axelniklasson, line 873:
```suggestion
    @JsonProperty("disabled")
```

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -866,6 +867,38 @@ data class StopRecordingCommand(
...
+    Disable,
+}
+
+data class SetAirplaneModeCommand(
+    val value: AirplaneValue,
+) : Command {
+    override fun description(): String {
+        return when (value) {
+            AirplaneValue.Enable -> "Enable Airplane mode"
+            AirplaneValue.Disable -> "Disable Airplane mode"
```
axelniklasson, line 883:
```suggestion
            AirplaneValue.Enable -> "Enable airplane mode"
            AirplaneValue.Disable -> "Disable airplane mode"
```

axelniklasson (approved):
Nice, this works well for me! Can you also raise a PR in https://github.com/mobile-dev-inc/maestro-docs as well as adding another test case to the integration tests? Thanks!

NyCodeGHG:
@axelniklasson 
I'm not sure on the integration test case
Can you elaborate?

axelniklasson:
Thanks @NyCodeGHG! Merging this now but please do also raise a PR in https://github.com/mobile-dev-inc/maestro-docs documenting the new commands - we'll merge it when we make the next release 👍 

Thanks again for your contribution!

NyCodeGHG:
will do!

PR #1721, opened 2024-04-19
Make flow execution output work with JUNIT output format

NyCodeGHG:
## Proposed Changes

This PR allows the flow execution output to work while choosing the JUnit output format.
It also adds a new flag `--hide-execution`, which allows hiding of the execution output.

## Testing

Ran the android advanced example flow, as well as an internal test suite, using with and without the `--output=junit` flag.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -23,6 +23,7 @@ import maestro.cli.App
 import maestro.cli.CliError
 import maestro.cli.DisableAnsiMixin
 import maestro.cli.api.ApiClient
+import maestro.cli.model.TestExecutionSummary
 import maestro.cli.report.TestDebugReporter
 import maestro.cli.runner.TestRunner
 import maestro.cli.runner.resultview.AnsiResultView
@@ -41,7 +42,7 @@ import java.util.concurrent.Callable
         "Render a beautiful video of your Flow - Great for demos and bug reports"
     ]
 )
-class RecordCommand : Callable<Int> {
+class RecordCommand : Callable<TestExecutionSummary.FlowResult> {
 
     @CommandLine.Mixin
     var disableANSIMixin: DisableAnsiMixin? = null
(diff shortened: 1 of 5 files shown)


axelniklasson:
@NyCodeGHG could you share some more examples on what this change does? Some console output with and without the `--hide-execution` would be appreciated, thanks!

axelniklasson:
@lucaswiechmann will help get this merged as I'll be OOO starting next week. Please let him know when you've updated the PR @NyCodeGHG!

NyCodeGHG:
@axelniklasson / @lucaswiechmann 
are you sure changing this behaviour is a good idea?
with this pr it's still different as it would show nothing at all, instead of the output that a flow is running.
So the default user experience would be worse

lucaswiechmann:
> @axelniklasson / @lucaswiechmann are you sure changing this behaviour is a good idea? with this pr it's still different as it would show nothing at all, instead of the output that a flow is running. So the default user experience would be worse

Hey @NyCodeGHG 

let's keep the current output/behavior that we have today
image

And then add two flags:

* one to show extended output like this
image

* Another (we could keep the hide-execution) to not show any output at all

NyCodeGHG:
@lucaswiechmann 
I like your suggestion, but I'm not sure if thats easily possible, as the current behaviour/output is an entirely different codepath.

chrisaydat:
is it possible to add the flow execution results to the output junit xml file?

bartekpacia:
@chrisaydat Could you clarify what you mean?

NyCodeGHG:
I'm not interested in this anymore as we moved away from maestro.
If someone else wants to pick this up, feel free to create a follow up PR.

chrisaydat:
Screenshot 2024-08-19 at 11 46 59 AM
@bartekpacia I mean something like this, whether it is the labels or the results as is on both HTML and Junit reports

bartekpacia:
Review note:
- I very much like the reduced code duplication across `TestRunner` and `TestSuiteInteractor`. Related: #1918 
- I don't agree with adding a new  `NoopResultsView`

bartekpacia:
My reasoning: specifying `--format` (whether JUnit or HTML) should have no impact on Maestro CLI output. It's an unrelated option.

Therefore, I don't think the `--hide-execution` is needed.

bartekpacia:
Ahh, there's a big merge conflict in `TestCommand`, because of #1732

Fishbowler:
Looks like this is abandoned. Some of that's definitely our fault in being tardy on reviews. Sorry about that. If you get energy to resubmit, please do, and please tag me.

PR #1723, opened 2024-04-23
Added KillApp option for launchApp command

galex:
## Proposed Changes

This PR allows to test a screen against what is called a [System-initiated Process Death](https://galex.dev/posts/how-to-detect-process-death-issues/) which is something very specific to the Android platform, but gives a very strong proposal value to Maestro for Android developers :+1: 

To trigger that properly, we need the `launchApp` command to call
```bash
adb shell am kill <package name>
```
Instead of 
```bash
adb shell am force-stop <package name>
```
I suggest doing so by adding a `killApp` parameter to the `LaunchApp` command that takes precedence to `stopApp`:

```yaml
appId: com.example.app
---
- launchApp:
    killApp: true
```

## Testing
- Added an unit test for reading a new `killApp` parameter on the `LaunchApp` command.
- Added an integration test
- Tested against a real [use case in a demo project](https://github.com/galex/process-death-demo-project/blob/main/maestro/process-death-flow.yaml)

Let me know what you think or if something missing! :blush: 
Thank you :pray:

### DIFF
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -41,6 +41,8 @@ interface Driver {
 
     fun stopApp(appId: String)
 
+    fun killApp(appId: String)
+
     fun clearAppState(appId: String)
 
     fun clearKeychain()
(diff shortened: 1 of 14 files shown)


galex:
Replaced by https://github.com/mobile-dev-inc/maestro/pull/1727

PR #1727, opened 2024-04-25
Added KillApp command to reproduce System-initiated Process Death on Android

galex:
## Proposed Changes

This PR allows to test a screen against what is called a [System-initiated Process Death](https://galex.dev/posts/how-to-detect-process-death-issues/) which is something very specific to the Android platform, but gives a very strong proposal value to Maestro for Android developers :+1: 

Currently, only `stopApp` is available as a command to try to reproduce it but it fails to as it uses the `force-stop` adb command:
```bash
adb shell am force-stop <package name>
```

To trigger System-initiated Process Death properly, Android developers need to be able to call the `kill` command on adb:
```bash
adb shell am kill <package name>
```

I suggest doing so by adding a `killApp` command:

```yaml
appId: com.example.app
---
- killApp
```
Different behaviors per platform:
- Android: Calls `adb shell am kill` under the hood
- iOS: Calls `stopApp` (synonym command) as there is no such thing as a Process Death on iOS 
- Web: Calls `stopApp` (synonym command) as there is no such thing as a Process Death on Web

Using this new command, an Android developer can really test for this use case by doing the following:
```yaml
appId: com.example.app
---
# Puts the app into the background
- pressKey: Home
# Kills the app (adb shell am kill) 
- killApp
# Relaunches the app
- launchApp:
    stopApp: false # Without adb shell am force-stop
```
A test to detect the Process Death issue [ can be found there](https://github.com/galex/process-death-demo-project/blob/main/maestro/detect-process-death-issue.yaml) and running it does manage to detect the issue properly:

Screenshot from 2024-04-25 17-32-54

## Testing
- Added an unit test for reading a new `killApp` command
- Added an integration test
- Tested against a real [use case in a demo project](https://github.com/galex/process-death-demo-project/blob/main/maestro/detect-process-death-issue.yaml)

Let me know what you think or if something missing! :blush: 
Thank you :pray:


amanjeetsingh150 (approved):

amanjeetsingh150:
@galex can you help raising PR on maestro-docs. For adding docs around what this solves and usage?

https://github.com/mobile-dev-inc/maestro-docs

galex:
@amanjeetsingh150 Sure!

PR #1742, opened 2024-05-15
change test-report output directory

depapp:
### what?
I propose these changes to move the test-result file output directory to `/Users/user/.maestro/tests/*/report.xml` instead of the root folder.

### why?
Since the generated files from test execution are stored at `/Users/user/.maestro/tests/*`, I think it'll be easier to find the test result if it's stored in the same location.

### result
Screenshot 2024-05-15 at 13 13 17

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -154,7 +154,7 @@ class TestCommand : Callable<Int> {
                     env = env,
                     reportOut = format.fileExtension
                         ?.let { extension ->
-                            (output ?: File("report$extension"))
+                            File(debugOutputPath.toFile(), "report$extension")
                                 .sink()
                                 .buffer()
                         },


Fishbowler:
I missed that an old PR I had for this got closed waiting for some feedback from me.

I've reopened it as #1748 which makes exactly the same change as you've got here, but makes it optional, and avoids accidentally deleting the contents if you set a custom path like `$HOME`

felipevolpone:
@depapp I'm not sure if I followed your goal with this PR. We already have the `--output` flag that you can use to save the report file. Isn't that enough for your use case?

This PR actually removes the functionality of that flag, so I don't get it

depapp:
> @depapp I'm not sure if I followed your goal with this PR. We already have the `--output` flag that you can use to save the report file. Isn't that enough for your use case?
> 
> This PR actually removes the functionality of that flag, so I don't get it

@felipevolpone Actually, the goal of this PR is to change the default location of the generated test report to the same location as other generated files (like screenshots and maestro.log). 

I am aware that Maestro has the `--output` option to save the report file. However, if I use that option, I cannot place the test report file in the same folder as those other files because the folder name includes a timestamp.

With this change, we can ensure that all related outputs are in the same folder, which greatly simplifies organization and accessibility.

If this PR doesn't make sense, feel free to close it 🙇🏻

PR #1445, opened 2023-09-13
Add command-line switch to omit folders from debug output

Fishbowler:
Fixes #1427 

Adding a `--flatten-debug-output` switch that will remove the datetime folder, as well as `/.maestro/tests/` from the path, leaving the debug output exactly where the user defined.

### <samp>🤖 Generated by Copilot at 972f615</samp>

Added a new option `--flatten-debug-output` to the `test` command in `maestro-cli` to simplify the test output folder structure for CI environments. Updated the `TestDebugReporter` class to handle this option and generate the output path accordingly.

## Testing

No params:
```
 ⇒ ./maestro test ~/maestro/scratch.yaml 
Running on iPhone 14 Pro - iOS 16.4 - 1FCE17AC-A5FF-496C-9EDB-7C66E9D0FBE3
                                                                                
 ║                                                                              
 ║  > Flow                                                                      
 ║                                                                              
 ║    ❌  1 is the same as 2                                                     
 ║                                                                              
                                                                                
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/Users/danc/.maestro/tests/2023-09-13_174225
```

Just the debug option:
```
 ⇒ ./maestro test --debug-output /tmp/maestro ~/maestro/scratch.yaml                    
Running on iPhone 14 Pro - iOS 16.4 - 1FCE17AC-A5FF-496C-9EDB-7C66E9D0FBE3
                                                                                
 ║                                                                              
 ║  > Flow                                                                      
 ║                                                                              
 ║    ❌  1 is the same as 2                                                     
 ║                                                                              
                                                                                
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/tmp/maestro/.maestro/tests/2023-09-13_174126
```

With the new option:
```
 ⇒ ./maestro test --debug-output /tmp/maestro --flatten-debug-output  ~/maestro/scratch.yaml
Running on iPhone 14 Pro - iOS 16.4 - 1FCE17AC-A5FF-496C-9EDB-7C66E9D0FBE3
                                                                                
 ║                                                                              
 ║  > Flow                                                                      
 ║                                                                              
 ║    ❌  1 is the same as 2                                                     
 ║                                                                              
                                                                                
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/tmp/maestro
```

## Issues Fixed

#1427

PR #1748, opened 2024-05-17
Add command-line switch to omit folders from debug output

### LINKED ISSUE #1427: [Feature Request] Be able to define actual folder name of debug output rather than using datetime so we can access screenshots on CI runs
Unable to access screenshots/logs on CI due to the folder name being created using datetime at runtime

Be able to define the actual folder the debug output screenshots/logs gets saved in, currently you can define the root folders but it always adds a folder at the end with the datetime of the run. This means it’s very difficult to be able to access these within CI if ever there’s a failure, as we have no idea what the folder will be called prior to the run.

Fishbowler:
Continuing #1445 
Fixes #1427 

Adding a `--flatten-debug-output` switch that will remove the datetime folder, as well as `/.maestro/tests/` from the path, leaving the debug output exactly where the user defined.

(Updated with feedback from previous PR to remove 

## Proposed Changes

Added a new option `--flatten-debug-output` to the `test` command in `maestro-cli` to simplify the test output folder structure for CI environments. Updated the `TestDebugReporter` class to handle this option and generate the output path accordingly.

## Testing

I'm using a simple flow to force an error:
```
appId: com.example
---
- assertTrue: 
    condition: ${ 1 === 2 }
    label: 1 is the same as 2
```

No params:
```
 ⇒ ./maestro test scratch.yaml 
Running on emulator-5554      
                              
 ║                            
 ║  > Flow                    
 ║                            
 ║    ❌  Does 1 equal 2?     
 ║                            
                              
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/Users/danc/.maestro/tests/2024-05-17_150239
```

Just the debug option:
```
 ⇒ ./maestro test --debug-output /tmp/maestro scratch.yaml                    
Running on emulator-5554        
                                
 ║                              
 ║  > Flow                      
 ║                              
 ║    ❌  1 is the same as 2    
 ║                              
                                
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/tmp/maestro/.maestro/tests/2024-05-17_150426
```

With the new option:
```
 ⇒ ./maestro test --debug-output /tmp/maestro --flatten-debug-output  scratch.yaml
Running on emulator-5554        
                                
 ║                              
 ║  > Flow                      
 ║                              
 ║    ❌  1 is the same as 2    
 ║                              
                                
Assertion is false: false is true

==== Debug output (logs & screenshots) ====

/tmp/maestro
```

Full file listing of /tmp/maestro after above:
```
 ⇒ find /tmp/maestro 
/tmp/maestro
/tmp/maestro/screenshot-❌-1715954725865-(scratch.yaml).png
/tmp/maestro/commands-(scratch.yaml).json
/tmp/maestro/.maestro
/tmp/maestro/.maestro/tests
/tmp/maestro/.maestro/tests/2024-05-17_150426
/tmp/maestro/.maestro/tests/2024-05-17_150426/commands-(scratch.yaml).json
/tmp/maestro/.maestro/tests/2024-05-17_150426/maestro.log
/tmp/maestro/.maestro/tests/2024-05-17_150426/screenshot-❌-1715954675177-(scratch.yaml).png
/tmp/maestro/maestro.log
```

## Issues Fixed

#1427

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -167,7 +167,9 @@ class TestCommand : Callable<Int> {
                     debugOutputPath = debugOutputPath
                 )
 
-                TestDebugReporter.deleteOldFiles()
+                if(!flattenDebugOutput){
+                    TestDebugReporter.deleteOldFiles()
+                }
                 if (suiteResult.passed) {
                     0
                 } else {
@@ -176,7 +178,9 @@ class TestCommand : Callable<Int> {
                 }
             } else {
                 if (continuous) {
-                    TestDebugReporter.deleteOldFiles()
+                    if(!flattenDebugOutput){
+                        TestDebugReporter.deleteOldFiles()
+                    }
                     TestRunner.runContinuous(maestro, device, flowFile, env)
                 } else {
                     val resultView = if (DisableAnsiMixin.ansiEnabled) AnsiResultView() else PlainTextResultView()
@@ -184,7 +188,9 @@ class TestCommand : Callable<Int> {
                     if (resultSingle == 1) {
                         printExitDebugMessage()
                     }
-                    TestDebugReporter.deleteOldFiles()
+                    if(!flattenDebugOutput){
+                        TestDebugReporter.deleteOldFiles()
+                    }
                     return@newSession resultSingle
                 }
             }
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -88,6 +88,12 @@ class TestCommand : Callable<Int> {
     private var debugOutput: String? = null
 
     @Option(
+        names = ["--flatten-debug-output"],
+        description = ["Don't create subfolders or timestamps for each run. Useful for CI."]
+    )
+    private var flattenDebugOutput: Boolean = false
+
+    @Option(
         names = ["--include-tags"],
         description = ["List of tags that will remove the Flows that does not have the provided tags"],
         split = ",",
@@ -130,7 +136,7 @@ class TestCommand : Callable<Int> {
 
         env = env.withInjectedShellEnvVars()
 
-        TestDebugReporter.install(debugOutputPathAsString = debugOutput)
+        TestDebugReporter.install(debugOutputPathAsString = debugOutput, flattenDebugOutput = flattenDebugOutput)
         val debugOutputPath = TestDebugReporter.getDebugOutputPath()
         
         return MaestroSessionManager.newSession(parent?.host, parent?.port, deviceId) { session ->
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -125,10 +127,11 @@ object TestDebugReporter {
     fun getDebugOutputPath(): Path {
         if (debugOutputPath != null) return debugOutputPath as Path
 
-        val dateFormat = "yyyy-MM-dd_HHmmss"
-        val dateFormatter = DateTimeFormatter.ofPattern(dateFormat)
-        val folderName = dateFormatter.format(LocalDateTime.now())
-        val debugOutput = Paths.get(debugOutputPathAsString ?: System.getProperty("user.home"), ".maestro", "tests", folderName)
+        val preamble = if(flattenDebugOutput) arrayOf("") else arrayOf(".maestro", "tests")
```
felipevolpone, line 130: @Fishbowler Instead of having this if in every line, let's create a method that returns the `debugOutput` when `flatterDebugOutput` is `true` and another method for when it's `false`

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -87,6 +87,12 @@ class TestCommand : Callable<Int> {
     )
     private var debugOutput: String? = null
 
+    @Option(
+        names = ["--flatten-debug-output"],
+        description = ["Don't create subfolders or timestamps for each run. Useful for CI."]
```
felipevolpone, line 92: Let's improve this description, something like:
```
All file outputs from the test case are created in the folder without subfolders or timestamps for each run. It can be used with --debug-output. Useful for CI
```

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -161,7 +167,9 @@ class TestCommand : Callable<Int> {
                     debugOutputPath = debugOutputPath
                 )
 
-                TestDebugReporter.deleteOldFiles()
+                if(!flattenDebugOutput){
```
felipevolpone, line 170: nit: the code format seems wrong, there is no space between `if` and the `(` and between `)` and the `{`.

Fishbowler:
Thanks for the thorough review! Will rebase and address.

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
Fishbowler: Far prefer your wording, switching it to a positive stance.

Fishbowler:
I've pushed new changes and rerun the same tests as above - all three give the same results.

Rather than split into 2 separate methods, I've split off some of the noisier stuff and refactored what's there. I could've reverted the stuff I split off to the one-liner that existed for this previously, but I think that this is more comprehendable for ongoing maintenance.

felipevolpone:
@Fishbowler thanks for handling the comments and changing the code. The PR LGTM. Between today and tomorrow, I'll get the branch locally, test it and if it's all good, I'll merge it. Thanks again 🙌

felipevolpone (approved):

felipevolpone:
Great work @Fishbowler. Thank you for your contribution

PR #1751, opened 2024-05-20
chore: use correct app id in wikipedia advanced example

axelniklasson:
## Proposed Changes

copilot:summary

### DIFF
--- a/samples/wikipedia-android-advanced/auth/login.yml
+++ b/samples/wikipedia-android-advanced/auth/login.yml
@@ -1,4 +1,4 @@
-appId: org.wikimedia
+appId: org.wikipedia
 ---
 - tapOn: "More"
 - tapOn: "LOG IN.*"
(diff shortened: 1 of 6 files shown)


samples/wikipedia-android-advanced/auth/login.yml
```diff
@@ -1,4 +1,4 @@
-appId: org.wikimedia
+appId: org.wikipedia
```
amanjeetsingh150, line 1: If we changed the app id. Wouldn't we need to change the app binaries?

axelniklasson: This was incorrect, it should be `org.wikipedia` and not `org.wikimedia`. The reason this was uncaught is that these are used as subflows so I'm mainly updating this for correctness

amanjeetsingh150 (approved):

PR #1770, opened 2024-06-22
Add Gradle distribution verification to project

asos-edgeorge:
## Proposed Changes

It's a security best practice to verify the Gradle distribution used within the project to protect against a supply-chain-style attack. These have been observed in the wild previously and are easily mitigated through the use of the `distributionSha256Sum` Gradle property [^1]

This PR adds the expected Gradle distro verification for Gradle Wrapper 7.6 by calling the `gradlew wrapper` task as follows:

```sh
./gradlew wrapper --gradle-version=7.6 --distribution-type=bin --gradle-distribution-sha256-sum=7ba68c54029790ab444b39d7e293d3236b2632631fb5f2e012bb28b4ff669e4b
```

The checksums for Gradle distributions are available via Gradle's website[^2] and the above can be verified.

> [!IMPORTANT]
> Post-merging, whenever the Gradle version is updated, the above command should be repeated with the checksum for the new version

## Testing
Verified the project builds correctly and downloads Gradle Wrapper with the expected checksum

[^1]: https://www.spght.dev/articles/23-07-2023/gradle-security
[^2]: https://gradle.org/release-checksums/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,15 +11,26 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
           java-version: 8
 
-      - uses: gradle/gradle-build-action@v2
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v3
 
       - name: Test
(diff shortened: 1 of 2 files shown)


bartekpacia (commented):
Thanks for the contribution!

~~Let's also use [this GitHub action](https://github.com/gradle/actions/tree/v3?tab=readme-ov-file#the-wrapper-validation-action) for added safety. ~~ Done.

bartekpacia (approved):

PR #1771, opened 2024-06-25
Add ios 17 to start device command

vbmaarten:
## Proposed Changes

Add ios 17 to start device command

## Testing

Run the cli, and validate the output


bartekpacia (approved):
This only updates the help message.

PR #1794, opened 2024-07-12
Collect more analytics data

bartekpacia:
## Proposed Changes

Adds collection of more data about the environment, so we can make more informed decisions.

## Testing

Done locally, I'm pretty confident it's OK.

## Issues Fixed

[Notion doc](https://www.notion.so/mobile-dev/Better-analytics-in-Maestro-CLI-69517eeaa5c14e82a5ced8e35b7aa9f6?pvs=4) (internal only)

Accompanying server PR: https://github.com/mobile-dev-inc/monorepo/pull/3405

# Questions

## 1

Right now, analytics data is sent added to every request (using interceptor [`SystemInformationInterceptor`](https://github.com/mobile-dev-inc/maestro/blob/c59b550686ac315c34fb789581dd1207297755ab/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt#L507-L521)).

I think we could/should introduce a new analytics endpoint and split responsibilities.

I suggest we don't change existing logic (i.e. keep using `SystemInformationInterceptor` to attach `X-UUID`, `X-OS`, etc. to all requests).

This leads to point 2:

## 2

Right now, analytics data is sent in every request, but it's collected for BigQuery only in the `maestro/version` endpoint handler.

[The query to `maestro/version` happens on every run of the CLI](https://github.com/mobile-dev-inc/maestro/blob/dc1c0fa878b3cfde4dd3bf9de62c356c80f647cb/maestro-cli/src/main/java/maestro/cli/App.kt#L88), I think it's a good thing.

There's [some code in the monorepo](https://github.com/mobile-dev-inc/monorepo/blob/f71eb52125c06dba4aed1d8cd315e0c91c8c1a0f/mobile.dev/api/api-client/src/main/kotlin/api/client/MobiledevApiClient.kt#L269-L288) that also pings `maestro/version` endpoint. ([Endpoints setup code](https://github.com/mobile-dev-inc/monorepo/blob/a8750a326943aa6b54212ee9614f8f25b0f2e0d1/mobile.dev/api/api/src/main/java/api/server/ApiServer.kt#L295-L298))

I'd like to split responsibilities and introduce a new `maestro/analytics` endpoint.

## 3

There's already [`EnvUtils.getJavaVersion`](https://github.com/mobile-dev-inc/maestro/blob/d51400108ccfe64a156d029f184e56c1e79d3a56/maestro-cli/src/main/java/maestro/cli/util/EnvUtils.kt#L45-L57), but it shells out to run the `java -version`. Maybe I should also get rid of that and use `java.version` from JVM system properties?

**Update: I did it.**


gradle/libs.versions.toml
bartekpacia, line 1: (these funny changes are there because I autoformatted this file in IntelliJ)

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
bartekpacia, line 1: There's some `logger.debug` sprinkled in. I'll remove them before merging for once, but I had this thought: maybe Maestro CLI should accept a `--verbose` flag, which would print DEBUG logs?

(that'd be in a different PR)

igorsmotto (approved):

bartekpacia:
@amanjeetsingh150 gave verbal approval, going forward 🎉

PR #1806, opened 2024-07-17
Fix exit code for single test

rasyid7:
## Proposed Changes
change `result` to `resultSingle` for return and calculate `passed` test

## Testing
Run single maestro test
`maestro test testFile.yaml`

## Issues Fixed
since `passed` counts the number of the exit code of `resultSingle` `val passed = results.sumOf { it.first ?: 0 }`,
if the test is passed, the exit code will be 0 and failed 1.
the calculation will be incorrect. hence if we revert the value, the calculation will be correct.


bartekpacia:
Thanks, I reproduced this issue locally @ master (0610d07379698c0d157ff79a746e2328370cf8d9). Then I tested this PR locally and it does fix this issue.

bartekpacia (approved):

Issue #1221, opened 2023-07-05
[Feature Request] Device rotation


mdethlefs-ffw:
We would like to have a way to change the device rotation for our flows

axelniklasson:
Somewhat related to #565

bartekpacia:
Hey! could you please clarify this feature request?

- Do you want a Maestro command `rotate` that would rotate the device? Caveat: it would not work on physical devices. I know it's easy to do on Android emulator, using emulator console:

<details><summary>rotating AVD from shell</summary>

    ```
    $ telnet localhost 5554
    Trying ::1...
    Connected to localhost.
    Escape character is '^]'.
    Android Console: Authentication required
    Android Console: type 'auth <auth_token>' to authenticate
    Android Console: you can find your <auth_token> in
    '/Users/bartek/.emulator_console_auth_token'
    OK
    auth uHnk/fBaIEhkA6jx
    Android Console: type 'help' for a list of commands
    OK
    rotate 90
    OK
    rotate -90
    OK
    rotate -90
    OK
    rotate 90
    OK
    ```
    
    Docs: https://developer.android.com/studio/run/emulator-console
    
    </details> 


- Or maybe you want Maestro Cloud to support this feature? E.g:

    ```
    $ maestro cloud \
      --android-api-level 33 \
      --orientation landscape-left `# new hypothetical option` \
      build/app/outputs/apk/debug/app-debug.apk \
      .maestro
    ```

mdethlefs:
Hey @bartekpacia 
We want to change orientation inside an flow. We don't want to start a flow with a certain orientation.

For example our app has a videoplayer and when you rotate your phone to landscape the fullscreen automatically activates and the player has some additional UI. When changing the orientation back the fullscreen player closes.

So we basically need something like:
`- rotateHorizontal`
and 
`- rotateVertical`

bartekpacia:
Thank you for a very quick response. I see your use case now.

I did some more googling and maybe we can also change rotation of physical Android devices with `adb`. [StackOverflow link](https://stackoverflow.com/q/25864385/7009800)

As for iOS, it seems much more restricted in this area. Couldn't find anything.

Would you OK with this feature being Android only?

mdethlefs:
I mean it would be awesome if you could do it with iOS too. But we are happy for anything in this direction - so only Android would be cool already

TheAlmightyBob:
[...]

PR #1809, opened 2024-07-18
Fix iOS landscape mode

### LINKED ISSUE #565: Landscape support for iOS
iOS testing currently does not support landscape mode. (I am not sure about the current status of landscape support with Android)

As a quick test, download the samples:

```
maestro download-samples
cd ./samples
unzip sample.zip
xcrun simctl install Booted Wikipedia.app
```

And update the iOS flow to:

```
appId: org.wikimedia.wikipedia
---
- launchApp
- tapOn: "Search Wikipedia"
- inputText: "tacos"
```

Launch the simulator and rotate into landscape mode, and then run:

```
maestro test ios-flow.yaml
```

The test reports that it passes, but it does not navigate into the search screen to type into the text.

Running `maestro studio` shows a mismatch where the overlay content appears to be laid out in landscape, but the image of the app's screen is rotated 90 degrees:

Screen Shot 2023-01-03 at 10 27 16 AM

NOTE: This was discussed in a couple of slack threads, but I'm adding a ticket here so myself and anyone else interested can follow along for status updates or to be notified when support is available.
- https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1671032621327279
- https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1671642307201909

bartekpacia:
## Proposed changes

Aims to fix #565 by mapping coordinates according to current simulator orientation.

## Testing

Tested locally.

I would like to have automated tests as well, but not sure yet how to go about it.


bartekpacia:
I think I got it working:

https://github.com/user-attachments/assets/6aaec26f-9232-406a-9960-bc57a0affcdf

Current limitations:
- it still breaks if orientation is changed during flow execution

bartekpacia:
Another approach idea: instead of mapping input coordinates (e.g in tapOn), let's map output coordinates when querying the view hierarchy.

This will probably work better with Maestro Studio.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/TouchRouteHandler.swift
```diff
@@ -11,21 +11,34 @@ struct TouchRouteHandler: HTTPHandler {
...
             return AppError(type: .precondition, message: "incorrect request body provided for tap route").httpResponse
         }
         
+        let (width, height) = ScreenSizeHelper.physicalScreenSize()
+        
+        let (x, y) = switch (orientation) {
+        case .portrait: (requestBody.x, requestBody.y)
+        case .landscapeLeft: (width - requestBody.y, requestBody.x)
+        case .landscapeRight: (requestBody.y, height - requestBody.x)
+        default: fatalError("Not implemented yet")
```
amanjeetsingh150, line 29: nit: indentation

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
```diff
@@ -0,0 +1,25 @@
+import XCTest
+
+struct ScreenSizeHelper {
+    static func physicalScreenSize() -> (Float, Float) {
```
amanjeetsingh150, line 4: Would we need to account this for other endpoints like swipe?

amanjeetsingh150:
What do you think about following?

1. How would a maestro user test this? Does this unblock all the recent requests that were inbound for landscape mode testing?
2. Is android working on landscape? We don't need any change there?
3. What do you think about this configuration to cloud?
5. The scenario we discussed about mid way changing the orientation? For this lets' keep that scope out of this pr we can first roll this out and see how users react and then plan to pick that. WDYT?

maestro-ios-driver/src/main/resources/maestro-driver-ios-config.xctestrun
```diff
@@ -20,6 +20,8 @@
 		<string>maestro-driver-iosUITests</string>
 		<key>BlueprintProviderName</key>
 		<string>maestro-driver-ios</string>
+		<key>BlueprintProviderRelativePath</key>
```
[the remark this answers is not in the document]
bartekpacia: good catch, it was autogenerated, so let's keep it out.

bartekpacia:
> How would a maestro user test this? Does this unblock all the recent requests that were inbound for landscape mode testing?

I believe it unblocks ~80% of the users who complained about landscape mode.

> Is android working on landscape? We don't need any change there?

That's right, Android landscape has always worked.

> What do you think about this configuration to cloud?

I consider this a quick-fix to unblock some people, who run tests locally. AFAIK Maestro Cloud doesn't allow for handling orientation changes. Maestro doesn't have a command for that as well (see #1221).

> The scenario we discussed about mid way changing the orientation? For this lets' keep that scope out of this pr we can first roll this out and see how users react and then plan to pick that. WDYT?

I agree, yes!

### Sum up

After this PR gets merged, there will be still 3 problems:

- (Android & iOS) Both Maestro & Maestro Studio break if device is rotate mid-flow
  - **possible fix**: Don't cache deviceSize
- (iOS) breaks when simulator is "physically" in landscape, but app is locked to portrait. I consider this a rare edge case and suggest we are fine with it.
  - **possible fix**: find iOS/XCUITest APIs that allow us to detect this situation and handle accordingly
- (iOS) Maestro Studio is broken when device is in horizontal mode
  - **possible fix**: [explained here](https://github.com/mobile-dev-inc/maestro/pull/1809#issuecomment-2242859449)

But I believe this PR gets us the most value for now.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/TouchRouteHandler.swift
bartekpacia: I'm honeslty not sure what's the right indentation here, I haven't been writing much Swift recently.

I created https://github.com/mobile-dev-inc/maestro/issues/1827 to make this situation better in the future:)

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
bartekpacia: With the current approach - yes.

I admit I forgot about this. I will do that (later today)

bartekpacia:
Last night I was trying out [this approach](https://github.com/mobile-dev-inc/maestro/pull/1809#issuecomment-2242859449), but it failed.

Yet another idea:

https://github.com/mobile-dev-inc/maestro/blob/1f7bbaae44d4c1a494a8cd55899bb6b03a7d3dae/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/TouchRouteHandler.swift#L26

Maybe it's enough to just pass the current orientation there?

amanjeetsingh150:
> Maybe it's enough to just pass the current orientation there?

Sure you can check if it works. Btw heads up this is a private API so you have to check if its even available as an option.

amanjeetsingh150:
I'm not sure if we discussed this. This could also be something we do in separate PR:

Right now maestro start-device doesn't,t support start device by default in landscape. I'm not even sure if that's possible. But it would be good if user also has an option to pass orientation while starting device.

What do you think?

bartekpacia:
> > Maybe it's enough to just pass the current orientation there?
> 
> Sure you can check if it works. Btw heads up this is a private API so you have to check if its even available as an option.


I did, but it doesn't work (admittedly though I didn't try very hard). [Here](https://github.com/mobile-dev-inc/maestro/pull/1832)'s the result of that small investigation (and [here](https://github.com/mobile-dev-inc/maestro/pull/1832/commits/ea61d57518c0083a6477908ef9ddf1c4af5ec919) is what didn't work)

> Right now maestro start-device doesn't,t support start device by default in landscape. I'm not even sure if that's possible. But it would be good if user also has an option to pass orientation while starting device.

That's right, `maestro start-device` doesn't have this functionality right now. Changing virtual device rotation is very easy (, but I understand it may be very useful for users using `maestro` directly on CI (directly=not on managed CI like Maestro Cloud).

I made a separate issue for this: #1833

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
bartekpacia: Done

amanjeetsingh150 (approved):

chiefchief:
doesn't work on iPad

bartekpacia:
@chiefchief Can you tell more? Did you build Maestro yourself locally?

This is not yet released.

chiefchief:
> @chiefchief Can you tell more? Did you build Maestro yourself locally?
> 
> This is not yet released.

as i understand, after the merge of this PR, new version was released, then i ran the command 
```bash
```
which install maestro from latest release 
and the issue still exist on iPad

bartekpacia:
So this is expected. New version (1.37.7) was released before this PR was merged.

chiefchief:
> So this is expected. New version (1.37.7) was released before this PR was merged.

Got it, thank you, i'll check once new version will be released

PR #1813, opened 2024-07-19
maestro-test: fix Kotlin/Java version mismatch

bartekpacia:
## Proposed changes

To make this warning no more:

```
> Task :maestro-test:compileKotlin
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
By default will become an error since Gradle 8.0+! Read more: https://kotl.in/gradle/jvm/target-validation
Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

## Testing

Should not have any runtime behavior.

### DIFF
--- a/examples/apps/android-app/build.gradle
+++ b/examples/apps/android-app/build.gradle
@@ -28,7 +28,7 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
     kotlinOptions {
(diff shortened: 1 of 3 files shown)


igorsmotto (approved):
lol, nice catch

PR #1832, opened 2024-07-26
Another take at landscape mode

bartekpacia:


bartekpacia:
Note for the future:

### What went wrong here

The orientation returned was changing, even though the simulator was still.

For example, at the beginning of the flow, the first few calls to it `XCUIDevice.shared.orientation` return portrait, but after those first calls, it starts to returns horizontal.

That was breaking everything.

It sounds weird, but that's what I saw.

### Possible action item

Create a minimal repro app and call `XCUIDevice.shared.orientation` many times and try to reproduce it.

IMHO it's unlikely that it was a problem/bug in Maestro. I think it's a problem in XCUIDevice.shared.orientation.

Issue #1827, opened 2024-07-25
Set up ktlint and swift-format


bartekpacia:
### Aim

To avoid wasting time during PRs for mundane things like indentation and other stuff that don't matter.

- ktlint is enforced for all kotlin code in PR github actions
- swift-format is enforced for all swift code in PR github actions

### Scope

- Apply ktlint plugin to every project
  - perhaps create a convention plugin to not copy code too much
- Run both ktlint and swift-format on the whole codebase
  - add that commit to `.git-blame-ignore-revs`

This will ensure our codebase is consistent (desirable for such a big open-source project)

bartekpacia:
+ detekt, + SwiftLint

PR #1836, opened 2024-07-27
Start adding detekt to plugins {} blocks

bartekpacia:
related #1827 

Currently not sure how to proceed because of https://github.com/detekt/detekt/discussions/7489


bartekpacia:
To run detekt on all subprojects:

```
./gradlew detekt --continue --auto-correct
```

To run detekt on all subprojects with autocorrect:

```
./gradlew detekt --continue --auto-correct
```

PR #1848, opened 2024-07-30
Add `MAESTRO_CLI_NO_ANALYTICS` flag

### LINKED ISSUE #1846: CICD asks for analytics, no way to opt out

### Steps to reproduce

We are pushing a RN update in our app. We use maestro for running visual regression screenshot tests, and the branch that has the update is getting stuck when maestro asks us to opt into analytics: 
image

### Actual results

The CICD workflow gets stuck after bundling the app and running maestro, as maestro asks for us to opt into analytics:
image

### Expected results

We would like to know if we can modify the command to run maestro to automatically opt out of analytics (we do not wish to opt in at this point)

### About app

The app is closed source. The issue with maestro is happening only when running out visual regression suites via github actions CICD runners.

### About environment

The workflow runs on ubuntu, I don't know which specific version

### Logs

<details>
<summary>Logs</summary>

```
BUILD SUCCESSFUL in 11m 46s
983 actionable tasks: 983 executed

Build successful!
Starting emulator...
Waiting for emulator to start...
Installing APK on emulator...
Performing Streamed Install
Success

App installed successfully!
Waiting for the simulator to start...
===== Run E2E Attempt: 1 ====
Maestro CLI would like to collect anonymous usage data to improve the product.
Enable analytics? [Y/n] 
Error: The operation was canceled.
```

</details>

### Maestro version

1.33.1

### Anything else?

This is the step where we are installing maestro in our CICD machine:
- name: Install Maestro
        run: |
          export MAESTRO_VERSION="1.33.1"

bartekpacia:
fix #1846


maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -73,6 +77,13 @@ object Analytics {
     fun maybeAskToEnableAnalytics() {
         if (hasRunBefore) return
 
+        // Fix for https://github.com/mobile-dev-inc/maestro/issues/1846
+        if (CiUtils.getCiProvider() != null) {
```
amanjeetsingh150, line 81: Let's maybe give DISABLE_ANALYTICS_ENV_VAR false by default if its CI?

amanjeetsingh150: We only need analytics if user is sourcing from CLI

bartekpacia: Are you sure? We set `uuid` to name of the CI service if we're running on CI:

https://github.com/mobile-dev-inc/maestro/blob/21ae6eb6eae669a397174dc69bedef203384413e/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt#L191-L193

which then gets uploaded to our backend:

https://github.com/mobile-dev-inc/maestro/blob/21ae6eb6eae669a397174dc69bedef203384413e/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt#L167

I think it can be useful knowledge to know what CI service is most used*, so we would have some data if we will have to build integrations with CIs in the future.

---

*A problem here may be that we don't have UUIDs of specific CI installations. We would need to have separate "UUID" and "CI" values in the AnalyticsReport.

https://github.com/mobile-dev-inc/maestro/blob/21ae6eb6eae669a397174dc69bedef203384413e/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt#L204-L220

Otherwise analytics from various runs on CI get merged together.

amanjeetsingh150 (approved):

PR #1854, opened 2024-07-30
Add installApp command (#1665)

### LINKED ISSUE #1665: Command to install APK or APP files from a URL
# Feature Request

There should be a command like `installApp` - that allows you to install an APP or APK file from a URL

## Why it is needed

Sometimes you want to test the interaction between different apps - `installApp` would provide a convenient way to install these apps.

Right now I am downloading APKs via Chrome, and I have to toggle a bunch of security settings to get it to work.

boranfrkn:
## Proposed changes

Added “installApp” command.

## Testing

- Put an apk file in the desired directory.
- Add the command - installApp: “yourPath” to your test.

## Issues fixed

Resolve #1665

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -454,6 +454,10 @@ class IOSDriver(
         LOGGER.warn("Airplane mode is not available on iOS simulators")
     }
 
+    override fun installApk(apk: File) {
+        LOGGER.warn("Install app is not available on iOS simulators")
+    }
+
     private fun addMediaToDevice(mediaFile: File) {
         val namedSource = NamedSource(
             mediaFile.name,
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -920,6 +920,20 @@ data class ToggleAirplaneModeCommand(
     }
 }
 
+data class InstallApplicationCommand(
+    val apkPath: String? = null,
+    val label: String? = null,
+): Command {
+    override fun description(): String {
+        return label ?: "Installing $apkPath"
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+
+}
+
 internal fun tapOnDescription(isLongPress: Boolean?, repeat: TapRepeat?): String {
     return if (isLongPress == true) "Long press"
     else if (repeat != null) {
--- a/maestro-test/src/test/resources/116_install_app.yaml
+++ b/maestro-test/src/test/resources/116_install_app.yaml
@@ -1,0 +1,3 @@
+appId: com.example.app
+---
+- installApp: "maestro-client/src/main/resources/maestro-app.apk"
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -102,4 +102,6 @@ interface Driver {
     fun isAirplaneModeEnabled(): Boolean
 
     fun setAirplaneMode(enabled: Boolean)
+
+    fun installApk(apk: File)
 }
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -595,6 +595,11 @@ class Maestro(private val driver: Driver) : AutoCloseable {
         driver.setAirplaneMode(enabled)
     }
 
+    fun installApk(apkPath: String?) {
+        val apkFile = File(apkPath.orEmpty())
+        driver.installApk(apkFile)
+    }
+
     companion object {
 
         private val LOGGER = LoggerFactory.getLogger(Maestro::class.java)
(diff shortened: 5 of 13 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -454,6 +454,10 @@ class IOSDriver(
         LOGGER.warn("Airplane mode is not available on iOS simulators")
     }
 
+    override fun installApk(apk: File) {
+        LOGGER.warn("Install app is not available on iOS simulators")
+    }
+
```
bartekpacia, line 460: We want Maestro to be cross-platform wherever possible, so this won't fly. Let's add support for installing on iOS simulator as well.

We can use `xcrun` for that:

```
xcrun simctl install <device> <path-to-app>
```

for example:

```
xcrun simctl install booted build/ios/iphonesimulator/Runner.app
```

boranfrkn: Hello @bartekpacia thanks for your review! I also added for iOS.

bartekpacia:
We want this feature to also work on Maestro Cloud - so we must enforce the app binary to be inside Maestro workspace folder.

boranfrkn:
> We want this feature to also work on Maestro Cloud - so we must enforce the app binary to be inside Maestro workspace folder.

If I understand correctly, we already make it enforced. @bartekpacia 

`if (appFile != null && !appFile.exists()) throw CliError("App file does not exist: ${appFile.absolutePathString()}")`

boranfrkn:
The pipeline got an error, there seems to be no authentication key, but idk why

bartekpacia:
Don't worry about the pipeline, it's expected. It only fails for external contributors, and it's okay, because we don't want to expose our backend secret key to external people.

boranfrkn:
I see, shall we merge it as it is then?

bartekpacia:
Let's also add docs for this new command in `mobile-dev-inc/maestro-docs`

bartekpacia:
re: [this](https://github.com/mobile-dev-inc/maestro/pull/1854#issuecomment-2274908893)

It's only checking if the app file exists, right? But it has to be inside Maestro workspace folder (together with all your flows, etc), so that when it's uploaded to Maestro Cloud (MC), it can work there as well.

If the app file is outside the Maestro workspace folder, it won't get uploaded and the flow will fail on MC.

boranfrkn:
All done 🥳 @bartekpacia 
https://github.com/mobile-dev-inc/maestro-docs/pull/68

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -69,6 +68,7 @@ class CloudInteractor(
     ): Int {
         if (appBinaryId == null && appFile == null) throw CliError("Missing required parameter for option '--app-file' or '--app-binary-id'")
         if (!flowFile.exists()) throw CliError("File does not exist: ${flowFile.absolutePath}")
+        if (appFile?.toPath()?.startsWith(flowFile.toPath().absolute())?.not() == true) throw CliError("App file does not exist: ${flowFile.absolutePath}")
```
bartekpacia, line 71: Let's also print the relative path, I can image in might help in debugging. No strong opinions though.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -920,6 +920,20 @@ data class ToggleAirplaneModeCommand(
     }
 }
 
+data class InstallApplicationCommand(
```
bartekpacia, line 923: let's rename it to `InstallAppCommand` (for consistency with existing `LaunchAppCommand`)

maestro-test/src/test/resources/116_install_app.yaml
```diff
@@ -0,0 +1,3 @@
+appId: com.example.app
+---
+- installApp: "maestro-client/src/main/resources/maestro-app.apk"
```
bartekpacia, line 3: ultra-nit: let's add newline

bartekpacia (approved):
Overall this LGTM, but I'm holding off because there are some concerns internally.

maestro-test/src/test/resources/116_install_app.yaml
boranfrkn: done

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
bartekpacia: Cool. Your call👍

bartekpacia:
Het @boranfrkn, we like this feature and will merge the PR soon, but first we need to make some changes to our backend platform.

Tentative plan is to release this in Maestro v1.39. Thanks for understanding and sorry about the delay!

bartekpacia:
The test failure is a flake.

It already occured previously [here](https://github.com/mobile-dev-inc/maestro/pull/1995#issuecomment-2322927088).

<details><summary>Screenshot</summary>

screenshot-❌-1725268971709-(Fill out form)


</details> 

Rerunning.

boranfrkn:
I see, thanks a lot Bartek. its passed 👍

tokou:
FYI @bartekpacia https://github.com/boranfrkn/maestro/pull/1

boranfrkn:
Hey @Fishbowler @bartekpacia any update here?

Fishbowler:
You only merged the change a few minutes ago, give us a chance! 😂

PR #1858, opened 2024-07-31
Ignore all failures while uploading analytics data

### LINKED ISSUE #1850: Simulator runtimes not found. java.lang.NullPointerException: list must not be null

### Steps to reproduce

Not clear how to reproduce, could be related to a new way xcode manages/installs runtimes.

### Actual results

```
❯ maestro --version
Exception in thread "main" java.lang.NullPointerException: list must not be null
        at maestro.cli.util.IOSEnvUtils.getSimulatorRuntimes(IOSEnvUtils.kt:19)
        at maestro.cli.analytics.Analytics.maybeUploadAnalyticsAsync(Analytics.kt:147)
        at maestro.cli.AppKt.main(App.kt:99)
```

### Expected results

no error

### About app

not relevant

### About environment

- Maestro version: 1.37.4
- Xcode version: 16 beta
- macOS version: 15.0
- Installation method: both Homebrew and curl

### Logs

<details>
<summary>Logs</summary>
❯ curl -Ls "https://get.maestro.mobile.dev/" | bash
* Create distribution directories...
* Downloading...
######################################################################## 100.0%
* Checking archive integrity...
* Extracting archive...
* Copying archive contents...
* Cleaning up...

Installation was successful!
Please open a new terminal OR run the following in the existing one:

    export PATH="$PATH":"$HOME/.maestro/bin"

Then run the following command:

    maestro

Welcome to Maestro!
❯ which maestro
/Users/filipe/.maestro/bin/maestro
❯ maestro
Exception in thread "main" java.lang.NullPointerException: list must not be null
        at maestro.cli.util.IOSEnvUtils.getSimulatorRuntimes(IOSEnvUtils.kt:19)
        at maestro.cli.analytics.Analytics.maybeUploadAnalyticsAsync(Analytics.kt:147)
        at maestro.cli.AppKt.main(App.kt:99)
-------
rm -rf ~/.maestro
rm -rf ~/.local/state/maestro
brew tap mobile-dev-inc/tap
brew install maestro
------
❯ maestro
Maestro CLI would like to collect anonymous usage data to improve the product.
Enable analytics? [Y/n] Y
Usage data collection enabled. Thank you!
Exception in thread "main" java.lang.NullPointerException: list must not be null
        at maestro.cli.util.IOSEnvUtils.getSimulatorRuntimes(IOSEnvUtils.kt:19)
        at maestro.cli.analytics.Analytics.maybeUploadAnalyticsAsync(Analytics.kt:147)
        at maestro.cli.AppKt.main(App.kt:99)
❯ maestro --version
Exception in thread "main" java.lang.NullPointerException: list must not be null
        at maestro.cli.util.IOSEnvUtils.getSimulatorRuntimes(IOSEnvUtils.kt:19)
        at maestro.cli.analytics.Analytics.maybeUploadAnalyticsAsync(Analytics.kt:147)
        at maestro.cli.AppKt.main(App.kt:99)
❯ which maestro
/opt/homebrew/bin/maestro
</details>

### Maestro version

1.37.4

Homebrew

### Anything else?

Slack thread https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1722335707236339

After updating to latest version, Maestro is unable to access simulator runtimes, resulting in a NullPointerException. This occurs even when trying to check the Maestro version 🤯

```
❯ xcrun simctl runtime list
== Disk Images ==
-- iOS --
iOS 17.4 (21E213) - 821D4CFB-369B-47B5-A913-B35EBED48A54 (Ready)

Total Disk Images: 1 (6.7G)
```

`/Library/Developer/CoreSimulator/Profiles/Runtimes`
Is empty
Under user folder there is a "runtimeMap" `/Users/<myusername>/Library/Developer/CoreSimulator/RuntimeMap.plist`
Could be something with xcode beta 16

Using `sudo maestro` works

```
❯ sudo maestro --version
1.37.4
❯ sudo which maestro
/opt/homebrew/bin/maestro
```

bartekpacia:
**Could not reproduce** #1850, but this should fix it anyway.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
@@ -158,10 +158,8 @@ object Analytics {
             // We don't care that much about analytics to bug user about it.
             return
         } catch (e: Exception) {
-            // This is less fine. Don't crash, but ask the user to report this.
-            println("Exception ocurred while uploading analytics. This is not a fatal issue, but we ask you to report it.")
-            println(e.message)
-            println(e.stackTraceToString())
+            // This is also fine. We don't want to bug the user.
+            // See discussion at https://github.com/mobile-dev-inc/maestro/pull/1858
         }
     }
 
--- a/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
@@ -160,6 +160,7 @@ object Analytics {
         } catch (e: Exception) {
             // This is also fine. We don't want to bug the user.
             // See discussion at https://github.com/mobile-dev-inc/maestro/pull/1858
+            return
         }
     }
 
--- a/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
@@ -113,49 +113,55 @@ object Analytics {
      * Uploads analytics if there was a version update.
      */
     fun maybeUploadAnalyticsAsync() {
-        if (!hasRunBefore) {
-            logger.trace("First run, not uploading")
-            return
-        }
-
-        if (analyticsDisabledWithEnvVar) {
-            logger.trace("Analytics disabled with env var, not uploading")
-        }
-
-        if (!analyticsState.enabled) {
-            logger.trace("Analytics disabled with config file, not uploading")
-            return
-        }
-
-        if (!uploadConditionsMet) {
-            logger.trace("Upload conditions not met, not uploading")
-            return
-        }
-
-        val report = AnalyticsReport(
-            uuid = analyticsState.uuid,
-            freshInstall = !hasRunBefore,
-            cliVersion = EnvUtils.CLI_VERSION?.toString() ?: "Unknown",
-            os = EnvUtils.OS_NAME,
-            osArch = EnvUtils.OS_ARCH,
-            osVersion = EnvUtils.OS_VERSION,
-            javaVersion = EnvUtils.getJavaVersion().toString(),
-            xcodeVersion = IOSEnvUtils.xcodeVersion,
-            flutterVersion = EnvUtils.getFlutterVersionAndChannel().first,
-            flutterChannel = EnvUtils.getFlutterVersionAndChannel().second,
-            androidVersions = AndroidEnvUtils.androidEmulatorSdkVersions,
-            iosVersions = IOSEnvUtils.simulatorRuntimes,
-        )
-
-        logger.trace("Will upload analytics report")
-        logger.trace(report.toString())
-
         try {
+            if (!hasRunBefore) {
+                logger.trace("First run, not uploading")
+                return
+            }
+
+            if (analyticsDisabledWithEnvVar) {
+                logger.trace("Analytics disabled with env var, not uploading")
+            }
+
+            if (!analyticsState.enabled) {
+                logger.trace("Analytics disabled with config file, not uploading")
+                return
+            }
+
+            if (!uploadConditionsMet) {
+                logger.trace("Upload conditions not met, not uploading")
+                return
+            }
+
+            val report = AnalyticsReport(
+                uuid = analyticsState.uuid,
+                freshInstall = !hasRunBefore,
+                cliVersion = EnvUtils.CLI_VERSION?.toString() ?: "Unknown",
+                os = EnvUtils.OS_NAME,
+                osArch = EnvUtils.OS_ARCH,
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -113,49 +113,55 @@ object Analytics {
...
         } catch (e: ConnectException) {
-            // This is fine. We don't care that much about analytics to bug user about it.
+            // This is fine. The user probably doesn't have internet connection.
+            // We don't care that much about analytics to bug user about it.
             return
+        } catch (e: Exception) {
+            // This is less fine. Don't crash, but ask the user to report this.
+            println("Exception ocurred while uploading analytics. This is not a fatal issue, but we ask you to report it.")
+            println(e.message)
+            println(e.stackTraceToString())
```
amanjeetsingh150, line 164: Why do we want to print it on cli?

I don't think this would have any meaning for user. For us yes then maybe log this in maestro.log or some log file instead.

bartekpacia: We want to make sure analytics data is flowing to us. If we will just swallow the exception, we will never learn if we silently break sth in analytics.

amanjeetsingh150: I don't really see value of showing this to user. Maybe there is a better option to provide this observability then?

Users don't really care about analytics, we do. So, maybe we opt another way for this observability WDYT?

bartekpacia: ~~I think the best solution would be to connect Sentry (or any other log reporting service)?~~ That would be a lot of work though.

Actually, we already have an `ErrorReporter`. I just took a look at BigQuery, and it looks pretty solid. ([query link](https://console.cloud.google.com/bigquery?ws=!1m7!1m6!12m5!1m3!1sperf-dev-289002!2sus-west1!3s750c7a13-e1c7-422d-9d27-f4507a1faf3f!2e1) - internal only)

The problem is that [we don't wrap our whole app with the ErrorReporter](https://github.com/mobile-dev-inc/maestro/blob/9e0f64257ec576022a85829cb5d6a5b2d03ea046/maestro-cli/src/main/java/maestro/cli/App.kt#L106), so not all errors are reported. This is a separate issue though.

For now I agree with you, let's remove this message for the good of our users.

amanjeetsingh150: Yes you can create a separate issue for this. Its definitely useful though!

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -113,49 +113,53 @@ object Analytics {
...
             ApiClient(EnvUtils.BASE_API_URL).sendAnalyticsReport(report)
             updateAnalyticsState()
         } catch (e: ConnectException) {
-            // This is fine. We don't care that much about analytics to bug user about it.
+            // This is fine. The user probably doesn't have internet connection.
+            // We don't care that much about analytics to bug user about it.
             return
+        } catch (e: Exception) {
+            // This is also fine. We don't want to bug the user.
+            // See discussion at https://github.com/mobile-dev-inc/maestro/pull/1858
```
amanjeetsingh150, line 162: return?

amanjeetsingh150 (approved):

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
bartekpacia: it's end of the function anyway, but yeah, I added the return.

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -113,48 +113,53 @@ object Analytics {
...
-        try {
             ApiClient(EnvUtils.BASE_API_URL).sendAnalyticsReport(report)
             updateAnalyticsState()
         } catch (e: ConnectException) {
-            // This is fine. We don't care that much about analytics to bug user about it.
+            // This is fine. The user probably doesn't have internet connection.
+            // We don't care that much about analytics to bug user about it.
+            return
+        } catch (e: Exception) {
+            // This is also fine. We don't want to bug the user.
```
herval, line 161: you don't even want to log this?

quietly dropping exceptions sounds like a bad idea...

amanjeetsingh150: We are gonna do it on separate PR as discussed above.

bartekpacia: created #1858 to track this

Issue #1853, opened 2024-07-30
Unable to run parallel/sharded tests on Android


simon-gilmurray:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

It always fail except for on one device when running Android in parallel. Things I've tried:
- Creating devices from Maestro CLI and using them
- Using Android Studio created devices
- Calling a basic command `maestro test --shards 3 maestro/`
- Passing in explicit device names `maestro --device 'emulator-5554,emulator-5556,emulator-5558' test -s 3 maestro/ `

### Actual results

Only one device seems to run, 2/3 of the tests appear to fail straight away with 0s:

<img width="929" alt="Screenshot 2024-07-30 at 11 57 37" src="https://github.com/user-attachments/assets/9aecd386-1e6f-4752-a180-51e4dbd8c3b2">

Logs:

```
[ERROR] m.cli.runner.TestSuiteInteractor - Failed to complete flow
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
    at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:271)
    at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:252)
    at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:165)
    at maestro_android.MaestroDriverGrpc$MaestroDriverBlockingStub.deviceInfo(MaestroDriverGrpc.java:634)
...
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/[0:0:0:0:0:0:0:1]:7006
Caused by: java.net.ConnectException: Connection refused
```
[maestro.log](https://github.com/user-attachments/files/16430851/maestro.log)

### Expected results

Tests should run in parallel, as they do on iOS which is working fine

### About app

React Native

### About environment

macOS Sonoma 14.4.1 M2 Pro

### Logs

<details>
<summary>Logs</summary>

```
[maestro.log](https://github.com/user-attachments/files/16430871/maestro.log)
```

</details>

### Maestro version

1.37.4

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

### Anything else?

Slack thread:

https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1722341761341199

bartekpacia:
[...]

PR #1867, opened 2024-08-01
Fix Android Parallel/Shards Race Condition

rasyid7:
## Proposed changes

I think this is because of the shards race condition
So by adding a delay will fix the race condition

## Issues fixed
https://github.com/mobile-dev-inc/maestro/issues/1853 Unable to run parallel/sharded tests on Android

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -74,6 +74,7 @@ object MaestroSessionManager {
         val heartbeatFuture = executor.scheduleAtFixedRate(
             {
                 try {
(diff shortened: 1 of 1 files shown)


bartekpacia:
Hey @rasyid7, thanks a lot for this PR. I tested it and it does fix the issue. That said I don't want to merge this as is, because the code is already complex. I'd prefer we find a root cause and fix that instead. Would you be up to do that?

rasyid7:
hi @bartekpacia , surely this is just hotfix. but if there's someone able to find the root cause, feel free to create a PR.
my capability is not good enough and maybe will take more time 🙏

bartekpacia:
I see. To unblock users, I'm going to merge it as it, because it does fix the issue.

That said, I will try to get back as soon as I have some free cycles, and get it fixed properly.

fyi @amanjeetsingh150

Issue #1585, opened 2023-11-10
[1.34.1] MaestroDriverStartupException$IOSDriverTimeoutException


nicolas-acl:
**Describe the bug**
The issue occurs when running a Flow.

**To Reproduce**
Steps to reproduce the behavior :

1. Start the simulator
2. Run `maestro test ./flows` // (flows is a folder with all my .yml)
3. The Flow fails at Maestro iOS driver did not start up in time.

**Expected behavior**
The error is not occurs

**Screenshots**
maestro.MaestroDriverStartupException$IOSDriverTimeoutException: Maestro iOS driver did not start up in time
	at maestro.drivers.IOSDriver.awaitLaunch(IOSDriver.kt:476)
	at maestro.drivers.IOSDriver.open(IOSDriver.kt:65)
	at maestro.Maestro$Companion.ios(Maestro.kt:585)
	at maestro.cli.session.MaestroSessionManager.createIOS(MaestroSessionManager.kt:309)
	at maestro.cli.session.MaestroSessionManager.createMaestro(MaestroSessionManager.kt:154)
	at maestro.cli.session.MaestroSessionManager.access$createMaestro(MaestroSessionManager.kt:49)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:82)
	at maestro.cli.session.MaestroSessionManager$newSession$session$1.invoke(MaestroSessionManager.kt:81)
	at maestro.cli.db.KeyValueStore.withExclusiveLock(KeyValueStore.kt:37)
	at maestro.cli.session.SessionStore.withExclusiveLock(SessionStore.kt:74)
	at maestro.cli.session.MaestroSessionManager.newSession(MaestroSessionManager.kt:81)
	at maestro.cli.session.MaestroSessionManager.newSession$default(MaestroSessionManager.kt:58)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:136)
	at maestro.cli.command.TestCommand.call(TestCommand.kt:46)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
	at picocli.CommandLine.access$1200(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
[...]

PR #1885, opened 2024-08-03
Add E2E test for iOS

bartekpacia:
follow up to #1837


bartekpacia:
@amanjeetsingh150 said that he'll fix it by fixing #1585

bartekpacia:
Status: `test-ios` workflow is still failing and I wasn't able to fix it. That said this PR contains logging improvements I made in the process, and the E2E test of the standalone XCTest Runner. I think it's worth merging it and solving the core problem later.

PR #1902, opened 2024-08-09
Add start session command to increase speed of tests

### LINKED ISSUE #1096: [Feature Request] Make Maestro faster by giving the option to swip APK install/uninstall
When running on lower end Android devices, Maestro can take a long time to start and to finish the test, since it's installing the APKs + uninstalling at the end

For instance, we use Flashlight to generate e2e performance reports based on several e2e test iterations and we recommend heavily the use of Maestro (https://docs.flashlight.dev/test/maestro)
However the time bottleneck of running multiple test iterations is those install/uninstall

It'd be nice to have the option to skip APK install/uninstall to speed things up
For instance
```bash
maestro test test.yaml --skipApkInstall
``` 
**I'd be happy to raise a PR for that!**

We could also add a `maestro setup` task that does the necessary setup before running a test

Almouro:
## Proposed changes

feat:Add start session command to increase speed of tests

When running on lower end Android devices, Maestro can take a long time to start and to finish the test, since it's installing the APKs + uninstalling at the end (can take ~20s on low end devices)

For instance, we use Flashlight to generate e2e performance reports based on several e2e test iterations and we [recommend heavily the use of Maestro](https://docs.flashlight.dev/test/maestro).
However the time bottleneck of running multiple test iterations is those install/uninstall.

This PR introduces a `start-session` command:
- you start it in one tab and it should do the setup necessary, including uninstalling/installing APKs
- then if you do `maestro test` in another tab, setup won't be done again, and the tests will be fast ⚡

## How it works

I was surprised by how easy it was to do and kudos for that 👏 You guys did a fantastic job there ♥️
The `start-session` command allows us to keep an active session open. Subsequent test command will find the session active in `SessionStore` and use it.
Then `openDriver` will be set to false, and the `AndroidDriver` won't run the `open()` function which does all the heavy setup.
Sadly this isn't the case for iOS, so we don't have it working out of the box.

## Potentially breaking shards? 

⚠️ I had to remove 2 lines of code: I have to be honest, I have to be honest, I didn't manage to understand their purpose 😓. So I'd love some help from @sdfgsdfgd or @igorsmotto  to know why they're needed. 🙏

For reference, I've run a "shard" test with `./maestro test -s 2 tests` including some `launchApp` and it worked for me. Likely I've missed a use-case.

## Testing

- Tested locally + in our AWS Device farm CI
- Also added an e2e test script

| Normal behavior | With active session ⚡ |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/36147a16-6130-4a99-ae00-77062b05915e" /> | <video src="https://github.com/user-attachments/assets/ee8c3c41-e9b7-470f-a9f9-a4670b5a4d52" /> | 

## Issues fixed

Fixes #1096

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -194,9 +194,6 @@ class AndroidDriver(
         launchArguments: Map<String, Any>,
         sessionId: UUID?,
     ) {
-        if(!open) // pick device flow, no open() invocation
-            open()
-
         if (!isPackageInstalled(appId)) {
             throw IllegalArgumentException("Package $appId is not installed")
         }
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -89,6 +89,7 @@ jobs:
             ./download_apps
             ./install_apps
             ./run_tests
+            ./run_start_session_test
 
       - name: Upload ~/.maestro artifacts
         uses: actions/upload-artifact@v4
--- a/e2e/run_start_session_test
+++ b/e2e/run_start_session_test
@@ -1,0 +1,69 @@
+#!/usr/bin/env sh
+
+set -eu
+
+# The point of this test is to ensure that the 'maestro start-session' works correctly
+# - Running it should reinstall Maestro APKs on device
+# - Running 'maestro test' while a 'maestro start-session' is open should not reinstall Maestro APKs
+
+command -v maestro >/dev/null 2>&1 || { echo "maestro is required" && exit 1; }
+
+[ "$(basename "$PWD")" = "e2e" ] || { echo "must be run from e2e directory" && exit 1; }
+
+get_maestro_apk_install_time() {
+  adb shell dumpsys package dev.mobile.maestro | grep "lastUpdateTime" | awk -F= '{print $2}'
+}
+
+start_maestro_session() {
+  LOG_FILE=$(mktemp)
+  MAX_TIMEOUT=60
+  START_TIME=$(date +%s)
+
+  # Start the Maestro session in the background and redirect output to logs.txt
+  echo "Starting Maestro session..."
+  maestro --device emulator-5554 start-session > "$LOG_FILE" 2>&1 &
+  MAESTRO_START_SESSION_PID=$!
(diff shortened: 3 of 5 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -194,9 +194,6 @@ class AndroidDriver(
         launchArguments: Map<String, Any>,
         sessionId: UUID?,
     ) {
-        if(!open) // pick device flow, no open() invocation
-            open()
```
Almouro, line 198: As mentioned, I had to remove 2 lines of code: I have to be honest, I didn't understand their purpose. 
So I'd love some help from @sdfgsdfgd or @igorsmotto  to know why they're needed. 🙏

Here the issue is that since `maestro test` has found an active session, it didn't "open" the Android driver since it doesn't need to. 
But this condition would force it to reopen when using `launchApp` which defeats the purpose of `maestro start-session`

bartekpacia:
What a great pull request @Almouro! 🎉

Thanks a ton! I'll take a look at it the next week.

bartekpacia:
Do I understand correctly that this is Android-only atm?

Almouro:
@bartekpacia indeed, probably a bit more work would be needed on iOS, sinc
e the iOS driver isn't the one responsible for starting the XCTestInstaller it seems

bartekpacia:
Thanks for confirming! Feature parity across platforms is very important for us, so we cannot merge it as is.

rtharston:
@bartekpacia, if someone were to look into implementing the iOS side of this, is the PR still in an acceptable state on the Android side of things?

Fishbowler:

sdfgsdfgd:
@Almouro 


 Hey, I finally dug back through this with current upstream.

  You were right about removing this from launchApp():

  if (!open) // pick device flow, no open() invocation
      open()

  open is only local state on that AndroidDriver object. With start-session, a later maestro test process can have open == false while the driver is already running in the session process.

  So launchApp() calling open() there restarts/reopens instrumentation and defeats session reuse.

  The extra missing piece was ports: shards now use dynamic driver ports. So “active session exists for this device” is not enough; reconnecting has to use the same driver port as the live session, or it skips startup but talks to the wrong gRPC port.

  I opened a fix that keeps your launchApp() point and adds the missing session-port reuse:

  https://github.com/mobile-dev-inc/Maestro/pull/3307

Issue #1801, opened 2024-07-14
Allow for use of latest Java and Gradle (without affecting compatibility)


bartekpacia:
We want to:
- keep support for running Maestro on JDK8
- use latest JDK & Gradle during development

This is possible to do. There are 2 great blogpost from Jake Wharton:
- [https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea](https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea)
- [Build on latest Java, test through lowest Java](https://jakewharton.com/build-on-latest-java-test-through-lowest-java)

Using Java 21 requires bumping Gradle [to a version that supports it](https://docs.gradle.org/current/userguide/compatibility.html). This in turn will require bumping Android Gradle Plugin to a version 8.x, which will require at least Java 17 for building. But it will have no impact on users of Maestro CLI.

### Action items

- update GitHub Actions to compile on latest JDK 21 and Gradle 8.7
- use Gradle toolchains to run tests on JDK 8, 11, 17 and 21

bartekpacia:
These 2 dependencies might be problematic though, if their latest versions require JDK 17:

https://github.com/mobile-dev-inc/maestro/blob/dc1c0fa878b3cfde4dd3bf9de62c356c80f647cb/gradle/libs.versions.toml#L53-L54

Fishbowler:
Maestro's now at Java 17 🎉

PR #1909, opened 2024-08-12
Enforce compatibility with JDK 8 when compiling on later JDKs

bartekpacia:
follow-up of #1880

aimed at #1801

With these changes I aim to:
- make it possible to build Maestro on all Java 11 and higher. Compilation is faster with each release, and [has other advantages](https://jakewharton.com/build-on-latest-java-test-through-lowest-java)
- keep minimum Java required to _run_ Maestro at 8. Note: for building, JDK11 will be the new minimum.

Notes:
- bumped Kotlin 1.8.0 -> 1.8.22 to [fix this bug](https://youtrack.jetbrains.com/issue/KT-52815/Compiler-option-Xjdk-release-fails-to-compile-mixed-projects)


bartekpacia:
Removing `SensitivityWatchEventModifier.HIGH` from `FileWatcher.kt` does not seem to introduce any noticeable delay to what we had before.

If we do discover some regression in the future though, we can try using [kfswatch](https://github.com/irgaly/kfswatch) instead of depending on hidden JDK APIs.

.github/workflows/test-e2e.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: zulu
-          java-version: 8
+          java-version: 11
```
igorsmotto, line 116: > - make it possible to build Maestro on all Java 11 and higher. Compilation is faster with each release, and [has other advantages](https://jakewharton.com/build-on-latest-java-test-through-lowest-java)
> - keep minimum Java required to run Maestro at 8. Note: for building, JDK11 will be the new minimum.

Following your comment, this test-e2e/test should work with java-version: 8 right? Otherwise, this means that maestro is unable to run at 8.

bartekpacia: > this test-e2e/test should work with java-version: 8 right?

This is correct. (This is actually one of the things I wanted to accomplish with the `test-e2e.yaml` - to always verify we work on Java 8)

Now, this change (in `test-cloud` job) is actually a mistake, it should've been 8.

On Android (the `test-local` job) it's currently not 8, because none of the "setup Android & start emulator" GitHub Actions support installing an older SDK Command-Line Tools release (newer require either Java 11 or 17). [I'm hoping this will change soon](https://github.com/Malinskiy/action-android/issues/86).

igorsmotto: So that means we can only test if `maestro cloud` works on java 8? Actually using maestro locally it's still TBD?

bartekpacia: > Actually using maestro locally it's still TBD? 

It already works as well, but the workaround I used is to compile on Java 8, and then run on Java 11 (because Android tools require at least Java 11). But now we will not be able to compile anymore on Java 8, so we have to find a way run on Java 8.

> Perhaps we could use devices from Maestro Cloud and connect them remotely using adb and therefore use the maestro cli?

We could, but this sounds a bit complex (but I'd love to discuss more if you think it makes sense here). For now though I'd prefer to go with the simplest possible solution - find a way to use old Android Command-Line Tools+Java 8 and run emulator on GitHub Actions.

bartekpacia:
Constantly changing between JDK versions is really annoying. I gave this PR a good test (run `maestro test samples/wikipedia-android-advanced/run-test.yml` from `maestro download-samples`.

In addition to that, I'm pretty confident that `-Xjdk-release` is reliable and we can depend on it to guarantee compatibility with JDK 8.

Going forward with the merge.

PR #1935, opened 2024-08-23
Mark all initFlow-related constructs as @Deprecated

### LINKED ISSUE #1921: `initFlow` should be deprecated and removed
Code in question:

https://github.com/mobile-dev-inc/maestro/blob/d23a25f669736dcc915f60080be66773c55b8ca1/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt#L146-L170

citing @igorsmotto on Slack:

> IIRC that's legacy, and we removed from the docs so people won't use it. But we didn't remove because there might be some customer still using it, so we don't want to break for them
> 
> The use case itself is the same one after being reintroduced as onFlowStart and onFlowComplete that you can find in the docs

Plan:
- Add warning whenever `initFlow` is provided
- Remove related code from `maestro` and `monorepo`
- Afte some time (2 weeks? a month?) remove support completely. If someone complains, guide them to either use an older version (not recommended), or migrate to `onFlowStart`

bartekpacia:
fix #1921


bartekpacia:
Actually, I think it's OK to remove this API, since the deprecation warning has been there for more than a year now.

It'll simplify Orchestra code a bit, which is a welcome change.

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -95,18 +86,10 @@ object TestRunner {
                 // Restart the flow if anything has changed
                 if (commands != previousCommands || initFlow != previousInitFlow) {
                     ongoingTest = thread {
-                        // If previous init flow was successful and there were no changes to the init flow,
```
bartekpacia, line 98: This variable was not used anywhere.

amanjeetsingh150 (approved):

PR #1937, opened 2024-08-23
Remove all mentions of idb

bartekpacia:
Maestro doesn't use `idb` in any way since [v1.31.0](https://github.com/mobile-dev-inc/maestro/blob/main/CHANGELOG.md#1310).


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -217,20 +216,6 @@ object MaestroSessionManager {
         }
     }
 
-    private fun isIOS(host: String?, port: Int?): Boolean {
```
bartekpacia, line 220: This function was marked by IntelliJ as unused

PR #1945, opened 2024-08-24
Add filename to env

### LINKED ISSUE #1616: Make the filename of the current test available as environment variable
Hi, is there a way to get the running test's filename?
one of the use case is I want to name the recording as same as the name of the file so that when I run multiple tests I don't have to worry for overwritten recording name

```
- startRecording: ${FILENAME}
```

tokou:
## Proposed changes

- Set `FILENAME` env value to the flow file's name
- This happens only in the context of the test and does not alter the system env

## Testing

I updated 2 of the integration tests to use the `FILENAME` env var and check their correct naming

Run ./gradlew :maestro-test:test
Run ./gradlew :maestro-orchestra:test
Run ./gradlew :maestro-orchestra-models:test

## Issues fixed

Fixes #1616

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
@@ -23,6 +23,7 @@ import org.slf4j.LoggerFactory
 import java.io.File
 import java.nio.file.Path
 import kotlin.concurrent.thread
+import maestro.orchestra.util.Env.withDefaultEnvVars
 
 /**
  * Knows how to run a single Maestro flow (either one-shot or continuously).
@@ -52,7 +53,7 @@ object TestRunner {
 
         val result = runCatching(resultView, maestro) {
             val commands = YamlCommandReader.readCommands(flowFile.toPath())
-                .withEnv(env)
+                .withEnv(env.withDefaultEnvVars(flowFile.nameWithoutExtension))
 
             YamlCommandReader.getConfig(commands)?.name?.let {
                 aiOutput = aiOutput.copy(flowName = it)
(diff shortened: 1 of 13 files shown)


bartekpacia:
Hey, thanks a ton Tarek. Overall this LGTM!

Two small things:
- let's change name to `MAESTRO_FILENAME`
- how does it behave in subflows? Is the name of the root flow printed? Is there any difference between JS engines?

PS In the longer term, I think we'll need to have a list of all Maestro-provided env variables, so people will not override them when they pass their own `MAESTRO_FILENAME` vars.

bartekpacia (approved):

bartekpacia:
@tokou I rebased with master (to make github checks ✅) and renamed the variable.

If you have some free time (and will!), tests would be great to see:)

tokou:
I noticed with the tests that we populate the env var with subflow file names in the subflows. I think it makes more sense for it to always be the top level flow name. So I fixed it and added some tests.

bartekpacia:
> I think it makes more sense for it to always be the top level flow name.

Agreed, thanks a ton!

After the conflicts are resolved this is good to go.

bartekpacia:
One last ask, if I may - could we keep `withEnv()` the way it was before? To minimize changed lines + make code more explicit.

```
.withUserProvidedEnv()
.withSystemEnv(...)
```

makes it obvious that "system env vars" (like `MAESTRO_FILENAME`) take precedence – which is what we want.

Lmk what you think. You already did a lot and I don't want you to lose enthusiasm :)

bartekpacia:
On the other hand, this is good enough. Thanks a lot!

PR #1946, opened 2024-08-25
Add optional field to commands

### LINKED ISSUE #1778: [Feature Request] SoftAsserts
I would like to be able to specify one by one in the commands if I want that in the case of a fail in that specific command (or flow or whatever) the test ends or not. Right now to option affects to all the workflow.

For example, If I have a process where I have to fill a form that takes 5 minutes to complete, and after that I have a view where I get a resume of everything, something like:
- name
- surname
- age
- date
- etc

I would like that the test ends if there is some failure in the fill form stage, but not in the resume view, where I validate all the fields one by one. For example, if I have a failure in the name validation, it will only show me that (cause the other validations didn't happen). I would have to fix it and replay to test only to discover that it also fails in the surname and so on.

If I have soft asserts there, in the first run, after failing in the name validation, it would continue and would tell me that it also fails in the surname and any other field. This would reduce the number of executions and would help in automation pipelines.

### LINKED ISSUE #1969: The `Command` interface should have a `String label` property
Since all Maestro commands now accept a `label`, it makes sense to add it to

https://github.com/mobile-dev-inc/maestro/blob/ba6927385c9d6305aef7eae5d21b49f190ee538a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt#L31-L39

so we never forget about it.

### Additional context

Mentioned by @amanjeetsingh150 in https://github.com/mobile-dev-inc/maestro/pull/1906/files/ba045503e5eb0c46e26e6279d6130c1a25954311#r1735133326

tokou:
## Proposed changes

- Adds an `optional` flag to all commands without exception
- Adds a catch at the level of `executeCommand` in Orchestra to throw a `CommandSkipped` exception in case of an error
- Cleans up some compiler warnings

I'm not sure of the direction taken here and have some questions:
- Is it really pertinent to add `optional` to every single command? I feel like some command never fail 🤔 
- Should we get rid of the `optional` in `ElementSelector`? Or is there a case where it makes sense to have it?
- I used the `CommandSkipped` exception but is the skipped state the same as an optional command failing? Should I create a separate state?

## Testing

- I added some `optional: true` commands to `076_optional_assertion.yml`
- `./gradlew test`

## Formats

I ran `./maestro-cli/build/install/maestro/bin/maestro test maestro-test/src/test/resources/076_optional_assertion.yaml`

```
Running on emulator-5556

 ║
 ║  > Flow
 ║
 ║    ⚠️ Scrolling DOWN until id: not_found is visible. (warned)
 ║    ⚠️ Assert that false is true (warned)
 ║    ⚠️ Assert that id: not_found is visible (warned)
 ║    ⚪️ Assert that (Optional) "Button" is visible (skipped)
 ║
```

## Issues fixed

Fixes #1778 
Fixes #1969


bartekpacia:
Hey @tokou, thanks a ton! I love the work done here!

> I used the CommandSkipped exception but is the skipped state the same as an optional command failing? Should I create a separate state?

I was also thinking about it recently. I think a new state `CommandWarned` should be added, because we do want to differentiate between the optional `runFlow` and optional `tapOn` - the former should be "skipped", while the latter should "warn". wdyt?

maestro-studio/web/package-lock.json
bartekpacia, line 1: this seems unrelated

maestro-studio/server/src/main/java/maestro/studio/DeviceService.kt
```diff
@@ -93,10 +93,7 @@ object DeviceService {
                 call.respond(response)
             }
         }
-        routing.static("/screenshot") {
-            staticRootFolder = SCREENSHOT_DIR.toFile()
-            files(".")
-        }
+        routing.staticFiles("/screenshot", SCREENSHOT_DIR.toFile())
```
bartekpacia, line 96: This is unrelated to this PR.

maestro-orchestra/src/test/java/maestro/orchestra/MaestroCommandSerializationTest.kt
```diff
@@ -78,7 +79,8 @@ internal class MaestroCommandSerializationTest {
                 "retryIfNoChange" : false,
                 "waitUntilVisible" : true,
                 "longPress" : false,
-                "label" : "My TapOnPoint"
+                "label" : "My TapOnPoint",
+                "optional" : false
```
bartekpacia, line 83: please use trailing commas (they haven't been used in this repo before but we're trying to migrate to using them everywhere)

(this applies to all code in this PR)

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroConfig.kt
```diff
@@ -26,13 +26,13 @@ data class MaestroConfig(
 }
 
 data class MaestroOnFlowComplete(val commands: List<MaestroCommand>) {
-    fun evaluateScripts(jsEngine: JsEngine): MaestroOnFlowComplete {
+    fun evaluateScripts(@Suppress("UNUSED_PARAMETER") jsEngine: JsEngine): MaestroOnFlowComplete {
```
bartekpacia, line 29: please don't add code that is not required for this PR, such as those `@Suppress`es

if you want to do cleanup, thanks! but do so in another PR

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -200,6 +205,7 @@ class Orchestra(
     private fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?): Boolean {
         val command = maestroCommand.asCommand()
 
+        @Suppress("DEPRECATION")
```
bartekpacia, line 208: ditto

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -286,6 +286,7 @@ class WebDriver(val isStudio: Boolean) : Driver {
         swipe(direction, durationMs)
     }
 
+    @Suppress("UNUSED_PARAMETER")
```
bartekpacia, line 289: ditto

maestro-cli/src/main/java/maestro/cli/report/HtmlTestSuiteReporter.kt
```diff
@@ -115,7 +115,7 @@ class HtmlTestSuiteReporter : TestSuiteReporter {
                   }
                 }
               }
-              script(src = "https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/js/bootstrap.min.js", content = "")
+              script(src = "https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/js/bootstrap.min.js") {}
```
bartekpacia, line 118: Is this required for this PR's functionality?

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -246,11 +246,11 @@ object DeviceService {
     fun isDeviceAvailableToLaunch(deviceName: String, platform: Platform): Device.AvailableForLaunch? {
         return if (platform == Platform.IOS) {
             listIOSDevices()
-                .filterIsInstance(Device.AvailableForLaunch::class.java)
+                .filterIsInstance<Device.AvailableForLaunch>()
```
bartekpacia, line 249: what are those changes for?

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -190,7 +190,7 @@ class TestCommand : Callable<Int> {
             val chunkPlans = plan.flowsToRun
                 .withIndex()
                 .groupBy { it.index % shards }
-                .map { (shardIndex, files) ->
+                .map { (_, files) ->
```
bartekpacia, line 193: If you want to do cleanup, please do so in another PR.

bartekpacia (changes requested):
I think it's ok to add `optional` to all existing commands. We already have `label` argument on all commands.

Also, we won't spend time deciding if a specific command can fail or not.

maestro-orchestra/src/test/java/maestro/orchestra/MaestroCommandSerializationTest.kt
tokou: This is JSON which does not support trailing commas I think 🤔 
I used trailing commas in all Kotlin code. Is that what you meant?

bartekpacia: Yes, I meant Kotlin code. Sorry for the mistake.

bartekpacia:
Restarting tests because of a flake:

<details><summary>Details</summary>


Screenshot 2024-09-02 at 16 58 42


</details> 

Also occurred in:
- https://github.com/mobile-dev-inc/maestro/pull/1995#issuecomment-2322927088
- https://github.com/mobile-dev-inc/maestro/pull/1854#issuecomment-2324388637

bartekpacia:
That [workflow run failure](https://github.com/mobile-dev-inc/maestro/actions/runs/10667970365/job/29571313868?pr=1946) is a false positive from #2005, safe to ignore for now.

bartekpacia (approved):

bartekpacia (approved):

bartekpacia:
@tokou This gets messed up with `optio

Take this flow as an example:

```yaml
# This flow is to ensure that commands with optional flag are not failing the flow.
appId: com.example.example
tags:
    - android
    - passing
---
- launchApp:
    clearState: true
- assertTrue:
    condition: ${ false }
    label: Warn
    optional: true
- tapOn:
    id: non-existent-id
    optional: true
- doubleTapOn:
    id: non-existent-id
    optional: true
- longPressOn:
    id: non-existent-id
    optional: true
- copyTextFrom:
    id: non-existent-id
    optional: true
- launchApp:
    appId: non.existend.app.id
    optional: true
- tapOn:
    id: non-existent-id
    repeat: 3
    delay: 500
    optional: true
- scrollUntilVisible:
    element:
      id: non-existent-id
    optional: true
```

Results in:

```yaml
./maestro-cli/build/install/maestro/bin/maestro test e2e/workspaces/demo_app/commands_optional_tournee.yaml

Running on emulator-5554

 ║
 ║  > Flow
 ║
 ║    ✅  Launch app "com.example.example" with clear state
 ║    ⚠️ Warn (warned)
 ║    ✅  Tap on (Optional) id: non-existent-id
 ║    ✅  Double tap on (Optional) id: non-existent-id
 ║    ✅  Long press on (Optional) id: non-existent-id
 ║    ⚠️ Copy text from element with (Optional) id: non-existent-id (warned)
 ║    ⚠️ Launch app "non.existend.app.id" (warned)
 ║    ✅  Tap x3 on (Optional) id: non-existent-id
 ║    ⚠️ Scrolling DOWN until (Optional) id: non-existent-id is visible. (warned)
 ║
```

bartekpacia:
### Possible breaking change investigation

Removing `optional` from Selectors could have an unintended consequence of breaking flows that use `optional` in selectors at deep levels using `childOf`.

I created the following flow to test this hypothesis:

```yaml
appId: com.example.example
tags:
    - android
    - passing
---
- launchApp:
    clearState: true
- tapOn: Nesting Test
- assertVisible:
    id: level-0
- assertVisible:
    id: level-0
    containsChild:
      id: level-1
      containsChild:
        id: level-2
        containsChild:
          id: level-3
          optional: true
```

and the following hierarchy:

```
level-0 \
  level-1 \
    level-2
```

**without level-3**

The above flow **fails** on latest Maestro v1.38.1, which is not what I expected. I expected it to pass, given that `optional` should work on selector level. But it doesn't.

bartekpacia:
Same goes for `rightOf` with `optional: true` (and presumably others). This fails, whereas I'd expect it to pass (i.e. be skipped)

```yaml
# This flow demonstrates how removal of "optional" property from Selector-level will break a rare case of nested selectors.
appId: com.example.example
tags:
    - android
    - passing
---
- launchApp:
    clearState: true
- tapOn: Nesting Test
- assertVisible:
    id: level-0
- assertVisible:
- 
    id: level-0
    containsChild:
      id: level-1
      containsChild:
        id: level-2
      rightOf:
        text: nothing
        optional: true
```

nicolaskopp:
possible bug: "optional" parameter cannot be parameterized, in other words, value has to be true or false, but cannot be a variable that sets optional to true or false on a central place.

Fishbowler:
That's not a bug, @nicolaskopp - that'd be a new feature request. I'd take a look through and see if it's already logged somewhere.

nicolaskopp:
> That's not a bug, @nicolaskopp - that'd be a new feature request. I'd take a look through and see if it's already logged somewhere.

my bad, documentation is a bit [interpretable](https://docs.maestro.dev/advanced/parameters-and-constants#accessing-variables-from-the-shell) on this one, it raises the impression that parameters work basically everywhere. Thanks for looking into it!

PR #1952, opened 2024-08-26
Platform specific appId

### LINKED ISSUE #1472: Pass different, platform-specific `appId`s in a single flow
### Use case

My app is cross-platform, and its UI and accessibility labels are almost 100% the same across Android and iOS. The only thing that differs is the application ID (Android) and bundle ID (iOS), for historical reasons.

I'd like to be able to run the same flow on both platforms easily. Currently I have to change the `appId` manually inside the flow, or [pass appId as env var](https://maestro.mobile.dev/troubleshooting/frequently-asked-questions), or split my single flow into 3 flows: 2 platform-specific flows and 1 common flow, containing all the test steps.

I think each of these solutions add a bit too much friction for something so basic.

### Proposal

```yaml
appId:
  android: pl.baftek.discoverrudy
  ios: pl.baftek.discoverRudy
---
- clearState
- launchApp
- tapOn: Increment
# ...
```

tokou:
## Proposed changes

- Introduces `MaestroAppId` model containing appIds for each platform
- Tricky parts
  - Ensuring backwards compatibility in deserializing commands
  - Finding a way to have only the current appId in descriptions without passing around Platform too deeply

## Testing

- Added integration tests
- Ran a flow locally and inspected logs and console output

## Issues fixed

Fixes #1472

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -144,7 +144,7 @@ class TestCommand : Callable<Int> {
     private fun isWebFlow(): Boolean {
         if (!flowFile.isDirectory) {
             val config = YamlCommandReader.readConfig(flowFile.toPath())
-            return Regex("http(s?)://").containsMatchIn(config.appId)
+            return Regex("http(s?)://").containsMatchIn(config.appId.web ?: "")
         }
 
         return false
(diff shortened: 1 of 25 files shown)


tokou:
The E2E test works fine for me locally. Maybe we should retrigger it? 🤔

Fishbowler:
Code LGTM, but I've not tested it.
I can see real value in this, even in the simple Wikipedia app examples.

Fishbowler:
Argh, this has been stale too long - @tokou are you able to rebase?

Fishbowler:

PR #1954, opened 2024-08-27
Add back CLI --platform option

tokou:
## Proposed changes

- I saw that the `--platform` option existed in the past and has been deprecated in [`1.11.0`](https://github.com/mobile-dev-inc/maestro/blob/main/CHANGELOG.md#1110---2022-10-26)
- I was wondering of the reason behind it as I think it might be very useful in local development scenarios
  - I often have one iOS and one Android device running as I develop a new feature
  - We often write single flows that run on both platforms
  - I'm often working on fixing code for a flow on a single platform at a time (generally Android then iOS)
  - I find myself having to always select a device while using `--platform` would make things much simpler for me
- The way I suggest adding it back here is in an optional fashion
  - No `--platform` works as before
  - `--platform` gets ignored in case `--device` is set
  - `--platform` set and a single device from that platform is connected: We select that device without prompt
  - `--platform` and multiple devices from that platform are connected: We prompt the user with only the relevant platform devices displayed
- If there's an issue with reverting it at the `App` level that I did not notice. May I suggest adding it to the `TestCommand` only in that case?

## Testing

- Run `--platform` alongside `--device`
- Run `--platform` with a single device from each platform running
- Run `--platform` with multiple devices from different platforms running

## Issues fixed

Linked #314


tokou:
I'm more and more thinking that this should be added to the Test command only 🤔  wdyt @bartekpacia ?

bartekpacia:
There's also `maestro hierarchy` command which should also work with `--udid` and `--platform` options.

But overall yes, I agree that `--platform` (and `--udid`) shouldn't be top-level options on the `maestro` command. But to keep consistency, let's continue doing it.

In a separate PR you might want do add those options to individual commands (and deprecate those top-level options, so we can remove them later).

PR #1958, opened 2024-08-27
GitHub Actions: install Android SDK and set up AVD manually

bartekpacia:
Both GitHub Actions ([ReactiveCircus'](https://github.com/ReactiveCircus/android-emulator-runner) and [Malinskiy's](https://github.com/Malinskiy/action-android) for setting Android emulator don't work for us, because they don't allow for specifying a concrete version of Android Command-line Tools to install. And we nee to install Android CLT v8, because it's the latest version that works on Java 8, which is Maestro's minimal support target.


bartekpacia:
whoooah it works, nice

PR #1961, opened 2024-08-28
test if e2e test actually works

bartekpacia:
misunderstood pull_request_target


bartekpacia:
it doesn't.

fixed in #1963

PR #1962, opened 2024-08-28
Fix samples

bartekpacia:
The samples were outdated. The flows in the samples don't pass.

This PR attempts to:
- update the samples so all tests pass
- automate updating the samples

In very near future I want to:
- hook up samples into our e2e testing pipeline


bartekpacia:
Stuff missing:
- `platform` specifier in the flow frontmatter. Workaround: separate `android` and `ios` directories

Issue #1914, opened 2024-08-14
An unknown network error occurred while communicating with the XCUITest server


gtokman:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

We're running Maestro on CI on a `macos-14` runner (GitHub actions) and are experiencing XCUITest not starting the test. I added some logs of the failure below.

### Actual results

The server is not starting and running the tests on CI.

### Expected results

We are expecting it to start correctly.

### About app

React Native - 0.74.5 - iOS 

### About environment

Xcode 15.4, arm64

### Logs

<details>
<summary>Logs</summary>

Running on iPhone 15 Pro - iOS 17.5 - B23095CC-E2ED-4301-A857-BCC36C176ACB
 > Flow
Run flow...
ℹ️ Retrying connection to the XCUITest server for 1...
ℹ️ Retrying connection to the XCUITest server for 2...
ℹ️ Retrying connection to the XCUITest server for 3...
ℹ️ Retrying connection to the XCUITest server for 4...
ℹ️ Retrying connection to the XCUITest server for 5...
Error: ⚠️ Error: An unknown network error occurred while communicating with the XCUITest server. If the issue persists, consider raising a GitHub issue with the error message and any available logs for further assistance.
⚠️ Error: An unknown network error occurred while communicating with the XCUITest server. If the issue persists, consider raising a GitHub issue with the error message and any available logs for further assistance.
Error: Request for http://[::1]:7001/deviceInfo failed, because of XCUITest server got crashed/exit, body: {"errorMessage":"An unknown network error occurred while communicating with the XCUITest server. If the issue persists, consider raising a GitHub issue with the error message and any available logs for further assistance.","code":"network-error"}
kotlinx.coroutines.JobCancellationException: BlockingCoroutine is cancelling; job=BlockingCoroutine{Cancelling}@35a3f96
❌ Error: BlockingCoroutine is cancelling
Caused by: NetworkError(errorResponse=Request for http://[::1]:7001/deviceInfo failed, because of XCUITest server got crashed/exit, body: {"errorMessage":"An unknown network error occurred while communicating with the XCUITest server. If the issue persists, consider raising a GitHub issue with the error message and any available logs for further assistance.","code":"network-error"})
	at xcuitest.XCTestDriverClient.handleExceptions(XCTestDriverClient.kt:280)
[...]

PR #1974, opened 2024-08-29
Reland iOS landscape mode support

bartekpacia:
Context for the reviewer:

- This PR is an attmept to reland #1809
- #1809 was reverted in #1916 because of a critical breakage in #1914
- The cause of critical breakage [was this](https://github.com/mobile-dev-inc/maestro/pull/1974#issuecomment-2321215588), and it's fixed in this PR
- additionally, I found a bigger problem #2031 which I don't think is worth solving (1) right now and (2) in this specific PR


bartekpacia:
The problem is that [`XCUIDevice.shared.orientation`](https://developer.apple.com/documentation/xctest/xcuidevice/1619046-orientation) returns [`UIDeviceOrientation.unknown`](https://developer.apple.com/documentation/uikit/uideviceorientation/unknown) when the iOS simulator is first started. This effectively makes it impossible to query the iOS simulator for its orientation.

**Plot twist**

When the iOS simulator's orientation is manually changed (screenshot below), then `[`XCUIDevice.shared.orientation`]` starts returning correct value:

Screenshot 2024-08-30 at 14 47 04

`[`XCUIDevice.shared.orientation`]` also starts returning correct value when it's assigned a value:

```swift
XCUIDevice.shared.orientation = .portrait
```

**Possible fix 1**

If `XCUIDevice.shared.orientation` returns [`UIDeviceOrientation.unknown`](https://developer.apple.com/documentation/uikit/uideviceorientation/unknown), let's assume that orientation is [`UIDeviceOrientation.portrait`](https://developer.apple.com/documentation/uikit/uideviceorientation/portrait) .

**Possible fix 2**

Introduce a new `setOrientation` command that will set orientation programatically, as shown above. Educate users that they have to call that command to change their app to a different orientation.

bartekpacia:
[I asked a question on StackOverflow](https://stackoverflow.com/q/78932288/7009800)

bartekpacia:
Good thing is that (from my testing) all iOS (and iPadOS) simulator always startup in portrait orientation, so defaulting to `portrait` when `XCUIDevice.shared.orientation` returns `unknown` seems like a good-enough workaround.

Screenshot 2024-09-09 at 10 43 18

bartekpacia:
Demo:

https://github.com/user-attachments/assets/9f6c1e99-ae56-4fbb-9a76-463e3ea46dd9

herval:
> The problem is that [`XCUIDevice.shared.orientation`](https://developer.apple.com/documentation/xctest/xcuidevice/1619046-orientation) returns [`UIDeviceOrientation.unknown`](https://developer.apple.com/documentation/uikit/uideviceorientation/unknown) when the iOS simulator is first started. This effectively makes it impossible to query the iOS simulator for its orientation.
> 
> **Plot twist**
> 
> When the iOS simulator's orientation is manually changed (screenshot below), then `[XCUIDevice.shared.orientation]` starts returning correct value:
> 
> Screenshot 2024-08-30 at 14 47 04
> `[XCUIDevice.shared.orientation]` also starts returning correct value when it's assigned a value:
> 
> ```swift
> XCUIDevice.shared.orientation = .portrait
> ```
> 
> **Possible fix 1**
> 
> If `XCUIDevice.shared.orientation` returns [`UIDeviceOrientation.unknown`](https://developer.apple.com/documentation/uikit/uideviceorientation/unknown), let's assume that orientation is [`UIDeviceOrientation.portrait`](https://developer.apple.com/documentation/uikit/uideviceorientation/portrait) .
> 
> **Possible fix 2**
> 
> Introduce a new `setOrientation` command that will set orientation programatically, as shown above. Educate users that they have to call that command to change their app to a different orientation.

how about option 3 - if it returns unknown, set it to portrait (the difference from that to option 1 is it would correct the behavior from there on, so if there's anything else that requires device orientation, it'd receive the correct value)

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -43,25 +43,26 @@ class Maestro(
...
-
-    private var screenRecordingInProgress = false
-
-    fun deviceName(): String {
-        return driver.name()
-    }
+    val deviceName: String
+        get() = driver.name()
+
+    private var _cachedDeviceInfo: DeviceInfo? = null
```
herval, line 49: aren't these two (this and the original code) functionally the same? My understanding is this is what `by lazy` does behind the scenes

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -160,8 +163,10 @@ class Maestro(
     }
 
     fun swipeFromCenter(swipeDirection: SwipeDirection, durationMs: Long) {
+        val deviceInfo = deviceInfo()
```
herval, line 166: this doesnt seem to use the cached info - is that the intent?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -529,3 +507,14 @@ class IOSDriver(
         private const val SCREEN_SETTLE_TIMEOUT_MS: Long = 3000
     }
 }
+
+private fun Double.asPercentOf(total: Int): Int {
```
herval, line 511: what's the advantage of this over the regular private functions from before?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
```diff
@@ -0,0 +1,47 @@
...
+        let screenSize = springboardApp.frame.size
+        return (Float(screenSize.width), Float(screenSize.height))
+    }
+
+    private static func actualOrientation() -> UIDeviceOrientation {
+        let orientation = XCUIDevice.shared.orientation
+        if orientation == .unknown {
+            // If orientation is "unknown", we assume it is "portrait" to
+            // work around https://stackoverflow.com/q/78932288/7009800
+            return UIDeviceOrientation.portrait
```
herval, line 16: what would be the implications of setting it to portrait as a side-effect here, when it's unknown?

maestro-client/src/main/java/maestro/Maestro.kt
bartekpacia: Yes.

If we use the cached version, then swipe will completely break if the device is rotated mid-flow (or when running `maestro test` in `--continuous` mode).

maestro-client/src/main/java/maestro/Maestro.kt
bartekpacia: It is.

I changed it away from using `lazy` because I needed `_cachedDeviceInfo` to be a `var` instead of `val` for $reasons which are no longer true because I changed the code and forgot to bring it back.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
bartekpacia: As of now, I'm not aware of any. See ["good things"](https://github.com/mobile-dev-inc/maestro/pull/1974#issuecomment-2337502676) above.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
bartekpacia: Conciseness, mainly.

If I were to keep using `widthPercentToPoint(percent: Double)` I'd have to refactor it to `widthPercentToPoint(percent: Double, width: Double)` which becomes a bit long and repetitive.

bartekpacia:
### Updated caveats

Consider [this older list of caveats](https://github.com/mobile-dev-inc/maestro/pull/1809#issuecomment-2249917209) to be out-of-date and not true.

Here's the actual list of caveats, as of now:

- (iOS) breaks when simulator is "physically" in landscape, but app is locked to portrait. I consider this a rare edge case and suggest we are fine with it.
  - **possible fix**: find iOS/XCUITest APIs that allow us to detect this situation and handle accordingly
- (iOS) Maestro Studio always displays the device in vertical orientation.
- (iOS) Taps made with Maestro Studio don't work (likely cause: coordinates aren't translated correctly).

PR #1949, opened 2024-08-25
Update graaljs

tokou:
## Proposed changes

- Update GraalJs
- Enable host access and class lookup

## Testing

- Update integration test 102 to add running the sample script

## Issues fixed

Fixes #1388

PR #1996, opened 2024-08-31
Enable class lookup in Graaljs

### LINKED ISSUE #1388: Update GraalJS engine to latest version
Currently the version used seems to be 22.0.0

Check:
https://github.com/mobile-dev-inc/maestro/blob/5fcec04d7d52e0134a934605082a92829e7e6981/gradle/libs.versions.toml#L29

As you can see here, a lot of cool stuff has been implemented: https://github.com/oracle/graaljs/blob/master/CHANGELOG.md

Take a look here too:
https://github.com/oracle/graaljs/blob/master/docs/user/JavaScriptCompatibility.md

If I try to run the following code with the current version it doesn't work, there are some things I would like to try, I don't even know if they work with the latest current version.
```js
var BigDec = Java.type('java.math.BigDecimal');
var bd = new BigDec("0.1");
console.log(bd.add(bd).toString());
```

Update the version to the latest.

tokou:
## Proposed changes

- I wanted to try and revert the GraalJS version but keep the other changes that seem to be already compatible with the older version
- These changes allow for users to load platform (i.e. Java) classes into JS like in the example 
- This seems to satisfy the request in the opened issue

## Testing

- All tests run fine locally. Let's see on the CI

## Issues fixed

Linked #1949
Fixes #1388


bartekpacia:
Hey again Tarek, thank you for this PR :)

Overall I love it, but I'm afraid there are some security concerns related to how this feature would affect Maestro Cloud (I'm double checking this with our backend guys and will get back).

Merging this as-is would allow for arbitrary code execution on Maestro Cloud's macOS machines (i.e. break the sandbox). Someone could write in their Maestro JS file `var File = Java.type('java.io.File')`, (or call `Runtime.getRuntime()` and then `rm -rf` or similar) and do nasty things.

tokou:
Yeah I understand the security risk @bartekpacia 
It's something we should consider
We can maybe make this option configurable through an env var or something?

bartekpacia:
Yes, let's guard this new functionality with env vars:

### Host access

```kotlin
allowHostAccess(HostAccess.ALL)
```

Call only when `MAESTRO_CLI_DANGEROUS_GRAALJS_ALLOW_HOST_ACCESS` is set (to any value, e.g. `1` or `true`).

### Host 

```kotlin
allowHostClassLookup { true }
```

Call only when `MAESTRO_CLI_DANGEROUS_GRAALJS_ALLOW_HOST_CLASS_LOOKUP` is set (to any value, e.g. `1` or `true`).

bartekpacia:
@tokou fyi I'm happy to merge this feature (assuming it'll be guarded with the env vars, as described above)

PR #1997, opened 2024-08-31
Fix evaluate script bug

### LINKED ISSUE #1788: Using integer value from JavaScript `outputs` results in a deserialization error
**Describe the bug**
When passing paramaters, outputs are passed as strings so this does not make possible to use speed and timeout values as parametric in scrollUntilVisible command.

**To Reproduce**
I would like to pass speed and timeout value for scrollUntilVisible command from a .js file called "contants.js". 

1. Start the simulator
2. Run `maestro test ...`
3. The Flow fails at parsing file and gives this error:

  ```
  Cannot deserialize value of type int from String "${output.speed.slow_speed}": not a valid int value
  ```

**Expected behavior**

I need to pass `int` values as parameter.

**Workspace (if applicable)**

**.maestro/test.yaml**

```yaml
appId: appId
---
- runScript: constants.js
- launchApp
- scrollUntilVisible:
    element:
      id: ${output.favorites.imageViewVariantPickerArrow}
    direction: DOWN
    speed: ${output.speed.slow_speed}
    timeout: ${output.speed.long_timeout}
```

**.maestro/constants.js**

```js
output.speed = {
  slow_speed: 50,
  default_speed: 70,
  fast_speed: 80,
}
```

boranfrkn:
## Proposed changes

Fix evaluate script bug

fixes #1788

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -95,7 +95,7 @@ data class SwipeCommand(
 data class ScrollUntilVisibleCommand(
     val selector: ElementSelector,
     val direction: ScrollDirection,
-    val scrollDuration: Long,
+    val scrollDuration: String = DEFAULT_SCROLL_DURATION,
     val visibilityPercentage: Int,
     val timeout: Long = DEFAULT_TIMEOUT_IN_MILLIS,
     val centerElement: Boolean,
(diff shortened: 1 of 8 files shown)


bartekpacia:
Hey @boranfrkn, thanks a lot.

Could you also add a test to verify the fix?

boranfrkn:
Hey @bartekpacia I think it's okay, what you think?

bartekpacia (approved):
This looks great, thanks a ton!

PR #2007, opened 2024-09-02
debug maestro-e2e-output not being present when tests fail

bartekpacia:
The artifacts stopped appearing.


bartekpacia:
Looks like it's caused by https://github.com/mobile-dev-inc/maestro/pull/2007. Sigh, GitHub Actions, the hopeless abomination.

PR #2011, opened 2024-09-03
Make MaestroCommand toString more concise

tokou:
## Proposed changes

- Make `MaestroCommand#toString` way shorter and thus all logs and test results more readable
- Basically only keep the arg name for the single non null `Command` inside the `MaestroCommand`

## Testing

- Added unit tests

## Issues fixed

None


bartekpacia:
Oh my, this is awesome, and something I wanted to fix for quite some time!

Thanks!

PR #2018, opened 2024-09-05
Add changelog for the new, not-yet-released version

bartekpacia:


axelniklasson (approved):

bartekpacia:
@tokou fyi: the changelog test if failing

bartekpacia:
@tokou Ah, it's failing because I retroactively changed the changelog for v1.38.1. So it's all fine.

bartekpacia:
I don't like that changelog must now be written in a quite rigid way, to make sure it displays correctly in the CLI.

I believe a link to the latest version should suffice.

tokou:
> I don't like that changelog must now be written in a quite rigid way, to make sure it displays correctly in the CLI.

@bartekpacia I back tested it with a lot of the older versions and it still shows something decent
Basically, it takes only the first level bullet points and ignores anything else, and changelog will always have those (and they're the most meaningful info)

bartekpacia:
Yeah, I thought about this more and I agree with you.

PR #2021, opened 2024-09-05
Commonize Gradle build logic with convention plugins

bartekpacia:
Source of inspiration: [now in android app](https://github.com/android/nowinandroid/tree/5a382e3a45e82e11260f56772aac600489168baa/build-logic)


bartekpacia:
Blocker: this requires Gradle 8

PR #2027, opened 2024-09-07
Fix input text speed on Android

### LINKED ISSUE #2005: `inputText` sometimes repeats individual characters

### Steps to reproduce

This is a flake. It sometimes occur in [our E2E action](https://github.com/mobile-dev-inc/maestro/blob/a9575c5b1ce8ee13d549449e6887e5a1ec4caae4/.github/workflows/test-e2e.yaml).

### Actual results

So far I noticed 3 occurences of this:
- https://github.com/mobile-dev-inc/maestro/pull/1995#issuecomment-2322927088
- https://github.com/mobile-dev-inc/maestro/pull/1854#issuecomment-2324388637
- https://github.com/mobile-dev-inc/maestro/pull/1946#issuecomment-2324950805

### Expected results

`inputText` always inputs expactly the text specified.

### About app

It's our [demo_app](https://github.com/mobile-dev-inc/demo_app).

### About environment

The environment on [our GitHub Action](https://github.com/mobile-dev-inc/maestro/blob/main/.github/workflows/test-e2e.yaml).

### Logs

<details>
<summary>Logs</summary>

```
```

</details>

### Maestro version

1.38.1

### Anything else?

_No response_

tokou:
## Proposed changes

- Group key codes from input text by state of modifiers
- Call `UiDevice#pressKeyCodes` with bunches of keyCodes instead of `UiDevice#pressKeyCode` with a single keyCode at a time
- Remove the `Thread.sleep(75)`

The speedup is amazing 🤩 

## Testing

- Went from 54s to 17s to input 445 chars of Lorem Ipsum 

| Before | After |
|-------|-------|
| input_text_old | input_text |

- Used this flow to compare on an API 34 and an API 25 emulator
```yml
appId: com.google.android.apps.messaging
name: Input text
env:
  INPUT: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
---
- launchApp:
    clearState: true

- tapOn:
    id: "com.google.android.apps.messaging:id/secondary_action_button"
    optional: true
- tapOn:
    id: "com.google.android.apps.messaging:id/(start_chat_fab|start_new_conversation_button)"
- inputText: "911"
- tapOn:
    id: "com.google.android.apps.messaging:id/(contact_name|recipient_text_view)"
- tapOn:
    id: "com.google.android.apps.messaging:id/compose_message_text"

- startRecording: input_text
    optional: true
- inputText: ${INPUT}
- extendedWaitUntil:
    visible: ${INPUT}
    timeout: 60_000
- stopRecording
    optional: true
```
- I had to change the timeout here to 60s as it took longer than that to input all the text
https://github.com/mobile-dev-inc/maestro/blob/6ac16c21bb3157fcd9dc06632bf5984cf1d474e7/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt#L72 

## Issues fixed

Fixes #2005


bartekpacia:
Hey @tokou! Thanks a ton for taking a stab and fixing this very annoying problem.

The speedup is amazing indeed, but I'm not sure if we want it, given that we want Maestro to behave like a normal user. Anyway, we could make this new behavior the default, and add a configuration option for `inputText` (for command and workspace level) in the future if someone requests the old way of how `inputText` worked (i.e. entering text character-by-character).

I like all the changes you made in this PR, but I'm curious whether you know what's the change that actually fixed the problem? Was it the removal of that `Thread.sleep(75)`?

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -403,63 +409,87 @@ class Service(
...
-                '?'.code -> keyPressShiftedToEvents(uiDevice, KEYCODE_SLASH)
-                '~'.code -> keyPressShiftedToEvents(uiDevice, KEYCODE_GRAVE)
-                '_'.code -> keyPressShiftedToEvents(uiDevice, KEYCODE_MINUS)
-            }
+    private fun setText(text: String) = text
+        .map { it.toKeyCodeAndMeta() }
+        .filterNotNull()
+        .fold(emptyList<Pair<List<Int>, Int>>()) { acc, elem ->
+            if (acc.isEmpty() || acc.last().second != elem.second) acc + listOf(listOf(elem.first) to elem.second)
+            else acc.dropLast(1) + listOf(acc.last().let { (it.first + listOf(elem.first)) to it.second })
```
bartekpacia, line 417: Let's add a comment explaining what this code does and/or simplify it.

bartekpacia (approved):

bartekpacia:
Oh, and let's also add an entry to CHANGELOG.md

Leland-Takamine:
Nice work on this! Some context on the delays added in the original implementation: We were seeing some issues where sometimes characters would be skipped. This was difficult to reproduce consistently, but we should ensure that this does not regress before making this the default behavior.

tokou:
@bartekpacia I added the comment. Hope it clears things up.
I also redid testing with the 3 cases : old, just removing the sleep and sending bunched up key presses

| Old (00:54) | No sleep (00:24) | Bunch keys (00:17) |
|-----|-------|-------|
| input_text_old | input_text_no_sleep | input_text_bunch |


@Leland-Takamine 
> We were seeing some issues where sometimes characters would be skipped.

I think this is due to the rate of commands sent to `UiDevice`. That's why the bunching up of commands through `pressKeyCodes` could be better than just blasting n calls to `pressKeyCode`

The only potential issue is that the documentation of `pressKeyCodes` states that it's more intended for multiple key presses at the same time instead of consecutive key presses 🤔 

> pressKeyCodes
> Presses one or more keys. For example, you can simulate taking a screenshot on the device by pressing both the power and volume down keys.

— https://developer.android.com/reference/androidx/test/uiautomator/UiDevice#pressKeyCodes(int[],int)

In practice, it seems to be working fine though.

tokou:
@bartekpacia Unfortunately, after adding the test on the demo app, I got some random failures when running it on my emulator. I think it's in part because of the demo app being a Flutter app that makes the issue happen more often.

amanjeetsingh150:
Hey @tokou is this PR ready for review? I see that you observed some random failures, let me know if you need help on this.

Fishbowler:

PR #2032, opened 2024-09-09
Stop checking in binary artifacts in git

### LINKED ISSUE #1822: Remove iOS XCUITest runner artifacts from the repo
Every commit that changes those files (e.g. running `./maestro-ios-xctest-runner/build-maestro-ios-runner.sh`) increases size of our git repo.

I suggest we gitignore those files. (And make sure we still keep packaging them for distribution)

bartekpacia:
fix #1822


bartekpacia:
Note: this will likely break [the `maestro@head` formula](https://github.com/mobile-dev-inc/homebrew-tap/blob/master/Formula/maestro%40head.rb)

Fishbowler:
Rebased.

@amanjeetsingh150 I think you've been near here recently. Got thoughts?

PR #2071, opened 2024-09-30
fix: optional commands in subFlow

amanjeetsingh150:
## Proposed changes

Optional command now throw command warned exception. This code was missing for handling sub flows which caused a regression on cloud.

## Testing

Locally and cloud

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -265,7 +265,8 @@ data class TapOnElementCommand(
 ) : Command {
 
     override fun description(): String {
-        return label ?: "${tapOnDescription(longPress, repeat)} on ${selector.description()}"
+        val optional = if (optional) "(Optional) " else ""
+        return label ?: "${tapOnDescription(longPress, repeat)} on $optional${selector.description()}"
     }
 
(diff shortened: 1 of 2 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -265,7 +265,8 @@ data class TapOnElementCommand(
 ) : Command {
 
     override fun description(): String {
-        return label ?: "${tapOnDescription(longPress, repeat)} on ${selector.description()}"
+        val optional = if (optional) "(Optional)" else ""
+        return label ?: "${tapOnDescription(longPress, repeat)} on $optional ${selector.description()}"
```
axelniklasson, line 269:
```suggestion
        val optional = if (optional) "(Optional) " else ""
        return label ?: "${tapOnDescription(longPress, repeat)} on $optional${selector.description()}"
```

axelniklasson: Nit: to avoid having two spaces in a row when the command is not optional we can do it like this

axelniklasson (approved):

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
amanjeetsingh150: Makes sense! I'll do that

PR #2073, opened 2024-10-02
fix: don't allow negative swipe duration

axelniklasson:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -3135,6 +3135,42 @@ class IntegrationTest {
         )
     }
 
+    @Test
+    fun `Case 118 - Scroll until view is visible - no negative values allowed`() {
+        // Given
+        val commands = readCommands("118_scroll_until_visible_negative")
+        val expectedDuration = "40"
+        val expectedTimeout = "20000"
+        val info = driver { }.deviceInfo()
+
+        val elementBounds = Bounds(0, 0 + info.heightGrid, 100, 100 + info.heightGrid)
+        val driver = driver {
+            element {
+                id = "maestro"
+                bounds = elementBounds
+            }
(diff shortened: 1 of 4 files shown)


maestro-test/src/test/resources/118_scroll_until_visible_negative.yaml
```diff
@@ -0,0 +1,8 @@
+appId: com.example.app
+---
+- scrollUntilVisible:
+    element:
+      id: "maestro"
+    direction: DOWN
+    speed: 110
```
axelniklasson, line 7: Setting speed > 100 yields a negative duration so this is intended

igorsmotto (approved):

maestro-test/src/test/resources/118_scroll_until_visible_negative.yaml
amanjeetsingh150: Hmmm, might be some change that we have setting on duration for this maestro release.

I believe this one: 

https://github.com/mobile-dev-inc/maestro/pull/1997/files

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -3135,6 +3135,42 @@ class IntegrationTest {
         )
     }
 
+    @Test
+    fun `Case 118 - Scroll until view is visible - no negative values allowed`() {
```
amanjeetsingh150, line 3139: Nice!

maestro-test/src/test/resources/118_scroll_until_visible_negative.yaml
axelniklasson: Yes that is suspicious but I couldn't figure out how this would yield a negative duration _without_ setting the speed in a flow.

amanjeetsingh150 (approved):

PR #2075, opened 2024-10-03
removing gpt deprecated versions

felipeduartea:
## Issues fixed

Just removed some gpt -4 versions with the specified versions in it


Fishbowler (approved):
I don't get a say, but this LGTM

PR #2080, opened 2024-10-08
chore: add e2e tests for optional

amanjeetsingh150:
There were 2 issues with optional commands, this PR creates tests for the issues:

1. The optional flag on the selector was removed that broke deserialization and the change was not backward compatible.  The optional flag on the selector was removed from the test itself which prevented catching this regression. After observing this issue, the optional flag from selector is now deprecated.

2. Optional flags now throw command skipped exception. This was handled for the commands present in main flow but was missing in composite commands like runFlow. The addition to subFlow is suppose to catch those problem.

### DIFF
--- a/e2e/workspaces/wikipedia/subflows/onboarding-android.yaml
+++ b/e2e/workspaces/wikipedia/subflows/onboarding-android.yaml
@@ -3,6 +3,9 @@ appId: org.wikipedia
 - launchApp:
     clearState: true
 - tapOn:
+      text: "Non existent view"
+      optional: true
+- tapOn:
     id: "org.wikipedia:id/fragment_onboarding_forward_button"
 - tapOn:
     id: "org.wikipedia:id/fragment_onboarding_forward_button"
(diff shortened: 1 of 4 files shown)


e2e/workspaces/wikipedia/subflows/onboarding-android.yaml
```diff
@@ -2,6 +2,9 @@ appId: org.wikipedia
 ---
 - launchApp:
     clearState: true
+- tapOn:
+      text: "Non existent view"
+      optional: true
```
Fishbowler, line 7: What's the value in this test compared to the one in android-advanced-flow.yaml?

amanjeetsingh150: It's the same do you feel to add optional: false as well?

Fishbowler: I don't think that an explicit test for `optional: false` is necessary - it's well tested in the serialisation that false is the default option.

My point was more ... Why have the same test in both locations?

amanjeetsingh150: Running tap from main flow and from composite command like runFlow are 2 different tests. We want to test for command present in both main flow and runFlow.

Fishbowler (approved):

Fishbowler:
Rebased.

Fishbowler:
Rebased

PR #2097, opened 2024-10-16
Refactor LocalXCInstaller to handle retries properly and cleanup

amanjeetsingh150:
## Proposed changes

copilot:summary

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,13 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+    </JetCodeStyleSettings>
+    <MarkdownNavigatorCodeStyleSettings>
+      <option name="RIGHT_MARGIN" value="72" />
+    </MarkdownNavigatorCodeStyleSettings>
+    <codeStyleSettings language="kotlin">
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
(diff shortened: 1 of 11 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/api/OkHttpClientInstance.kt
```diff
@@ -0,0 +1,16 @@
+package xcuitest.api
+
+import okhttp3.ConnectionPool
+import okhttp3.OkHttpClient
+import java.util.concurrent.TimeUnit
+
+object OkHttpClientInstance {
```
herval, line 7: this will lead to multiple places using the same configuration for http connection, which is bad (some situations might require larger timeouts, some might require smaller pools, etc. I dont think it's a good idea

maestro-ios-driver/src/main/kotlin/xcuitest/api/OkHttpClientInstance.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    fun get(): OkHttpClient {
+        return OkHttpClient.Builder()
+            .connectionPool(ConnectionPool(225, 10, TimeUnit.MINUTES))
```
herval, line 11: why 225?

amanjeetsingh150: Going to remove this

maestro-ios-driver/src/main/kotlin/xcuitest/api/OkHttpClientInstance.kt
amanjeetsingh150: Going to use the HttpClientBuilder you created on the other PR. Removing this

PR #2115, opened 2024-10-26
Introducing the DataFaker library along with new features

Matheeusb:
## Proposed changes

The purpose of this PR is to add the **DataFaker** library, which provides various resources for random data (https://www.datafaker.net).

Three new random data resources have been added: `inputRandomCityName`, `inputRandomCountryName`, and `inputRandomColorName`. Additionally, a refactor was done on the existing resources, replacing the `InputRandomTextHelper` class with the DataFaker implementation.

copilot:summary

## Testing

Unit tests were added for the `Commands` class, validating that the methods are returning random texts.


Fishbowler:
Unless someone is supporting a version of datafaker for Java 11, this might be tricky. I don't think the project is gonna jump to 17 quite yet.
https://www.datafaker.net/releases/2.0.0/

Fishbowler:
Although, to be clear, I absolutely LOVE this contribution. There are loads of folks who want the ability to do inline require/import in the JS engine, purely to import faker. A version of this that removes some/all of that need would be a massive uptick in capability for a lot of testers!

Fishbowler:
Looking at the changelog for v2.0.0, it's possible that a simple swap-in for v1.9.0 might be possible.
There are no known CVEs for datafaker (seemingly, ever?)

Reckon there's a way to expose this in a more generic pattern too, to allow folk to reach the other providers based on an input key?

Fishbowler:
This experiment didn't work out?

Fishbowler:
@Matheeusb If I had some time, would you have any objections to me building on this?

Issue #1601, opened 2023-11-22
clearState doesn't reliably clear UserDefaults on iOS


AtomicGeorge:
**Describe the bug**
Launching an app with `clearState` or calling `clearState` directly doesn't reliably clear `UserDefaults` in an iOS app.

**To Reproduce**
1. Write a value to `UserDefaults` in an iOS app.
2. Quit the app
3. Launch an app with maestro using
```
- launchApp:
    appId: "com.example.app"
    clearState: true
```
or `-clearState` followed by `-launchApp`
4. Observe that `UserDefaults` still contains the previously written value.
**Expected behavior**
`userDefaults` should be cleared and the app should be launched as though it had not been used previously

**Environment information (please complete the following information):**
- Maestro version: 1.34.1
- Platform: iOS 17.0.1
- Framework: Native iOS
- Device model and OS version: iPhone 15 iOS 17.0.1
- Simulator
- Host: Mac M1

**Workspace (if applicable)**
[MaestroUserDefaultsTestApp.zip](https://github.com/mobile-dev-inc/maestro/files/13439069/MaestroUserDefaultsTestApp.zip)

**Additional context**
A workaround I've found is using `-clearState` before running `launchApp` with the `clearState` option set to `true` as far as I can tell this should delete everything twice which seems to ensure that the app is fairly reliably launched with a completely clear state.

Losiowaty:
@brandenbyers @AtomicGeorge 
Folks, do you by chance use shared user defaults (in an app group)?

What I've found is that running the `idb` command found in `clearState` docs ([here](https://maestro.mobile.dev/api-reference/commands/clearstate)) does clean the "regular" UserDefaults just fine, but doesn't touch the shared ones.

AtomicGeorge:
> @brandenbyers @AtomicGeorge Folks, do you by chance use shared user defaults (in an app group)?
> 
> What I've found is that running the `idb` command found in `clearState` docs ([here](https://maestro.mobile.dev/api-reference/commands/clearstate)) does clean the "regular" UserDefaults just fine, but doesn't touch the shared ones.

@Losiowaty The example project just uses standard user defaults, no sharing or anything fancy at all.

PR #2118, opened 2024-10-30
fix: reinstall ios app to clear state reliably

axelniklasson:
## Proposed changes
Reinstall the app and then launch it instead of manually trying to delete app state folders, which should improve the experience when relying on clear state to have a completely blank slate in the app.

Relates to #1601

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
@@ -5,10 +5,14 @@ import com.fasterxml.jackson.module.kotlin.readValue
 import maestro.utils.MaestroTimer
 import org.rauschig.jarchivelib.ArchiveFormat
 import org.rauschig.jarchivelib.ArchiverFactory
+import org.slf4j.LoggerFactory
 import util.CommandLineUtils.runCommand
 import java.io.File
 import java.io.InputStream
 import java.lang.ProcessBuilder.Redirect.PIPE
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.Path
 import kotlin.io.path.createTempDirectory
 
 object LocalSimulatorUtils {
@@ -16,6 +20,8 @@ object LocalSimulatorUtils {
     data class SimctlError(override val message: String) : Throwable(message)
 
     private val homedir = System.getProperty("user.home")
+
+    private val logger = LoggerFactory.getLogger(LocalSimulatorUtils::class.java)
 
     private val allPermissions = listOf(
         "calendar",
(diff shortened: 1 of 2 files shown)


herval:
how do we make sure this is working?

herval (approved):

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -171,30 +177,112 @@ object LocalSimulatorUtils {
...
+            logger.info("Copying app binary from $pathToBinary to $tmpBundlePath")
+            Files.copy(pathToBinary, tmpBundlePath)
+            copyDirectoryRecursively(pathToBinary, tmpBundlePath)
+
+            logger.info("Reinstalling and launching $bundleId")
+            uninstall(deviceId, bundleId)
+            install(deviceId, tmpBundlePath)
+            launch(deviceId, bundleId, emptyList(), null)
+
+            ensureRunning(deviceId, bundleId)
```
igorsmotto, line 255:
```suggestion
```
I tested this without launching the app locally, and it works as expected.

Reasoning:
clear state can be used either as a standalone command (`- clearState`) or as part of the `- launchApp` command. In the former case, we don't want to launch the app because it's not specified by the command. In the latter case, the app will be launched automatically after clearState, so there's no need to relaunch it.

WDYT?

axelniklasson: Good point, I missed that `- clearState` does not automatically launch the app again 👍

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -171,30 +177,109 @@ object LocalSimulatorUtils {
...
-            "Documents",
-            "Library",
-            "Library/Caches",
-            "Library/Preferences",
-            "SystemData",
-            "tmp"
-        )
+        // reinstall the app as that is the most stable way to clear state
+        reinstallApp(deviceId, bundleId)
+    }
```
amanjeetsingh150, line 269: This could break the current maestro flows for all the users right? How do we plan to roll this out? 

Do we wait to bring on cloud/robin or we do it right away?

Should this be in next release or we include in this current version of CLI as 1.39.1?

axelniklasson: I am torn - the fact is that clearState is not working properly today, so the fact that we fix it means that flows that implicitly rely on the broken state of clearState might stop working. However, it's a bug fix and not a breaking change so doing a major version bump doesn't feel right. Thoughts?

igorsmotto: About versioning, to me it's a bug fix. Clear state being flaky since forever **is a bug**

amanjeetsingh150: > Wait, how? Like, do you have any idea? We could potentially test this more to reduce this risk

People would have made their flows adapting to the current bug. So, the flows could break when this is actually fixed.


> I am torn - the fact is that clearState is not working properly today, so the fact that we fix it means that flows that implicitly rely on the broken state of clearState might stop working. However, it's a bug fix and not a breaking change so doing a major version bump doesn't feel right. Thoughts?

TBH, I am in favour of fully rolling this out without any hacks on backend. Given we emphasize that it could potentially break flows, I think its ok. 

Around the rollout, I feel this not requires a major version bump but we should not maybe just release (1.39.1) instead of waiting this to get in 1.40.0. Get this in cloud, robin and CLI as a quick follow up instead of waiting for 1.40.0 WDYT?

PR #2131, opened 2024-11-12
fix: use noop insights by default on maestro

amanjeetsingh150:
## Proposed changes

`Insights` is a CLI-only concept. We've seen this implementation causing problems on the cloud. This PR provides a noop implementation to Insights by default so that we don't use it on cloud.

For CLI the real implementation is passed so that behavior of CLI remains unchanged

## Testing

Locally running a command that raises insight. Tapping on optional view that's not there.

### DIFF
--- a/maestro-utils/src/main/kotlin/Insight.kt
+++ b/maestro-utils/src/main/kotlin/Insight.kt
@@ -1,31 +1,20 @@
 package maestro.utils
 
-object Insights {
+object Insights: InsightsInterface {
 
     private var insight: Insight = Insight("", Insight.Level.NONE)
     private val listeners = mutableListOf<(Insight) -> Unit>()
 
-    fun report(insight: Insight) {
-        this.insight = insight
+    override fun report(insight: Insight) {
+        Insights.insight = insight
         listeners.forEach { it.invoke(insight) }
     }
 
-    fun onInsightsUpdated(callback: (Insight) -> Unit) {
+    override fun onInsightsUpdated(callback: (Insight) -> Unit) {
         listeners.add(callback)
     }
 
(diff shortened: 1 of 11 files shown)


axelniklasson (approved):

igorsmotto (approved):

maestro-utils/src/main/kotlin/Insight.kt
```diff
@@ -1,31 +1,20 @@
 package maestro.utils
 
-object Insights {
+object Insights: InsightsInterface {
```
herval, line 3: IIUC, this doesn't do anything, other than invoking a bunch of listeners. Why do you need a noop version in that case?

maestro-utils/src/main/kotlin/InsightsInterface.kt
```diff
@@ -0,0 +1,38 @@
+package maestro.utils
+
+interface InsightsInterface {
```
herval, line 3: nit: interfaces shouldn't include "Interface" on their name.

maestro-utils/src/main/kotlin/Insight.kt
amanjeetsingh150: The Insights are only used for CLI, the noop is default implementation that we pass for cloud versions and only override for CLI. Earlier the implementations were coupled inside Orchestra and hence the cloud code.

maestro-utils/src/main/kotlin/InsightsInterface.kt
amanjeetsingh150: Good point, I'll fix this

amanjeetsingh150: Done ✅

PR #2143, opened 2024-11-18
rightsizing connections to speed up xctest process

herval:
- large connectTimeouts are only useful when you're dealing with a flaky network - eg trying to connect to something over 3G. In this case, having the timeouts at 40 means that _if xctestserver isn't running on the other side_, the connection will hang for 40 seconds, then die. We use those connections inside a retry loop, which means the retry loop is 40x slower than it should be

- readtimeout is set to 100 seconds, but there are cases (eg getting a flutter view hierarchy) where it actually takes _longer_ than that to complete. Bumping it up.

same applies to the installer:
- if connection is taking more than 1 seconds, it's an indication the installer is hung, so it doesn't make sense to wait more.


lucaswiechmann (approved):

amanjeetsingh150:
Let's merge this once CLI 1.39.2 is released.

PR #2144, opened 2024-11-18
[RFC] measurement for device commands

herval:
The idea here is to introduce micrometer management across the board, starting w/ individual commands on ios and android. This will allow us to identify potential bottlenecks, errors at the command level, etc

The default implementation is a noop - it won't collect or send metrics anywhere. Users of maestro (eg robin itself) can hook them up so we get metrics for tests that run in the platform.

The idea is to also auto-instrument http requests (using okhttp), such that we can tell cases where retries are happening more than they should, etc

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -36,6 +36,7 @@ import maestro.android.AndroidAppFiles
 import maestro.android.AndroidLaunchArguments.toAndroidLaunchArguments
 import maestro.utils.BlockingStreamObserver
 import maestro.utils.MaestroTimer
+import maestro.utils.MetricsProvider
 import maestro.utils.ScreenshotUtils
 import maestro.utils.StringUtils.toRegexSafe
 import maestro_android.*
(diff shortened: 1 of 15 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -60,10 +62,13 @@ class AndroidDriver(
     private val dadb: Dadb,
     hostPort: Int? = null,
     private var emulatorName: String = "",
-) : Driver {
+    private val metricsProvider: Metrics = MetricsProvider.getInstance(),
+    ) : Driver {
```
amanjeetsingh150, line 66: What would be the value here for CLI driver?

maestro-utils/src/main/kotlin/Metrics.kt
```diff
@@ -0,0 +1,81 @@
...
+import io.micrometer.core.instrument.MeterRegistry
+import io.micrometer.core.instrument.Tag
+import io.micrometer.core.instrument.Timer
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry
+import java.util.concurrent.TimeUnit
+
+
+// singleton to provide a metric manager across maestro code since there's so many singleton objects and passing it around would be a massive change
+object MetricsProvider {
+    private var metrics: Metrics = NoOpMetrics()
```
amanjeetsingh150, line 13: Ohh I see this is default so that's what would be there on cli

amanjeetsingh150 (approved):

PR #2153, opened 2024-11-25
fix: add waitToSettleTimeOut on swipe commands

amanjeetsingh150:
## Proposed changes

Adding `waitToSettleTimeoutMs` to the swipe commands that do waitForAppToSettle call. This is to control the wait time if there are repeatable animations on the screen. 

Using this option would help users to wait less on the screens where app is settled but have repeatable animations.

## Testing

- [ ] Locally


axelniklasson (approved):
Nice! Let's also make sure we update the docs with this

amanjeetsingh150:
Created issue for this:

https://github.com/mobile-dev-inc/maestro-docs/issues/102

PR #2156, opened 2024-11-26
Improved web support

dmitry-zaitsev:
- Fix: Maestro can test web pages again (it was broken)
- Fix: WebDriver was reporting invalid screen size
- Web: support cases where a new tab is opened from the page
- Web: screen recording support (via JCodec for now, but we could add ffmpeg later)
- Web: fake geolocation support
- Studio: better layout for wide aspect-ratio screens (i.e. web pages or tablets)


maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
```diff
@@ -113,31 +117,36 @@ object DeviceService {
         }
     }
 
-    fun listConnectedDevices(): List<Device.Connected> {
-        return listDevices()
+    fun listConnectedDevices(includeWeb: Boolean = false): List<Device.Connected> {
```
amanjeetsingh150, line 120: Why a boolean here? Shouldn't we by default list web devices?

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -252,11 +289,10 @@ class WebDriver(val isStudio: Boolean) : Driver {
     }
 
     override fun isKeyboardVisible(): Boolean {
-        TODO("Not yet implemented")
+        return false;
```
amanjeetsingh150, line 292: nit:
```suggestion
        return false
```

amanjeetsingh150 (approved):

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -203,6 +208,20 @@ class WebDriver(val isStudio: Boolean) : Driver {
...
+    private fun detectWindowChange() {
+        // Checks whether there are any new window handles available and, if so, switches Selenium driver focus to it
+        val driver = ensureOpen()
+
+        if (lastSeenWindowHandles != driver.windowHandles) {
+            val newHandles = driver.windowHandles - lastSeenWindowHandles
+            lastSeenWindowHandles = driver.windowHandles
+
+            if (newHandles.isNotEmpty()) {
+                driver.switchTo().window(newHandles.first())
```
axelniklasson, line 231: Could we log something here so it is clear that selenium automatically shifted focus to a new tab?

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -168,6 +171,8 @@ class WebDriver(val isStudio: Boolean) : Driver {
     override fun contentDescriptor(excludeKeyboardElements: Boolean): TreeNode {
         ensureOpen()
 
+        detectWindowChange()
```
axelniklasson, line 185: Whenever focus shifts to a new tab, how can we shift focus back?

maestro-web/src/main/kotlin/maestro/web/record/WebScreenRecorder.kt
```diff
@@ -0,0 +1,102 @@
...
+        seleniumDevTools.createSession()
+
+        seleniumDevTools.send(Page.enable())
+
+        seleniumDevTools.send(
+            Page.startScreencast(
+                Optional.of(Page.StartScreencastFormat.JPEG),
+                Optional.of(80),
+                Optional.of(1280),
+                Optional.of(1280),
```
axelniklasson, line 69: Will it still use the dimensions of the current window? Looks like it's the max width / height according to the API, but curious if we should instead pull the dimensions from the current window to pipe in here.

axelniklasson (approved):

dmitry-zaitsev:
> Are there any additional deps to run this locally or to set up a trivial e2e test? e.g. does this need ChromeDriver, or is that a bundled dependency?

I _believe_ it does need anything - chrome driver will be downloaded automatically on the first launch. The idea is that it should be as simple as specifying `url` instead of `appId`. If it is not - we have more work to do 😄

maestro-web/src/main/kotlin/maestro/web/record/WebScreenRecorder.kt
dmitry-zaitsev: Yes, these are indeed the max dimensions so the actual recording will keep the proper aspect ratio.

I was originally using the full device size, but this results in a substantially slower-to-process video at the end (since we are capturing quite a lot of frames), so I settled for some reasonable default.

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
dmitry-zaitsev: Unfortunately there is no way to detect that at the moment, to the best of my knowledge. Selenium does not report what the currently active tab is.

What we could do (later) is to allow user to switch to a preferred tab manually.

axelniklasson: Sounds reasonable, thanks Dima!

maestro-cli/src/main/java/maestro/cli/device/DeviceService.kt
dmitry-zaitsev: That would be rather confusing for mobile engineers because then we will always try to run the test in a browser if no mobile devices are connected

PR #2166, opened 2024-12-02
feat: added element bounds to matadata, with interactive commands

proksh:


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -861,8 +866,10 @@ class Orchestra(
         retryIfNoChange: Boolean,
         waitUntilVisible: Boolean,
         config: MaestroConfig?,
-    ): Boolean {
+    ): Bounds {
```
amanjeetsingh150, line 869: Can we change it back to boolean and call metadata update inside tap?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1207,6 +1214,7 @@ class Orchestra(
         val evaluatedCommand: MaestroCommand? = null,
         val logMessages: List<String> = emptyList(),
         val insight: Insight = Insight("", Insight.Level.NONE),
+        val bounds: Bounds? = null,
```
amanjeetsingh150, line 1217: We would need in other commands that have element selector as well

amanjeetsingh150 (approved):

amanjeetsingh150 (approved):

Issue #2128, opened 2024-11-08
1.39.0 logs have DEBUG log level info and can't be changed to INFO, which makes them 1000x bigger


mikamikaWolt:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Run any flow on a local machine on 1.36.0 and 1.39.0, then compare maestro.log from two runs.

### Actual results

maestro has test suite logs in /.maestro/tests/<timestamp>/maestro.log
Now it contains DEBUG logs, mostly from io.grpc.netty.NettyClientHandler.log.
It can't be turned off, but this makes the log file much bigger.
For example, my short sign in to the app test on 1.36.0 has 8KB log, but with 1.39.0 it is a whopping 4 MB.
This makes it too expensive to store logs as an artifact in CI or send them with test reports somewhere.

### Expected results

The default log level should be INFO as before.
Log level should be available to configure.

### About app

Any mobile app: problem is with maestro logs, not with the app.

### About environment

macOS 15.1

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

1.39.0

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

Fishbowler:
The ask is to make log levels configurable.
Perhaps we can go one better, by redirecting the Netty GRPC logging to a separate file?
Would need a bit of juggling in how LogConfig.kt is called, but looks achievable.

amanjeetsingh150:
This should be out in 1.39.3 release

PR #2170, opened 2024-12-03
fix: switch off netty handler logs

amanjeetsingh150:
## Proposed changes

I got annoyed with netty logs for the GRPC channel we have between the client and UI test on android. They are not useful haven't used them debugging issue. Fills up maestro.log file and makes it hard to debug actual issue.

They were enabled I believe after we bumped GRPC in a release.

This PR disables it, we can bring it back later if it's useful. The large size of logs is a problem for infra running the tests as well.

See: https://github.com/mobile-dev-inc/maestro/issues/2128

## Testing
Locally tested

## Issues fixed
https://github.com/mobile-dev-inc/maestro/issues/2128


Fishbowler (approved):
I use GRPC, but infrequently. If I want it, I'll add proper lines in for the right info.

Fishbowler:
Rebased

PR #2171, opened 2024-12-03
fix: assert condition command timeout parsing

rwtnb:
## Proposed changes

Removing private access modifier from timeout field

## Issues fixed

Private modifier on timeout field is preventing the field to be populated on deserialization.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -372,7 +372,7 @@ data class AssertCommand(
 
 data class AssertConditionCommand(
     val condition: Condition,
-    private val timeout: String? = null,
+    val timeout: String? = null,
     override val label: String? = null,
     override val optional: Boolean = false,
 ) : Command {


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -372,7 +372,7 @@ data class AssertCommand(
 
 data class AssertConditionCommand(
     val condition: Condition,
-    private val timeout: String? = null,
+    val timeout: String? = null,
```
amanjeetsingh150, line 375: I'm not too sure how changing scope of variable fixes the issue?

Can you write a test for it as well, so that it not breaks again?

Fishbowler:
Rebased

Fishbowler:
How do you reproduce the problem?

Fishbowler:
Locally, I tried (using the release version):

```
appId: com.example.example
---
- launchApp
- extendedWaitUntil:
    visible: "2"
    timeout: 60000
```

Which times out after 60s, as it should.

Is the serialisation problem only a problem for Robin?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
felipeduartea: no sure how this test would look, since it happens only on Robin. I can add a the test there two if you want

felipeduartea (approved):

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
herval: > I'm not too sure how changing scope of variable fixes the issue?
> 
> Can you write a test for it as well, so that it not breaks again?

private variables don't get filled out by json parsing. If this is being parsed (it seems to be the case), it'll always be null

herval (approved):
we should probably review _all_ the commands to see which ones have private vals and make them public, since this might be leading to other issues we're not seeing yet

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
amanjeetsingh150: I feel I might be missing something, the deserialization of Yaml commands to Maestro commands works through same deserializer here in [YamlCommandReader](https://github.com/mobile-dev-inc/maestro/blob/fd06096e43d9976298afd213b74f36097c355a47/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCommandReader.kt#L43) for both Robin and CLI. 

If it's working on CLI and not on Robin, it sounds like we must be doing something wrong on Robin. Not sure it's clear from here what.

felipeduartea: it seams ot me that this is due to some library version, as the code implementations is the same in both codes. Still figuring out which one is it. Should I proceed with the release?

Also kept thinking why was the condition not private already, do you have context on that? @amanjeetsingh150

amanjeetsingh150: You can go ahead with the release no worries! 

But since that can be an issue with private variables. I think more importantly let's ensure there are no other private variables in maestro commands that could also face this issue.

On the version thing probably it makes sense to confirm that we are using same version of serialzers for robin for jackson or some dependency is not missing?

felipeduartea: > But since that can be an issue with private variables. I think more importantly let's ensure there are no other private variables in maestro commands that could also face this issue.
Checked this, it seems that this was the last private val that we had on the MaestroCommand

> On the version thing probably it makes sense to confirm that we are using same version of serialzers for robin for jackson or some dependency is not missing?
Also think that it's something related to this, will take a look as soon as I free up

PR #2172, opened 2024-12-03
Add null check in childNafCheck to handle case when child node is null

Leland-Takamine:
Seeing this NPE show up in Maestro Cloud and Robin. Incorporating this AOSP fix that didn't exist when the code was copied over into Maestro: https://android-review.googlesource.com/c/platform/frameworks/support/+/2631315/2/test/uiautomator/uiautomator/src/main/java/androidx/test/uiautomator/AccessibilityNodeInfoDumper.java

### DIFF
--- a/maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
+++ b/maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
@@ -220,6 +220,7 @@ object ViewHierarchy {
         val childCount = node.childCount
         for (x in 0 until childCount) {
             val childNode = node.getChild(x)
+            if (childNode == null) continue;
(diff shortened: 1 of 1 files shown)


dmitry-zaitsev (approved):

amanjeetsingh150:
Nice 🙌  Can you run:

```
./gradlew assemble
./gradlew assembleAndroidTest
```

That should copy the new servers to the directories itself. We should probably automate this, will create a ticket for this.

amanjeetsingh150 (approved):

Fishbowler:
When you get those binaries in, the E2E Test should pass, right?

amanjeetsingh150:
> When you get those binaries in, the E2E Test should pass, right?

Correct right now the driver is not updated. Good idea to update it and see if it's not there anymore in E2E tests. I see that its actually reproducible here as well.

Fishbowler (approved):

PR #2175, opened 2024-12-06
[MA-2561] Bad requests for inputText and eraseText if keyboard is not present

amanjeetsingh150:
## Proposed changes

When the keyboard is not present on the screen the call to inputText and eraseText leads to a timeout which then becomes an infra error.

This change:

1. Raises a bad request if keyboard is not present on the screen during the operations
2. Catching them as known exception so that command doesn't fail.

## Testing

- [x] Locally
- [ ] Cloud


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -415,8 +417,11 @@ class IOSDriver(
 
     override fun inputText(text: String) {
         metrics.measured("operation", mapOf("command" to "inputText")) {
-            // silently fail if no XCUIElement has focus
-            runDeviceCall("inputText") { iosDevice.input(text = text) }
+            try {
+                runDeviceCall("inputText") { iosDevice.input(text = text) }
+            } catch (badRequest: XCUITestServerError.BadRequest) {
+                LOGGER.info("Bad request for input text, reason: ${badRequest.message}")
```
herval, line 423: why logger.info instead of logger.error?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -434,7 +439,11 @@ class IOSDriver(
 
     override fun eraseText(charactersToErase: Int) {
         metrics.measured("operation", mapOf("command" to "eraseText")) {
-            runDeviceCall("eraseText") { iosDevice.eraseText(charactersToErase) }
+            try {
+                runDeviceCall("eraseText") { iosDevice.eraseText(charactersToErase) }
```
herval, line 443: these will quietly fail. How do we know if eraseText failed, if the exception is caught here?

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Fixed

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: I've failed the command now instead instead of silently failing.

proksh (approved):

PR #2178, opened 2024-12-09
don't try to rm folder that doesn't exist

herval:
In some cases (eg a fresh ios16 simulator), the Keychains folder doesn't actually exist, so deleting it returns a non-zero status code (which bubbles up as an exception).

This prevents that by not deleting if it doesn't exist


lucaswiechmann (approved):

felipeduartea (approved):

proksh (approved):
🙌

PR #2180, opened 2024-12-09
Fix eof for setLocale

lucaswiechmann:
## Proposed changes

Fix eof when setting device locale. Exception was not being propagated which probably was keeping connections hanging until socket timeout.

setLocaleWith could potentially throwns an EOFException, and if it does, connection from dadb would hang until timeout

### DIFF
--- a/maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
+++ b/maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
@@ -7,18 +7,15 @@ import android.util.Log
 import dev.mobile.maestro.handlers.LocaleSettingHandler
 import org.apache.commons.lang3.LocaleUtils
 import java.util.*
-import org.slf4j.LoggerFactory
 import java.lang.Exception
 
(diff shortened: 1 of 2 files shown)


maestro-android/src/main/java/dev/mobile/maestro/receivers/LocaleSettingReceiver.kt
```diff
@@ -25,39 +29,48 @@ class LocaleSettingReceiver : BroadcastReceiver(), HasAction {
...
+                    return
                 }
-                resultCode = RESULT_LOCALE_NOT_VALID
-                resultData = "Failed to set locale $locale, the locale is not valid"
-                return
             }
+        } catch (e: Exception) {
+            Log.e(TAG, "Failed to validate device locale", e)
+            resultCode = RESULT_LOCALE_NOT_VALID
+            resultData = "Failed to set locale $locale, the locale is not valid"
```
herval, line 73: The reason in the case of an exception isn’t “the e locale is not valid”. It’s whatever the exception says…

lucaswiechmann: updated it

herval (approved):

PR #2184, opened 2024-12-11
[MA 2747] Check keyboard existence in all foreground apps

amanjeetsingh150:
## Proposed changes

This PR fixes 2 things:

1. Checks existence of keyboard in all the foreground apps instead of just first one.
2. Use running apps + installed apps to get the installed apps.


maestro-ios-driver/src/main/kotlin/util/XCRunnerCLIUtils.kt
```diff
@@ -44,7 +44,7 @@ object XCRunnerCLIUtils {
         val mapper = jacksonObjectMapper()
         val appsMap = mapper.readValue(json, Map::class.java) as Map<String, Any>
 
-        return appsMap.keys
+        return appsMap.keys + runningApps(deviceId).keys
```
amanjeetsingh150, line 47: Included the runningApps as well since appsMap was not giving all the list of installed apps.

herval (approved):
Candidate for Fix of the Year 🏆

DavidREntwistle:
We seem to be experiencing the `Keyboard not presented within 1 second timeout for input command` error in the latest 1.39.4 release. We perform two commands, first we `tapOn` the input field text field and then we perform the `inputText`.

The previous 1.39.2, didn't hit this `if !isKeyboardPresented` error, with the keyboard displaying for both the `tapOn` input field and `inputText`. Is there a reason why this might now be causing following this change?

PR #2191, opened 2024-12-14
Extract text with AI command

dmitry-zaitsev:
Introducing a `extractTextWithAI` command that uses LLM to extract text form a screenshot. Some situations where it is useful:

- Finding dynamic elements on the screen (i.e. as a result of a search)
- Finding elements by their semantic meaning rather than hardcoded strings or IDs
- Bypassing Captcha

## Usage Example

```
- extractTextWithAI: CAPTCHA value
- inputText: ${aiOutput}
```

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -84,7 +84,7 @@ class Orchestra(
     private val onCommandStart: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandComplete: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandFailed: (Int, MaestroCommand, Throwable) -> ErrorResolution = { _, _, e -> throw e },
-    private val onCommandWarned: (Int, MaestroCommand) -> Unit = { _,  _ -> },
+    private val onCommandWarned: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandSkipped: (Int, MaestroCommand) -> Unit = { _, _ -> },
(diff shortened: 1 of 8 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -84,7 +84,7 @@ class Orchestra(
     private val onCommandStart: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandComplete: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandFailed: (Int, MaestroCommand, Throwable) -> ErrorResolution = { _, _, e -> throw e },
-    private val onCommandWarned: (Int, MaestroCommand) -> Unit = { _,  _ -> },
```
dmitry-zaitsev, line 87: There is a bit of auto-reformatting in this class. The main relevant change is `extractTextWithAICommand` method

dmitry-zaitsev:
Not sure why detekt pushed so many styling changes 🤷 Perhaps these classes were not touched since the style changed?

felipeduartea (approved):
looks good to me! 😄

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -413,6 +415,26 @@ class Orchestra(
         false
     }
 
+    private fun extractTextWithAICommand(command: ExtractTextWithAICommand): Boolean = runBlocking {
+        // Extract text from the screen using AI
+        if (ai == null) {
```
felipeduartea, line 420: nit: Although this `ai` variable was not added now, it probably deserves a better name

PR #2217, opened 2024-12-24
chore: added log for null hierarchy

proksh:
When there is null hierarchy, 
```val error = mapper.readValue(responseBodyAsString, Error::class.java)```
fails, and we don't get the status code and error message. Added a warn log before it so that we can know the status code and error message.

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
+++ b/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
@@ -80,8 +80,12 @@ struct ViewHierarchyHandler: HTTPHandler {
     }
 
     func getHierarchyWithFallback(_ element: XCUIElement) throws -> AXElement {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by amanjeetsingh150 — APPROVED** [review]

**Comment by amanjeetsingh150** [comment]
Let's also add logs on ViewHierarchyHandler and update driver.

File: maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift [hunk 1]
```diff
@@ -118,11 +122,17 @@ struct ViewHierarchyHandler: HTTPHandler {
                 }
 
                 let alerts = logger.measure(message: "Fetch alert hierarchy") {
-                    fullScreenAlertHierarchy(element)
+                    fullScreenAlertHierarchy(element) {
```
amanjeetsingh150, line 125: Is this call taking a function argument? I don't think this will compile can you try executing:

```
sh maestro-ios-xctest-runner/build-maestro-ios-runner.sh
```
Let's actually add a log on line 40 and 42 before we return because in case the endpoint returns error we would have the logs.

PR #2226, opened 2025-01-06
Improve set permissions logs and avoid affecting location permissions

amanjeetsingh150:
## Proposed changes

Adds finer-grained logs for the set permissions operations performed using simctl and applesimutils. We've observed instances where applesimutils fails to set permissions, which can lead to another issue of not setting location permissions. 

This PR also decouples the simctl and applesimutil permission so that they don't couple each other.

## Testing

- [x] Locally
- [ ] Cloud


Fishbowler (approved):
Looks sensible

PR #2235, opened 2025-01-09
fix: update maestro test to respect --host and --port arguments

### LINKED ISSUE #2233: Setting the `host` and `port` parameters does not work

### Steps to reproduce

You can see the full history on Slack [here](https://mobile-dev-inc.slack.com/archives/C041FU72T54/p1736342869766629?thread_ts=1736324123.917629&cid=C041FU72T54)

Reproducible example:

1. Setup ADB and connect a device
2. Setup the following Docker image and run it, while exposing port 5037
<details>
  <summary>Docker image definiton</summary>

```Docker
# Use an official OpenJDK image as the base
FROM openjdk:11-jdk

# Set environment variables for the Android SDK
ENV ANDROID_SDK_ROOT=/usr/local/android-sdk
ENV PATH=$PATH:$ANDROID_SDK_ROOT/cmdline-tools/latest/bin:$ANDROID_SDK_ROOT/platform-tools

# Install required packages
RUN apt-get update && apt-get install -y --no-install-recommends \
    wget \
    unzip \
    libglu-dev \
    && rm -rf /var/lib/apt/lists/*

# Download and install the Android SDK Command-line Tools
RUN mkdir -p $ANDROID_SDK_ROOT/cmdline-tools
WORKDIR /tmp
RUN wget https://dl.google.com/android/repository/commandlinetools-linux-9477386_latest.zip -O cmdline-tools.zip \
    && unzip cmdline-tools.zip \
    && rm cmdline-tools.zip \
    # Move the extracted "cmdline-tools" folder to "latest"
    && mv cmdline-tools $ANDROID_SDK_ROOT/cmdline-tools/latest

# Accept all Android SDK licenses
RUN yes | sdkmanager --licenses

# Install essential SDK packages, including platform-tools (for adb)
RUN sdkmanager --update \
    && sdkmanager "platforms;android-33" "build-tools;33.0.0" "platform-tools"

# Install Maestro CLI (native Linux binary)
ENV MAESTRO_VERSION=1.39.7
RUN mkdir -p /opt/maestro && \
    wget -q -O /tmp/${MAESTRO_VERSION} \
      "https://github.com/mobile-dev-inc/maestro/releases/download/cli-${MAESTRO_VERSION}/maestro.zip" && \
    unzip -q /tmp/${MAESTRO_VERSION} -d /opt/ && \
    rm /tmp/${MAESTRO_VERSION}

# Add Maestro to PATH
ENV PATH=/opt/maestro/bin:${PATH}

# Expose any ports you might need for Maestro / adb
EXPOSE 7000

# Set the default entrypoint to the Maestro CLI
ENTRYPOINT ["maestro"]
CMD ["--help"]

```
</details>
3. Now try to run the container's Maestro and make it use the host's ADB via `maestro  --host host.docker.internal   --port 5037 test test.yml`

### Actual results

Maestro cannot detect ADB, and will start its own ADB server

Even though ADB is reachable (you can manually do `adb devices` and see that it talks to the host's)

### Expected results

Maestro connect to the host's ADB

### About app

There is no app

### About environment

M4 Mac running docker

### Logs

<details>
<summary>Logs</summary>

```
```

</details>

### Maestro version

1.39.7

Homebrew

### Anything else?

_No response_

Leland-Takamine:
Resolves #2233

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
@@ -61,7 +61,7 @@ class PrintHierarchyCommand : Runnable {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = parent?.port,
+            driverHostPort = null,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
         ) { session ->
--- a/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
@@ -66,7 +66,7 @@ class QueryCommand : Runnable {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = parent?.port,
+            driverHostPort = null,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
         ) { session ->
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -108,7 +108,7 @@ class RecordCommand : Callable<Int> {
         return MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = parent?.port,
+            driverHostPort = null,
             deviceId = deviceId,
             platform = parent?.platform,
         ) { session ->
--- a/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
@@ -52,7 +52,7 @@ class StudioCommand : Callable<Int> {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = parent?.port,
+            driverHostPort = null,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
             isStudio = true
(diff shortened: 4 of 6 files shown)


maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -61,7 +61,7 @@ class PrintHierarchyCommand : Runnable {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = parent?.port,
+            driverHostPort = null,
```
Leland-Takamine, line 64: @sdfgsdfgd Your sharding change seemed to overload the --port option on the CLI to also set the driverHostPort which didn't seem correct. This would mean that --port 5555 would set both the Dadb port and the driver port which wouldn't work (they are different ports). If you see anything wrong with this change setting the driverHostPort to null (meaning it would use the default driver host port in these case) please let me know!

Fishbowler (approved):
Approved as "working" rather than as "nice code" - I don't know this area well.

On a clean windows machine, I've followed the docs, and successfully run a flow in WSL, and interacted with Maestro Studio. I've intentionally not tried any of the sharding aspects, given there's a chance that there be changes afoot :grin:

amanjeetsingh150 (approved):

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
avinash-bharti: @Leland-Takamine @amanjeetsingh150 
- By removing the driverHostPort value setting there is one issue that has been poped up.
- **Issue:** Multiple instances of maestro cli trying to communicate over same android driver port even when device id is different.
-- This issue was resolvable by passing custom --port values until version 1.39.7.
-- This particular change has overridden the behaviour.
-- Do you suggest adding separate ports for driverHost and dadb? (should I create seperate issue for this?)

sdfgsdfgd: @Leland-Takamine I've been away for a while, sorry !  I see there has been multiple changes and some new issues with sharding, I'll find the time to sit down and go through the open issues soon, but the patches so far seem great, on point. The overload was actually necessary at the time, although I don't really remember why, I think it was preventing the race maybe that's why I kept the overload at the time. If you found a better solution, that's perfect. I'll go through the new patches for some context. 👍

Issue #2113, opened 2024-10-24
iOS miss `-` Launch Arguments when value type is Bool


littlema0404:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

```yaml
appId: xxx
- launchApp:
    clearState: true
    permissions:
      userTracking: deny
    arguments:
        argumentA: true
        argumentB: 1
        argumentC: "abc"

```

### Actual results

```swift
CommandLine.arguments // ["argumentA", "true", "-argumentB", "1", "-argumentC", "abc"]
```

### Expected results

```swift
CommandLine.arguments // ["-argumentA", "true", "-argumentB", "true", "-argumentC", "abc"]
```

### About app

Closed source iOS native app

### About environment

Java version
```
openjdk version "17.0.10" 2024-01-16 LTS
OpenJDK Runtime Environment Corretto-17.0.10.7.1 (build 17.0.10+7-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.10.7.1 (build 17.0.10+7-LTS, mixed mode, sharing)
```

OS and its version
```
macOS 14.6.1（23G93）
```

Processor architecture 
```
M2 arm64
```

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

1.38.1

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

### Anything else?

`-` This is the specification for iOS launch arguments. Currently, only when the YAML value type is Bool, the - sign is not added. However, if the value type is Int or String, the - sign is automatically added.

Fishbowler:
Looks like the intent in #1018 was that you'd access booleans differently?

littlema0404:
Yeh, for here a little bit confused.

like example from #1018 

```swift
 ProcessInfo.processInfo.arguments.contains("isMaestro")
```

The contains method only indicates whether the ‘isMaestro’ launch argument is present, but it doesn’t allow me to retrieve its boolean value.

johntmcintosh:
I ran into this today as was/am confused as well. 

In my case, I was passing one launch argument that was an explicit boolean, and another that was a boolean passed through the yaml environment:

```
appId: ${APP_ID}
---
- launchApp:
    arguments: 
        argA: true
        argB: ${ARG_B_VALUE}
```

and then I'm calling this flow with

```
[...]

PR #2242, opened 2025-01-13
Fix iOS miss launch arguments with boolean 

littlema0404:
## Proposed changes

add prefix `-` at key for all iOS launch arguments

## Testing

none

## Issues fixed

link https://github.com/mobile-dev-inc/maestro/issues/2113


Fishbowler:
This seems to have gotten left behind. Sorry about that.

I've read up on the linked issue, and this makes sense to me.

Does this have the potential to break the workflows of existing folks?

PR #2254, opened 2025-01-17
Implement setTimeout/clearTimeout in Maestro's JavaScript Environment

JulienLeal:
## Proposed changes
This PR implements the standard JavaScript timer functions (`setTimeout`/`clearTimeout`) in Maestro's GraalJS environment, which previously didn't include these functions.

### GraalJsTimer.kt
- Implemented new `GraalJsTimer` class to handle JavaScript timer functionality
- Added thread-safe timer management using `AtomicInteger`
- Implemented proper timer tracking with `CountDownLatch`
- Added cleanup handling for timer resources

### GraalJsEngine.kt
- Added timer bindings to the JavaScript context
- Exposed `setTimeout` and `clearTimeout` functions
- Added timer cleanup in engine close

### Orchestra.kt
- Added support for waiting on active timers in script execution
- Improved script condition handling with proper null safety
- Added timer completion checks before finishing script execution

## Implementation Details
The implementation follows JavaScript's standard timer behavior:
- `setTimeout(callback, delay, ...args)` returns a numeric timer ID
- `clearTimeout(timeoutId)` cancels a pending timeout
- Timer management is thread-safe and properly cleaned up
- Test execution waits for pending timeouts

## Why This Matters
This implementation aligns Maestro's JavaScript environment with the JavaScript standard ecosystem, enabling essential asynchronous patterns in Maestro scripting. While the immediate use case involves polling mechanisms, it unlocks numerous scenarios including:

- Waiting for an OTP code by periodically checking with setTimeout
- Handling network response delays
- Implementing retry logic with backoff
- Creating non-blocking delays in test sequences

## Testing

Tested with various scenarios:
- Short and long timeouts
- Multiple concurrent timeouts
- Timer cancellation
- Resource cleanup

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsTimer.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsTimer.kt
@@ -1,0 +1,84 @@
+package maestro.js
+
+import org.graalvm.polyglot.Value
+import org.graalvm.polyglot.proxy.ProxyExecutable
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
+
+class GraalJsTimer {
+    private val scheduler: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
+    private val timeouts = ConcurrentHashMap<Int, ScheduledFuture<*>>()
+    private var timeoutCounter = 0
+    private var activeTimersCount = AtomicInteger(0)
+    private var activeTimers = CountDownLatch(0)
+
+    val setTimeout = ProxyExecutable { args ->
+        val callback = args[0] as Value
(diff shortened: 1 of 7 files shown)


maestro-client/src/main/java/maestro/js/GraalJsTimer.kt
```diff
@@ -0,0 +1,84 @@
...
+                activeTimers.countDown()
+            }
+        }
+    }
+
+    fun waitForActiveTimers(timeout: Long = 30_000) {
+        if (activeTimersCount.get() > 0) {
+            activeTimers.await(timeout, TimeUnit.MILLISECONDS)
+        }
+    }
```
Fishbowler, line 72: Is this worth an integration test, that a 31s setTimeout never fires? I hate the idea of long running tests, but I can't think of a better way to test that this doesn't leave long running activities.

I'm not so worried about local running - I'm thinking about whether this change could provide any risk to the Maestro hosted infrastructure by a malicious or careless user.

JulienLeal:
Hi @Fishbowler I'll take a look on these comments on the next weekend. I was on vacations, sorry my delay

neiker:
@JulienLeal is it possible to merge this PR?

Fishbowler:

PR #2259, opened 2025-01-20
Adds shake command

### LINKED ISSUE #1164: Shake Simulation
I want to simulate the bugs report and it's always by shaking the device

- shake
could be a great command

ivanacostarubio:
## Proposed changes

It implements the shake command. It closes https://github.com/mobile-dev-inc/maestro/issues/1164 

### AI

copilot:summary

## Testing

I was able to test this using my local iOS simulator. However, I'm unsure why it's not working on the Android emulator. I can see the shake happening when I use the terminal to run the adb commands.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -252,6 +252,11 @@ class AndroidDriver(
         }
     }
 
+    override fun shake() {
+        dadb.shell("emu sensor set acceleration 100:100:100")
+        dadb.shell("emu sensor set acceleration 0:0:0")
+    }
+
     override fun clearAppState(appId: String) {
         metrics.measured("operation", mapOf("command" to "clearAppState", "appId" to appId)) {
             if (!isPackageInstalled(appId)) {
--- a/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
+++ b/maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
@@ -176,6 +176,44 @@ class XCTestDriverClient(
         installer.close()
     }
 
+    fun shake(){
+        logger.trace("Executing shake gesture on iOS simulator")
+
+        try {
+            // Create ProcessBuilder for executing AppleScript
+            val processBuilder = ProcessBuilder(
+                "osascript",
+                "-e", """
+                tell application "Simulator"
+                    activate
+                    delay 0.5 -- Give simulator time to come to foreground
+                    tell application "System Events"
+                        tell process "Simulator"
+                            -- Access the Hardware menu and click Shake Gesture
+                            click menu item "Shake" of menu "Device" of menu bar 1
+                        end tell
+                    end tell
+                end tell
+            """.trimIndent()
+            )
+
+            // Execute the command
+            val process = processBuilder.start()
+            val exitCode = process.waitFor()
+
+            if (exitCode != 0) {
+                val errorStream = process.errorStream.bufferedReader().readText()
+                logger.error("Failed to execute shake gesture. Exit code: $exitCode, Error: $errorStream")
+                throw IOException("Failed to execute shake gesture on simulator. Exit code: $exitCode")
+            }
+
+            logger.trace("Successfully executed shake gesture")
+        } catch (e: Exception) {
+            logger.error("Error executing shake gesture", e)
+            throw XCUITestServerError.UnknownFailure("Failed to execute shake gesture on simulator: ${e.message}")
+        }
+    }
+
     fun setPermissions(permissions: Map<String, String>) {
         executeJsonRequest("setPermissions", SetPermissionsRequest(permissions))
     }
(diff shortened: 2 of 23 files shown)


maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -170,6 +170,44 @@ class XCTestDriverClient(
...
+            )
+
+            // Execute the command
+            val process = processBuilder.start()
+            val exitCode = process.waitFor()
+
+            if (exitCode != 0) {
+                val errorStream = process.errorStream.bufferedReader().readText()
+                logger.error("Failed to execute shake gesture. Exit code: $exitCode, Error: $errorStream")
+                throw IOException("Failed to execute shake gesture on simulator. Exit code: $exitCode")
```
rtharston, line 201: The catch block below will re-throw this with this message wrapped in the same message again:
"Failed to execute shake gesture on simulator: Failed to execute shake gesture on simulator. Exit code: $exitCode"

ivanacostarubio: Thanks for the review. https://github.com/mobile-dev-inc/Maestro/pull/2259/commits/68212e09855366bfc31d91ae4c3e56b2c9cafd48

maestro-ios-driver/src/main/kotlin/xcuitest/XCTestDriverClient.kt
```diff
@@ -170,6 +170,43 @@ class XCTestDriverClient(
...
+            // Execute the command
+            val process = processBuilder.start()
+            val exitCode = process.waitFor()
+
+            if (exitCode != 0) {
+                val errorStream = process.errorStream.bufferedReader().readText()
+                logger.error("Failed to execute shake gesture. Exit code: $exitCode, Error: $errorStream")
+            }
+
+            logger.trace("Successfully executed shake gesture")
```
rtharston, line 203: Now if exitCode != 0 you will log the error, but don't do anything about it and you'll log this 'Successful' message and return normally.

Sorry I wasn't more clear in the earlier review. I'd suggest bringing the previous throw back, so the function exits properly on error. But something else needs to change to avoid the double message.

I thought of two options.
1. cut down the message: `throw IOException("Exit code: $exitCode")` and the outer catch/throw will use that as the error message.
2. Move the code that checks the exitCode outside the try block (everything from 'if' on 198 to 'logger.trace' on 203). That way so the `throw IOException` can throw directly to the calling function. That means the calling function will get the exception as IOException instead of XCUITestServerError.UnknownFailure.

rtharston: Actually, there is a different issue at play.

When I mentioned the caller I decided to look at the rest of the file for other examples, and I noticed that this function is different from all the others that call `executeJsonRequest`. Looks like you need to move all this code to a new Handler file in `/Maestro/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/` (and add a new case to the Rouse enum in XCTestHTTPServer.swift and handle that case in createRouteHandler in RouteHandlerFactory.swift).

Then instead of throwing exceptions you'll want to return HTTP errors.

Of course, this is all from looking at this file for a few minutes, I'm not one of the original developers, so we'll want one of them to chime in as well.

rtharston (changes requested):
While I was writing the other comment about the deleted 'throw' I decided to look at the rest of the XCTestDriverClient file for other examples of throwing exceptions, and I noticed that this function is different from all the others that call executeJsonRequest. Looks like you need to move all this code to a new Handler file in /Maestro/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ (and add a new case to the Rouse enum in XCTestHTTPServer.swift and handle that case in createRouteHandler in RouteHandlerFactory.swift).

Then instead of throwing exceptions you'll want to return HTTP errors.

Of course, this is all from looking at this file for a few minutes, I'm not one of the original developers, so we'll want one of them to chime in as well.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -235,6 +235,11 @@ class AndroidDriver(
         }
     }
 
+    override fun shake() {
+        dadb.shell("emu sensor set acceleration 100:100:100")
+        dadb.shell("emu sensor set acceleration 0:0:0")
```
herval, line 240: have you tested this? 

I've seen something like this on a SO thread:
```
adb emu sensor set acceleration 100:100:100; sleep 1; adb emu sensor set acceleration 0:0:0
```

curious how we make sure this sort of function actually works

Fishbowler: @ivanacostarubio I've done a rebase (and pushed a commit to fix the drift problems) and after sorting out some automated e2e testing (that I've not committed yet) it turns out that this doesn't work through dadb as expected

```
Shell response (127):
/system/bin/sh: emu: inaccessible or not found
```

Fishbowler: I see the same thing at the CLI

```
danc@Mac demo_app % adb shell "emu sensor set acceleration 100:100:100"                                            
/system/bin/sh: emu: inaccessible or not found
```

Fishbowler (changes requested):
See thread above - the AndroidDriver implementation doesn't work at present.

Let me know if you need a hand with this.

Fishbowler:
@ivanacostarubio Do you have any interest in moving this forwards?

ivanacostarubio:
Not at this moment @Fishbowler 🙃

PR #2265, opened 2025-01-22
feat(analyze): Adds Analyze option to test command

luistak:
## Proposed changes

This PR:
- Adds new `--analyze`  flag to optionally generate AI insights analysis after the test command
- Adds a new notification to new users about the Analyze feature
- Adds new debug parameters for Login Command

## Testing

- [x] Local


maestro-ai/src/main/java/maestro/ai/Prediction.kt
```diff
@@ -270,4 +270,4 @@ object Prediction {
         return response.text ?: ""
     }
 
-}
+}
```
luistak, line 273: some of these styles and changes were related to inteliJs linter, but if I need to adjust something feel free to comment 🙏

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -158,6 +158,19 @@ class TestCommand : Callable<Int> {
     )
     private var headless: Boolean = false
 
+    @Option(
+        names = ["--analyze"],
+        description = ["[Beta] Enhance the test output analysis with AI Insights"],
+    )
+    private var analyze: Boolean = false
+
+    @Option(names = ["--apiUrl"], description = ["[Beta] API base URL"])
```
herval, line 167: other params use `dashes-between-words` instead of `camelCase`

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -158,6 +158,19 @@ class TestCommand : Callable<Int> {
...
+    private var apiUrl: String = "https://api.copilot.mobile.dev"
+
+
+    @Option(names = ["--apiKey"], description = ["[Beta] API key"])
```
herval, line 171: don't we have this already? `--api-key`?

luistak: Not in this command, neither in parent command like `--verbose`

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
luistak: Oh i'll adjust to kebab-case since every command in this file is following this pattern, but I inspired these options from the `UploadCommand` do I change them there as well or do i open an issue later for consistency across all commands?

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
```diff
@@ -0,0 +1,142 @@
...
+    val textFiles: List<Pair<ByteArray, Path>>
+)
+
+class TestAnalysisManager(private val apiUrl: String, private val apiKey: String?) {
+    private val apiclient by lazy {
+        ApiClient(apiUrl)
+    }
+
+    fun runAnalysis(debugOutputPath: Path): Int {
+        val flowFiles = processFilesByFlowName(debugOutputPath)
```
amanjeetsingh150, line 32: Is this flow file or debug files?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -440,6 +458,58 @@ class ApiClient(
...
+
+            flowFile.textFiles.forEach { (textData, path) ->
+                val textName = path.fileName.toString()
+                logs.add(Pair(textName, textData))
+            }
+        }
+
+        val requestBody = mapOf(
+            "screenshots" to screenshots,
+            "logs" to logs
```
amanjeetsingh150, line 484: We are sending both commands hierarchy and maestro logs as logs?

How would server know which text file is for what?

I think it would be good to keep those as a different field in request.

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        return Pair(imageFiles, textFiles)
+    }
+
+    /**
+     * The analytics system for Test Analysis.
+     *  - Uses configuration from $XDG_CONFIG_HOME/maestro/analyze-analytics.json.
+     */
+    companion object Analytics {
```
amanjeetsingh150, line 93: We already have analytics class for this why not use that?

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
```diff
@@ -89,6 +90,9 @@ object MaestroCommandRunner {
         }
 
         refreshUi()
+        if (analyze) {
+            ScreenshotUtils.takeDebugScreenshotByCommand(maestro, debugOutput, CommandStatus.PENDING)
+        }
```
amanjeetsingh150, line 95: Why an extra screenshot here?

amanjeetsingh150 (commented):
A potential improvement you can do for cleaning the similar screenshots would be to use this logic from maestro:

https://github.com/mobile-dev-inc/maestro/blob/f69bf5b7997b974195b5af7dc555741eb13ab6bb/maestro-client/src/main/java/maestro/utils/ScreenshotUtils.kt#L86

We can only send the screenshots which are considerably different from each other.

amanjeetsingh150 (commented):
nit: Can you also add the command change that we add with this on maestro cli?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -440,6 +458,58 @@ class ApiClient(
...
+            .post(body)
+            .build()
+
+        val response = client.newCall(request).execute()
+
+        response.use {
+            if (!response.isSuccessful) {
+                val errorMessage = response.body?.string().takeIf { it?.isNotEmpty() == true } ?: "Unknown"
+                throw CliError("Analyze request failed (${response.code}): $errorMessage")
+            }
```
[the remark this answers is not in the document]
luistak: initially on the backend, since the whole generation and edge cases are abstracted on robin api

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
luistak: debug, good point I'll rename it to be more consistent

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
luistak: Because we don't differentiate them for now, I thought it would be more flexible to allow us to send more logs in the future, like a custom `ai-(flow-file).json` and future files without a change in the API contract

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
luistak: Analytics was the inspiration initially but rethinking the purpose of this class I think that the naming is kinda wrong, the use of the analytics file rn is related to our tracking of local data, and the whole purpose of this class is to manage the TestAnalysis Notification on the CLI, we are not tracking any data, do you think it would be better to rename this class to `Notification`?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
amanjeetsingh150: I see the flexibility point you’re making here, but I’m coming at this from the perspective of using this to communicate with any AI agent. Having a distinction here, especially from the client side, would make it easier for the server to handle proper mappings of Maestro logs and hierarchy separately.

This separation could simplify things when processing or transforming data before sending it downstream.

maestro-cli/src/main/java/maestro/cli/command/LoginCommand.kt
```diff
@@ -25,8 +26,11 @@ class LoginCommand : Callable<Int> {
     @CommandLine.Mixin
     var showHelpMixin: ShowHelpMixin? = null
 
+    @Option(names = ["--api-url", "--apiUrl"], description = ["API base URL"])
+    private var apiUrl: String = "https://api.copilot.mobile.dev"
+
     private val auth by lazy {
-        Auth(ApiClient("https://api.copilot.mobile.dev/v2"))
+        Auth(ApiClient("$apiUrl/v2"))
```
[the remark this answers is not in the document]
luistak: to allow local development easily like `maestro upload --apiUrl="http://localhost:8080" ...`

maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
luistak: Btw I added the suggestion to filter similar images and not it's ignoring the unnecessary cookies programmatically

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
```diff
@@ -0,0 +1,179 @@
...
+
+        private val notificationState: AnalysisNotificationState
+            get() = JSON.readValue<AnalysisNotificationState>(notificationStatePath.readText())
+
+        fun maybeNotify() {
+            if (shouldNotNotify) return
+
+            println(
+                listOf(
+                    "Tryout our new Analyze with Ai feature.\n",
```
luistak, line 153:
```suggestion
                    "Try out our new Analyze with Ai feature.\n",
```

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
```diff
@@ -0,0 +1,179 @@
...
+            if (shouldNotNotify) return
+
+            println(
+                listOf(
+                    "Try out our new Analyze with Ai feature.\n",
+                    "See what's new:",
+                    "> https://maestro.mobile.dev/cli/test-suites-and-reports#analyze",
+                    "Analyze command:",
+                    "$ maestro test flow-file.yaml --analyze | bash\n",
+                    "To disable this notification, set $DISABLE_NOTIFICATION_ENV_VAR environment variable to \"true\" before running Maestro."
```
herval, line 158: the correct env var name is $MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED, not $DISABLE_NOTIFICATION_ENV_VAR

maestro-cli/src/main/java/maestro/cli/insights/TestAnalysisManager.kt
```diff
@@ -0,0 +1,179 @@
...
+                ).joinToString("\n").box()
+            )
+            ack();
+        }
```
luistak, line 162:
```suggestion
        private const val MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED = "MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED"
        private val disabled: Boolean
            get() = System.getenv(MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED) == "true"

        private val notificationStatePath: Path = EnvUtils.xdgStateHome().resolve("analyze-notification.json")

        private val JSON = jacksonObjectMapper().apply {
            registerModule(JavaTimeModule())
            enable(SerializationFeature.INDENT_OUTPUT)
        }

        private val shouldNotNotify: Boolean
            get() = disabled || notificationStatePath.exists() && notificationState.acknowledged

        private val notificationState: AnalysisNotificationState
            get() = JSON.readValue<AnalysisNotificationState>(notificationStatePath.readText())

        fun maybeNotify() {
            if (shouldNotNotify) return

            println(
                listOf(
                    "Try out our new Analyze with Ai feature.\n",
                    "See what's new:",
                    "> https://maestro.mobile.dev/cli/test-suites-and-reports#analyze",
                    "Analyze command:",
                    "$ maestro test flow-file.yaml --analyze | bash\n",
                    "To disable this notification, set $MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED environment variable to \"true\" before running Maestro."
                ).joinToString("\n").box()
            )
            ack();
        }
```

PR #2022, opened 2024-09-05
Add allow host access to Graaljs under flags

tokou:
## Context

https://github.com/mobile-dev-inc/maestro/pull/1996#issuecomment-2324679424

## Proposed changes

- I also added `MAESTRO_CLI_DANGEROUS_GRAALJS_ALLOW_ALL_ACCESS` for `allowAllAccess()`
- I added a disabled integration test for the case where the env var needs to be set as I don't think it's possible to mock System.env.

## Testing

- Unit tests for truthness of env values
- Integration tests

## Issues fixed

Linked https://github.com/mobile-dev-inc/maestro/pull/1949
Linked https://github.com/mobile-dev-inc/maestro/pull/1996
Fixes https://github.com/mobile-dev-inc/maestro/issues/1388

PR #2271, opened 2025-01-24
#1637 [Graaljs] allow import

### LINKED ISSUE #1637: [v1.34.1] Importing CommonJS or ECMAScript modules doesn't work with graaljs
**Describe the bug**
Graaljs supports importing CommonJS or ECMAScript modules. However, due to the `GraalJsEngine` implementation, no module imports work currently.

This most likely happens as `Context.newBuilder("js")` is missing `.allowIO(true)` here: https://github.com/mobile-dev-inc/maestro/blob/main/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt#L80

See more info on module imports in the graalVM reference manual: https://www.graalvm.org/latest/reference-manual/js/Modules/#ecmascript-modules-esm

**To Reproduce**
1. Create two js files, e.g. `main.js` and `env.mjs`

```
// main.js
import { env } from "./two.mjs";

const testImport = () => {
  console.log(`${env.URL}`);
  return env.URL;
};

output.url = testImport();
```
```
// env.mjs
export const env = {
  URL: "https://example.com",
  ANOTHER_URL: "https://another-example.com",
};
```
2. Include `main.js` in e.g. `main.yaml`
```
appId: com.example.myapplication
jsEngine: graaljs
---
- runScript: main.js
- evalScript: ${console.log(output.url)}
```
3. Run `maestro test main.yaml`

**Expected behavior**
Importing CommonJS or ECMAScript modules should work.

**Screenshots**
Currently maestro tries to evaluate the import as a script.
Screenshot 2023-12-30 at 0 28 29

tokou:
## Proposed changes

This PR adds the option to use `import` directive from within JS scripts.
This can be helpful to better organize large scripts and avoid duplication between flows.
This is done through either the env var `MAESTRO_GRAALJS_ALLOW_IO` or the config `graalJsAllowIo`.

## Specifications

- Scripts that use imports (wether importing or exporting) should be defined with the `mjs` extension
- The importing path is relative to the execution path of Maestro
- See GraalJs documentation about [Modules](https://github.com/oracle/graaljs/blob/master/docs/user/Modules.md#ecmascript-modules-esm)

## Testing

Two integration tests have been added, in order to test cases where it's allowed and not allowed.

## Issues fixed

#1637

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
@@ -29,6 +29,7 @@ class GraalJsEngine(
         protocols = listOf(Protocol.HTTP_1_1)
     ),
     platform: String = "unknown",
+    private val allowIo: Boolean = false,
 ) : JsEngine {
 
     private val openContexts = HashSet<Context>()
@@ -84,6 +85,7 @@ class GraalJsEngine(
         }
 
         val context = Context.newBuilder("js")
+            .allowIO(allowIo)
             .option("js.strict", "true")
             .logHandler(NULL_HANDLER)
             .out(outputStream)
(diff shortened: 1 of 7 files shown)


tokou:
@Fishbowler I don't understand why the E2E test fails. Would appreciate any help.

Fishbowler:
Does this replace #2022?

tokou:
@Fishbowler No, they are 2 different features of the underlying JS engine.
One lets you call platform (i.e. JVM) classes from JS and interop with them
The other lets you allow access to files and thus use import and modules (which are enabled as part of the ECMAScript but impossible to use due to io restrictions in GraalJS)

PR #2275, opened 2025-01-29
[MA 2992] Prebuilt iOS driver without xcodebuild

amanjeetsingh150:
## Proposed changes

1. **Faster Driver Setup on iOS**: This change improves setup time by skipping xcodebuild and directly launching the runner on the simulator. By default, this remains disabled (false).
2. **Fixing Error Logging in runCommand**: Previously, all errors were redirected to /dev/null, making logs inaccessible. This update redirects errors to a pipe, ensuring they are captured properly.

## Testing

- [x] Locally
- [x] Cloud


herval:
Instead of passing booleans around everywhere, could that be a xctestinstaller flag? Or maybe a different xctestinstaller class?

amanjeetsingh150:
> Instead of passing booleans around everywhere, could that be a xctestinstaller flag? Or maybe a different xctestinstaller class?

Done ✅

PR #2288, opened 2025-02-04
feat: start trial option if not started

proksh:
Screenshot 2025-02-04 at 5 23 44 PM

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -32,6 +32,7 @@ import okio.IOException
 import okio.buffer
 import java.io.File
 import java.nio.file.Path
+import java.util.Scanner
 import kotlin.io.path.absolutePathString
 import kotlin.io.path.exists
 import kotlin.time.Duration.Companion.minutes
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -368,6 +386,25 @@ class ApiClient(
         }
     }
 
+    private fun startTrial(authToken: String, baseUrl: String): Boolean {
+        println("Starting your trial...")
+        val body = "".toRequestBody(null)
```
amanjeetsingh150, line 391: Weird why is the request for this null? It can be the default to empty body IMO.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -368,6 +386,25 @@ class ApiClient(
...
+            .header("Authorization", "Bearer $authToken")
+            .url(trialActivationUrl)
+            .post(body)
+            .build()
+
+        try {
+            val response = client.newCall(trialRequest).execute()
+            return response.isSuccessful;
+        } catch (e: IOException) {
+            println("Network error while starting trial: ${e.message}")
```
amanjeetsingh150, line 403: Hmmm, how are we handling IO exceptions in other places?

I recall that we're retrying in some cases. Should this failure also be retried, or would it be better to refine this message to guide the user on the appropriate action when this happens?

WDYT?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -351,6 +352,23 @@ class ApiClient(
...
+                if (response.code == 403 && errorMessage.contains("Your trial has not started yet", ignoreCase = true)) {
+                    println("\u001B[31mYour trial has not started yet.\u001B[0m")
+                    print("Would you like to start your trial now? (Y/N): ")
+                    val scanner = Scanner(System.`in`)
+                    val userInput = scanner.nextLine().trim().lowercase()
+                    if (userInput == "y") {
+                        val isTrialStarted = startTrial(authToken, baseUrl)
+                        if (isTrialStarted) {
+                          println("\u001B[32mTrial successfully started. Enjoy your 7-day free trial!\u001B[0m")
+                          return retry("Trial has started, retrying the upload")
```
amanjeetsingh150, line 364: Umm, why is this retry block here? If the trial is started can't we call the upload api directly.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
proksh: I don't think we need to retry here. Since we are showing appropriate error message here, and start trial is one time only thing for any user.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
proksh: Done

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
amanjeetsingh150: Sounds good, IMO we should tweak the message to have better actionable for user: we're experiencing connectivity issues, please try again in sometime, reach out to the slack channel in case if this doesn't work. Something along those lines.

amanjeetsingh150 (approved):

PR #2302, opened 2025-02-11
Revert "Revert "fixing upload on Robin""

felipeduartea:
Reverts mobile-dev-inc/Maestro#2300


Fishbowler:
@felipeduartea Is this still needed?

PR #2326, opened 2025-02-19
fix: Missmatch between screenshotBasedTap and hierarchyBasedTap

MarcellDr:
## Proposed changes

Assign tab-based function with its correct implementation.
Previous PR reference: https://github.com/mobile-dev-inc/Maestro/pull/787

## Testing

Tested both on Android and IOS.
IOS now using screenshot-based and Android is using hierarchy-based.
Both are still working as expected.

The performance has improved for Android, with the time per tap reducing by about one-third (previously 3 seconds, now 2 seconds).
While IOS remains the same.

## Issues fixed

-


Fishbowler:
I hope this pans out - this is gonna save a lot of people a lot of execution time!

There's a couple of failing tests here - are you able to take a look?

MarcellDr:
Hi,
I resolved the issues by removing FAST_HIERARCHY from the FakeDriver. I checked that FakeDriver.deviceInfo is set to iOS, ensuring that it should not use FAST_HIERARCHY.

amanjeetsingh150:
> Hi,
> I resolved the issues by removing FAST_HIERARCHY from the FakeDriver. I checked that FakeDriver.deviceInfo is set to iOS, ensuring that it should not use FAST_HIERARCHY.

FakeDriver, is not used on the production CLI code. Its just for tests. I'm not sure what exactly has changed for the android or iOS driver with this? Can you help elaborating the changes this means for android and iOS driver.

MarcellDr:
Sure, I'll start by providing some background.

I have an app that performs some e-commerce validations (PIN, amount, etc). 
None of these validations use the soft keyboard; instead, they rely on custom on-screen keyboard.

I tested the app on both Android and iOS and found that Android was lagging behind in performance.

Upon investigation inside Maestro repo, I discovered that Android is using additional screenshot checks, while iOS only checks its hierarchy. Furthermore, I also noticed a mismatch between the screenshot and hierarchy implementations. It seems that the current approach doesn't align with the intended design (from this PR https://github.com/mobile-dev-inc/Maestro/pull/787).

MarcellDr:
In maestro-client/src/main/java/maestro/Maestro.kt, within the performTap function.

We see that FAST_HIERARCHY is using an additional screenshot comparison, even though its function name is hierarchyBasedTap. On the other hand, screenshotBasedTap only checks the hierarchy.

So what I did change to address this inconsistency, 
I just swapped the function names for screenshotBasedTap and hierarchyBasedTap. 

As a result, my Android app's performance improved by approximately one-third, without the need for the screenshot check.

Fishbowler (approved):

Fishbowler:
Rebased

Fishbowler:
Thanks @MarcellDr! This is gonna be awesome!

Issue #2278, opened 2025-01-30
Maestro CLI errors on all commands - Fedora 41


leonardohofling:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

1. Install latest version of maestro cli on Fedora 41
2. Run maestro command on terminal

### Actual results

leonardohofling@DESKTOP-BRUUJDN ➜  ProjectX git:(feature/add-redux) ✗ **maestro**                            
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'int org.fusesource.jansi.internal.CLibrary.isatty(int)'
	at org.fusesource.jansi.internal.CLibrary.isatty(Native Method)
	at maestro.cli.DisableAnsiMixin$Companion.applyCLIMixin(DisableAnsiMixin.kt:43)
	at maestro.cli.DisableAnsiMixin$Companion.executionStrategy(DisableAnsiMixin.kt:21)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at maestro.cli.AppKt.main(App.kt:143)

leonardohofling@DESKTOP-BRUUJDN ➜  ProjectX git:(feature/add-redux) ✗ **maestro test e2e/register-flow.yaml** 
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'int org.fusesource.jansi.internal.CLibrary.isatty(int)'
	at org.fusesource.jansi.internal.CLibrary.isatty(Native Method)
	at maestro.cli.DisableAnsiMixin$Companion.applyCLIMixin(DisableAnsiMixin.kt:43)
	at maestro.cli.DisableAnsiMixin$Companion.executionStrategy(DisableAnsiMixin.kt:21)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at maestro.cli.AppKt.main(App.kt:143)

Alternative way to execute without error:
leonardohofling@DESKTOP-BRUUJDN ➜  ProjectX git:(feature/add-redux) ✗ **maestro --no-ansi --no-color**                                      
option '--no-color' should be specified only once
Usage: maestro [-hv] [--[no-]ansi] [--verbose] [-p=<platform>] [--udid=<deviceId>] [COMMAND]
  -h, --help      Display help message
      --[no-]ansi, --[no-]color
                  Enable / disable colors and ansi output
  -p, --platform=<platform>
                  (Optional) Select a platform to run on
      --udid, --device=<deviceId>
[...]

PR #2330, opened 2025-02-24
fix: use environment variables to check for terminal

mchisolm0:
Issue #2278 finds CLibrary.isatty gives an error when any `maestro` command is run unless both the --no-color and --no-ansi flags are given. This fix switches to environment variables because on CachyOS Linux and Fedora Linux users consistently got those errors.

## Proposed changes

I switched from using the CLibrary.isatty to using environment variables to check if the call is from a terminal. 

Based on the issue, it seems this may be a Linux specific problem. If the team would rather solve this another way, I am happy to help if you would like to direct me how you want it solved. Otherwise, if the team would prefer to go another direction without my help, that is also understandable. 

## Testing
I ran the `./maestro-cli/build/install/maestro/bin/maestro` command with no flags and got expected output. 
I also got the demo_app in the `e2e/` directory installed on an Android Emulator and ran the tests provided. The output is below.
Ran the `./maestro-cli/build/install/maestro/bin/maestro --device emulator-5554 test e2e/workspaces/demo_app` command.
<details>
  <summary>Output from tests</summary>

```console
Maestro on  fix-cli [!?]
➜ ./maestro-cli/build/install/maestro/bin/maestro --device emulator-5554 test e2e/workspaces/demo_app

Waiting for flows to complete...
[Passed] ai_complex (4s)
[Passed] ai_simple (0s)
[Passed] commands_optional_tournee (40s)
[Failed] commands_tour (1m 8s) (Assertion is false: id: com.google.android.inputmethod.latin:id/key_pos_shift is visible)
[Failed] fail_fast (1s) (Assertion is false: false is true)
[Failed] fail_launchApp (0s) (Unable to launch app com.nonexistent: Package com.nonexistent is not installed)
[Failed] fail_launchApp_nonDefault (0s) (Unable to launch app com.nonexistent: Package com.nonexistent is not installed)
[Failed] fail_not_found (18s) (Element not found: Id matching regex: non-existent-id)
[Failed] fail_visible (18s) (Assertion is false: id: non-existent-id is visible)
[Failed] fail_visible_extended (2s) (Assertion is false: id: non-existent-id is visible)
[Passed] fill_form (16s)
[Passed] long_input_text (49s)
[Passed] relatives (20s)
[Passed] scrollUntilVisible_timeout (3s)
[Passed] swipe (8s)

7/15 Flows Failed

```
</details>

## Issues fixed
Possible fix for Issue #2278 

## Feedback welcome
I am new to Maestro, so I would appreciate any feedback on how I solved the issue, if I should delete the comments in the code, and anything else that would help the team or project better.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/DisableAnsiMixin.kt
+++ b/maestro-cli/src/main/java/maestro/cli/DisableAnsiMixin.kt
@@ -35,6 +35,13 @@ class DisableAnsiMixin {
             return null
         }
 
+        private fun isWindowsTerminalColorCapable(): Boolean {
+            // Check for modern Windows terminals that support ANSI
+            return System.getenv("PSVersionTable") != null || // PowerShell
+                   System.getenv("WT_SESSION") != null || // Windows Terminal
+                   System.getenv("TERM_PROGRAM") == "vscode" // VS Code terminal
+        }
+
         private fun applyCLIMixin(parseResult: CommandLine.ParseResult) {
             // Find the first mixin for which of the enable-ansi parameter was specified
             val parserWithANSIOption = findFirstParserWithMatchedParamLabel(parseResult, "<enableANSIOutput>")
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/DisableAnsiMixin.kt
```diff
@@ -40,9 +40,12 @@ class DisableAnsiMixin {
             val parserWithANSIOption = findFirstParserWithMatchedParamLabel(parseResult, "<enableANSIOutput>")
             val mixin = parserWithANSIOption?.commandSpec()?.mixins()?.values?.firstNotNullOfOrNull { it.userObject() as? DisableAnsiMixin }
 
-            val stdoutIsTTY = CLibrary.isatty(CLibrary.STDOUT_FILENO) != 0
+            // Instead of using CLibrary.isatty, use environment variables to detect terminal
+            val forceDisable = System.getenv("MAESTRO_DISABLE_ANSI")?.toBoolean() ?: false
+            val isTTY = !forceDisable && System.getenv("TERM") != null
```
Fishbowler, line 45: I absolutely love the idea of dropping a dependency and simplifying the solution.

This will work on a Mac, but will this work on Windows?

mchisolm0: From my research, it looks like the answer is no, it does not. I have added a few more checks specific for windows based on my research, but the only one I have actually been able to verify returns a boolean is `($PSVersionTable -ne $null)`. I ran into issues building on my first Windows machine (seemed to be a problem with Java, somehow, but I could not figure it out.

I am working on getting my other Windows machine ready to build it and see if I can verify any of it works.

PR #2334, opened 2025-02-26
Changes iOS bundle ID to end with iosOfficial

dmitry-zaitsev:
## Proposed changes

copilot:summary


maestro-ios-driver/src/main/resources/maestro-driver-iosUITests-Runner.zip
amanjeetsingh150, line 1: Ummm is this deleted now?

Are we planning to change this location as well?

dmitry-zaitsev: I don't think it is deleted, maybe GH shows it as such? I did rebuild artifacts and made sure they were added

PR #2335, opened 2025-02-27
fix: remove cloud upload

proksh:


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -244,8 +241,8 @@ class ApiClient(
         disableNotifications: Boolean,
         deviceLocale: String? = null,
         progressListener: (totalBytes: Long, bytesWritten: Long) -> Unit = { _, _ -> },
-        projectId: String? = null,
-    ): UploadResponse {
+        projectId: String,
+    ): RobinUploadResponse {
```
Fishbowler, line 245: Is removing Robin references something we should save for another PR?

proksh: Any reason for that?

proksh: Yes, agree and updated!! Thanks!

maestro-ios-xctest-runner/maestro-driver-iosUITests/maestro_driver_iosUITests.swift
```diff
@@ -32,6 +32,13 @@ final class maestro_driver_iosUITests: XCTestCase {
         maestro_driver_iosUITests.logger.info("Will start HTTP server")
         try await server.start()
     }
+    
+    func testSomething() async throws {
+        let s = try! await XCUIApplication(bundleIdentifier: "de.komoot.berlinbikeapp")
+        let snap = s.snapshot().dictionaryRepresentation
+        
+        print(snap)
+    }
```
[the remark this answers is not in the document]
proksh: Not sure how this got included. Removed!

amanjeetsingh150 (approved):

PR #2341, opened 2025-03-02
Add DataFaker support

Fishbowler:
## Proposed changes

- Replaces the backing of existing random commands with DataFaker
- Adds some new random data commands `InputRandomCityName`, `InputRandomCountryName`, `InputRandomColorName`
- Add support for using DataFaker in JavaScript expressions (GraalJS only). This works just like it would in Java ([docs](https://www.datafaker.net/documentation/getting-started/#usage)) supports the full range of [providers](https://www.datafaker.net/documentation/providers/).

### Examples
```yaml
appId: com.example.app
jsEngine: graaljs
---
- inputRandomCityName
- inputRandomCountryName
- inputRandomColorName
```

```yaml
appId: com.example.app
jsEngine: graaljs
---
- evalScript: ${output.thisName = faker.name().firstName()}
- evalScript: ${output.thisCreditCard = faker.expression('#{finance.creditcard}')} // same as faker.finance().creditcard()
- evalScript: ${console.log(output.thisName + ' has a credit card number ' + output.thisCreditCard)}

- evalScript: ${console.log(faker.lordOfTheRings().character() + ' is in ' + faker.lordOfTheRings().location())}

- evalScript: ${output.thisNumber = faker.expression("#{number.numberBetween '1' '10'}")}
- evalScript: ${console.log('Random number between 1 and 10 = ' + output.thisNumber)}
```

which outputs:
```
 ║    ✅   Run ${output.thisName = faker.name().firstName()}                                                                             
 ║    ✅   Run ${output.thisCreditCard = faker.expression('#{finance.creditcard}')} // same as faker.finance().creditcard()              
 ║    ✅   Run ${console.log(output.thisName + ' has a credit card number ' + output.thisCreditCard)}                                    
 ║         Log messages:                                                                                                                
 ║           Erna has a credit card number 4849-2585-7581-8357                                                                          
 ║    ✅   Run ${console.log(faker.lordOfTheRings().character() + ' is in ' + faker.lordOfTheRings().location())}                        
 ║         Log messages:                                                                                                                
 ║           Aragorn is in Elostirion                                                                                                   
 ║    ✅   Run ${output.thisNumber = faker.expression("#{number.numberBetween '1' '10'}")}                                              
 ║    ✅   Run ${console.log('Random number between 1 and 10 = ' + output.thisNumber)}                                                  
 ║         Log messages:                                                                                                               
 ║           Random number between 1 and 10 = 5     
```

## Testing

Lots of unit and integration tests added.

I also had a good galumph with local flows.

## Issues fixed

Apparently not, beyond mitigating the requests for folks wanting the JS environment to have imports so that they can have faker.js

## Anything else

* h/t to @Matheeusb who started this work.
* This still needs docs.


Leland-Takamine:
Any way to have the JS faker API just match the DataFaker API so we don't have to reimplement and redocument each custom helper method?

Fishbowler:
> Any way to have the JS faker API just match the DataFaker API so we don't have to reimplement and redocument each custom helper method?

It largely already does. We've added some extra helpers to match our top-level commands, as a convenience. We don't expose things like `datafaker.person.personName()` - I looked at it, but couldn't work out how to expose top level objects in that way. The helpers could be dropped if they're too much or confusing, but I don't see them growing over time. The only other difference is using `generate` rather than `evaluate` for a bit of user-friendliness.

Leland-Takamine:
> We don't expose things like datafaker.person.personName() - I looked at it, but couldn't work out how to expose top level objects in that way.

If there's a way to accomplish this, would that be a preferable API?

Your call what you want think makes the most sense from the API standpoint

Leland-Takamine (approved):

Fishbowler:
> If there's a way to accomplish this, would that be a preferable API?

Yeah, I think so. It'd be a better match for how folks would use it in 'regular' JavaScript.

Leland-Takamine:
You should be able to do something like this:
```kotlin
class NameFaker {
    @HostAccess.Export
    fun fullName(): String {
        return "John Doe"
    }
}

class GraalFaker {
    @HostAccess.Export
    fun name() = NameFaker()
}

context.getBindings("js").putMember("datafaker", GraalFaker())
```

Which enables the JS API which matches the [datafaker API](https://www.datafaker.net/documentation/usage/)
```js
datafaker.name().fullName()
```

Fishbowler:
Having to map every function of every provider would be a lot of work and maintenance.

Would love to be able to do something like:

```kotlin
class GraalFaker {
    @HostAccess.Export
    fun faker() = Faker()
}

context.getBindings("js").putMember("faker", GraalFaker.faker)
```
then
```yaml
- evalScript: ${output.thisName = faker().name().fullName()}
```

...but I don't think JS bindings allow method access like that.

Leland-Takamine:
What about using reflection?

```kotlin
private fun HostAccess.Builder.allowAllPublicOf(clazz: Class<*>) = apply {
    clazz.methods.filter { it.declaringClass != Object::class.java && java.lang.reflect.Modifier.isPublic(it.modifiers) }
        .forEach { allowAccess(it) }
    clazz.fields.filter { java.lang.reflect.Modifier.isPublic(it.modifiers) }
        .forEach { allowAccess(it) }
}
```

maestro-client/src/main/java/maestro/js/GraalJsHttp.kt
```diff
@@ -5,12 +5,27 @@ import okhttp3.Headers
...
+import org.graalvm.polyglot.HostAccess
 import org.graalvm.polyglot.HostAccess.Export
 import org.graalvm.polyglot.proxy.ProxyObject
 
 class GraalJsHttp(
     private val httpClient: OkHttpClient
-) {
+) : GraalHostAccessible {
+
+    override fun configureHostAccess(builder: HostAccess.Builder) {
```
Leland-Takamine: Don't think this is necessary if you use `.allowAccessAnnotatedBy(HostAccess.Export::class.java)`

Fishbowler: Oh, that's _much_ tidier. Testing...

maestro-client/src/main/java/maestro/js/GraalFaker.kt
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    private fun HostAccess.Builder.allowAllPublicOf(clazz: Class<*>) {
+        if (clazz in publicClasses) return
+        publicClasses.add(clazz)
+        clazz.methods.filter {
+            it.declaringClass != Object::class.java &&
+            it.declaringClass != AbstractProvider::class.java &&
+            java.lang.reflect.Modifier.isPublic(it.modifiers)
+        }.forEach { method ->
```
Fishbowler: Took some time to narrow this down, and to review what was left to ensure nothing hairy was exposed. The hairiest thing I could find was [this](https://javadoc.io/doc/net.datafaker/datafaker/latest/net/datafaker/providers/base/BaseFaker.html#addPath(java.util.Locale,java.nio.file.Path)), but couldn't see any way of constructing the correct objects to pass in from inside the sandbox. There's nothing that takes strings or other simple types and results in FS access.

maestro-client/src/main/java/maestro/js/GraalJsHttp.kt
Fishbowler: Fixed.

PR #2350, opened 2025-03-04
Add option to use Chrome dev tools for Android WebView hierarchy

Leland-Takamine:
Add an option to use Chrome dev tools to build the Android WebView hierarchy.

eg.
```
appId: com.example.app
androidWebViewHierarchy: devtools
---
- launchApp
```

Also adds an option to studio CLI command to do the same:

```
maestro studio --android-webview-hierarchy=devtools
```

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
@@ -1,7 +1,6 @@
 package maestro.cli.command
 
 import maestro.cli.App
-import maestro.cli.CliError
 import maestro.cli.DisableAnsiMixin
 import maestro.cli.ShowHelpMixin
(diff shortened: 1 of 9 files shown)


dmitry-zaitsev (approved):

Fishbowler:
Needs accompanying docs changes

CDRussell:
Discussing this on the Maestro slack (https://mobile-dev-inc.slack.com/archives/C05EVEUTKRP/p1741860861973389; think this PR has introduced regressions re `aria` labels. 

- We have tests which use a `WebView` and makes use of `tapOn` / `assertVisible` for text which is contained in an `aria` label (different from the _visible_ text on screen).
- This has been working fine until today when all of these tests started failing
- Can’t reproduce it locally; only failing on `Robin`, and suspicion is maybe on this PR (assuming this is used by `Robin` already)

PR #2354, opened 2025-03-06
Add note about Xcode version used in GitHub Action build step.

rtharston:
## Proposed changes

I added a note in the contributing file about which version of Xcode is used when building the iOS XCTest runner app.
In the past I have run into issues where our CI build used an older version of Xcode (as is the case here) and contributors used a newer version and would inadvertently add changes that would build locally but not in CI.
This note makes contributors aware of this issue.

## Testing

N/A

## Issues fixed

Hole in documentation.


Fishbowler (approved):

Fishbowler:
Thanks! That's a nice tip!

PR #2364, opened 2025-03-12
fix: properly set closing driver state

amanjeetsingh150:
## Proposed changes

We were not properly close variable state that skips closing driver if grpc server shuts down suddenly in between of test execution. This is already done on all the calls except the viewHierarchy call.

## Testing

- [ ] Locally
- [ ] Cloud

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -398,9 +398,23 @@ class AndroidDriver(
             blockingStubWithTimeout.viewHierarchy(viewHierarchyRequest {})
         } catch (throwable: StatusRuntimeException) {
             val status = Status.fromThrowable(throwable)
-            if (status.code == Status.Code.DEADLINE_EXCEEDED) {
-                LOGGER.error("Timeout while fetching view hierarchy")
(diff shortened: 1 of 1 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -398,9 +398,23 @@ class AndroidDriver(
...
-            if (status.code == Status.Code.DEADLINE_EXCEEDED) {
-                LOGGER.error("Timeout while fetching view hierarchy")
-                throw MaestroException.DriverTimeout("Android driver unreachable")
+            when (status.code) {
+                Status.Code.DEADLINE_EXCEEDED -> {
+                    LOGGER.error("Timeout while fetching view hierarchy")
+                    closed = true
+                    throw MaestroException.DriverTimeout("Android driver unreachable")
+                }
+                Status.Code.UNAVAILABLE -> {
```
herval, line 407: unavailable means the server is busy processing some other request, so it's technically retryable. Is that not the case in our case?

amanjeetsingh150: Yup, this is retryable in this case. The only scenario where it’s not retryable is when the status is unavailable and cause is an IO exception. That indicates the server is no longer reachable, so we set close = true in that case.

PR #2367, opened 2025-03-12
fix(cli): proper android start and listing

felipeduartea:
## Proposed changes

- Started using the device name in my connectedDevices description (for android)
- Added an await to the start device process to finish, to avoid inconsistent information from dadb 
- Started using the last element from dadb list instead of the first (the most recent one is always the one I'm creating now)


trindadetiago (commented):
good 👍

PR #2391, opened 2025-03-22
properly closing maestro sessions

felipeduartea:
## Proposed changes

Noticed that the close function in maestroSessionManager wasn't really closing sessions. That happened because the SessionStore creates the session again with the heartbeat.

To handle this correctly we need to cancel the hearbeat before closing the session


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -70,13 +71,16 @@ object MaestroSessionManager {
         )
         val sessionId = UUID.randomUUID().toString()
 
+        val isHeartbeatActive = AtomicBoolean(true) //Controlling heartbeat with an atomic bool
```
felipeduartea, line 74: I'm not really sure about wether this is a good way to handle this, so feedbacks appreciated

PR #2414, opened 2025-04-07
Android WebView devtoools fixes

Leland-Takamine:
* Use custom socket for chrome devtools so we can leverage OkHttpClient
* Handle expected IOExceptions for WebView race conditions


maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
```diff
@@ -49,23 +67,36 @@ private data class DevToolsResponse<T>(
...
 
-class DadbChromeDevToolsClient(private val dadb: Dadb) {
+class DadbChromeDevToolsClient(private val dadb: Dadb): Closeable {
 
     private val json = jacksonObjectMapper().configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
 
-    private val httpClient = DadbHttpClient(dadb)
+    private val okhttp = OkHttpClient.Builder()
+        .dadb(dadb)
+        .build()
```
amanjeetsingh150, line 76: We have a HttpClient object that has a okhttp client that we can leverage.

amanjeetsingh150 (approved):

PR #2421, opened 2025-04-11
Launch app through xctest request instead of simctl #2374

Sanskargautam:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
+++ b/maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
@@ -89,6 +89,8 @@
 		94A90DDE298AE72A006EB769 /* XCUIElement+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 94A90DDD298AE72A006EB769 /* XCUIElement+Extensions.swift */; };
 		9811C9092C49751D00DDACA0 /* ScreenSizeHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9811C9082C49751D00DDACA0 /* ScreenSizeHelper.swift */; };
 		F328D3E62A2A98E7000546D3 /* StringExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = F328D3E52A2A98E7000546D3 /* StringExtensions.swift */; };
+		F4C52E462D9E612D008E0A1A /* LaunchAppHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4C52E452D9E612B008E0A1A /* LaunchAppHandler.swift */; };
+		F4C52E482D9E8DF1008E0A1A /* LaunchAppRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4C52E472D9E8DE9008E0A1A /* LaunchAppRequest.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -280,6 +282,8 @@
 		94A90DDD298AE72A006EB769 /* XCUIElement+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCUIElement+Extensions.swift"; sourceTree = "<group>"; };
 		9811C9082C49751D00DDACA0 /* ScreenSizeHelper.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ScreenSizeHelper.swift; sourceTree = "<group>"; };
 		F328D3E52A2A98E7000546D3 /* StringExtensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StringExtensions.swift; sourceTree = "<group>"; };
+		F4C52E452D9E612B008E0A1A /* LaunchAppHandler.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LaunchAppHandler.swift; sourceTree = "<group>"; };
+		F4C52E472D9E8DE9008E0A1A /* LaunchAppRequest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LaunchAppRequest.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -520,6 +524,7 @@
 		9407362C2940C9F900A72E99 /* Models */ = {
 			isa = PBXGroup;
 			children = (
+				F4C52E472D9E8DE9008E0A1A /* LaunchAppRequest.swift */,
 				521C1F132D86D3840024EE42 /* TerminateAppRequest.swift */,
 				9407362D2940CA1600A72E99 /* GetRunningAppRequest.swift */,
 				32762AF92966DC8300FB69BD /* SwipeRequest.swift */,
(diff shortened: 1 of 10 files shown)


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/LaunchAppRequest.swift
```diff
@@ -0,0 +1,5 @@
+struct LaunchAppRequest: Codable {
+    let bundleId: String
+    let arguments: [String]?
```
amanjeetsingh150, line 3: Is this not passed as map here:

```
data class LaunchAppRequest(val appId: String, val launchArguments: Map<String, Any>, val maestroSessionId: UUID?)
```

How is this converted to array? Also not sure why we need environment here.

We just need support launch arguments with all type of values mentioned in docs here:

https://docs.maestro.dev/api-reference/commands/launchapp#sending-launch-arguments

amanjeetsingh150 (commented):
How do you plan to test these changes? It would be helpful to include an E2E test that we can run locally.

You can create a new directory under e2e for this. The tests should cover the following scenarios:

- App launch works on real iOS devices
- App launch works on simulators
- App launch arguments work correctly for all of these types: String, Boolean, Double, and Integer

maestro-ios-xctest-runner/maestro-driver-iosUITests/LaunchAppHandlerTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        request.httpBody = try JSONEncoder().encode(launchRequest)
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+
+        // Make HTTP call to trigger app launch
+        let (_, response) = try await URLSession.shared.data(for: request)
+        XCTAssertEqual((response as? HTTPURLResponse)?.statusCode, 200)
+
+        // Verify app is running and arguments are displayed
+        let app = XCUIApplication(bundleIdentifier: launchRequest.appId)
+        app.launch()
```
amanjeetsingh150, line 45: I'm a bit unclear on how this test works — do you actually have the app (com.example.example) running locally?

The test is supposed to verify that the app launches, but it looks like we're launching it ourselves in the test, which feels off.

It also seems to test the request payload rather than verifying that the app was launched with the correct arguments.

What I was expecting from this test:

* The app is actually launched
* It’s launched with the correct arguments from the request

Right now, it feels more like we’re testing the request structure, not the actual implementation behavior.

maestro-ios/src/main/java/ios/LocalIOSDevice.kt
```diff
@@ -102,9 +102,9 @@ class LocalIOSDevice(
 
     override fun launch(
         id: String,
-        launchArguments: Map<String, Any>,
-    ) {
-        deviceController.launch(id, launchArguments)
+        launchArguments: Map<String, Any>
+    ): Result<Unit, Throwable> {
```
amanjeetsingh150, line 106: Why is this return type added here? This doesn't exist on the main

maestro-ios-driver/src/main/kotlin/device/IOSDevice.kt
```diff
@@ -88,7 +88,7 @@ interface IOSDevice : AutoCloseable {
     fun launch(
         id: String,
         launchArguments: Map<String, Any>,
-    )
+    ): Result<Unit, Throwable>
```
amanjeetsingh150, line 91: This should not be changed. Not sure I understand intent behind this?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/LaunchAppHandler.swift
```diff
@@ -1,26 +1,51 @@
...
-        
-        return HTTPResponse(statusCode: .ok)
+        // Set launch arguments if available
+        let arguments = requestBody.launchArguments
+        let flattenedArgs = arguments.flatMap { key, value -> [String] in
+            switch value.value {
+            case let bool as Bool:
+                return bool ? ["--\(key)"] : []
+            default:
+                return ["--\(key)", "\(value)"]
```
amanjeetsingh150, line 29: I am not sure I got this flatmap logic? 

Why do we want to check the bool arguments separately here. Can't we just pass arguments as is?

Sanskargautam: The reason we're using flatMap with a special case for Bool is because of how launchArguments work in XCUIApplication. It's not a dictionary but an array of strings, similar to command-line arguments.

For boolean flags:

If the value is true, we include it as a standalone flag like --featureFlagA.

If the value is false, we omit it entirely (which is the typical convention for CLI-like flags).

maestro-ios-xctest-runner/driver-verification-tests.xctestplan
```diff
@@ -0,0 +1,28 @@
...
+        "maestro_driver_iosUITests\/testHttpServer()"
+      ],
+      "target" : {
+        "containerPath" : "container:maestro-driver-ios.xcodeproj",
+        "identifier" : "52049BF2293503A200807AA3",
+        "name" : "maestro-driver-iosUITests"
+      }
+    }
+  ],
+  "version" : 1
```
amanjeetsingh150, line 27: Why do we need this?

maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
```diff
@@ -1111,6 +1139,14 @@
 				version = 0.22.0;
 			};
 		};
+		F4FD416D2DCA33440063BDB3 /* XCRemoteSwiftPackageReference "AnyCodable" */ = {
```
amanjeetsingh150, line 1142: What’s the intent behind this new library? Do we already have an internal API or utility that could be reused instead of adding a new dependency?

maestro-ios-xctest-runner/maestro-driver-iosUITests/LaunchAppHandlerTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        let app = XCUIApplication(bundleIdentifier: launchRequest.appId)
+        XCTAssertTrue(app.wait(for: .runningForeground, timeout: 5), "App did not launch")
+
+        let expected = [
+                    "--featureFlagA",
+                    "--username", "maestroUser",
+                    "--retryCount", "3"]
+        
+        let actual = flattenLaunchArguments(launchRequest.launchArguments)
+                XCTAssertEqual(actual, expected)
```
amanjeetsingh150, line 53: This is still asserting from request right? How is this asserting what is passed on app com.example.example?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/LaunchAppHandler.swift
amanjeetsingh150: Ummm, not sure I still got fully. For example, if we omit it and if someone is expecting false value on their app, will it give them nil or false?

I think right now your test also doesn't seem to cover this case. If you can ensure we have all the types covered in the test that would help here.

maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
Sanskargautam: The intent behind using AnyCodable is to support the flexible structure of the launchArguments payload that Maestro sends during the launchApp request. Specifically, Maestro sends this field as a Map<String, Any>, where the values can be of varying types (String, Bool, Int, etc.).

Since Swift’s Codable doesn’t natively support [String: Any], AnyCodable was introduced to cleanly decode this structure without writing extensive boilerplate or custom decoding logic.

I'm happy to switch to an internal utility if we already have something that supports decoding heterogeneous dictionaries like this — let me know if there’s a recommended alternative!

PR #2425, opened 2025-04-14
maestro: making AI commands work

trindadetiago:
## Proposed changes

Creating an interface for AIEnginePrediction. Using the CloudAIEnginePrediction implementation to call maestro's api to get the AI commands.

Pass the API Key from Run Tests - if present - to use in the API call

## Testing

Tested running a workflow with AI commands both before and after the changes, with and without setting the MAESTRO_CLOUD_API_KEY, and passing the --api-key param.

## Issues fixed

Allow for maestro.dev to use the AI commands

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/auth/Auth.kt
+++ b/maestro-cli/src/main/java/maestro/cli/auth/Auth.kt
@@ -73,11 +73,11 @@ class Auth(
     private val apiClient: ApiClient
 ) {
 
-    fun getAuthToken(apiKey: String?): String {
+    fun getAuthToken(apiKey: String?, triggerSignIn: Boolean = true): String {
         return apiKey // Check for API key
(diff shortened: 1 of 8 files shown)


maestro-cli/src/main/java/maestro/cli/auth/Auth.kt
```diff
@@ -72,6 +73,20 @@ class Auth(
...
 ) {
 
+    fun getAuthToken(apiKey: String?): String {
+        return apiKey // Check for API key
+            ?: getCachedAuthToken() // Otherwise, if the user has already logged in, use the cached auth token
+            ?: EnvUtils.maestroCloudApiKey() // Resolve API key from shell if set
+            ?: triggerSignInFlow() // Otherwise, trigger the sign-in flow
+    }
+
+    fun getAuthTokenSimple(apiKey: String?): String? {
```
herval, line 83: why two functions? You can have the same w/ a flag instead

herval (approved):

PR #2428, opened 2025-04-15
adding metadata logs to AI commands

trindadetiago:
## Proposed changes

Adding logs for AI commands by updating its metadata

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -209,7 +209,7 @@ class Orchestra(
 
                 try {
                     try {
-                        executeCommand(evaluatedCommand, config)
+                        executeCommand(evaluatedCommand, config, command)
                         onCommandComplete(index, command)
                     } catch (e: MaestroException) {
                         val isOptional =
@@ -275,7 +275,7 @@ class Orchestra(
     /**
      * Returns true if the command mutated device state (i.e. interacted with the device), false otherwise.
      */
-    private fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?): Boolean {
+    private fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?, originalCommand: MaestroCommand): Boolean {
         val command = maestroCommand.asCommand()
 
         return when (command) {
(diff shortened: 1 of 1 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -393,6 +396,9 @@ class Orchestra(
             onCommandGeneratedOutput(command, defects, imageData)
 
             val word = if (defects.size == 1) "defect" else "defects"
+            val message = "Found ${defects.size} possible $word: ${defects.joinToString { it.reasoning }}"
+            updateMetadata(maestroCommand, metadata.copy(logMessages = metadata.logMessages + message))
```
amanjeetsingh150, line 400: Why is the AI reasoning treated as logMessage when it can be a different field on metadata. Log messages are support to be console logs on js scripts inside maestro.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -382,6 +382,9 @@ class Orchestra(
             throw MaestroException.CloudApiKeyNotAvailable("`MAESTRO_CLOUD_API_KEY` is not available. Did you export MAESTRO_CLOUD_API_KEY?")
         }
 
+        val maestroCommand = MaestroCommand(command)
```
amanjeetsingh150, line 385: Why are we creating another copy object of maestro command isn't this same thing passed from executeCommand function?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -411,15 +417,18 @@ class Orchestra(
...
```
amanjeetsingh150, line 420: Same seems like we can have it from executeCommand function right?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -275,7 +275,7 @@ class Orchestra(
     /**
      * Returns true if the command mutated device state (i.e. interacted with the device), false otherwise.
      */
-    private fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?): Boolean {
+    private fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?, originalCommand: MaestroCommand): Boolean {
```
amanjeetsingh150, line 278: What is difference between maestro command and originalCommand. Can't we use maestroCommand here?

PR #2429, opened 2025-04-16
[MA-3186] Setting up iOS drivers for both real iOS device and simulators

amanjeetsingh150:
## Proposed changes

1. Write script to generate build products for iPhoneSimulator and iOS device.
2. Rough launch app implementation on iOS driver XCUITest.
3. Use localhost (127.0.0.1) instead of IPV6 implementation.
4. Introduce device type to configure the installation of drivers for real devices and simulators.
5. Remove bunch of unused stuff:
- maestro sessionId during launch, we don't use that anymore
- `enableXCTestOutputFileLogging` config on LocalXCTestInstaller. Its true always.

## Testing and Pending things

- [x] Distinguish android emulator and physical devices
- [x] DeviceCtl list on session manager.
- [x] Driver works on simulator
- [x] Close calls migrate from simctl command to real device support.
- [x] Driver works on real iOS device
- [x] Change launchApp back to deviceController for now.
- [x] E2E test run on iOS locally
- [x] E2E test run on android locally
- [x] Multiple device connected case: multiple android and multiple iOS


dmitry-zaitsev (approved):

abedolinger:
Thanks @amanjeetsingh150, this is huge!

AksAman:
@amanjeetsingh150 Quick follow-up after the merge - can you help with running Maestro on a real iPhone?

I ran `driver-setup` with our Apple ID (drivers installed successfully),
```bash
maestro driver-setup --apple-team-id FCHWS325SB
✅ Building iOS driver for generic/platform=iphoneos...

✅ Drivers successfully set up for destination generic/platform=iphoneos
```

 then tried:

```bash
maestro --device $DEVICE_ID test $TEST_FILE
```

but get:

```
Device $DEVICE_ID[redacted] was requested, but it is not connected.
```

Meanwhile `xcrun xctrace list devices` shows the device as connected, and I can deploy/run the React Native app on that device fine. We need a real device because the tests use Bluetooth libraries (simulators aren’t an option).

For context I’m on Maestro CLI v2.0.5. Any pointers, flags, or docs you can share to get Maestro to recognize the device would be super helpful — thanks!

Issue #1679, opened 2024-02-19
[Feature Request] Support Orientation Change in Maestro Studio


mdethlefs:
**Is your feature request related to a problem? Please describe.**
When there is an orientation change in the app (for example if you open a video player) maestro studio won't update and you are unable to inspect the screen.

**Describe the solution you'd like**
Meastro Studio should update it's state according to the orientation of the device. You should be able to inspect all UI elements.

**Describe alternatives you've considered**
I tried restarting Maestro Studio after the orientation change. This only seems to work on Android in my case. When I try this on iOS it doesn't work.

verbalking:
Is there any work round? 😭

bartekpacia:
@verbalking Not for now, but we're aware of the problem. It's in our backlog. Stay tuned!

abedolinger:
@verbalking it's not ideal, but the touch areas on the device in Maestro Studio are actually correct, even though the visuals/element outlines are wrong.  So you can tap where the element "really" is (refer to your device) and it will be selected.

@bartekpacia +1 for this feature, it would be super helpful 😄

PR #2431, opened 2025-04-16
Add rotate simulator feature for iOS

abedolinger:
## Proposed changes

Add `rotateDevice` command for iOS simulator, so Maestro can properly see the screen if (for example) a portrait-display app changes orientation programmatically to landscape.

copilot:summary

video:

https://github.com/user-attachments/assets/3c1119b3-a183-4e95-b212-74083e3cdfd7

## Testing

1. Tested rotation of iOS simulator
2. Tested no-op on Android

## Issues fixed

https://github.com/mobile-dev-inc/Maestro/issues/1679

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ users still use Java 8.
 
 For development, you need to use Java 11 or newer.
 
-If you made changes to the CLI, rebuilt it with `./gradlew :maestro-cli:installDist`. This will generate a startup shell
+If you made changes to the CLI, rebuild it with `./gradlew :maestro-cli:installDist`. This will generate a startup shell
 script in `./maestro-cli/build/install/maestro/bin/maestro`. Use it instead of globally installed `maestro`.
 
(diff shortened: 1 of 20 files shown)


herval:
we recently merged a `setOrientation` command - it should work for you?

abedolinger:
> we recently merged a `setOrientation` command - it should work for you?

Hi @herval, thanks for the pointer, and glad this was merged! It doesn't fit my use case unfortunately. Our app changes orientation programmatically. Then Maestro can't "see" half the screen. I'll try to explain in ascii

```
------------
|    |     | 
|    |     | - The window Maestro "sees" - it seems it expects the 
------------   simulator to be oriented to the side
|    |
|    | - Simulator window: still oriented upright, but the app is
------   landscape. Selectors can't find content in this region
```
When the simulator window is rotated (Device -> Rotate Right/Left), all the screen content is selectable again.

Does that make sense?

christos:
We are also coming up against this issue. Our (React Native) app launches content (games/videos) and it programmatically rotates the device orientation in order to present them in landscape format. 

Maestro Studio doesn't pick up the change, instead is sees only a partial view just as @abedolinger described above.

Fishbowler:
We've made some relevant fixes in the area you've described.
If there are problems when the device rotates programmatically, do please open an issue.
Given `setOrientation`, I think this is obsolete?

PR #2449, opened 2025-04-30
fix: studio mode should not foreground running app

amanjeetsingh150:
## Proposed changes

We introduced prebuilt mode on iOS to make driver setup faster and more consistent. While it's working as expected, it caused a regression in the Maestro Studio flow by sending the foreground app to the background.

This PR fixes that by:
1. **Avoiding backgrounding the app**: We switch back to using xcodebuild for flows where keeping the app in the foreground is necessary (e.g., Maestro Studio).
2. **Fixing infinite loop in device detection**: The recent changes for iOS physical device detection logic causes an infinite loop during local installation because user input isn't possible in the current setup. This PR disables that behavior in local mode.

## Testing

Locally

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
@@ -70,6 +70,7 @@ object MaestroSessionManager {
         isStudio: Boolean = false,
         isHeadless: Boolean = isStudio,
         reinstallDriver: Boolean = true,
+        prebuiltIOSRunner: Boolean,
         block: (MaestroSession) -> T,
     ): T {
         val selectedDevice = selectDevice(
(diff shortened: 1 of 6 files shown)


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -291,14 +296,16 @@ object MaestroSessionManager {
         openDriver: Boolean,
         driverHostPort: Int,
         reinstallDriver: Boolean,
+        isStudio: Boolean
```
herval, line 299: Why do we need to introduce a new flag, and why call it this?

herval:
I don’t think this fixes the issue - it’s just sidestepping it. We need to fix the prebuiltRunner mode, no?

PR #2459, opened 2025-05-02
flag to skip interactive device selection by picking a device-index

herval:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
@@ -81,6 +81,13 @@ class PrintHierarchyCommand : Runnable {
     )
     private var compact: Boolean = false
 
+    @CommandLine.Option(
+        names = ["--device-index"],
+        description = ["The index of the device to run the test on"],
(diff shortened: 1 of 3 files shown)


maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -81,6 +81,13 @@ class PrintHierarchyCommand : Runnable {
     )
     private var compact: Boolean = false
 
+    @CommandLine.Option(
+        names = ["--device-index"],
+        description = ["The index of the device to run the test on"],
+        hidden = true
+    )
+    private var deviceIndex: Int? = null
+
```
amanjeetsingh150, line 90: Do we just need this on hierarchy command?

amanjeetsingh150 (approved):

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
herval: `test` uses the first device, `start-device` requires a device. So I'm guessing so, yes

PR #2473, opened 2025-05-08
[MA 3181] iOS device info frame creating incorrect hierarchy.

amanjeetsingh150:
## Proposed changes

In large devices and some screens app frame doesn't account window insets from springboard, this fixes the logic by adjusting the app hierarchy.

This impacted hierarchy by giving incorrect bound for some elements affecting all the commands that uses hierarchy.

## Testing

- [x] Locally


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/AXElement.swift
```diff
@@ -16,7 +21,7 @@ extension AXFrame {
 
 struct AXElement: Codable {
     let identifier: String
-    let frame: AXFrame
+    var frame: AXFrame
```
herval, line 24: why var?

maestro-ios-xctest-runner/maestro-driver-iosUITests/ViewHierarchyHandleTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        request.httpMethod = "POST"
+        request.httpBody = try JSONEncoder().encode(ViewHierarchyRequest(appIds: [], excludeKeyboardElements: false))
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        
+        let springboardApp = await XCUIApplication(bundleIdentifier: "com.apple.springboard")
+        let testApp = await XCUIApplication(bundleIdentifier: "org.wikimedia.wikipedia")
+        let springboardFrame = await springboardApp.frame
+        let testAppFrame = await testApp.frame
+        
+        let offsetX = springboardFrame.width - testAppFrame.width
```
herval, line 34: it would make more sense to have the expected numbers hardcoded in the test - the way it is right now, this could mean that the actual implementation is wrong, and the test would quietly accept it?

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Models/AXElement.swift
amanjeetsingh150: Changed back to immutable.

maestro-ios-xctest-runner/maestro-driver-iosUITests/ViewHierarchyHandleTests.swift
amanjeetsingh150: In general, hardcoding expected values is a solid way to catch regressions. But in this case, we avoided that because the offset between the springboard and app frames varies by simulator/device (e.g., iPhone 14 Pro vs SE). Hardcoding values like +30 or +12 would require forcing a specific destination, which we want to avoid for test portability.

We're asserting that the actual app frame reflects the runtime-calculated offset between springboard and app dimensions. So if the implementation is wrong (e.g. offset isn’t applied), this test still fails. The test is dynamic but still strict since we assert the calculation here.

PR #2474, opened 2025-05-08
[MAE 167] breaking deserialization

amanjeetsingh150:
## Proposed changes

Made os version number and name in device properties nullable. We got 2 reports where we got to know that we might not have name and os number.

Updated the unit tests as well.

## Testing

- [x] Locally

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/IOSDevice.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/IOSDevice.kt
@@ -19,7 +19,7 @@ data class DeviceCtlResponse(
     @JsonIgnoreProperties(ignoreUnknown = true)
     data class Device(
         val identifier: String,
-        val deviceProperties: DeviceProperties,
+        val deviceProperties: DeviceProperties?,
         val hardwareProperties: HardwareProperties,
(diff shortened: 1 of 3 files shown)


maestro-ios-driver/src/main/kotlin/util/IOSDevice.kt
```diff
@@ -19,7 +19,7 @@ data class DeviceCtlResponse(
     @JsonIgnoreProperties(ignoreUnknown = true)
     data class Device(
         val identifier: String,
-        val deviceProperties: DeviceProperties,
+        val deviceProperties: DeviceProperties?,
```
herval, line 22: Not sure I understand the cause or the fix. It’s making 3 properties nullable. Why weren’t they before? Those aren’t new attributes afaik, when were they introduced?

amanjeetsingh150: These fields were added along with physical device support and are used to show the connected device's description. In my local testing, all three values were always present, so I overlooked the possibility that any of them could be null. This update now handles those cases properly.

I'm also in touch with a customer to look this.

PR #2487, opened 2025-05-15
[MA 3258] Introducing debug info on element not found exceptions

amanjeetsingh150:
## Proposed changes

1. Introduce a debug info on maestro errors to send debug messages on failures. This is to help users guide how they show be changing commands if something errors out on assertions specifically scrollUntilVisible.
2. scrollUntilVisible command also doesn't show other attributes like speed, timeout which are even there by default. This PR includes them in description

Screenshot 2025-05-15 at 3 28 36 PM

## Testing

- [x] Locally

### DIFF
--- a/maestro-client/src/main/java/maestro/Errors.kt
+++ b/maestro-client/src/main/java/maestro/Errors.kt
@@ -19,6 +19,10 @@
 
 package maestro
 
+data class DebugInfo (
+    val debugMessage: String
+)
+
 sealed class MaestroException(override val message: String) : RuntimeException(message) {
 
     class UnableToLaunchApp(message: String) : MaestroException(message)
@@ -32,12 +36,14 @@ sealed class MaestroException(override val message: String) : RuntimeException(m
     open class AssertionFailure(
         message: String,
         val hierarchyRoot: TreeNode,
+        val debugInfo: DebugInfo,
     ) : MaestroException(message)
 
     class ElementNotFound(
         message: String,
         hierarchyRoot: TreeNode,
-    ) : AssertionFailure(message, hierarchyRoot)
+        debugInfo: DebugInfo,
+    ) : AssertionFailure(message, hierarchyRoot, debugInfo)
 
     class CloudApiKeyNotAvailable(message: String) : MaestroException(message)
 
--- a/maestro-client/src/main/java/maestro/Errors.kt
+++ b/maestro-client/src/main/java/maestro/Errors.kt
@@ -19,10 +19,6 @@
 
 package maestro
 
-data class DebugInfo (
-    val debugMessage: String
-)
-
 sealed class MaestroException(override val message: String) : RuntimeException(message) {
 
     class UnableToLaunchApp(message: String) : MaestroException(message)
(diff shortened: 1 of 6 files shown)


maestro-client/src/main/java/maestro/Errors.kt
```diff
@@ -19,6 +19,10 @@
 
 package maestro
 
+data class DebugInfo (
+    val debugMessage: String
```
herval, line 23: what do we expect this object to contain, other than a message?

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -410,39 +410,10 @@ class Maestro(
...
-                "No element that matches regex: $regex",
-                viewHierarchy().root
-            )
-    }
-
     fun viewHierarchy(excludeKeyboardElements: Boolean = false): ViewHierarchy {
         return ViewHierarchy.from(driver, excludeKeyboardElements)
     }
 
-    fun findElementByIdRegex(regex: Regex, timeoutMs: Long): UiElement {
```
herval, line 427: are these not used anywhere?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1046,6 +1079,7 @@ class Orchestra(
             ) ?: throw MaestroException.ElementNotFound(
                 "Element not found: $description",
                 parentViewHierarchy.root,
+                DebugInfo("Element with $description not found. Check the UI hierarchy in debug artifacts to verify if the element exists.")
```
herval, line 1082: can this include any helpful information? Eg if the selector is wrong, is there something on the screen with a similar name, perhaps?

maestro-client/src/main/java/maestro/Maestro.kt
amanjeetsingh150: Nope

maestro-client/src/main/java/maestro/Errors.kt
amanjeetsingh150: Yeah I have changed it back to string for now, from the discussion yesterday. Want to keep it simple.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -371,6 +371,7 @@ class Orchestra(
             throw MaestroException.AssertionFailure(
                 message = "Assertion is false: ${command.condition.description()}",
                 hierarchyRoot = maestro.viewHierarchy().root,
+                debugInfo = DebugInfo("Assertion '${command.condition.description()}' failed. Check the UI hierarchy in debug artifacts to verify the element state and properties.")
```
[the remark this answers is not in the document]
amanjeetsingh150: I still see this as a useful improvement — many users don’t realize they should be using the view hierarchy to debug issues, so surfacing it more clearly is definitely a step up from the current experience.

When it comes to assertion failures, it’s often due to a selector property changing or the element not being visible at all. Since property changes might represent real regressions, it’s hard to generalize suggestions. So IMO, the best approach is to guide users toward inspecting the hierarchy — that way, we’re giving them the right tool and context to debug different types of failures more effectively.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
amanjeetsingh150: > Eg if the selector is wrong, is there something on the screen with a similar name, perhaps?

I thought about this but problem with this is we might be asking them to hide real regressions. What if accessibility text broke and we ask them to change, but on contrary its actually a regression 🤔

amanjeetsingh150: How about:

```
Assertion '${command.condition.description()}' failed. 
Check the UI hierarchy in debug artifacts to verify the element state and properties.

Possible causes:
```

PR #2488, opened 2025-05-15
make run flows cancellable if coroutine scope is cancelled + pause and resume features

herval:
The goal here is to make a Maestro Orchestra coroutine-cancellable - if the context where it's running is cancelled, it'll skip heavy setup steps (eg initializing the js engine). It'll still go through all the commands on the flow, but marking them all as SKIPPED instead of executing them.


amanjeetsingh150 (commented):
Direction looks great, I'm trying to think what would be best way to add a test for this to ensure, cancelling coroutine:

1. Skips rest of commands
2. Expect to release resources which were acquired

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -175,10 +181,20 @@ class Orchestra(
...
-        initAndroidChromeDevTools(config)
+        if (!coroutineContext.isActive) {
+            logger.info("Flow cancelled, skipping initAndroidChromeDevTools...")
+        } else {
+            initAndroidChromeDevTools(config)
+        }
 
         commands
             .forEachIndexed { index, command ->
+                if (!coroutineContext.isActive) {
```
amanjeetsingh150, line 192: I'm thinking how this should look like for cli, should there be any controls there? I believe not useful, because users usually press control + c to break from flows.

Although I'm not sure about CI

herval: ctrl+c kills the entire Java process, so it's not gonna impact any of this. Coroutine cancellation is for when you have a process still running and you want to cancel it

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -175,10 +181,20 @@ class Orchestra(
...
+                    logger.info("Flow cancelled, skipping command...")
+                    onCommandSkipped(index, command)
+                    return@forEachIndexed
+                }
```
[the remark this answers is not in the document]
herval: > Let's say if coroutine context is cancelled where should we be releasing resources that got acquired during execution?

on the same place they get released without cancellation?

is there anything on this PR _not_ being released with this change?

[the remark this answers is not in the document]
herval: (I'll adjust the jsEngine thing to not use lateinit, not sure it makes sense)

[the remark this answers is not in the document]
herval: hmm I see some funkiness with the jsEngine thing being a lateinit. I'll keep it being initialized even if the routine is cancelled to avoid some unexpected behavior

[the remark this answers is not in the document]
herval: verified that screenrecording.close will be called regardless of cancellation, so that one seems safe?

trindadetiago:
Proposed additional changes:

- Adding PAUSE and RESUME options through the FlowController;
- Skip commands when cancelling also through the executeCommand function:
```
private suspend fun executeCommand(maestroCommand: MaestroCommand, config: MaestroConfig?): Boolean {
        val command = maestroCommand.asCommand()

        if (!coroutineContext.isActive) {
            throw CommandSkipped
        }
```
This is important for sub-flows; Otherwise, when running a sub-flow and cancelling mid-run, it will go all the way through until finished.


---

Created 3 more integration tests! 1 for cancelling (apart from the already existent one) and 2 for the pause and resume features

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -72,6 +75,42 @@ sealed class CommandOutput {
...
+    override fun resume() { 
+        _isPaused = false
+    }
+    
+    override val isPaused: Boolean get() = _isPaused
+
+    override fun setJsEngine(engine: JsEngine?) {
+        _jsEngine = engine
+    }
+
```
trindadetiago: When pausing and resuming, the JsEngine is lost. This allow for it to continue

example/src/main/kotlin/Main.kt
```diff
@@ -57,6 +58,8 @@ private fun executeAndroidCommands() {
     )
 
     maestro.use {
-        Orchestra(it).executeCommands(commands)
+        runBlocking {
+            Orchestra(it).executeCommands(commands)
+        }
     }
```
amanjeetsingh150: I feel we can even delete this, nobody uses this

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -252,6 +303,7 @@ class Orchestra(
         } else {
             httpClient?.let { RhinoJsEngine(it, platform) } ?: RhinoJsEngine(platform = platform)
         }
+        (flowController as? DefaultFlowController)?.setJsEngine(jsEngine)
```
amanjeetsingh150: Not sure why to cast here?, just inject this with default value.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
herval: why is it lost?

trindadetiago: It’s due to the pause/resume behavior.

When we pause with delay, the coroutine switches threads when resuming, but the JavaScript engine is tied to the thread it was created on. That thread switch causes it to lose context.

I fixed this now by using Thread.sleep instead of delay, which keeps the engine on the same thread during pause and resume, avoiding the issue. It is a better approach

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -72,6 +75,36 @@ sealed class CommandOutput {
...
+
+class DefaultFlowController : FlowController {
+    private var _isPaused = false
+    
+    override suspend fun waitIfPaused() {
+        while (_isPaused) {
+            if (!coroutineContext.isActive) {
+                break
+            }
+            Thread.sleep(500)
```
amanjeetsingh150, line 93: The right way to suspend coroutine is delay, I don't think this should be needed

example/src/main/kotlin/Main.kt
trindadetiago: Delete what? This whole piece of code? Or the file

amanjeetsingh150: The whole file

PR #2489, opened 2025-05-15
Do no retry taps by default

### LINKED ISSUE #1202: Android double tap issues - impacting Android 13 a lot
**Describe the bug**
There is a known issue of Android double tap:

https://maestro.mobile.dev/troubleshooting/known-issues#android-accidental-double-tap

This seems to be impacting Android 13 a lot more noticeably than previous Android versions.

A number of my tests would start sporadically failing in random places, and it occurs when a tap on an element has occurred, and then Maestro thinks the tap wasn't successful so taps again, and since the animation has already taken place and proceeded to the next screen, it ends up tapping a different element and causing the test to fail.

The specified solution of `retryTapIfNoChange: false` does work, but it's not great having to add these to every tapOn command just in case it should happen. Would be great if a solution could be looked in to.

**To Reproduce**
Run tests on Android, specifically 13 and sporadic failures can start to happen for taps

**Expected behavior**
Double taps shouldn't happen, and I shouldn't have to add `retryTapIfNoChange: false` to every tapOn command just in case

- Maestro version v1.29.0
- Platform: Android, specifically 13
- Framework: Native Android
- Device model and OS version: -
- Simulator, emulator or physical device: Both
- Host: Mac

Leland-Takamine:
**WARNING:** This is a breaking change for flows that are benefitting from the previous default behavior of retrying taps when Maestro determines the UI has not changed.

It's very difficult to know whether the previous behavior was actually benefitting anyone, but there are certainly lots of complaints of the default behavior causing problems.

I propose ripping the bandaid off with this PR and changing this behavior to an opt-in.

Resolves #1202

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -174,7 +174,7 @@ class Maestro(
     fun tap(
         element: UiElement,
         initialHierarchy: ViewHierarchy,
-        retryIfNoChange: Boolean = true,
+        retryIfNoChange: Boolean = false,
         waitUntilVisible: Boolean = false,
         longPress: Boolean = false,
         appId: String? = null,
@@ -228,7 +228,7 @@ class Maestro(
     fun tapOnRelative(
         percentX: Int,
         percentY: Int,
-        retryIfNoChange: Boolean = true,
+        retryIfNoChange: Boolean = false,
         longPress: Boolean = false,
         tapRepeat: TapRepeat? = null,
         waitToSettleTimeoutMs: Int? = null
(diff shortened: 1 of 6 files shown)


amanjeetsingh150 (approved):

CodingItWrong:
UPDATE: the below was an error on my end; I had not updated our CI machine to Maestro 1.41 🤦 - working great now

I've updated to Maestro 1.41 to get this fix, but I'm still seeing cases where a double tap seems to be happening on Android unless I specify `retryTapIfNoChange: false`. For example, the attached test recording. At 0:48 there is a tap on the "Add credit card" button:

```yaml
- tapOn:
    id: add-credit-card-button
```

In my mocked testing environment one tap in that fills in "Visa XXXX XXXX XXXX 1234", and if you tap it again you are prompted to remove it.

However, in the recording at 0:49 the remove button is tapped again automatically. It can be fixed by changing the tap command to:

```yaml
- tapOn:
    id: add-credit-card-button
    retryTapIfNoChange: false
```

<video width="300" src="https://github.com/user-attachments/assets/ca99621a-23f3-4370-82c2-aaa9e2df1d2a" />

This is test is running on a M4 Mac mini on macOS Sequoia 15.6 and the latest Android Studio.

PR #2490, opened 2025-05-16
Feat/maestro/raw description

trindadetiago:
## Context

The goal of this PR is to expose the properties of the original description of the file - the one that would exist without the label.

What do I mean by that?

When we have a command, we have the properties:
- label
- description()

The description represents the human-readable form of the command. Each command has a corresponding "rule" or formula to create such a format. However, when the command has a label, this formula is not accessible: we always return the label in the description() function when it is present.

So the goal is to allow us to see the original Description (the one without the label)

## High-level reason for why do so
-> so we can store, in Maestro Cloud (and possibly other scenarios) both the labeled description and the unlabeled one, allowing us to choose which one to see.

## Proposed changes

Create the property originalDescription: exactly the same as description(), except it ignores the label
Make the description be a choice between label and originalDescription: if label is present, return label

## Testing

Wrote test cases to confirm the correct accessibility of the properties.


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -30,7 +30,11 @@ import maestro.orchestra.util.InputRandomTextHelper
 
 sealed interface Command {
 
-    fun description(): String
+    /**
+     * Returns the description of the command.
+     * @param useLabel if true, returns the label if present, otherwise returns the raw description
+     */
+    fun description(useLabel: Boolean = true): String
```
amanjeetsingh150, line 37: This feels like a weird contract — why should the consumer need to understand or handle the concept of a "label" directly?

Wouldn't it be cleaner to just return the label as a separate field? We already have a label field, so we could simply introduce an additional field.

That way, consumers have access to all three clearly separated:

1. originalDescription — the original/raw description

2. label — the static label (if present)

3. description — the final version used for display/debug/ with evaluated label or JS.

trindadetiago: Makes sense

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -457,9 +443,7 @@ data class ExtractTextWithAICommand(
     override val optional: Boolean = true,
     override val label: String? = null
 ) : Command {
-    override fun description(): String {
-        if (label != null) return label
-
+    override fun originalDescription(): String {
```
amanjeetsingh150, line 446: Won't this break? I didn't imagine to actually change the current description implementation?


I thought we would have description that either resolves to label or evaluate description (display description). On other hand, the original description simply holds the original value of description?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -30,7 +30,16 @@ import maestro.orchestra.util.InputRandomTextHelper
 
 sealed interface Command {
 
-    fun description(): String
+    /**
+     * Returns the raw parsed description of the command, ignoring any labels
+     */
+    fun originalDescription(): String
```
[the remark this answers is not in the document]
trindadetiago: Some command descriptions are a little more complex than simple text and are built dynamically.

ScrollUntilVisible, for example:

```
override fun originalDescription(): String {
        val baseDescription = "Scrolling $direction until ${selector.description()} is visible"
        val additionalDescription = mutableListOf<String>()
        additionalDescription.add("with speed $originalSpeedValue")
        additionalDescription.add("visibility percentage $visibilityPercentage%")
        additionalDescription.add("timeout $timeout ms")
        waitToSettleTimeoutMs?.let {
            additionalDescription.add("wait to settle $it ms")
        }
        if (centerElement) {
            additionalDescription.add("with centering enabled")
        } else {
            additionalDescription.add("with centering disabled")
        }
        return "$baseDescription ${additionalDescription.joinToString(", ")}"
    }
```

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
amanjeetsingh150: Can you write unit test case for these 3 cases:

- original description having the original constructed description
- description evaluating to label if its non null or the description (with **_evaluated env. variable_**). The description also has values that are evaluated in runtime by js engine.

Checkout `YamlCommandReaderTest`, that should help

maestro-orchestra/src/test/java/maestro/orchestra/CommandDescriptionTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        val jsEngine = RhinoJsEngine(platform = "ios")
+        jsEngine.putEnv("username", "John")
+
+        val inputCommand = commands[2] as InputTextCommand
+        val evaluatedInput = inputCommand.evaluateScripts(jsEngine)
+        assertThat(evaluatedInput.description()).isEqualTo("Input text John")
+
+        val assertCommand = commands[3] as AssertConditionCommand
+        val evaluatedAssert = assertCommand.evaluateScripts(jsEngine)
+        assertThat(evaluatedAssert.description()).isEqualTo("Assert that \"Hello John\" is visible")
```
amanjeetsingh150, line 49: Shouldn't we actually also assert original description and description both?

There are cases when both are same and both would be different the tests should mention that for both cases right?

maestro-orchestra/src/test/java/maestro/orchestra/CommandDescriptionTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    fun `description uses label when available`(
+        @YamlFile("028_command_descriptions.yaml") commands: List<Command>
+    ) {
+        val tapCommand = commands[1] as TapOnElementCommand
+        assertThat(tapCommand.description()).isEqualTo("Scroll to find the maybe-later button")
+
+        val inputCommand = commands[2] as InputTextCommand
+        assertThat(inputCommand.description()).isEqualTo("Input text \${username}")
```
amanjeetsingh150, line 33: Shouldn't we assert original description here as well?

amanjeetsingh150 (approved):

PR #2507, opened 2025-05-29
feat: add ability to switch and assert themes on iOS and Android: setDarkMode, toggleDarkMode, assertDarkMode & assertLightMode

markrickert:
## Proposed changes

This is a new feature that adds the ability to set the `light`/`dark` mode on iOS and Android platforms.

It adds the following new functionality:

- `setDarkMode` - possible values: `enabled` / `disabled`
- `toggleDarkMode` - swaps between light/dark mode
- `assertDarkMode` - fails the flow if dark mode is not currently enabled
- `assertLightMode` - fails the flow if dark mode is currently enabled

It achieves this by running commands for iOS/Android:

|   | iOS | Android |
|--------|--------|--------|
| `setDarkMode: enabled` | `XCUIDevice.shared.appearance = .dark` (via a new `setAppearance` endpoint on the XCUITest runner) | `adb shell 'cmd uimode night yes'` |
| `setDarkMode: disabled` | `XCUIDevice.shared.appearance = .light` (via the same endpoint) | `adb shell 'cmd uimode night no'` | 

iOS requires iOS 15+ (`XCUIDevice.appearance` isn't available below that); the endpoint returns a precondition error on older runtimes.

Web (`CdpWebDriver`) is a no-op for now — `isDarkModeEnabled`/`setDarkMode` are stubbed out since there's no dark-mode signal to drive on a headless/CDP-controlled browser yet.

## Testing

`./gradlew :maestro-test:test && ./gradlew test` passes 🟢 

I then "ignited" an app that [has a default maestro test in it called `FavoritePodcast.yaml`](https://github.com/infinitered/ignite/blob/master/boilerplate/.maestro/flows/FavoritePodcast.yaml) and replaced the contents with the following, adding calls to `setDarkMode`, `toggleDarkMode`, `assertDarkMode`, and `assertLightMode`

```yaml
# flow: run the login flow and then navigate to the demo podcast list screen, favorite a podcast, and then switch the list to only be favorites.

appId: ${MAESTRO_APP_ID}
env:
  FAVORITES_TEXT: "Switch on to only show favorites" # en.demoPodcastListScreen.accessibility.switch
onFlowStart:
  - runFlow: ../shared/_OnFlowStart.yaml
---
- runFlow: ../shared/_Login.yaml
- setDarkMode: disabled
- assertLightMode
- toggleDarkMode
- toggleDarkMode
- assertLightMode
- tapOn: "Podcast"
- assertVisible: "React Native Radio episodes"
- tapOn:
    text: ${FAVORITES_TEXT}
- setDarkMode: disabled
- assertVisible: "This looks a bit empty"
- tapOn:
    text: ${FAVORITES_TEXT}
    # https://maestro.mobile.dev/troubleshooting/known-issues#android-accidental-double-tap
    retryTapIfNoChange: false
- toggleDarkMode
- assertDarkMode
- repeat:
    times: 2
    commands:
      - scroll
- copyTextFrom:
    text: "RNR .*" # assumes all podcast titles start with RNR
    index: 2 # grab the third one, others might not be fully visible
- toggleDarkMode
- longPressOn: ${maestro.copiedText}
- scrollUntilVisible:
    element:
      text: ${FAVORITES_TEXT}
    direction: UP
    timeout: 50000
    speed: 90
    visibilityPercentage: 100
- tapOn:
    text: ${FAVORITES_TEXT}
- assertVisible: ${maestro.copiedText}
```

Then to test this, i ran the associated application on the iOS simulator and from this projects' directory i ran:

```bash
./maestro record -e MAESTRO_APP_ID=com.foo ../Foo/.maestro/flows/FavoritePodcast.yaml
```

Here is the resulting video showing the interface swapping back and forth from light to dark:

https://github.com/user-attachments/assets/d52da13c-9a39-48d9-8be7-ccfe37745b6f

## Issues fixed

n/a

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -535,6 +535,18 @@ class IOSDriver(
         LOGGER.warn("Airplane mode is not available on iOS simulators")
     }
 
+    override fun isDarkModeEnabled(): Boolean {
+        return metrics.measured("operation", mapOf("command" to "isDarkModeEnabled")) {
+            runDeviceCall("isDarkModeEnabled") { iosDevice.isDarkModeEnabled() }
+        }
+    }
+
+    override fun setDarkMode(enabled: Boolean) {
+        metrics.measured("operation", mapOf("command" to "setDarkMode")) {
+            runDeviceCall("setDarkMode") { iosDevice.setAppearance(if (enabled) "dark" else "light") }
+        }
+    }
+
     private fun addMediaToDevice(mediaFile: File) {
         metrics.measured("operation", mapOf("command" to "addMediaToDevice")) {
             val namedSource = NamedSource(
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -1243,6 +1243,30 @@ data class ToggleDarkModeCommand(
     }
 }
 
+data class AssertDarkModeCommand(
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+    override val originalDescription: String
+        get() = "Assert dark mode is enabled"
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+}
+
+data class AssertLightModeCommand(
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+    override val originalDescription: String
+        get() = "Assert dark mode is disabled"
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+}
+
 internal fun tapOnDescription(isLongPress: Boolean?, repeat: TapRepeat?): String {
     return if (isLongPress == true) "Long press"
     else if (repeat != null) {
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -1210,6 +1210,39 @@ data class ToggleAirplaneModeCommand(
     }
 }
 
+enum class DarkModeValue {
+    Enable,
+    Disable,
+}
+
+data class SetDarkModeCommand(
+    val value: DarkModeValue,
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+    override val originalDescription: String
+        get() = when (value) {
+            DarkModeValue.Enable -> "Enable dark mode"
+            DarkModeValue.Disable -> "Disable dark mode"
+        }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+}
+
+data class ToggleDarkModeCommand(
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+    override val originalDescription: String
+        get() = "Toggle dark mode"
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
+}
+
 internal fun tapOnDescription(isLongPress: Boolean?, repeat: TapRepeat?): String {
     return if (isLongPress == true) "Long press"
     else if (repeat != null) {
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
@@ -26,6 +26,8 @@ import maestro.Point
 import maestro.TapRepeat
 import maestro.orchestra.AddMediaCommand
 import maestro.orchestra.AssertConditionCommand
+import maestro.orchestra.AssertDarkModeCommand
+import maestro.orchestra.AssertLightModeCommand
 import maestro.orchestra.AssertNoDefectsWithAICommand
 import maestro.orchestra.AssertScreenshotCommand
 import maestro.orchestra.AssertWithAICommand
@@ -144,6 +146,8 @@ data class YamlFluentCommand(
     val toggleAirplaneMode: YamlToggleAirplaneMode? = null,
     val setDarkMode: YamlSetDarkMode? = null,
     val toggleDarkMode: YamlToggleDarkMode? = null,
+    val assertDarkMode: YamlAssertDarkMode? = null,
+    val assertLightMode: YamlAssertLightMode? = null,
     val retry: YamlRetryCommand? = null,
     @JsonIgnore val _sourceInfo: SourceInfo,
 ) {
@@ -502,6 +506,24 @@ data class YamlFluentCommand(
                     ToggleDarkModeCommand(
                         toggleDarkMode.label,
                         toggleDarkMode.optional
+                    )
+                )
+            )
+
+            assertDarkMode != null -> listOf(
+                MaestroCommand(
+                    AssertDarkModeCommand(
+                        assertDarkMode.label,
+                        assertDarkMode.optional
+                    )
+                )
+            )
+
+            assertLightMode != null -> listOf(
+                MaestroCommand(
+                    AssertLightModeCommand(
+                        assertLightMode.label,
+                        assertLightMode.optional
                     )
                 )
             )
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
@@ -58,6 +58,7 @@ import maestro.orchestra.RunScriptCommand
 import maestro.orchestra.ScrollCommand
 import maestro.orchestra.ScrollUntilVisibleCommand
 import maestro.orchestra.SetAirplaneModeCommand
+import maestro.orchestra.SetDarkModeCommand
 import maestro.orchestra.SetLocationCommand
 import maestro.orchestra.SetOrientationCommand
 import maestro.orchestra.SetPermissionsCommand
@@ -70,6 +71,7 @@ import maestro.orchestra.TakeScreenshotCommand
 import maestro.orchestra.TapOnElementCommand
 import maestro.orchestra.TapOnPointV2Command
 import maestro.orchestra.ToggleAirplaneModeCommand
+import maestro.orchestra.ToggleDarkModeCommand
 import maestro.orchestra.TravelCommand
 import maestro.orchestra.WaitForAnimationToEndCommand
 import maestro.orchestra.error.InvalidFlowFile
@@ -140,6 +142,8 @@ data class YamlFluentCommand(
     val addMedia: YamlAddMedia? = null,
     val setAirplaneMode: YamlSetAirplaneMode? = null,
     val toggleAirplaneMode: YamlToggleAirplaneMode? = null,
+    val setDarkMode: YamlSetDarkMode? = null,
+    val toggleDarkMode: YamlToggleDarkMode? = null,
     val retry: YamlRetryCommand? = null,
     @JsonIgnore val _sourceInfo: SourceInfo,
 ) {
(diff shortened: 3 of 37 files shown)


markrickert:
Adde docs PR to document new feature - https://github.com/mobile-dev-inc/maestro-docs/pull/133

morganick:
@markrickert does this handle specifying the device?

markrickert:
> @markrickert does this handle specifying the device?

@morganick On iOS, it should pass through the device id and if you don't pass one, it'll use "booted" as a default. I'm unsure how android handles this. 

https://github.com/mobile-dev-inc/Maestro/pull/2507/files#diff-25ebfb9b3069b37481b09445ea8c4458fdebdc760b55ef6e9af537f0fc1e9221R518-R527

morganick:
> > @markrickert does this handle specifying the device?
> 
> @morganick On iOS, it should pass through the device id and if you don't pass one, it'll use "booted" as a default. I'm unsure how android handles this.
> 
> https://github.com/mobile-dev-inc/Maestro/pull/2507/files#diff-25ebfb9b3069b37481b09445ea8c4458fdebdc760b55ef6e9af537f0fc1e9221R518-R527

@markrickert Yeah, I saw the `deviceId` portion on iOS, but trying to run your commands when I have more than one android device results in `adb: more than one device/emulator`. It needs the `-s <serial-number>` for the specified device. I didn't see how this was handled on the android side either.

markrickert:
@morganick I believe i've addressed the problem. I was using the regular `shell` command when i should have been using `dadb.shell` to reference the currently active emulator instance.

https://github.com/mobile-dev-inc/Maestro/pull/2507/commits/fd264a598d4ae8587a73d7408f8d1dbae11d7b82

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSetDarkMode.kt
```diff
@@ -0,0 +1,72 @@
...
+            else -> throwInvalidInputException(input)
+        }
+    }
+
+    private fun throwInvalidInputException(input: List<String>): Nothing {
+        throw IllegalArgumentException(
+            "setDarkMode command takes either: \n" +
+                    "\t1. enabled: To enable dark mode\n" +
+                    "\t2. disabled: To disable dark mode\n" +
+                    "\t3. value: To set dark mode to a specific value (enabled or disabled) \n" +
```
Leland-Takamine, line 59: I was a bit confused about this line

Leland-Takamine:
> @morganick I believe i've addressed the problem. I was using the regular `shell` command when i should have been using `dadb.shell` to reference the currently active emulator instance.
> 
> [fd264a5](https://github.com/mobile-dev-inc/Maestro/commit/fd264a598d4ae8587a73d7408f8d1dbae11d7b82)

`shell` method was actually correct - it's just a helper that checks the exit code as well. You'd just pass in the adb shell command without the `adb shell` part since it's using Dadb.

```
    private fun shell(command: String): String {
        val response: AdbShellResponse = try {
            dadb.shell(command)
        } catch (e: IOException) {
            throw IOException(command, e)
        }

        if (response.exitCode != 0) {
            throw IOException("$command: ${response.allOutput}")
        }
        return response.output
    }
```

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -515,6 +515,16 @@ class IOSDriver(
         LOGGER.warn("Airplane mode is not available on iOS simulators")
     }
 
+    override fun isDarkModeEnabled(): Boolean {
+        val deviceId = iosDevice.deviceId ?: "booted"
+        return XCRunnerCLIUtils.isDarkModeEnabled(deviceId)
```
Leland-Takamine, line 520: Is there a way to do this via xcuitest? We're trying to move away from simctl commands since they won't work for physical devices

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSetDarkMode.kt
markrickert: I was copying the airplane mode function as it allows this syntax:

```yaml
- setDarkMode enabled
```

or

```yaml
- setDarkMode
    value: enabled
    label: "Something"
```

I'm definitely open to suggestions on how to improve this!

Leland-Takamine: Oh yeah I think the syntax makes sense. The 3rd line of the error message is what was confusing me - seems redundant?

```
1. enabled: To enable dark mode
2. disabled: To disable dark mode
3. value: To set dark mode to a specific value (enabled or disabled)
```

Leland-Takamine: Oh ok I think I understand now. Maybe it would be clearer to show those two styles of syntax as examples

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
markrickert: I'll look into this tomorrow to see if there's a way to do it. Not sure at the moment.

markrickert: @Leland-Takamine I found this reference already inside the repo's source code, but i have no idea how i would go about using it 🤷  https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-ios-xctest-runner/maestro-driver-iosUITests/PrivateHeaders/XCTest/XCUIDevice.h#L18-L22

amanjeetsingh150: Hey @markrickert thanks for contribution, I think this is would be a new endpoint in our XCUITest driver here:

https://github.com/mobile-dev-inc/Maestro/tree/main/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers

You can check any one handler there for example. On implementation we would receive the appearence and then as you shared:

```swift
switch requestBody.appearance.lowercased() {
   case "dark":
          XCUIDevice.shared.appearance = .dark
   case "light":
          XCUIDevice.shared.appearance = .light
   default: return AppError(type: .precondition, message: "Invalid appearance value. Use 'dark' or 'light'").httpResponse
}
            
```

And this endpoint would be reached from XCTest client here in maestro: https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-ios/src/main/java/ios/xctest/XCTestIOSDevice.kt

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
```diff
@@ -478,6 +482,25 @@ data class YamlFluentCommand(
                 )
             )
 
+            setDarkMode != null -> listOf(
```
amanjeetsingh150, line 494: I wonder are there cases where it is useful to set appearance mid flow execution? 

Is this more useful as a command or flow based or upload level config?

maestro-orchestra/src/test/resources/YamlCommandReaderTest/023_labels.yaml
```diff
@@ -124,6 +124,11 @@ appId: com.example.app
       label: "Turn on airplane mode for testing"
 - toggleAirplaneMode:
       label: "Toggle airplane mode for testing"
+- setDarkMode:
+      value: enabled
+      label: "Turn on dark mode for testing"
+- toggleDarkMode:
+      label: "Toggle dark mode for testing"
```
amanjeetsingh150, line 134: What is the difference between these two commands?

Can't I toggle with reverting the value passed in setDarkMode command? 🤔

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -3240,6 +3244,20 @@ class IntegrationTest {
...
+        val commands = readCommands("121_dark_mode")
+        val driver = driver { }
+
+        Maestro(driver).use {
+            orchestra(it).runFlow(commands)
+        }
+
+        // Then
+        // No test failures
+        driver.assertNoInteraction()
```
amanjeetsingh150: This feels like incomplete we do want to assert the appearance event here though right?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -1026,6 +1026,39 @@ data class ToggleAirplaneModeCommand(
...
+    override val optional: Boolean = false,
+) : Command {
+    override val originalDescription: String
+        get() = when (value) {
+            DarkModeValue.Enable -> "Enable dark mode"
+            DarkModeValue.Disable -> "Disable dark mode"
+        }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
```
amanjeetsingh150, line 1230: Should we be allowing setting this variable with js? Not sure. Like runtime through JS. If yes then we might want to handle this.

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
Fishbowler: I can't think of a use-case for mid-flow. But given the JS comment later on down, I think I like the idea of being able to do it very early in execution, explicitly as part of your test setup, perhaps calling this before launchApp.

rursache:
any progress on this? we need the option to toggle the dark mode options, at least on iOS where `xcrun simctl ui booted appearance dark` is all it takes

maestro-orchestra/src/test/resources/YamlCommandReaderTest/023_labels.yaml
Fishbowler: Matches what we do for Airplane mode. I like the consistency. But I've never used the `toggle` variant.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
Fishbowler: I wanna set _everything_ via JS, because at some point, someone _will need this_.

Fishbowler:
@markrickert Any energy for looking at Amanjeet's few comments? 
This looks like it's a rebase and a couple of changes from done.

Importantly: How would someone using Maestro evaluate whether anything had changed? `assertScreenshot`?

maestro-orchestra/src/test/resources/YamlCommandReaderTest/023_labels.yaml
markrickert: Yeah, left it in for consistency with airplane mode.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
markrickert: I've updated the PR to utilize the XCTest runner framework instead of simctl for ios! 🎉

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
markrickert: yup! i've got the assertion commands all wired up now!

markrickert:
After a conversation with @Leland-Takamine at the Chain React after party last week, i decided to revive this PR and implement it with the XCTest framework instead of simctl for ios like i was asked about.

This branch is all up to date with the `main` branch with my old feature replayed on top of it and the methodology for ios changed.

Claude helped me with this task, but i have read every line of code, understand every line of code added/changed, and have manually verified that the tests are working in ios, android, and on web!

pedro18x (approved):
This is looking great, Mark, and thanks for seeing it through after Chain React. I went through the reworked commits end to end: the command plumbing, YAML models, Orchestra dispatch, and the assertion logic (real device read-back with a clear failure and debug message) all read well, and the integration cases cover the behavior. The iOS 15 precondition guard surfaces a clear "requires iOS 15+" message rather than an opaque failure, and keeping `setDarkMode` literal-only mirrors `setAirplaneMode` nicely for consistency.

One optional follow-up, definitely not a blocker: the web `CdpWebDriver` dark-mode path (`setDarkMode` / `isDarkModeEnabled`) isn't exercised by the integration cases since they run on `FakeDriver`. That's consistent with how the rest of `CdpWebDriver` is tested, so no objection to merging as-is, but a small test hitting the real prefers-color-scheme emulation and the `matchMedia` read-back would be a nice addition next time you're in here.

Approving. Nice work.

markrickert:
Thanks! This has been a long time coming and glad it wasn't too difficult to port over with hundreds of commits made to the project since i opened this PR!

I agree that we should defer the web driver test... everything else here is pure unit tests so this would be the first test that needs a chrome driver to spin up on CI so maybe on the next one :D

PR #2511, opened 2025-06-02
fix: fix build ios driver scripts

Rezisn:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-ios-xctest-runner/build-maestro-ios-runner.sh
+++ b/maestro-ios-xctest-runner/build-maestro-ios-runner.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


maestro-ios-xctest-runner/build-maestro-ios-runner.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-if [ "$(basename "$PWD")" != "Maestro" ]; then
+if [ "$(basename "$PWD")" != "maestro" ]; then
```
Fishbowler, line 4: Except, the repo is Maestro, not maestro

image

Perhaps a case insensitive match would be better?

Fishbowler (changes requested):

PR #2514, opened 2025-06-02
feat: stop functionality for studio

trindadetiago:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -113,6 +113,12 @@ class Orchestra(
 
     private val rawCommandToMetadata = mutableMapOf<MaestroCommand, CommandMetadata>()
 
+    private var isStopping = false
+
+    fun stopFlow() {
+        isStopping = true
+    }
+
     fun runFlow(commands: List<MaestroCommand>): Boolean {
         timeMsOfLastInteraction = System.currentTimeMillis()
 
@@ -151,6 +157,10 @@ class Orchestra(
         } catch (e: Throwable) {
             exception = e
         } finally {
+            if (isStopping) {
+                isStopping = false
+            }
+
             val onCompleteSuccess = config?.onFlowComplete?.commands?.let {
                 executeCommands(
                     commands = it,
@@ -178,6 +188,11 @@ class Orchestra(
 
         commands
             .forEachIndexed { index, command ->
+                if (isStopping) {
+                    onCommandSkipped(index, command)
+                    return true
+                }
+
                 onCommandStart(index, command)
 
                 jsEngine.onLogMessage { msg ->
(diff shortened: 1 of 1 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -113,6 +113,45 @@ class Orchestra(
 
     private val rawCommandToMetadata = mutableMapOf<MaestroCommand, CommandMetadata>()
 
+    private var isStopping = false
+    private var isPaused = false
+    private val pauseLock = Object()
+
+    fun stopFlow() {
+        isStopping = true
```
amanjeetsingh150, line 121: Can we use coroutine to manage these states? This is something that was already started here:

https://github.com/mobile-dev-inc/Maestro/pull/2488/files

I feel stopped is probably already implemented there. Since when coroutine is cancelled and coroutine becomes inactive we directly return from the function here:

https://github.com/mobile-dev-inc/Maestro/pull/2488/files#diff-00989d49f0eaa1f9e614580322e3b2ec79d9915a804ffa004e2f53ba36215698R188

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -113,6 +113,45 @@ class Orchestra(
...
+
+    fun resumeFlow() {
+        synchronized(pauseLock) {
+            isPaused = false
+            pauseLock.notifyAll()
+        }
+    }
+
+    private fun checkPauseState() {
+        synchronized(pauseLock) {
```
amanjeetsingh150, line 143: If we plan to merge the coroutine implementation we can do this via coroutine:

Something like this:

```kotlin
private suspend fun waitWhilePaused() {
        // suspend this block until the variable isPause is true
       while (isPaused) {
            delay(pauseCheckDelay) // or lock on object until paused
      }
}

    // Example of a coroutine with pause functionality
suspend fun executeCommands() {
      try {
          // command execution
          waitWhilePaused()
       } catch (e: CancellationException) {
          println("Coroutine was cancelled")
          throw e
       }
}
```

amanjeetsingh150 (commented):
Another question: where do we set these variable? Are these something that can be used anywhere or only studio?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -113,6 +113,45 @@ class Orchestra(
 
     private val rawCommandToMetadata = mutableMapOf<MaestroCommand, CommandMetadata>()
 
+    private var isStopping = false
+    private var isPaused = false
+    private val pauseLock = Object()
```
amanjeetsingh150, line 118: One problem with local variable like this also is, it makes things untestable can we think of way so that we can write integration test for this? 

We can inject those on Orchestra and configure them from test so that we can test this flow

PR #2519, opened 2025-06-04
Add important-for-accessibility and error attributes to the view hierarchy

Thomvis:
## Proposed changes

Add support for important-for-accessibility and error attributes to the Android view hierarchy. This enhancement improves accessibility information available in Maestro's view dumps by exposing these critical accessibility properties.

## Testing

- Functionality has been tested directly by invoking the CLI
- Tested by invoking it in a larger automated workflow

## Issues fixed

N/A - Enhancement to improve accessibility support


Fishbowler:
Would the same props also exist in iOS?

Thomvis:
@Fishbowler On iOS the app would not return accessibility elements for unimportant views. It does not have the notion of unimportant accessibility elements.

Thomvis:
I'd love to get this merged. `important-for-accessibility` is a great indicator for if a view is actually visible/interactable. E.g. views behind a drawer are returned with `important-for-accessibility: false`.

This PR lays the groundwork for a refinement related to https://github.com/mobile-dev-inc/Maestro/pull/2577. We should consider filtering out views with `important-for-accessibility == false` before returning the hierarchy to the LLM.

PR #2528, opened 2025-06-10
improving logs + actually finding errors

felipeduartea:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-client/src/main/java/maestro/Errors.kt
+++ b/maestro-client/src/main/java/maestro/Errors.kt
@@ -19,48 +19,51 @@
 
 package maestro
 
-sealed class MaestroException(override val message: String) : RuntimeException(message) {
+sealed class MaestroException(override val message: String, cause: Throwable? = null) : RuntimeException(message, cause) {
 
-    class UnableToLaunchApp(message: String) : MaestroException(message)
+    class UnableToLaunchApp(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class UnableToClearState(message: String) : MaestroException(message)
+    class UnableToClearState(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class AppCrash(message: String): MaestroException(message)
+    class AppCrash(message: String, cause: Throwable? = null): MaestroException(message, cause)
 
-    class DriverTimeout(message: String): MaestroException(message)
+    class DriverTimeout(message: String, cause: Throwable? = null): MaestroException(message, cause)
 
     open class AssertionFailure(
         message: String,
         val hierarchyRoot: TreeNode,
         val debugMessage: String,
-    ) : MaestroException(message)
+        cause: Throwable? = null,
+    ) : MaestroException(message, cause)
 
     class ElementNotFound(
         message: String,
         hierarchyRoot: TreeNode,
         debugMessage: String,
-    ) : AssertionFailure(message, hierarchyRoot, debugMessage)
+        cause: Throwable? = null,
+    ) : AssertionFailure(message, hierarchyRoot, debugMessage, cause)
 
-    class CloudApiKeyNotAvailable(message: String) : MaestroException(message)
+    class CloudApiKeyNotAvailable(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class DestinationIsNotWritable(message: String) : MaestroException(message)
+    class DestinationIsNotWritable(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class UnableToCopyTextFromElement(message: String): MaestroException(message)
+    class UnableToCopyTextFromElement(message: String, cause: Throwable? = null): MaestroException(message, cause)
 
     class InvalidCommand(
         message: String,
-    ) : MaestroException(message)
+        cause: Throwable? = null,
+    ) : MaestroException(message, cause)
 
-    class NoRootAccess(message: String) : MaestroException(message)
+    class NoRootAccess(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class UnsupportedJavaVersion(message: String) : MaestroException(message)
+    class UnsupportedJavaVersion(message: String, cause: Throwable? = null) : MaestroException(message, cause)
 
-    class MissingAppleTeamId(message: String): MaestroException(message)
+    class MissingAppleTeamId(message: String, cause: Throwable? = null): MaestroException(message, cause)
 
-    class IOSDeviceDriverSetupException(message: String): MaestroException(message)
+    class IOSDeviceDriverSetupException(message: String, cause: Throwable? = null): MaestroException(message, cause)
 }
 
-sealed class MaestroDriverStartupException(override val message: String): RuntimeException() {
-    class AndroidDriverTimeoutException(message: String): MaestroDriverStartupException(message)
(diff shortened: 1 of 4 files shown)


maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -290,8 +292,20 @@ object LocalSimulatorUtils {
                 bundleId,
             )
         ).start()
-
-        return String(process.inputStream.readBytes()).trimEnd()
+        if (!process.waitFor(5, java.util.concurrent.TimeUnit.MINUTES)) {
```
amanjeetsingh150: There is abstraction `CommandLineUtils.runCommand` we can use that shows error output right?

Whats missing in that function?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -933,7 +933,8 @@ class Orchestra(
             val permissions = command.permissions ?: mapOf("all" to "allow")
             maestro.setPermissions(command.appId, permissions)
 
-        } catch (e: Exception) {
+        } catch (e: Exception) { 
+            logger.error(e.stackTraceToString())
```
amanjeetsingh150:
```suggestion
            logger.error("Failed clearing app state", e)
```

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
felipeduartea: I don't think we have a way o capturing the return value on runCommand

amanjeetsingh150: Why do we need it?

amanjeetsingh150: Can't we just use the process object returned by it?

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -273,40 +269,53 @@ object LocalSimulatorUtils {
...
     private fun getAppBinaryDirectory(deviceId: String, bundleId: String): String {
-        val process = ProcessBuilder(
-            listOf(
-                "xcrun",
-                "simctl",
-                "get_app_container",
-                deviceId,
-                bundleId,
+        logger.info("[Start] getAppBinaryDirectory: $deviceId, $bundleId")
+        return runCatching {
```
herval: why do you need these catch-and-release blocks?

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -371,7 +380,7 @@ object LocalSimulatorUtils {
         )
     }
 
-    fun uninstall(deviceId: String, bundleId: String) {
+    fun uninstall(deviceId: String, bundleId: String) { //The error happens here
```
herval: remove comment?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -933,7 +933,8 @@ class Orchestra(
             val permissions = command.permissions ?: mapOf("all" to "allow")
             maestro.setPermissions(command.appId, permissions)
 
-        } catch (e: Exception) {
+        } catch (e: Exception) { 
+            logger.error("Failed clearing app state", e)
```
herval: passing `e` as the cause field for `UnableToClearState` would allow for proper stack traces (the way this is done right now, the stack trace ends here

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
felipeduartea: isn't that the best way for me to capture the exception in this block?
Saw it being used in the `terminate` function

herval: you can try/catch wherever you want to capture errors - ie wherever this function is called

felipeduartea: but shouldn't it be always expected to throw an error when this command fails? 

Not sure I'm following, but why would I add the error handling on every place I call this function instead of inside the function itself?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
felipeduartea: maestro exceptions don't really accept an additional parameter. You think it's a good idea to add it there or just use logger.error?

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
herval: how many places are you calling that function on?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
herval: you can add it?

```
sealed class MaestroException(override val message: String, cause: Throwable? = null) : RuntimeException(message, cause) {
```

logger.error and then rethrowing an exception will just give u a log output _somewhere_ (in this case wherever this logger writes to) and an unhandled exception without a root cause. Which is why it's usually so difficult to figure out where stuff is actually breaking in many places

maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
felipeduartea: honest question:

I know it's only used in one place, but I can't see why it would be better to add the error handling to the place where the function is called instead of inside the function.


When I run it to an app that doesn't exist I receive this in the terminal
```
xcrun simctl get_app_container 59D37691-F8B3-4A70-AAC3-8159799D79D5 app.wikipedia
An error was encountered processing the command (domain=NSPOSIXErrorDomain, code=2):
The operation couldn’t be completed. No such file or directory
No such file or directory
```

So for me it should always throw an exception when this happens. But open to understanding what you mean

herval: one reason is this is a framework, so generally it'd make sense to avoid logging that's not relevant

the other reason is this pattern leads to multiple try/catches in a row catching, logging and throwing the exception. Which makes the logs noisy for no reason

the cleanest method is usually to catch exceptions at the place you're handling them (and logging errors there as well)

the example u included above shows what happens with this sort of loggin: it has a redundant log line since it's printed + re-thrown in two places (`The operation couldn’t be completed. No such file or directory` + `No such file or directory`)

maestro-client/src/main/java/maestro/Errors.kt
```diff
@@ -19,48 +19,51 @@
 
 package maestro
 
-sealed class MaestroException(override val message: String) : RuntimeException(message) {
+sealed class MaestroException(override val message: String, cause: Throwable? = null) : RuntimeException(message, cause) {
```
herval, line 22: 🥳

herval (approved):
approving to unblock

PR #2530, opened 2025-06-11
[POC] Add MCP server to Maestro's CLI

steviec:
## Proposed changes

This proof-of-concept implements a Model Context Protocol (MCP) server for the Maestro CLI, enabling AI assistants to interact with mobile test automation. 

### Background

I've been trying to provide the Cursor IDE some MCP-based simulator feedback to increase my agentic coding loops. The existing MCP servers that connect to simulators are truly, truly terrible, and I dropped a few days trying to get 5 different ones working before I realized that Maestro would be the ideal place for something like this. Maestro is a battle-tested, cross-device automation framework that I already use heavily for e2e testing, and it seems like the perfect place to add an MCP-based AI automation interface to.  And frankly I just want Maestro to stay relevant in the coding agent world, and this seems like a great way to do it.

### Implementation:

This adds a new `McpCommand` to the CLI so that the server can be launched with `maestro mcp`. It spins up an MCP server listening over STDIO.

It contains the following:
- two simple tool abstractions for implementing tools (`McpTool` and `McpCommandTool`)
- a tool registry that tells the server which tools to provide for `tools/list` calls
- some test scripts for verifying that the tools are working correctly
- a starting set of useful tools, namely: ListDevices, StartDevice, LaunchApp, TakeScreenshot, TapOn, and InspectUI

## Testing

If you want to build the tool and test it, do:

```
./gradlew :maestro-cli:installDist
./maestro-cli/src/test/mcp/test-mcp-tool.sh
```

If you want to test it from an IDE, add this to your MCP servers config (in Cursor it's called `mcp.json`):

```
    "maestro-mcp": {
      "command": "<path_to_your_maestro_dir>/maestro-cli/build/install/maestro/bin/maestro",
      "args": ["mcp"]
    }
```

And then just start chatting with your LLM. [Here's a video](https://youtu.be/cWOa6LJ4xcY) showing an interaction with a coding agent where it learns what tools are available and then just starts using the appropriate one based on the task. It can "see" the simulator screen both by taking screenshots or by getting a compact version of the view hierarchy with the `inspect_ui` tool.

## Feedback requests

I'd really like feedback on this approach before dropping any more time making it pretty/more robust. It's possible that 1) the overall approach I'm using is a bad one or 2) the maestro maintainers don't think this is a useful addition at all.  I'm already using it heavily for local development (I use it to automate my React Native development flow), and it's really working well for me. 

I'll continue to use this locally for my own work, and I'd spend whatever time necessary bringing this up to par for the Maestro codebase, but I don't want to spend that time if this is unlikely to ever be accepted. Personally I think it would be a huge win for any developer who spends a lot of time developing against an iOS or Android simulator.


herval:
Hi Stevie! We actually started an implementation here: https://github.com/mobile-dev-inc/maestro-mcp

I like the idea of keeping it all on the maestro codebase though - would you be willing to port the [tools we had planned](https://github.com/mobile-dev-inc/maestro-mcp/blob/main/src/maestro_mcp/cli.py)?

steviec:
Ooh, thanks for sharing that link with me! I'd be happy to port the tools, but let's discuss the approach a little. I want to make sure that the approach I'm taking is the optimal one.

First, I want to confirm the tools in your implementation:
- get_cheat_sheet
- run_code
- run_flow_files
- get_hierarchy
- query_docs
- check_syntax

There's a key difference in your approach compared to the one I took. In your approach, you just tell the LLM what the Maestro syntax is (via `get_cheat_sheet`), write the YAML code to a file and then hand it off to the maestro CLI. That's pretty interesting; it allows you to get full access to the Maestro capabilities without having to hand roll individual tools. 

My approach is more deeply integrated with the Maestro internals, which gives me more flexibility in defining the tool parameters and making the output more LLM friendly, but it requires specifying each tool individually, which is a lot more maintenance overhead. 

I'm trying to think through the tradeoffs of the two approaches. Do you have a preference? I can make another PR that attempts to use your approach, but still keep it built into the CLI. Also, your implementation relies on access to the Maestro API (for cheat_sheet and query_docs), which I'd prefer not to have as a requirement for this server.

steviec:
One more thought: [this video](https://youtu.be/eeOANluSqAE?si=_OZQqSB_-7CFALcx&t=255) talks a bit about the tradeoffs in giving an LLM full API access vs creating more custom-built tools that leverage an internal API. These points are interesting:

1. LLMs are terrible at choosing from a long list of tools. They do much better with a focused set of tools.
2. API descriptions for typical APIs are not well written for LLMs. They need concise explanations and clear examples. Writing for an LLM is different than writing for a human.
3. APIs are typically designed for low-level automation, which is not well-suited for LLMs. It's far more powerful to provide higher level, task-oriented tools that map to the way LLMs work.

In the video he provides an example of exposing the Neon postgres service to an LLM. The anti-pattern he shows is to expose a `run_sql` command that allows the LLM to execute arbitrary SQL and let the LLM figure out how to do a migration. 

I could argue that the `run_code` approach is similar to this anti-pattern; we'd be putting a lot of responsibility on the LLM to figure out the exact flow syntax to construct to get the job done (but note that Neon actually _does_ expose a `run_sql` command almost as a fallback, but also provide more purpose-built, LLM-oriented tools). After looking through the list of commands in the [API reference](https://docs.maestro.dev/api-reference/commands), I think only a subset of them are actually useful for an LLM to do automation, and providing unnecessary tools is a surefire way to confuse an LLM.

That being said, it wouldn't be hard to additionally expose a `run_code` command to see if the LLM can consistently use it well, and using an eval harness could help to determine if LLMs are generally happy just using the cheatsheet + run_code combo.

herval:
so, the idea of the python mcp is that the LLM can write, validate and run code. That's why there is a run_code implemented. Ther eason why there's a cheat sheet is that no current LLM has a good enough understanding of Maestro (chatgpt, claude, etc), so they often produce code that doesn't even compile. Adding the cheat sheet to the context increases the success rate dramatically, and then adding the syntax checking makes it better for the cases it gets it almost right

those aren't really "direct api access" - there's no maestro api to run a single line of maestro code, for instance. What happens with this set of tools is that the LLM is able to iterate running individual commands, then puts together a full script in the end and creates a test file. Granted, I haven't benchmarked it extensively yet, so it might not be reliable all the time

I like your approach for cases where you want to _interact with a device but not write code_ - eg "install an app for me" - it doesnt have to be exposed as a run_code kind of operation. I think they're complementary and serve different needs.

do notice the python implementation is shelling out to the CLI since it's in python - it's best to bake it into the CLI and use the code directly, as in your implementation here. I faced the same issues w/ the official sdk and went w/ that approach as a v1 instead, but it'd definitely be easier to distribute as a `maestro mcp` command

herval:
with regards to query_docs - it's essentially a RAG tuned to answer questions regarding Maestrop. Without that, the LLM will try to do a web search when it can't figure something out, which will often produce irrelevant or outdated results in my tests. 

Cheat sheet is distributed as a remote resource as well since it's easier to keep it updated that way. But we can try baking it in the maestro codebase as well. _Without_ the cheat sheet, the MCP will be pretty bad at writing/executing maestro, but also at answering questions (eg "what's this code doing?")

herval:
finally, wrt defining a bunch of tools individually - claude currently supports a maximum of 50 tools, so having a ton of micro-tools is definitely going to hit a ceiling quickly (plus it's considerably more expensive in terms of tokens). I'm guessing this will grow (the github MCP is a monster with a ton of functions - it's quite literally the opposite of what the Neon guy suggests on his talk), so can't predict what will be better in the future here

herval:
tl;dr:
- what your PR does so far is useful in cases where the user wants to _interact with devices_ using Maestro - taking screenshots, tapping stuff, starting a device
- what the python MCP covers is the use cases where the user wants to _write, explain or validate maestro code_

I think both are useful and complementary, hence why I think we should merge them

steviec:
Thanks for the thoughtful response! I agree with all of your points; these are complementary use cases. Your feedback is enough to unblock me, I'll take a first stab at supporting both the "interact with devices" and the "write, explain or validate maestro code" use cases.

Thomvis:
I was also in need of an Maestro MCP and built my own, not aware of your PR: https://github.com/mobile-dev-inc/Maestro/compare/main...Thomvis:Maestro:feature/mcp.

I opted to use https://github.com/modelcontextprotocol/java-sdk/tree/main/mcp, so there's a lot less code.

Only implemented the hierarchy and test commands, enough for my use case.

steviec:
Checking out the Java reference SDK was top of my list, that's great that it worked well for you! I'll follow a similar pattern. I think the Java SDK is being updated far more regularly than the kotlin one.

Thomvis:
I went with the Java one because the Kotlin SDK seemed to require a bump in Kotlin language versions and I didn't want to touch that. Even the Java SDK had unintended consequences. It pulls in a newer version of slf4j, requiring some changes as well.

herval:
can we move forward with merging one of these, then adding the remaining functionality?

which one do y'all prefer?

steviec:
I prefer @Thomvis's underlying approach; relying on the Java SDK for tool/resource/prompt definition and service instantiation means that we'll be able to benefit from improvements made to that library. However, it's missing a bunch of the things that will enable an LLM to create the correct syntax and get feedback that you have in your python version. Give me a day to put together a hybrid version for an initial PR, and than I'll create a separate one to add more automation-specific tooling on top of it.

steviec:
Welp, things keep getting more interesting. The Java MCP reference implementation targets Java 17+ because it uses quite a few modern Java features. @Thomvis : I'm assuming you used openjdk17+ to build? 

Unless I'm misunderstanding something, we have 5 options, from most complex to least complex:
1. Bump the Java version to 17 for the `maestro` CLI so we can use the Java SDK (I'm assuming that's not going to happen)
2. Bump the Kotlin version to 2.21.0 for the `maestro` CLI so we can use the Kotlin SDK (also assuming this is not going to happen)
3. Roll our own "SDK" for our mcp server, like I did in my original PR, so we could still target Java 8. More of a headache to maintain, and wouldn't get new stuff (e.g. Streamable HTTP, MCP spec update related changes) for free.
4. Create a separate `maestro-mcp` standalone server that can target Java 17 (this would allow us to get all the benefits of using the reference SDK, but would be a separate binary that we'd create a separate homebrew formula template for)
5. Go back to the python version that is just a wrapper for the `maestro` CLI

If I had a vote, I'd vote for 4 or 3. But I don't, so I'll leave that to you guys :). I'm still happy to update the PR to follow whatever approach you all need, but don't want to dive in until this is answered.

herval:
bumping the java or kotlin versions of maestro is difficult right now (in particular because it'd be a breaking change for a lot of android devs), so 1 & 2 aren't gonna be easy

I'm game w/ 3, but my concern is the speed it's evolving (eg they just released some remote MCP capabilities today that change the underlying protocol to support oauth & remote MCPs, but I'm guessing it's all still backwards compatible).

4 comes with a lot of maintenance overhead

I'd really like to avoid 5 bc it's super wonky - it's slower than integrating it directly on the CLI like both of your implementations, we have to expose micro-commands to use w/ the MCP, etc

given all this, IMO the best alternative _right now_ would be number 3, since we could ship it today. And since the work on the framework is already done, adding the stuff @Thomvis did on his branch could be trivial

shall we go that route? If so, we can merge this, then have a subsequent PR adding the remaining tools, then we announce 🍰

herval:
(wrt @Thomvis approach - I do like that too, but bumping dependencies will break _a bunch_ of stuff - slf4j, for instance, broke everything last time we tried to upgrade, so I'm being a bit conservative until we have a good route to migrate to java 17+)

Thomvis:
My version with the MCP SDK works with Java 11. That's my conclusion given that this works:

```bash
➜ echo $JAVA_HOME                                 
/opt/homebrew/Cellar/openjdk@11/11.0.27/libexec/openjdk.jdk/Contents/Home
➜ ./gradlew :maestro-cli:installDist
Starting a Gradle Daemon (subsequent builds will be faster)
[...]
BUILD SUCCESSFUL in 30s
```

Wrt the risks of updating slf4j and logback, I'll gladly defer to your experience. It does *seem* to be working (once I updated logback to a compatible version), but I am definitely not an expert.

steviec:
Ah, that's because the `javac` compiler doesn't check the bytecode version of your dependencies, so it doesn't complain at compile time. Double check this on your side, but I bet the problem will only show up when the JVM tries to load the classes at runtime. 

```
» export JAVA_HOME='/opt/homebrew/opt/openjdk@11'
» ./gradlew :maestro-cli:installDist  
[...]
BUILD SUCCESSFUL in 50s
37 actionable tasks: 37 executed

 » ./maestro-cli/build/install/maestro/bin/maestro mcp
 Exception in thread "main" java.lang.UnsupportedClassVersionError: io/modelcontextprotocol/server/McpSyncServerExchange has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
	[...]
	
» export JAVA_HOME='/opt/homebrew/opt/openjdk@17' 
» ./maestro-cli/build/install/maestro/bin/maestro mcp
[Works!]
```

And @herval : option 3 sounds good to me. If staying on the cutting edge of MCP features is a priority, and it's too much of a pain to keep the DIY SDK up-to-date, you guys can re-evaluate this decision. I'll try to just get the code working in the current `maestro` CLI. I don't like adding this much code, but it seems like the least bad option right now.

steviec:
Ok, I have a beta version ready for testing. 

Some important notes:
- I spent an hour and was able to create (with Claude Code's help!) a Kotlin 1.8.22 version of the kotlin MCP sdk. Most of the changes were in the build dependencies, and minor changes in the syntax. But I did have to remove the SSE and websockets implementation, since it relies on a version of ktor that we can't use. I think this gets us most of the benefits of 2). Note that Maestro's build dependencies now rely on my forked branch, and if you like this approach we can put that in the mobile-dev-inc repo instead
- I changed some of the tool names to map to how I think the LLM would intuitively call them (e.g. run_flow instead of run_code, etc). Check out the names and the parameters for the tools for any feedback
- the test suite involves calling [MCP Inspector](https://github.com/modelcontextprotocol/inspector) to run the server and test the tools. you can test individual tools with the `test-mcp-tool.sh` script, or run the full set of tools with `mcp-server-test.sh`

Remaining todos:
- fix the failing checks
- any code changes requested
- more thorough manual testing from an MCP-enabled IDE or claude code to find edge cases
- clear instructions for people using this how to add to an mcpServers stanza (including the auth stuff)
- add better logging based on MCP best practices
- maybe: add eval testing to this. I think it would really benefit from automated evals to ensure that LLMs consistently call the right tools and do the right things.

Here's the current list of tools:

### Device Management
  **list_devices** - List all available devices for automation
  **start_device** - Start a simulator/emulator

  ### Screen Interaction
  **launch_app** - Launch an application
  **stop_app** - Stop an application
  **take_screenshot** - Capture device screen
  **tap_on** - Tap UI elements by text, ID, or other selectors
  **input_text** - Type into focused text field
  **back** - Press back button
  **inspect_view_hierarchy** - Get screen UI tree structure

  ### Flow Execution
  **run_flow** - Execute Maestro YAML commands
  **run_flow_files** - Run complete test files
  **check_flow_syntax** - Validate YAML syntax

  ### Documentation
  **cheat_sheet** - Get Maestro command reference
  **query_docs** - Search Maestro documentation

herval:
brilliant! 🙌 

I'm not sure if I'm doing something wrong, but it doesnt seem to run correctly for me:

```
% ./maestro-cli/src/test/mcp/mcp-server-test.sh
=== MCP Server Integration Tests ===
WARNING: MAESTRO_API_KEY not set. query_docs and cheat_sheet tools will be skipped.

Testing tools/list...

Failed to connect to MCP server: MCP error -32000: Connection closed
Failed to connect to MCP server: MCP error -32000: Connection closed
```

```
% java -version
openjdk version "11.0.27" 2025-04-15 LTS
OpenJDK Runtime Environment Corretto-11.0.27.6.1 (build 11.0.27+6-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.27.6.1 (build 11.0.27+6-LTS, mixed mode)
```

```
% ./maestro-cli/build/install/maestro/bin/maestro mcp 
Exception in thread "main" java.lang.UnsupportedClassVersionError: io/modelcontextprotocol/kotlin/sdk/server/Server has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1022)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:800)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:698)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:621)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:579)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
	at maestro.cli.mcp.McpServerKt.runMaestroMcpServer(McpServer.kt:37)
	at maestro.cli.command.McpCommand.call(McpCommand.kt:15)
	at maestro.cli.command.McpCommand.call(McpCommand.kt:7)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2314)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
	at maestro.cli.DisableAnsiMixin$Companion.executionStrategy(DisableAnsiMixin.kt:22)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at maestro.cli.AppKt.main(App.kt:156)
```

anything I'm doing wrong?

steviec:
Agh, sorry. I pushed my changes last night to `Maestro` without pushing my final changes to the `kotlin-sdk` fork, which changes the JVM toolchain version and compatibility target to the correct ones. Try rebuilding and it should pull in the updated the kotlin-sdk dependency and work correctly. If not, you might need to force it with `./gradlew :maestro-cli:installDist --refresh-dependencies`.  

I can't seem to get the Test workflow to pass; it keeps failing on the Java 11 tests on this test:

```
IntegrationTest > Case 121 - Cancellation before the flow starts skips all the commands() FAILED
    kotlinx.coroutines.TimeoutCancellationException at CoroutineDebugging.kt:42
        Caused by: kotlinx.coroutines.TimeoutCancellationException at Timeout.kt:188
```

When reviewing other test runs, this error seems to happen with other commits as well, so I'm not sure how much time to drop trying to fix it: https://github.com/mobile-dev-inc/Maestro/actions/workflows/test.yaml

steviec:
Note that I force-pushed to my branch to pickup changes from `main` in case it fixed these test failures, but it didn't. You'll need to fetch from the origin and do a `git reset --hard origin/steviec/mcp-server` to pick up my newest branch.

steviec:
(sorry for the insane comment thread)

I just did a battery of manual tests, and most clients were having a terrible time finding elements on the screen. I was able to fix a lot of the issues, but there are more to tackle. Here are some observations:
- inspect_view_hierarchy: JSON is far too verbose for dumping the view hierarchy. I did a [deep research pass](https://chatgpt.com/share/6855cc5f-a360-800c-977b-92e48342bfb7) on different ways to compress hierarchical data that 1) would reduce token count and 2) still be understandable by LLMs, and I came up with a pretty neat YAML schema that reduced the size of one example screen from a 123K json file to a 3K yaml file that was functionally lossless, and LLMs seem to understand it fine. I just pushed this change in the most recent commit.
- take_screenshot: certain MCP clients do not handle MCP image data well. In order to "use" an image returned by an MCP server, the client needs to either do separate analysis on the image and then give the context to the LLM, or feed it to a multi-modal model. Cursor does not do this; when Cursor uses the `take_screenshot` tool, the image will correctly show up in the chat window, but Cursor will completely hallucinate what it's seeing. The IDE must not actually feed the image to a multi-modal model. Claude Code handles it great, and is able to correctly identify all elements on the screen from the screenshot. I don't think there's anything we can do about it, but I'll do some testing to see if I can add description information that will more clearly have the calling client indicate whether or not it can actually "see" the results.
- tap_on: there's more work I can do here.

herval:
wrt view hierarchy - what I did on the python MCP was using the `--compact` flag, which renders it as a csv instead. Saves a gazillion tokens and seems to work well.

herval:
ok managed to run this time!

as I expected, the code writing is sorta flaky without the cheat sheet (it hallucinates a bunch). But that's somethign we can fix on a subsequent PR.

could you address the failing build? Then we can merge

herval:
merging! let's continue development on follow-on PRs, if needed

brilliant work @steviec ! Please reach out to me on the Slack community, I'd love to send you some swag 🥇

PR #2534, opened 2025-06-11
[Web] Adding support for <select> tag

dmitry-zaitsev:
`<select>` tags are handled differently than every other DOM element. For one - options are not rendered as part of the DOM but rather by the browser itself, so we do not get any `bounds` back.

image

To work around that, we pretend that there are visible `option` elements on the screen, just very far away (in what I call a "synthethic coordinate space"). If such element is clicked, we detect it and handle tap ourselves (without Selenium). That allows to keep the rest of the Maestro implementation untouched.

### DIFF
--- a/maestro-client/src/main/java/maestro/ViewHierarchy.kt
+++ b/maestro-client/src/main/java/maestro/ViewHierarchy.kt
@@ -62,7 +62,7 @@ value class ViewHierarchy(val root: TreeNode) {
         return matches[0]
     }
 
-    private fun getElementAt(
+    fun getElementAt(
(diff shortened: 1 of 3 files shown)


herval (approved):
A common (and I imagine less flaky) way people use select elements on playwright/selenium/etc is selecting by element value or by element text. Do we support either in any way? (also on mobile, I suppose)

dmitry-zaitsev:
@herval we allow selection by text. A usual select flow will look like this:

```
- tapOn: Select Value
- tapOn: Item B
```

PR #2537, opened 2025-06-15
[Web] CSS/DOM selector

dmitry-zaitsev:
Allowing elements to be selected via CSS/DOM query

```
- assertVisible:
    css: .myTag
```

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -2,15 +2,14 @@ package maestro.test
 
 import com.google.common.truth.Truth.assertThat
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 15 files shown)


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -3778,25 +3776,51 @@ class IntegrationTest {
...
+    @Test
+    fun `Case 125 - Assert visible by CSS selector`() {
+        // Given
+        val commands = readCommands("125_assert_by_css")
+
+        val driver = driver {
+            element {
+                bounds = Bounds(0, 0, 100, 100)
+                text = "Test Element"
+                matchesCssFilter = ".test"
```
herval, line 3835: Do #id and nth-child work? Would be good to have tests for them

dmitry-zaitsev: As discussed, they do

PR #2538, opened 2025-06-16
[MA 3384] Handling Timed out while evaluating UI query on iOS

amanjeetsingh150:
## Proposed changes

In cases where app screen is too complex for example loading too much data at once on table views. Snapshot call on iOS XCTest framework fails with following message:

```
Timed out while evaluating UI query
``` 

This PR handles these gracefully by raising DriverTimeout exceptions and showing debug message to user to simplify hierarchy.

Message:

```
Your app screen might be too complex.
                    
* This usually happens when the screen has very large view hierarchies, such as table views loading with large amount of data.
* Try loading fewer cells initially or implementing lazy loading to reduce the load during tests.
```

## Testing

- [x] Locally

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,6 +33,7 @@ jobs:
           cache: gradle
 
       - name: Test
+        id: unit-test
         run: ./gradlew test
 
       - name: Run Integration test
@@ -51,6 +52,15 @@ jobs:
         with:
           name: maestro-root-dir-android
           path: ~/.maestro
+          retention-days: 1
+          include-hidden-files: true
+
+      - name: Upload test report
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: maestro-test-report
+          path: ./maestro-test/build/reports/tests/test/index.html
           retention-days: 1
           include-hidden-files: true
 


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -544,6 +544,17 @@ class IOSDriver(
         } catch (appCrashException: IOSDeviceErrors.AppCrash) {
             LOGGER.error("Detected app crash during $callName command", appCrashException)
             throw MaestroException.AppCrash(appCrashException.errorMessage)
+        } catch (timeoutException: IOSDeviceErrors.OperationTimeout) {
+            throw MaestroException.DriverTimeout(
+                message = "Maestro driver timed out during $callName call",
+                operation = callName,
+                debugMessage = """
```
herval, line 550: there's a similar debug message somewhere else. Why duplicate it here?

maestro-client/src/main/java/maestro/Errors.kt
```diff
@@ -27,7 +27,7 @@ sealed class MaestroException(override val message: String) : RuntimeException(m
 
     class AppCrash(message: String): MaestroException(message)
 
-    class DriverTimeout(message: String): MaestroException(message)
+    class DriverTimeout(message: String, val operation: String? = null, val debugMessage: String? = null): MaestroException(message)
```
herval: exceptions are a weird place to put this IMO. Is there another way we can do this?

amanjeetsingh150: If you are referring to operation, I agree let me change that.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
amanjeetsingh150: Ohh yeah that other PR just commons it out, its not an duplicate:

https://github.com/mobile-dev-inc/Maestro/pull/2542/files#diff-25ebfb9b3069b37481b09445ea8c4458fdebdc760b55ef6e9af537f0fc1e9221R548

This one is base for the above PR.

maestro-client/src/main/java/maestro/Errors.kt
herval: operation AND debugMessage. Not sure why have them here..

amanjeetsingh150: Done ✅

PR #2544, opened 2025-06-16
Tests automation xcodebuild

amanjeetsingh150:
## Proposed changes

copilot:summary


.github/scripts/boot_simulator.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Specify the device type and runtime as per your requirements
+DEVICE_TYPE="iPhone 15 Pro"
+RUNTIME="iOS17.5"
```
Fishbowler, line 5: Any chance these tests could vary in behaviour on different runtimes?

amanjeetsingh150: We can, I think I would need to all the supported iOS version. Do we have docs somewhere whats the min iOS version? I think there was confusion it was iOS 14. I feel it has to be aligned with cloud which is on iOS 15 right now.

Issue #2012, opened 2024-09-03
A way for Maestro to read a variable easily


AnaLuisaFav:
### Use case

It should be simpler to Maestro read variable values. For example, for aesthetic and organizational reasons, I want to access flows that are in another folder. So I have to write
`- runFlow: ...Ubuntu\home\ana\app-syndic-automation\tests\login.yaml`
`- runFlow:...Ubuntu\home\ana\app-syndic-automation\flows\subflows\accessPixComponent.yaml`
`- runFlow:...Ubuntu\home\ana\app-syndic-automation\flows\subflows\accessPixKeyArea.yaml`

This visually looks ugly and gives an impression of disorganization. So I didn't sit still until I found a way to do it, having to write a variable processing script in python. See, something that seems so simple, I had to make a SCRIPT for it. 

And this is how it is now:

`- runFlow: {{ LOGIN }}/login.yaml`
`- runFlow: {{ CONDOMINIUM }}/accessPixComponent.yaml`
`- runFlow: {{ CONDOMINIUM }}/accessPixKeyArea.yaml`

So my suggestion is to find an easier way to do this.

### Proposal

![image](https://github.com/user-attachments/assets/2eaf3b6a-cc74-4fee-a1f6-7421a971b82b)

bartekpacia:
Hey @AnaLuisaFav 👋🏻 Thanks for taking time to share this feature request with us.

I thought you could already do something similar with JavaScript and `output` object, but turns out it's not possible.

**.maestro/demo.yaml**

```yaml
appId: com.example.example
---
- launchApp:
- runScript: setVars.js
- runFlow: ${output.FLOW_DIR}/fill_form.yaml
```

**.maestro/setVars.js**

```js
output.FLOW_DIR = './my/very/deep/dir/structure'
```

**.maestro/my/very/deep/dir/structure/fill_form.yaml**

```yaml
appId: com.example.example
---
- launchApp
```

It fails with:

```
Failed to parse file: /Users/bartek/projects/maestro/e2e/workspaces/demo_app/demo.yaml
Flow file does not exist: file:///Users/bartek/projects/maestro/e2e/workspaces/demo_app/$%7Boutput.FLOW_DIR%7D/fill_form.yaml
```

bartekpacia:
I'm gonna change title to make it possibe to allow JS variables interpolation within `runFlow`'s `path` argument.

InkAndSteel:
I also need this, implementation would be much appreciated!

[...]

Issue #1314, opened 2023-08-03
[Feature Request] Parameter interpolation should work for runFlow.file


solidcell:
**Is your feature request related to a problem? Please describe.**
In order to dry usage of nested flows, I'd like to define constants for their paths. However, parameters don't seem to work when used for `runFlow.file`.

For instance:

`$ export MAESTRO_EMAIL_LOGIN_SUBFLOW=/path/to/subflows/email-login.yaml`

```yaml
...
- runFlow:
    file: /${MAESTRO_SUBFLOW_EMAIL_LOGIN}
...
```

Note: I'm using a leading `/` so that the path isn't interpreted as a relative path. Ostensibly, this wouldn't be necessary if interpolation worked, since it would see there's already a leading `/`.

**Describe the solution you'd like**
I'd like for the above to interpolate correctly. However, it looks like interpolation doesn't occur. The text is taken literally.

I get a failure with the following error message:
`Flow file is invalid: /${MAESTRO_SUBFLOW_EMAIL_LOGIN}`

[The documentation](https://maestro.mobile.dev/advanced/parameters-and-constants) doesn't go into specifics about where interpolation is permitted, but from the examples and my experimentation, it seems like it only works on top-level nodes.

**Describe alternatives you've considered**

1. I could just continue to fully type out the paths, which is quite verbose and not DRY.
2. I could run some external preprocessor to interpolate it myself and not rely on Maestro's interpolation, however this doesn't seem worth it. The documentation should at least be specific about the intended usage and limitations.

ricardo7martins:
This would be super helpful. Interesting how this is a thing for actions, but not for runFlow/runScript. As pointed above, having to type the same path on several files is quite cumbersome. I was trying to create an output.paths for a while before coming here and seeing it's not a thing already. Sad

kris-luminar:
I'm encountering this same limitation. I agree with OP that this should be supported and if not, should be explicitly stated in the docs that file: doesn't accept variable input.

I tried a million different variations and nothing works.  I tried using a JavaScript variable too, which is my current use-case.

ramosjoel:
I'm running into the same thing as OP. The real issue is working with relative paths where a flow to relies on another file (yaml or JS).

For me, relative paths work locally but don't work in CI when I try to run tests in Robin Cloud via the `maestro cloud` command.

I'm trying to have separate iOS and Android tests rely on common JS scripts that are not platform-specific.

```
js/
- script.js
ios/
[...]

PR #2547, opened 2025-06-19
Add support for string interpolation in runScript file names

Dambakk:
## Proposed changes

Will enable `runScript` commands to have js string interpolation in their file names. Example:

```
appId: com.example.app
env:
    SCRIPT_FILE: 125_myScript.js
---
- runScript: 125_myScript.js
- runScript:
      file: ${SCRIPT_FILE}
- runScript: ${ (120 + 5) + '_myScript.js' }
- evalScript: ${ output.somePath = "125_myScript.js" }
- runScript: ${ output.somePath }
```

This change moves some file handling from the `orchestra` module to the `orchestra-models` module (including the corresponding `InvalidFlowFile` exception), to be able to read the script files _after_ the `jsEngine` has done its magic and applied any string interpolation. This, however, means that any "file not found" issues are now not reported until runtime instead of when the maestro command is ran. After quick discussion with Dan Caseley, we figured this would be an acceptable tradeoff.

## Testing

I've added a test to the integration test suite, and built the Maestro CLI locally and tested on some sample flow.

## Issues fixed

This change does not fully fix a single issue, but is a partial fix or the beginning on a fix to the following issues:
- Discussing runFlow and runScript: https://github.com/mobile-dev-inc/Maestro/issues/2012
- Discussing runFlow: https://github.com/mobile-dev-inc/Maestro/issues/1314


maestro-orchestra-models/src/main/java/maestro/orchestra/util/InvalidFlowFile.kt
```diff
@@ -1,8 +1,8 @@
-package maestro.orchestra.error
+package maestro.orchestra.util
```
Fishbowler: @amanjeetsingh150 Got opinions on this change? It seems like a necessary evil, in order to move file parsing after path parsing in evaluateScripts, but I don't know if it causes any issues elsewhere?

maestro-test/src/test/resources/125_run_script_with_string_interpolation.yaml
Fishbowler, line 1: I love the variation in this test ❤️

Fishbowler (commented):
Some first thoughts. Gonna do a rebase now, and see how it runs.

Dambakk:
Thanks for having a look at this, hope we can get it merged soon! 🙏

Fishbowler:
@Dambakk You'll note that I moved the tests from 125 to 131 in my previous rebase, and that the next rebase will need to increment them further. I hate the sequencing in that file...

Dambakk:
Yeah, that looks like a bit of a hazzle..

Dambakk:
Is there anything I can do to help get this through? @Fishbowler

Fishbowler:
I've had a quick crack at rebasing this, but I think the changes in #2917 need some consideration. Is that something you fancy picking up @Dambakk ?

maestro-orchestra-models/src/main/java/maestro/orchestra/util/InvalidFlowFile.kt
Dambakk: I've moved this file:
maestro-orchestra-models/src/main/java/maestro/orchestra/util/InvalidFlowFile.kt -> maestro-orchestra-models/src/main/java/maestro/orchestra/error/InvalidFlowFile.kt
to better match the package names (felt å bit strange to have an Exception in an utils directory). What do you think?

Dambakk:
Yes! I've rebased and pushed now. What do you think?

oyanyev:
Hi @Dambakk , @Fishbowler! I would greatly appreciate if this PR is merged.

PR #2553, opened 2025-06-24
Add `format` argument to the `inspect_view_hierarchy` MCP tool

Thomvis:
## Proposed changes
Add a `format` argument to the `inspect_view_hierarchy` MCP tool. This allows MCP clients to request the view hierarchy in JSON format. YAML remains the default format.

## Testing

Used [modelcontextprotocol/inspector](https://github.com/modelcontextprotocol/inspector)

## Issues fixed
N/A

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/InspectViewHierarchyTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/InspectViewHierarchyTool.kt
@@ -1,11 +1,12 @@
 package maestro.cli.mcp.tools
 
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import io.modelcontextprotocol.kotlin.sdk.*
 import io.modelcontextprotocol.kotlin.sdk.server.RegisteredTool
 import kotlinx.serialization.json.*
 import maestro.cli.session.MaestroSessionManager
 import maestro.TreeNode
-import kotlinx.coroutines.runBlocking
 
 object InspectViewHierarchyTool {
     fun create(sessionManager: MaestroSessionManager): RegisteredTool {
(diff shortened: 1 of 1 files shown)


herval:
Hi @Thomvis! What’s the benefit of using the json format in this case?

Thomvis:
@herval Hi! The primary reason for this PR is that I've been feeding json output of Maestro to a LLM over the past weeks and it has been working well. In an earlier prototype of that project I did use yml, for token reasons, but I decided to switch because tokens are not that expensive and LLMs have, at least in theory, better support for JSON. JSON is so common, there is just so much more of it in the training data and it is also trained to produce valid JSON. But to be honest, I have no strong evidence that it is better than yaml.

herval:
In my original python POC, I ended up outputting the view hierarchy as a CSV bc the JSON is enormous and seemed to confuse the LLM a lot. Wonder if there’s a way for us to compare the two modes here… @steviec was talking about creating an eval, perhaps we can do that?

steviec:
I did a deep research pass on this exact question, and [the results of that](https://chatgpt.com/share/6855cc5f-a360-800c-977b-92e48342bfb7) led me to the current implementation in the MCP server's `inspect_view_hierarchy` tool. But implementing based on the theoretically good format is very different than adapting to "real world" examples, so we should just test these out and see which is consistently the most effective (balancing minimal token count with LLM understandability).

I'm working on an eval harness right now, and that should allow us to try out some of these different formats with different prompts.

From my manual testing so far, I've actually found using `take_screenshot` with a multimodal LLM to be pretty darn effective, except with React Native, which has a difficult time tapping on input elements based on placeholder text. It's only able to correctly input text when it uses `inspect_view_hierarchy` and finds the `resource-id` and uses that to input the text.

steviec:
Update: now that the eval harness is live, I'll put together some tests during some free time tomorrow so we can compare. I'm assuming that our goals is to consume the smallest number of tokens while still retaining relatively "complete" understanding of the view hierarchy. If there's any specific tests that come to mind to you guys let me know; I was planning on just loading a somewhat complex screen and having the LLM describe it using various formats.

herval:
I think that should be enough, yes!

Curious about the results we’ll see

steviec:
Sorry for the slow turnaround on this, I had to pause to wrap up some other work, I'll get a full update by early next week.

I've tested CSV vs JSON vs YAML on about 10 different screen types with different elements, nesting, buttons, etc. And I'm seeing pretty similar levels of comprehension across modern LLMs! If that holds, then I think we should just go with whatever format uses the least tokens (likely CSV). Stay tuned.

steviec:
Update: I think we should just use CSV. I put the rationale in the PR here: https://github.com/mobile-dev-inc/Maestro/pull/2577

steviec:
@Thomvis , I'm going to close this one, since I think it's easier for LLMs to just have one format type instead of having to figure out which is best for their use case. The other PR is merged and will go out in the next release.

PR #2565, opened 2025-07-01
(fix) Fix path name resolution issues for MCP server's run_flow_files tool

steviec:
## Proposed changes

There's been a few complaints in the Slack forum about flow file paths not being resolved correctly when working with LLMs to create flow files. I've narrowed this down to a few distinct problems and corresponding solutions.

1. Vague flow requests don't succeed ("run my dashboard flow")
I was able to reproduce this using the evals framework consistently. Part of the problem is that some LLMs will just refuse the task, since it's not specific enough, and some will just try "dashboard.yaml" in the root directory. I've added some additional context that solves it in my local test cases, but we might need to add more context.

2. Cursor does not start the MCP server in the project's working directory
This is really, really annoying. When Cursor launches an MCP server, the working directory is the directory _of the mcp server's executable_, which means paths relative to the project root are unable to be resolved (unbelievable! [but](https://forum.cursor.com/t/how-get-the-correct-current-work-directory-in-mcp-server/99215) [true](https://forum.cursor.com/t/allow-workspacefolder-in-mcp-project-configration/74861)). The solve for this was to add a `--working-dir` param to the MCP server CLI and, in cursor's case, pass ".". Hat tip to [svallory for the suggestion](https://forum.cursor.com/t/allow-workspacefolder-in-mcp-project-configration/74861/6?u=steviec
). This should work until Cursor supports a more official way of passing in the working directory.

## Testing

I updated the MCP test harness's `mcp-server.json` config file to use the new `--working-dir` parameter, and then added some tests in the eval harnes that would only work if paths relative to the working directory were functioning.

## Issues fixed

None had been filed.


maestro-cli/src/test/mcp/mcp-server-config.json
```diff
@@ -2,7 +2,11 @@
   "mcpServers": {
     "maestro-mcp": {
       "command": "maestro-cli/build/install/maestro/bin/maestro",
-      "args": ["mcp"],
+      "args": [
+        "mcp",
+        "--working-dir",
```
herval, line 7: given the mcp configuration is editor-wide and working dirs will change depending on whatever codebase you're working on, will this work?

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunFlowFilesTool.kt
```diff
@@ -9,13 +9,14 @@ import maestro.orchestra.yaml.YamlCommandReader
...
 import java.nio.file.Paths
+import maestro.cli.util.WorkingDirectory
 
 object RunFlowFilesTool {
     fun create(sessionManager: MaestroSessionManager): RegisteredTool {
         return RegisteredTool(
             Tool(
                 name = "run_flow_files",
-                description = "Run one or more full Maestro test files. If no device is running, you'll need to start a device first.",
+                description = "Run one or more full Maestro test files. If no device is running, you'll need to start a device first. If the command fails using a relative path, try using an absolute path.",
```
herval, line 19: are there any env vars that cursor/windsurf/etc already inject that we could use as working dir too? For instance `$PWD` when using claude code...

maestro-cli/src/test/mcp/mcp-server-config.json
herval: (just read the description, guess it answers this)

steviec: Great question, I had the same initial concern. Cursor will automatically expand "." to the working directory of whatever Cursor instance is running, and since each Cursor instance will use it's own maestro MCP server process, that process will get the correct working directory. I added debug logging to the MCP server startup that I could inspect in Cursor's MCP Output log panel, and when launching Cursor from different repos, I confirmed that different working directories were set.

I'm still not 100% sure this is the right approach. Cursor specifically calls out this issue on their debugging page [here](https://modelcontextprotocol.io/docs/tools/debugging#working-directory), and unhelpfully they say to just use absolute paths. But in their example, they pass the path to the `server-filesystem` as a CLI parameter. See here:
```
{
  "command": "npx",
  "args": [
    "-y",
    "@modelcontextprotocol/server-filesystem",
    "/Users/username/data"
  ]
}
```

To make things more complicated, I found [this forum response](https://forum.cursor.com/t/workspace-level-mcp-json-execution-context/98599) from a Cursor engineer, where he says to use the `WORKSPACE_FOLDER_PATHS` env variable (which is not documented _anywhere_, so I'm leery about relying on it). This variable will not get auto-expanded in the mcp server config file, so we'd have to hard code the env var check in our MCP server, which is a bit of a smell. I think passing the path as a CLI param decouples our code from Cursor-specific logic so it felt like the least bad option.

steviec: (but I'm happy to change this to whatever if you prefer using WORKSPACE_FOLDER_PATHS)

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunFlowFilesTool.kt
steviec: See the above reply. Cursor has `WORKSPACE_FOLDER_PATHS`, but it feels brittle to hard code Cursor-specific syntax in the Maestro MCP codebase when I think we get the same benefit by passing it to `--working-dir` on the command line.

maestro-cli/src/test/mcp/mcp-server-config.json
herval: IMO a potential thing we could try is using a WORKSPACE_FOLDER_PATHS, then PWD, in sequence. That'd work w/ _some_ clients at least. 

What I did on the original mcp was also to explain in the tool documentations that it should always use full paths - another potential mitigation

steviec: I'd be fine taking that approach for now if you want to avoid additional params. Let me know! 

And I like the idea of just telling the LLM to always use absolute paths; it might make most of the problems just go away. I'll make a change in the `run_flow_files` so that it more closely reads like [your original description](https://github.com/mobile-dev-inc/maestro-mcp/blob/a827ffc3b19a08f5cb6fa32a9950002c4d743ae2/src/maestro_mcp/cli.py#L147).

herval: I guess we can iterate on those

PR #2567, opened 2025-07-01
Enhanced MCP Integration with Environment Variables and Proper Hook Execution

atakankarsli:
While integrating the new MCP feature you've added into our internal Maestro fork, I noticed some functionality wasn't working properly in our tests and ended up fixing them. I wanted to contribute these fixes back to the main repo as well.

### Environment Variables Weren't Working
MCP tools couldn't use environment variables like terminal commands (`-e APP_ID=com.example.app -e LANGUAGE=tr`). We start our tests with these envs to run them regionally.

### onFlowStart/onFlowComplete Hooks Weren't Executing  
JavaScript setup scripts, deeplinks, and wildcard imports (`*Users.js`) were silently failing. We use runScript wildcards inside onFlowStart to load them, to better utilize page object model in tests (this was added to runFlow).

### Port Conflicts
MCP and terminal commands were interfering with each other (same issue causes tests to not run when Maestro Studio is open). Both were using the same port (7001). Perhaps this could be addressed with a better approach for session cleanup.

## This MR fixes them by:

**Environment Variables:**
- Added `env` parameter to `run_flow`, `run_flow_files`, `launch_app`, `stop_app` tools
- `${VAR_NAME}` syntax works compatibly with terminal

**Hook Execution Fix:**
```kotlin
// Before (broken)
orchestra.executeCommands(commands) 

// After (fixed)  
orchestra.runFlow(commands)
```

**Port Isolation:**
- MCP: 7200+ port range
- Terminal: 7001 port
- `newMcpSession()` with dedicated session management

## Result
Now MCP tools work identically to terminal commands:
- ✅ Environment variables
- ✅ JavaScript hooks  
- ✅ Wildcard script imports
- ✅ Port isolation

Let me know what do you think about these issues and fixes  @steviec @herval


maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -57,10 +57,19 @@ object MaestroSessionManager {
     private const val defaultHost = "localhost"
     private const val defaultXctestHost = "127.0.0.1"
     private const val defaultXcTestPort = 22087
+    private const val mcpPortStart = 7200 // MCP uses different port range
```
herval: I dont think we should leak MCP implementation details into this class.

The `newMcpSession` impl below looks like the regular `newSession` too - why duplicate it?

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunFlowFilesTool.kt
```diff
@@ -71,37 +87,46 @@ object RunFlowFilesTool {
...
+                    val finalEnv = env
+                        .withInjectedShellEnvVars()
+                        .withDefaultEnvVars()
+                    
                     buildJsonObject {
                         put("success", results.all { (it["success"] as Boolean) })
                         put("device_id", deviceId)
                         put("total_files", flowFiles.size)
                         put("total_commands_executed", totalCommands)
+                        put("env_vars_count", finalEnv.size)
```
herval: why do you need this count?

atakankarsli: removed

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
atakankarsli: I've removed newMcpSession, set MCP sessions to use port 7200 and added logic to prevent terminal sessions from reusing existing MCP sessions to avoid conflicts.

maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
```diff
@@ -88,7 +88,7 @@ object TapOnTool {
                 val result = sessionManager.newSession(
                     host = null,
                     port = null,
-                    driverHostPort = null,
+                    driverHostPort = 7200, // Fixed port for MCP to avoid conflicts
```
herval: this should probably be a constant?

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
```diff
@@ -105,11 +104,12 @@ object MaestroSessionManager {
...
+            } else if (driverHostPort == 7200) {
+                // MCP sessions can reuse existing sessions for faster execution
+                SessionStore.hasActiveSessions(sessionId, selectedDevice.platform)
             } else {
-                SessionStore.hasActiveSessions(
-                    sessionId,
-                    selectedDevice.platform
-                )
+                // Terminal sessions should not reuse MCP sessions to avoid conflicts
+                false
```
herval, line 117: not sure I understand this change - IIUC this changes the default behavior of maestro cli from reusing to not reusing sessions. Why?

atakankarsli: Because after using mcp tools once its not possible to run any terminal commands right now 

This allow terminal commands to create new sessions and not get errors

maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
atakankarsli: yes it can be

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
herval: won't that get fixed by the change in port number for MCP? I'm not sure the cli behaviour should change

atakankarsli: It wasn’t enough — the terminal still tried to use an old session from mcp on a different port, which caused a new error 😢

I’m not saying this is the ideal way either. Maestro Studio has the same issue — once enabled, running tests throws appId = null errors for terminal commands. It’s a known pain.

I was hoping you’d have a cleaner suggestion for these session issues that I  I could apply 😄

herval: @amanjeetsingh150 do you know why we reuse sessions on maestro CLI? Is that necessary?

maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
herval: could you resolve the conflicts + address this constant? Waiting on @amanjeetsingh150 on the question around session reuse, seems to be the only pending point otherwise

atakankarsli: done

maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt
caffeineflo: I can confirm the same thing, the mcp won't be able to do any calls without this change, reusing a session doesn't seem to work

amanjeetsingh150: > @amanjeetsingh150 do you know why we reuse sessions on maestro CLI? Is that necessary?

The reason to reuse session was to prevent driver getting install again everytime by default when we trigger 2nd maestro command. 

We could change that if it breaks MCP, curious though @atakankarsli what's the error can you send me the steps to reproduce? Maybe we can fix that instead. I can suggest once I am able to reproduce this.

atakankarsli: 1. You should run some MCP tools first.
2. After that, try running the tests from the terminal.

It will throw an error like: appId = null when it starts running.
Also if you run maestro studio in one terminal, then try running tests from other terminal same error happens @amanjeetsingh150

steviec:
Thanks for the PR @atakankarsli! I just made some changes and created a new PR: https://github.com/mobile-dev-inc/Maestro/pull/2597

Here's the rationale for my changes:
- Env vars: I kept your changes for `run_flow_files` and `run_flow` tools, but didn't keep those for `launch_app` or `stop_app`. Env var replacement is conceptually just for flow files, and I don't want to leak that behavior into our "direct" tool calls. I'm assuming LLMs should still be able to handle launching/stopping fine, but let me know if I'm missing something.
- Hooks fix: great catch, thank you for finding this.
- Port Conflicts: I took the sessions/ports changes out of this PR. You've pinpointed a really confusing part of the system, and I'm opening up a new ticket to address this in a more sustainable way. I'll link that ticket here.

As a part of this, I added some tests to our harness and discovered that our MCP server was not setting `isError: true` if any flow failed during `run_flow_files`, so I fixed that, too.

PR #2577, opened 2025-07-09
Optimize view hierarchy's CSV format

steviec:
## Proposed changes

This PR makes the MCP server's `inspect_view_hierachy` tool return flattened CSV very similar to the `--compact` option in the CLI. The goal was to find a format that 1) used minimal tokens while 2) maintaining LLM interpretability. In the eval test harness, the CSV format performed basically as well as YAML and JSON with a smaller payload. Note that it performed _slightly_ less well using models from >1 year ago, but I don't think we should optimize for those models.

Additional improvements were made to the `--compact` format's output:
- removed duplicate attribute entries
- pulled "bounds" out into its own column to reduce verbosity

This PR also has the following changes:
- reorganized the MCP test suite for easier comprehension
- added the ability to leverage the e2e test suite's app setup for our evals harness
- pulled out the ViewHierarchy formatters into their own helper function and maintained the existing YAML/JSON output in case we want to return to them (or borrow some techniques from them, e.g. removing unnecessary nested views)

If this PR is accepted, we should close the related PR that adds additional formats: https://github.com/mobile-dev-inc/Maestro/pull/2553, since giving the LLM the ability to choose from different formats did not appear to improve performance.

## Testing

Ran the existing gradle test suite and the mcp-specific test suite (tools and evals).

## Additional improvements

These additional improvements could be added to this PR or done later:
- have `--compact` use the same formatter instead of using its own
- update the new CSV output to omit entries that have identical bounds to their children to avoid unnecessary duplicate of empty views


maestro-cli/src/test/mcp/evals/inspect-view-hierarchy-evals.yaml
```diff
@@ -0,0 +1,131 @@
...
+    - start_device
+    - launch_app
+  responseScorers:
+  - type: llm-judge
+    criteria: >
+      The assistant successfully uses the inspect_view_hierarchy tool with format=csv.
+      The assistant correctly identifies two buttons: "allow", and "don't allow" in the middle of the screen.
+    threshold: 0.8
+
+# NOTE: Below are old tests when we were doing evals on various format types. We might want to resuscitate these at some point
```
herval: why comment those out?

maestro-cli/src/test/mcp/modelcontextprotocol-inspector-0.15.0.tgz
herval, line 1: what's this?

maestro-cli/src/test/mcp/tool-tests/test-all-mcp-tools.sh
```diff
@@ -1,20 +1,21 @@
...
-CONFIG="maestro-cli/src/test/mcp/mcp-server-config.json"
+CONFIG="./mcp-server-config.json"
 SERVER="maestro-mcp"
 
 failures=0
 tests_run=0
 
 function inspector() {
-  npx @modelcontextprotocol/inspector --cli --config "$CONFIG" --server "$SERVER" "$@"
+  npx -y "./modelcontextprotocol-inspector-0.15.0.tgz" --cli --config "./mcp-server-config.json" --server "$SERVER" "$@"
```
herval: why can't we keep using the published npx one?

maestro-cli/src/test/mcp/evals/inspect-view-hierarchy-evals.yaml
steviec: Those evals were only for when I exposed a "format" property that the agent could use to test out the different formats. Now that we've settled on one for now, I removed the format property so prompts like `Using the inspect_view_hierarchy tool with the "yaml" format, describe the number of clickable elements on the screen, and their position on the screen. Do not take a screenshot.` would no longer work.

I could just remove those commented out lines entirely, but if we wanted to re-test any of these in the future it could be handy to have those in there.

maestro-cli/src/test/mcp/modelcontextprotocol-inspector-0.15.0.tgz
steviec: This is my forked version, since there is no "published" version of the inspector that has the evals framework in it so I needed to build it (using `npm pack`) and add it to the codebase. The PR I made to the upstream inspector repo was rejected (interesting thread [here](https://github.com/modelcontextprotocol/inspector/pull/568)), but I just finished pulling it out into a standalone CLI tester today. I'll update this PR with that version so we don't have to include this package.

maestro-cli/src/test/mcp/tool-tests/test-all-mcp-tools.sh
steviec: Since I was already having to bundle my forked version to support evals, I thought it was better to keep using it than to use the anthropic version of the `inspector` tool.  However, now that I finished my standalone version (doing final tests on it now) that can just be run with `npx mcp-server-tester` I'll update this PR to only use my fancy new one :).

steviec:
@herval : Ok, I've swapped out the old forked inspector and added my new `mcp-server-tester` tool with the new file formats. Take a look and let me know.

steviec:
This would only impact the CLI, so I'm going to move forward.

PR #2578, opened 2025-07-10
DriverBuilder: make the xcodebuild timeout configurable

jochen-testingbot:
## Proposed changes

Instead of the hardocded `process.waitFor(120, TimeUnit.SECONDS)` we propose a way to set this with an environment variable (`MAESTRO_XCODEBUILD_WAIT_TIME`), which could help on slower host machines.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/driver/DriverBuilder.kt
+++ b/maestro-cli/src/main/java/maestro/cli/driver/DriverBuilder.kt
@@ -8,7 +8,6 @@ import java.util.concurrent.TimeUnit
 import kotlin.io.path.pathString
 
 class DriverBuilder(private val processBuilderFactory: XcodeBuildProcessBuilderFactory = XcodeBuildProcessBuilderFactory()) {
-    private const val DEFAULT_XCODEBUILD_WAIT_TIME: Long = 120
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/driver/DriverBuilder.kt
```diff
@@ -8,6 +8,9 @@ import java.util.concurrent.TimeUnit
 import kotlin.io.path.pathString
 
 class DriverBuilder(private val processBuilderFactory: XcodeBuildProcessBuilderFactory = XcodeBuildProcessBuilderFactory()) {
+    private val XCODEBUILD_WAIT_TIME: Long by lazy {
+        System.getenv("XCODEBUILD_WAIT_TIME")?.toLongOrNull() ?: DEFAULT_XCODEBUILD_WAIT_TIME
```
herval: Can you make this prefixed with MAESTRO_? That way it’s namespace like the other env vars and isn’t likely to clash with other systems

jochen-testingbot: Changed in https://github.com/mobile-dev-inc/Maestro/pull/2578/commits/e2ee096904262d7dc3d9c3f27377aecf129eb36e

Issue #2587, opened 2025-07-17
Environment variables not working in Web Beta


CalvinNFT:
**About**

When running `maestro test --headless tests/test-suite.yaml` an error of `Want to use 0 devices, which is not enough to run 1 shards. Missing 1 device(s). [\n] Not enough devices connected (1) to run the requested number of shards (1).`

The issue arose after converting a hardcoded URL value to an environment variable. The issue persists with all methods of passing params / env variables to yaml files, including -e flag.

**Reproduction**
Test script setting the env:
```shell
#!/bin/bash

export MAESTRO_URL="https://example.com"
maestro test --headless tests/test.yaml
```

Test file:
```yaml
url: ${MASTRO_URL}
---
- launchApp
```

**Expected result**

Open and run tests in the browser (Chromium). 
NOTICE: If the `${MASTRO_URL}` is changed to `https://example.com` the test suite runs without error.

**Conclusion**
Environment variables do not work currently in the web beta due to a bug.

**Issue Logs and CLI version**
[2025-07-17_160048.zip](https://github.com/user-attachments/files/21296594/2025-07-17_160048.zip)
[2025-07-17_160052.zip](https://github.com/user-attachments/files/21296596/2025-07-17_160052.zip)
[2025-07-17_160103.zip](https://github.com/user-attachments/files/21296595/2025-07-17_160103.zip)
[2025-07-17_160314.zip](https://github.com/user-attachments/files/21296597/2025-07-17_160314.zip)
[2025-07-17_160420.zip](https://github.com/user-attachments/files/21296598/2025-07-17_160420.zip)
[2025-07-17_160528.zip](https://github.com/user-attachments/files/21296593/2025-07-17_160528.zip)

```shell
❯ maestro -v
1.41.0
```

Issue #2383, opened 2025-03-19
Maestro Web not detected when appId is environment variable


pandeiro:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Maestro appears to use the presence of (http|s) in the `appId`/`url` field to resolve whether flows are browser or native-targeted.

When a flow uses an environment variable for this value (eg `maestro test -e APP_ID="http://localhost:3000" flow.yaml`), the current Maestro version fails to resolve and run tests in the browser.

### Actual results

Maestro will report insufficient shards, as if it expects mobile simulators/emulators to run, despite the flows targeting an 'http*' appId.

### Expected results

Maestro should wait to resolve environment variables before any resolution/dispatch logic depending on the value of appId.

### About app

Using an environment variable for `appId` is a common pattern especially for testing universal apps.

### About environment

macOS Sonoma 14.5 (23F79)
openjdk version "23" 2024-09-17

### Logs

<details>
<summary>Logs</summary>

```
$ maestro test -e APP_ID='http://localhost:3000' flow.yaml

Want to use 0 devices, which is not enough to run 1 shards. Missing 1 device(s).
Not enough devices connected (1) to run the requested number of shards (1).
```

</details>

### Maestro version

1.39.13

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

CodingItWrong:
I'm seeing the same problem. It looks like a workaround that will work is installing a simulator/emulator on the CI machine, even though it's not actually necessary for the test. For example, adding this step on a macOS runner fixed it for me:

```
      - name: 🏗 Set up simulator to work around https://github.com/mobile-dev-inc/Maestro/issues/2383
        uses: futureware-tech/simulator-action@v2
        with:
          model: iPhone 16 Pro Max
```

CodingItWrong:
@Angelk90 that error message isn't the "insufficient shards" error that this bug relates to; it seems to be a different issue than this one. So I doubt my full workflow will help with that. But here is my workflow if it helps anyone:

```yaml
name: "Build Web"

on:
  pull_request:
    types: [opened, synchronize, reopened]
  push:
    branches:
      - main

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

jobs:
  maestro-web:
    runs-on: macos-latest

    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Setup Node Project
[...]

PR #2611, opened 2025-07-29
Better detection of web flows

steviec:
We've used a string regex in the `File.isWebFlow` method to determine if a flow is a web flow, but this fails when flows use JS expressions since it hasn't evaluated expressions yet. This has led some people to do some [wacky stuff](https://github.com/mobile-dev-inc/Maestro/issues/2383). Instead of this working:

```
url: ${APP_ID}
---
- launchApp
```

They do this:
```
# Temporary until this bug is fixed:
# https://github.com/mobile-dev-inc/Maestro/issues/2383
appId: http://localhost:4000
jsEngine: graaljs
---
- launchApp: ${APP_ID} # This works around the above issue, couldn't tell you why
```

That `appId` line is just there to let the system know it's a web flow with the presence of "https:", but that's all it does. Then when `launchApp` is called, the JS expression is evaluated and pulls in the env's APP_ID. Fun!

To solve, this commit preserves the `url:` field in the YamlConfig, but then maps it to `appId` in the MaestroConfig. This allows the YamlConfig to more accurately reflect what's in the raw yaml, and allows us to just use the presence of the `url:` field to determine if a flow is a web flow.

The only downside I can see is if people are using "appId: https://example.com" (like in the hacky workaround above). This will no longer work, but I'm not sure if that's important since this is in beta. If we like this commit, we should also change this line in the documentation:

`For readability, you should specify url instead of an appId in your test file (but they're synonyms behind the scenes).`

If we DON'T like this approach, then anytime `File.isWebFlow` is called we'll need to have a JS engine available to use.

Fixes: https://github.com/mobile-dev-inc/Maestro/issues/2587
Fixes: https://linear.app/mobile-dev/issue/MAE-231/environment-variables-not-working-in-web-beta

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
@@ -2,6 +2,7 @@ package maestro.orchestra.yaml
 
 import com.fasterxml.jackson.annotation.JsonAlias
 import com.fasterxml.jackson.annotation.JsonAnySetter
+import com.fasterxml.jackson.core.JsonLocation
 import maestro.orchestra.ApplyConfigurationCommand
 import maestro.orchestra.MaestroCommand
 import maestro.orchestra.MaestroConfig
@@ -9,16 +10,34 @@ import maestro.orchestra.MaestroOnFlowComplete
 import maestro.orchestra.MaestroOnFlowStart
 import java.nio.file.Path
 
+// Exception for config field validation errors
+class ConfigParseError(
+    val errorType: String,
+    val location: JsonLocation? = null
+) : RuntimeException("Config validation error: $errorType")
+
 data class YamlConfig(
     val name: String?,
-    @JsonAlias("url")
-    val appId: String,
+    @JsonAlias("appId") private val _appId: String?,
+    
+    val url: String?, // Raw url from YAML - preserved to distinguish web vs app configs
     val tags: List<String>? = emptyList(),
     val env: Map<String, String> = emptyMap(),
     val onFlowStart: YamlOnFlowStart?,
     val onFlowComplete: YamlOnFlowComplete?,
     private val ext: MutableMap<String, Any?> = mutableMapOf<String, Any?>()
 ) {
+
+    // Computed appId: uses url for web flows, _appId for mobile apps
+    // Preserving both fields allows detecting web vs app configuration contexts
+    val appId: String
+
+    init {
+        if (url == null && _appId == null) {
+            throw ConfigParseError("missing_app_target")
+        }
+        appId = url ?: _appId!!
+    }
 
     @JsonAnySetter
     fun setOtherField(key: String, other: Any?) {
@@ -27,7 +46,7 @@ data class YamlConfig(
 
     fun toCommand(flowPath: Path): MaestroCommand {
         val config = MaestroConfig(
-            appId = appId,
+            appId = appId,  // maestro-cli uses url as appId for web flows
             name = name,
             tags = tags,
             ext = ext.toMap(),
--- a/maestro-orchestra/src/test/resources/workspaces/e018_config_missing_appId/error.txt
+++ b/maestro-orchestra/src/test/resources/workspaces/e018_config_missing_appId/error.txt
@@ -1,4 +1,4 @@
-> Config Field Required: appId
+> Config Field Required
 
 /tmp/WorkspaceExecutionPlannerErrorsTest_workspace/workspace/Flow.yaml:2
 ╭────────────────────────────────────────────────────────────────────────╮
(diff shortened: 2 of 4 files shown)


maestro-orchestra/src/test/resources/workspaces/e018_config_missing_appId/error.txt
```diff
@@ -1,19 +1,27 @@
...
+> Config Field Required
 
 /tmp/WorkspaceExecutionPlannerErrorsTest_workspace/workspace/Flow.yaml:2
 ╭────────────────────────────────────────────────────────────────────────╮
 │ 1 | name: MyFlow                                                       │
 │ 2 | ---                                                                │
 │     ^                                                                  │
 │ ╭────────────────────────────────────────────────────────────────────╮ │
-│ │ The config section is missing a required field: `appId`. Eg.       │ │
+│ │ Either 'url' or 'appId' must be specified in the config section.   │ │
```
amanjeetsingh150, line 9: It makes it little difficult to understand if we'll have both url and appId. Shouldn't the solution actually be ensuring url values also gets evaluated through js engine?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
```diff
@@ -2,32 +2,51 @@ package maestro.orchestra.yaml
...
+    val location: JsonLocation? = null
+) : RuntimeException("Config validation error: $errorType")
+
 data class YamlConfig(
     val name: String?,
-    @JsonAlias("url")
-    val appId: String,
+    @JsonAlias("appId") private val _appId: String?,
+    
+    val url: String?, // Raw url from YAML - preserved to distinguish web vs app configs
```
amanjeetsingh150, line 23: The MaestroConfig file would also need a change:

https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroConfig.kt

I suggest we use both appId and url. Only one of them can be present at one time if someone tries using both our deserializer should error out gracefully.

maestro-orchestra/src/test/resources/workspaces/e018_config_missing_appId/error.txt
steviec: Ah, the URL values _do_ get evaluated, it just happens much deeper in the code. The reason I didn't want to instantiate a JS engine at this point is because we check the config to determine if it's a webflow in a LOT of places. My sense is that in many of these callers we won't want to have to have a jsengine available, and we should just use the presence of `url:` to determine if it's a webflow or not. If we use that, we don't have to have the jsEngine available during parsing


Here's the full list of where we detect the web flow:

studio-server, in 2 places:

When uploading a flow, we check isWebFlow before checking to make sure an appId is present:
https://github.com/mobile-dev-inc/copilot/blob/79c0e31ec5f085e191eaad08778f2cea2ed5b0aa/studio-server/src/main/java/maestro/studio/core/cloud/CloudService.kt#L153

When running a flow in studio, we _re-implement_ the isWebFlow "https://" regex to check if the running device matches the flow type:
https://github.com/mobile-dev-inc/copilot/blob/79c0e31ec5f085e191eaad08778f2cea2ed5b0aa/studio-server/src/main/java/maestro/studio/core/command/FlowExecutor.kt#L58


maestro-cli "cloud" command:

When uploading a flow, we check isWebFlow before checking to make sure an appId is present:
https://github.com/mobile-dev-inc/Maestro/blob/b38380726a214dfb56820909b9f312edd77dfacb/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt#L80

When scanning the workspace, we check isWebFlow to determine if the user should have passed an "--app-file" on the CLI
  https://github.com/mobile-dev-inc/Maestro/blob/b38380726a214dfb56820909b9f312edd77dfacb/maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt#L266

When creating the manifest from the workspace, we use it to filter our flows that are not app flows to find the appId:
https://github.com/mobile-dev-inc/Maestro/blob/b38380726a214dfb56820909b9f312edd77dfacb/maestro-cli/src/main/java/maestro/cli/web/WebInteractor.kt#L29


maestro-cli "record" command:

We use it to throw an error if they attempt to use "record" against a web flow:
https://github.com/mobile-dev-inc/Maestro/blob/b38380726a214dfb56820909b9f312edd77dfacb/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt#L109

maestro-cli "test" command:

We use it to filter out browsers from the connected device list:
https://github.com/mobile-dev-inc/Maestro/blob/b38380726a214dfb56820909b9f312edd77dfacb/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt#L270

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlConfig.kt
steviec: We could do that, but in the PR I thought the simplest approach was to preserve `url` _only_ in the YamlConfig. This way the yaml config would be a pure representation of the Yaml file (instead of aliasing url => appId). But then when we hand off the information to MaestroConfig, we use `appId` as the unified identifier for both web and device flows. I'd be happy to make it url "all the way down", but that will changing a bunch of different files (though I would prefer it for clarity).

PR #2612, opened 2025-07-30
feat: isolate environment variables between peer runFlow commands

steviec:
Environment variables from one runFlow command were persisting and being accessible to subsequent peer runFlow commands, violating expected scoping rules. This change ensures proper isolation while maintaining inheritance from parent flows to subflows.

Implementation:
- Add enterEnvScope/leaveEnvScope methods to JsEngine interface
- RhinoJS: leverage existing JavaScript scope mechanism
- GraalJS: implement manual stack-based environment management
- Orchestra: wrap runSubFlow execution with env scoping

Testing:
- Added a flow test for both graal and rhino engines to confirm the behavior works as expected (see the two yaml files; they demonstrate it clearly. Thanks to @Fishbowler for a good example repro in the ticket!)
- Added simple unit tests for the new behavior in the two engine implementations

Only concern: I don't know how much memory increase there is for the Rhino engine in this approach. If there are heavily nested subflows, this will add one new scope layer for each subflow (instead of just one new scope per subflow there will be two).

Fixes: [MA-3504](https://linear.app/mobile-dev/issue/MA-3504/env-variables-should-be-cleared-up-between-flow-runs)

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -911,28 +911,34 @@ class Orchestra(
         config: MaestroConfig?,
         subflowConfig: MaestroConfig?,
     ): Boolean {
-        executeDefineVariablesCommands(commands, config)
-        // filter out DefineVariablesCommand to not execute it twice
-        val filteredCommands = commands.filter { it.asCommand() !is DefineVariablesCommand }
+        // Enter environment scope to isolate environment variables for this subflow
+        jsEngine.enterEnvScope()
+        return try {
+            executeDefineVariablesCommands(commands, config)
+            // filter out DefineVariablesCommand to not execute it twice
+            val filteredCommands = commands.filter { it.asCommand() !is DefineVariablesCommand }
 
-        var flowSuccess = false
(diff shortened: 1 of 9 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -911,28 +911,34 @@ class Orchestra(
         config: MaestroConfig?,
         subflowConfig: MaestroConfig?,
     ): Boolean {
-        executeDefineVariablesCommands(commands, config)
-        // filter out DefineVariablesCommand to not execute it twice
-        val filteredCommands = commands.filter { it.asCommand() !is DefineVariablesCommand }
+        // Enter environment scope to isolate environment variables for this subflow
+        jsEngine.enterEnvScope()
```
herval, line 915: just to be sure - does this work with subflows? In other words, do subflows create their own Orchestra instance at any point - which would accidentally wipe out the env vars? Or  is it one single Orchestra per full execution?

steviec: I chatted with Dan, and added a few more interesting cases to the YAML test files, everything is looking good, merging.

Issue #2049, opened 2024-09-12
Deprecate RhinoJS engine


bartekpacia:
## Plan

We want to deprecate RhinoJS engine, thus making GraalJS the only choice.

### Phase 1 - communicate

- [] Communicate to community our intent to deprecate RhinoJS
  - In our public Slack workspace, in the `#announcements` channel
  - In v1.39.0 release Medium post
  - In v1.39.0 CHANGELOG.md
  - In our [JavaScript docs](https://maestro.mobile.dev/advanced/javascript)

### Phase 2 - soft warning

- [ ] Add warning (in both CLI and Maestro Cloud) if GraalJS is not used to execute a flow (i.e. `jsEngine: graaljs` is not set, or `MAESTRO_USE_GRAALJS` env var is not set).
- [ ] Ping the community once again about our intent to deprecate GraalJS

### Phase 3 - change default

- Default to GraalJS (in both CLI and Maestro Cloud)
- Display warning if  is used

### Phase 4 - remove RhinoJS

- Remove `RhinoJS` completely from Maestro codebase
- Display warning if `jsEngine` or `MAESTRO_USE_GRAALJS` are set - say that "these config options don't have any effect anymore"

## Notes

- GraalJS was implemented in PR #1147
- As we move ahead with phases, we will listen for user feedback and help them migrate to GraalJS, by fixing possible bugs in Maestro when GraalJS is used and improving docs.

Fishbowler:
I switched all of my tests from Rhino to GraalJS.
No issues encountered, no changes needed. :+1:

FelixLisczyk:
I have installed Maestro 1.39.0 on macOS 15.0.1 (24A348) using Homebrew. When I run any script using the GraalJS engine, I encounter the following exception immediately:

```
Exception in thread "main" java.lang.NoSuchMethodError: 'void sun.misc.Unsafe.ensureClassInitialized(java.lang.Class)'
	at com.oracle.truffle.api.library.LibraryFactory.ensureLibraryInitialized(LibraryFactory.java:384)
	at com.oracle.truffle.api.library.LibraryFactory.getUncached(LibraryFactory.java:364)
	at com.oracle.truffle.api.library.LibraryFactory.<init>(LibraryFactory.java:210)
	at com.oracle.truffle.api.interop.InteropLibraryGen.<init>(InteropLibraryGen.java:170)
[...]

PR #2631, opened 2025-08-11
MA-3547 Make Graal default JS engine

steviec:
Graal is now the default engine, but we allow rhino if specified via `jsEngine: rhino` in the test config.

There are subtle differences in the Graal implementation that _could_ affect users in edge case situations. Here are the ones I discovered while updating the tests:
- correctly resolves `${parseInt(3)}` to `3` (rhino would resolve to `3.0`)
- preserves newlines in ENV vars passed in via CLI or YAML (rhino would strip those out)
- correctly escapes double slashes (rhino would run two passes of escaping, reducing `////` to `/`)

## Testing

- ran all tests including with `./gradlew build`
- did not run e2e tests (assuming CI will do that)

## Issues fixed
MA-3547
https://github.com/mobile-dev-inc/maestro/issues/2049

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -290,13 +290,19 @@ class Orchestra(
         if (this::jsEngine.isInitialized) {
             jsEngine.close()
         }
-        val shouldUseGraalJs =
-            config?.ext?.get("jsEngine") == "graaljs" || System.getenv("MAESTRO_USE_GRAALJS") == "true"
+        val isRhinoExplicitlyRequested = config?.ext?.get("jsEngine") == "rhino"
+        
+        // Warn users about deprecated Rhino JS engine
+        if (isRhinoExplicitlyRequested) {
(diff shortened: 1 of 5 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -290,13 +290,19 @@ class Orchestra(
...
             jsEngine.close()
         }
-        val shouldUseGraalJs =
-            config?.ext?.get("jsEngine") == "graaljs" || System.getenv("MAESTRO_USE_GRAALJS") == "true"
+        val isRhinoExplicitlyRequested = config?.ext?.get("jsEngine") == "rhino"
+        
+        // Warn users about deprecated Rhino JS engine
+        if (isRhinoExplicitlyRequested) {
+            logger.warn("⚠️  The Rhino JS engine (jsEngine: rhino) is deprecated and will be removed in a future version. Please migrate to GraalJS (the default) for better performance and compatibility.")
+        }
```
amanjeetsingh150, line 298: I don't think this will be shown to users this just redirects to the maestro.log file on debug artifacts.

For showing to users we have Insights abstraction. You can use that to show this warning.

PR #2632, opened 2025-08-12
labels with JS injection can now be evaluated

jerriais:
## Proposed changes

Label attribute would only show a static string. Now it evaluates injected JS as described in the docs at https://docs.maestro.dev/advanced/javascript/run-javascript#inject 
copilot:summary

## Testing

I ran my full Maestro test suite using ./maestro from the root. It all worked and labels with JS injection were correctly evaluated.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -86,7 +86,8 @@ data class SwipeCommand(
         return copy(
             elementSelector = elementSelector?.evaluateScripts(jsEngine),
             startRelative = startRelative?.evaluateScripts(jsEngine),
-            endRelative = endRelative?.evaluateScripts(jsEngine)
+            endRelative = endRelative?.evaluateScripts(jsEngine),
+            label = label?.evaluateScripts(jsEngine)
         )
     }
 
@@ -150,6 +151,7 @@ data class ScrollUntilVisibleCommand(
             selector = selector.evaluateScripts(jsEngine),
             scrollDuration = scrollDuration.evaluateScripts(jsEngine).speedToDuration(),
             timeout = timeout.evaluateScripts(jsEngine).timeoutToMillis(),
+            label = label?.evaluateScripts(jsEngine)
         )
     }
 
@@ -161,7 +163,7 @@ data class ScrollUntilVisibleCommand(
     }
 }
 
-class ScrollCommand(
+data class ScrollCommand(
     override val label: String? = null,
     override val optional: Boolean = false,
 ) : Command {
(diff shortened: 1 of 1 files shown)


jerriais:
Just want to check one issue where a `- evalScript: ${console.log("Vehicle selection took "+(new Date().getTime() - vehicleSelectStartTime)/1000+" seconds")}` command is returning NaN instead of the time taken.

herval:
could you add a test case to `IntegrationTest.kt` with an example?

jerriais:
That was a weird one. When I run Maestro locally (using ./maestro test .... from root of the Maestro repo) it won't let me console log a variable I just declared in the previous line, it is undefined (first screenshot), but it will display it if I run the same test from the CLI (maestro test .... second screenshot) Maybe a timing issue?
Anyway, I see the issue on main branch of Maestro so at least I didn't cause the regression :face_exhaling:
image
image

herval:
> That was a weird one. When I run Maestro locally (using ./maestro test .... from root of the Maestro repo) it won't let me console log a variable I just declared in the previous line, it is undefined (first screenshot), but it will display it if I run the same test from the CLI (maestro test .... second screenshot) Maybe a timing issue? Anyway, I see the issue on main branch of Maestro so at least I didn't cause the regression 😮‍💨 image image

could you run the test with `jsEngine: rhino` and verify the behavior is the same as you have with local maestro? (you configure it on the [test header, like so](https://docs.maestro.dev/advanced/javascript/graaljs-support))

jerriais:
> could you run the test with `jsEngine: rhino` and verify the behavior is the same as you have with local maestro? (you configure it on the [test header, like so](https://docs.maestro.dev/advanced/javascript/graaljs-support))

I checked this with @Fishbowler on the community Slack and he confirmed it's because of the GraalJS update, so it is working as expected.

jerriais:
> could you add a test case to `IntegrationTest.kt` with an example?

I've just put in a commit with a basic test where I try to assert that a label on an assertTrue is correctly injecting the JS, but I can only find a way to assert a statement from a console.log. 
Is there a way I can assert the text output from a test run that includes the label text? Or a way to grab the label text and console log it so it appears in receivedLogs?

jerriais:
My integration test is now working. However test case 121 is failing, but that also fails on main so it's not a regression. Tried to fix it too but it would require a deeper dive than I have time for now.

jerriais:
@herval is there anything else that this PR needs before merging? It now has a working integration test (plus another integration test improved/updated, thanks @Fishbowler !) 
Test case 121 does fail but that appears to be currently failing on all PRs (mentioned here https://github.com/mobile-dev-inc/Maestro/pull/2635#issuecomment-3184056322 )

Fishbowler:
What I haven't tested is whether we can back out that Orchestra change and my test still passes. I'm not certain how often that bit is executed and what the cost is

Fishbowler:
I've rebased the changes, and taken a look at the Orchestra change and why it was necessary. I decided in this case it was better to have one fewer integration tests and save Orchestra the extra computing in non-test scenarios.

Fishbowler:
I've fixed one more thing I mucked up in my rebase, and added the test that @herval asked for.

Fishbowler:
Thanks @jerriais!

jerriais:
Thanks for the extra work to bring it across the finish line @Fishbowler !

Fishbowler:
Absolutely no problem. I think most of my effort was due me messing up a rebase 😂

PR #2638, opened 2025-08-14
fix: web flow check & mix sharding

proksh:
Fixed issues with web flow test related to this [PR](https://github.com/mobile-dev-inc/Maestro/commit/be97bc657a671577de110c8ee77cd4ad71989f1a) which cause some issues with `test` and `sharding`.

**Issue 1:**
No web flow in config, yet it stills open the web flow. Reason was we were checking all the files and if any of them included web flow we were adding that device. 

https://github.com/user-attachments/assets/c1a2a36a-2626-44de-99ec-4cc63a29414d

**Fix: Updated the function to check files that need to be executed.**

**Issue 2:**
Opened android and tried to do sharding between 2 devices. Workspace has both android and web flows. But it errors out saying only one device present. 

https://github.com/user-attachments/assets/ced42a6d-d690-4eba-8af9-a9bc9f7c7657

**Fix: In case web flow is present, add chromium to the device Id list** (we were just replacing the list with chromium)

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -204,6 +204,10 @@ class TestCommand : Callable<Int> {
         return plan.flowsToRun.any { it.toFile().isWebFlow() } || 
                plan.sequence.flows.any { it.toFile().isWebFlow() }
     }
+
+    private fun allFlowsAreWebFlow(plan: ExecutionPlan): Boolean {
+      return plan.flowsToRun.all { it.toFile().isWebFlow() }
+    }
   
     override fun call(): Int {
         TestDebugReporter.install(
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -200,11 +200,11 @@ class TestCommand : Callable<Int> {
     private val usedPorts = ConcurrentHashMap<Int, Boolean>()
     private val logger = LoggerFactory.getLogger(TestCommand::class.java)
 
-    private fun includesWebFlow(): Boolean {
-        return flowFiles.any { it.isWebFlow() }
+    private fun executionPlanIncludesWebFlow(plan: ExecutionPlan): Boolean {
```
proksh, line 203: Instead of checking entire workspace, we will now check only the files that are part of Execution Plan

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -487,17 +487,21 @@ class TestCommand : Callable<Int> {
...
-        val deviceIds = arguments
+    private fun getPassedOptionsDeviceIds(plan: ExecutionPlan): List<String> {
+        val userDeviceIds = parent?.deviceId
             .orEmpty()
             .split(",")
             .map { it.trim() }
             .filter { it.isNotBlank() }
-        return deviceIds
+        
+        return if (executionPlanIncludesWebFlow(plan)) {
```
proksh: In case of web flow:
Instead of just returning "chromium", we will add "chromium" to the previous list of device ids

amanjeetsingh150 (approved):

PR #2643, opened 2025-08-18
feat: Add passingPercentage to JUnit report

mdedetrich:
## Proposed changes

This PR adds a `passingPercentage` to the XML JUnit report. Currently the logic is that only tests that have `FlowStatus.SUCCESS` count as passed although it can be argued that `FlowStatus.WARNING` should also be included

## Testing

Tests have been updated to reflect changes

## Other considerations

It could be argued that to be consistent with the other XML attributes we should have `passed` as an int rather than `passingPercentage` and then its easy to calculate the percentage from that. Similarly it can also be argued that since we are already calculating `passingPercentage` we should add a `passed` int attribute as well.


mdedetrich:
Updated PR to also include warning as passed

mdedetrich:
Closing PR as not needed

PR #2647, opened 2025-08-18
[MA-3571] Cloud Upload Resolve Dependencies for Single Flow Upload

trindadetiago:
## Proposed Changes

**Enhance Cloud Upload with Automatic Dependency Discovery**

When uploading a single flow file to Maestro Cloud, the system now automatically discovers and includes all dependent files (subflows, scripts, and media) in the workspace ZIP. This eliminates the need for users to upload the workspace in order to run one flow that needs other files.

- **New `DependencyResolver` class**: Automatically discovers dependencies by parsing flow files for `runFlow`, `runScript`, and `addMedia` commands
- **Enhanced `WorkspaceUtils`**: Automatically detects single file uploads and includes dependencies, while maintaining existing directory upload behavior

### **User Experience Impact**
- **Before**: Users had to upload workspace in order to run one specific file with dependencies
- **After**: Users can upload a single flow file and all dependencies are automatically included

### **Example**
Uploading `android-advanced-flow.yaml` now works seamlessly, whereas before it would fail due to missing subflows, or scripts

## Testing
- Created comprehensive `DependencyResolverTest.kt` covering all dependency types
- Tests verify nested subflows, script dependencies, and media file discovery
- Tests include scenarios where subflows reference other subflows (nested dependencies)

## Maestro Studio

Changes will have the same effect there :) 

---

Error we get:

image

With this PR:

image

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/util/DependencyResolver.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/DependencyResolver.kt
@@ -1,0 +1,116 @@
+package maestro.cli.util
+
+import maestro.orchestra.yaml.YamlCommandReader
+import maestro.orchestra.yaml.YamlFluentCommand
+import maestro.orchestra.yaml.MaestroFlowParser
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.exists
+
+object DependencyResolver {
+
+    fun discoverAllDependencies(flowFile: Path): List<Path> {
+        val discoveredFiles = mutableSetOf<Path>()
+        val filesToProcess = mutableListOf(flowFile)
+        
+        while (filesToProcess.isNotEmpty()) {
+            val currentFile = filesToProcess.removeFirst()
+            
+            // Skip if we've already processed this file (prevents circular references)
+            if (discoveredFiles.contains(currentFile)) continue
+            
+            // Add current file to discovered set
+            discoveredFiles.add(currentFile)
+            
+            try {
+                // Only process YAML files for dependency discovery
+                if (!currentFile.fileName.toString().endsWith(".yaml") && !currentFile.fileName.toString().endsWith(".yml")) {
+                    continue
+                }
+                
+                val flowContent = Files.readString(currentFile)
+                val commands = MaestroFlowParser.parseFlow(currentFile, flowContent)
+                
+                // Discover dependencies from each command
+                val dependencies = commands.flatMap { maestroCommand ->
+                    val commandDependencies = mutableListOf<Path>()
+
+                    // Check for runFlow commands
+                    maestroCommand.runFlowCommand?.let { runFlow ->
+                        val flowPath = resolvePath(currentFile, runFlow.sourceDescription ?: "")
+                        if (flowPath.exists()) {
+                            commandDependencies.add(flowPath)
+                        }
+                    }
+                    
+                    // Check for runScript commands
+                    maestroCommand.runScriptCommand?.let { runScript ->
+                        val scriptPath = resolvePath(currentFile, runScript.sourceDescription ?: "")
+                        if (scriptPath.exists()) {
+                            commandDependencies.add(scriptPath)
+                        }
+                    }
+                    
+                    // Check for addMedia commands
+                    maestroCommand.addMediaCommand?.let { addMedia ->
+                        addMedia.mediaPaths.forEach { mediaPath ->
+                            val mediaFile = currentFile.fileSystem.getPath(mediaPath)
+                            if (mediaFile.exists()) {
+                                commandDependencies.add(mediaFile)
+                            }
+                        }
+                    }
+                    
+                    commandDependencies
+                }
+                
+                val newDependencies = dependencies.filter { it.exists() && !discoveredFiles.contains(it) }
+                filesToProcess.addAll(newDependencies)
+                
+            } catch (e: Exception) {
+                println("Warning: Could not parse dependencies for ${currentFile.fileName}: ${e.message}")
+            }
+        }
+        
+        return discoveredFiles.toList()
+    }
+    
+    private fun resolvePath(flowPath: Path, requestedPath: String): Path {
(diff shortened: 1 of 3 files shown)


Leland-Takamine:
Does this work when the initial file references a sub file that is in a parent directory (eg. `../subflow.yaml`)?

trindadetiago:
> Does this work when the initial file references a sub file that is in a parent directory (eg. `../subflow.yaml`)?

Yes! 

image

Fishbowler:
I might be wrong, but I _think_ your response doesn't quite answer Leland's question. Your example contains siblings and children in the directory structure.

Given a directory structure like this:
```
/
├── utility.yaml
└── tests
    └── onboarding
        └── finish-onboarding.yaml
```
If finish-onboarding.yaml is the test that's called, and it contains
```
- runFlow: ../../utility.yaml
```
then will the dependency resolver cope with that?

Leland-Takamine:
> I might be wrong, but I _think_ your response doesn't quite answer Leland's question. Your example contains siblings and children in the directory structure.
> 
> Given a directory structure like this:
> 
> ```
> /
> ├── utility.yaml
> └── tests
>     └── onboarding
>         └── finish-onboarding.yaml
> ```
> 
> If finish-onboarding.yaml is the test that's called, and it contains
> 
> ```
> - runFlow: ../../utility.yaml
> ```
> 
> then will the dependency resolver cope with that?

Yeah - this is exactly the situation I was wondering about ^

trindadetiago:
> I might be wrong, but I _think_ your response doesn't quite answer Leland's question. Your example contains siblings and children in the directory structure.
> 
> Given a directory structure like this:
> 
> ```
> /
> ├── utility.yaml
> └── tests
>     └── onboarding
>         └── finish-onboarding.yaml
> ```
> 
> If finish-onboarding.yaml is the test that's called, and it contains
> 
> ```
> - runFlow: ../../utility.yaml
> ```
> 
> then will the dependency resolver cope with that?

Hey. Sorry about the answer, it was kinda confusing. But the answer for that is Yes.

Let me explain a new example from scratch. I have this folder structure:

```
/Users/tiagotrindade
├─ Documents
│  └─ samples
│     ├─ android-advanced-flow.yaml
│     ├─ scripts
│     │  └─ getSearchQuery.js
└─ subflows
   └─ onboarding-android.yaml
```

The subflows are on the parent of the parent.

android-advanced-flow.yaml looks like this:

```
appId: org.wikipedia
tags:
  - android
  - passing
  - advanced
---
- runFlow: ../../subflows/onboarding-android.yaml
- tapOn:
    id: "org.wikipedia:id/search_container"
- tapOn:
    text: "Non existent view"
    optional: true
- runScript: scripts/getSearchQuery.js
- inputText: ${output.result}
- assertVisible: ${output.result}
- runFlow: ../../subflows/launch-clearstate-android.yaml
```

It runs fine: 
https://app.maestro.dev/project/proj_01jr9he0xyfwtbjt5hz7r6m5y6/maestro-test/app/app_01jzqaev4je9p875hkfm7h610f/upload/mupload_01k31gbjave3pr8jn458zs0v9k?sort=name


Other example more similar to yours:

```
samples
├─ main-folder
│  └─ main-again-to-be-sure
│     └─ **android-advanced-flow.yaml**
├─ subflows
│  ├─ onboarding-android.yaml
│  └─ launch-clearstate-android.yaml
└─ scripts
   └─ getSearchQuery.js
```

android-advanced-flow.yaml:

```
appId: org.wikipedia
tags:
  - android
  - passing
  - advanced
---
- runFlow: ../../subflows/onboarding-android.yaml
- tapOn:
    id: "org.wikipedia:id/search_container"
- tapOn:
    text: "Non existent view"
    optional: true
- runScript: ../../scripts/getSearchQuery.js
- inputText: ${output.result}
- assertVisible: ${output.result}
- runFlow: ../../subflows/launch-clearstate-android.yaml
```

Also runs fine:
https://app.maestro.dev/project/proj_01jr9he0xyfwtbjt5hz7r6m5y6/maestro-test/app/app_01jzqaev4je9p875hkfm7h610f/upload/mupload_01k31gpanjegxabvabnewy9tc1

The thing is that the workspace saved doesn't keep that nested structure. If you open this example above and view workspace, you will see this:

image

trindadetiago:
Is that what you guys are talking about or am I not understanding?

Leland-Takamine:
Yep that's the situation @trindadetiago thanks for confirming!

> The thing is that the workspace saved doesn't keep that nested structure

Are there any issues with this? One thing I can think of is that the workspace isn't valid when downloaded since the relative path will no longer will resolve correctly.

Edit: Surprised this this works on cloud actually since the relative paths don't match the saved workspace

Leland-Takamine:
^ Not a blocker btw as long as it runs correctly on cloud

Fishbowler:
I agree - we could solve for the workspace bit cloud-side if it ever becomes a customer problem. I don't think many folks outside of staff use that.

Fishbowler (approved):

trindadetiago:
The only problem is indeed downloading the workspace in Cloud, which doesn't return the original structure. But there isn't any edge case for where the upload itself would fail - at least for what I've tested and think of.

maestro-cli/src/main/java/maestro/cli/util/DependencyResolver.kt
```diff
@@ -0,0 +1,122 @@
...
+            
+            // Skip if we've already processed this file (prevents circular references)
+            if (discoveredFiles.contains(currentFile)) continue
+            
+            // Add current file to discovered set
+            discoveredFiles.add(currentFile)
+            
+            try {
+                // Only process YAML files for dependency discovery
+                if (!isYamlFile(currentFile)) {
```
herval, line 25: what if one of the files is using js?

trindadetiago: We ignore that file. For now, if a JS file references an external file, it won't be included in the upload

PR #2635, opened 2025-08-13
feat: Detect minimum JVM version

mdedetrich:
## Proposed changes

This PR modifies both the bash (for unix* systems) and bat (for Windows) so that it detects the minimum version (currently 17) of java before executing the main maestro-cli java process.

## Testing

The easiest way to test this is firstly clean the maestro build

```sh
./gradlew clean
```

Then lets create a package using jreleaser (make sure you are running java version 17 for this step)

```sh
export JRELEASER_GITHUB_TOKEN=not used
./gradlew :maestro-cli:jreleaserPackage --no-daemon --no-parallel
```

This will create the distributions at `maestro-cli/build/distributions` so lets unzip them and try out the start script

```sh
<@Maestro>cd maestro-cli/build/distributions/
<@Maestro/m/b/distributions>unzip maestro.zip 
Archive:  maestro.zip
   creating: maestro
   creating: maestro/lib
  inflating: maestro/lib/maestro-cli-1.41.0.jar  
  inflating: maestro/lib/server.jar  
  inflating: maestro/lib/maestro-orchestra.jar  
  inflating: maestro/lib/maestro-orchestra-models.jar  
  inflating: maestro/lib/maestro-client.jar  
  inflating: maestro/lib/maestro-ios.jar  
  inflating: maestro/lib/maestro-ios-driver.jar  
  inflating: maestro/lib/maestro-utils.jar  
  inflating: maestro/lib/maestro-ai.jar  
  inflating: maestro/lib/grpc-kotlin-stub-1.4.1.jar  
  inflating: maestro/lib/protobuf-kotlin-3.21.9.jar  
  inflating: maestro/lib/maestro-web.jar  
  inflating: maestro/lib/dadb-1.2.10.jar  
  inflating: maestro/lib/logging-interceptor-4.12.0.jar  
  inflating: maestro/lib/okhttp-4.12.0.jar  
  inflating: maestro/lib/kotlin-sdk-jvm-0.5.0-maestro1.jar  
  inflating: maestro/lib/ktor-client-cio-jvm-2.3.6.jar  
  inflating: maestro/lib/ktor-client-content-negotiation-jvm-2.3.6.jar  
  inflating: maestro/lib/ktor-client-core-jvm-2.3.6.jar  
  inflating: maestro/lib/ktor-server-cors-jvm-2.3.6.jar  
  inflating: maestro/lib/ktor-server-status-pages-jvm-2.3.6.jar  
[...]

PR #2648, opened 2025-08-19
Fix Java version parsing on Windows

hobindar:
## Proposed changes

Modify Java version check so that it passes on Windows

## Testing

Still builds. Ran modified batch script _outside_ independently of Maestro

## Issues fixed

Java version check consistently failing on Windows

### DIFF
--- a/maestro-cli/build.gradle.kts
+++ b/maestro-cli/build.gradle.kts
@@ -49,7 +49,7 @@ fun windowsMinimumJavaText(minimumJavaVersion: String): String = """
 rem Parses x out of 1.x; for example 8 out of java version 1.8.0_xx
 rem Otherwise, parses the major version; 9 out of java version 9-ea
 set JAVA_VERSION=0
-for /f "tokens=3" %%g in ('%JAVA_EXE% -Xms32M -Xmx32M -version 2^>^&1 ^| findstr /i "version"') do (
+for /f "tokens=3" %%g in ('"%JAVA_EXE%" -Xms32M -Xmx32M -version 2^>^&1 ^| findstr /i version') do (
   set JAVA_VERSION=%%g
 )
 set JAVA_VERSION=%JAVA_VERSION:"=%


Fishbowler:
Thanks! This will fix an instance of this problem, but we're seeing some feedback that the check might be fruitless - the output of `java -version` has quotes or it doesn't, and appears as the 2nd or 3rd token, all depending on the distribution (Oracle vs Adoptium vs Zulu vs Coretto, etc)

We're considering backing the check out completely. WDYT?

hobindar:
I believe the feedback you've received is from a colleague of mine (Dhrumil). In his case, the issue actually isn't with Adoptium's output. The current batch script is able to parse it just fine.

The problem was that the script cannot handle the space character in "Program Files." This should cause the check to fail on the vast majority of Windows systems for all JDKs, as all JDKs are typically installed inside "Program Files" on Windows.

Just an aside, the only reason the Stackoverflow answer (which the original check is based on) works at all, is because they're missing a % in the JAVA_HOME check which causes the script to always use the path instead of JAVA_HOME. Essentially, a bug was masking another bug, and when this was copied into Maestro, the first bug was fixed revealing the latter.

That said, if you think other JDKs could be problematic, this check could be removed as well. I'll leave that decision up to all of you. :)

mdedetrich:
Thanks for the PR! For now we have decided to revert the minimum version Java check (see https://github.com/mobile-dev-inc/Maestro/pull/2649).

We will work on a better solution, currently our plan is to use `java -XshowSettings:properties -version` to get the version from `java.version` Java system property as this looks like it produce a completely consistent versioning scheme amongst all platforms.

mdedetrich:
> The problem was that the script cannot handle the space character in "Program Files." This should cause the check to fail on the vast majority of Windows systems for all JDKs, as all JDKs are typically installed inside "Program Files" on Windows.

This version of the script worked fine on the Windows system where I tested which had this path

IMG_1513

You can see evidence of it working in the original PR at https://github.com/mobile-dev-inc/Maestro/pull/2635#issuecomment-3188293328

So unless I misunderstand something I don't think this is the cause (alone)

However the fix in this PR might work for all JDK's on Windows in combination with other fixes, but this is something we have to properly test (evidently).

Fishbowler:
Indeed. Dhrumils `java -version` output is

 ```
$ java --version
 openjdk 21.0.8 2025-07-15 LTS
 OpenJDK Runtime Environment Temurin-21.0.8+9 (build 21.0.8+9-LTS)
 OpenJDK 64-Bit Server VM Temurin-21.0.8+9 (build 21.0.8+9-LTS, mixed mode, sharing)
```

The existing check would've taken the third token, the date, as the java version to evaluate.

hobindar:
> Thanks for the PR! For now we have decided to revert the minimum version Java check (see #2649).
> 
> We will work on a better solution, currently our plan is to use `java -XshowSettings:properties -version` to get the version from `java.version` Java system property as this looks like it produce a completely consistent versioning scheme amongst all platforms.

Sounds good. I'll go ahead and close this PR then. :)

PR #2659, opened 2025-08-21
Dependency Resolver to include Retry, Repeat, onFlowStart, onFlowComplete and subcommands

trindadetiago:
## Proposed changes

- Included Retry and Repeat commands in the dependency extraction
- Included subcommands from composite commands in the dependency extraction
- Included commands from the `onFlowComplete` and `onFlowStart` hooks in the dependency extraction 

- Added sourceDescription for the `retry` command: now we save the external file it references, if it does (similar to how runFlow works)

## Testing

Created more test cases to include all possibilities from the mentioned above.

### DIFF
--- a/maestro-cli/src/test/kotlin/maestro/cli/util/DependencyResolverTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/util/DependencyResolverTest.kt
@@ -561,4 +561,37 @@ class DependencyResolverTest {
         assertThat(dependencies).contains(retryCleanup)    // From onFlowComplete -> retry -> runScript  
         assertThat(dependencies).contains(mainSubflow)     // From main flow -> runFlow
     }
+
+    @Test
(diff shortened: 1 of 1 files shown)


Fishbowler:
Found one more edge case? [Multipart Uploads](https://docs.maestro.dev/advanced/javascript/make-http-s-requests#post-request)

Fishbowler:
Oooh, one more - the mediaPaths for AddMedia is evaluated through `evaluateScripts` at runtime.

Fishbowler:
There's a chance we can't catch em all, and we accept these two edge cases as accepted. Maybe there's more that we can do, but we could split those into new issues instead?

trindadetiago:
> There's a chance we can't catch em all, and we accept these two edge cases as accepted. Maybe there's more that we can do, but we could split those into new issues instead?

I think is totally valid we don't cover ALL cases, specially these very specific ones

Fishbowler (approved):

PR #2665, opened 2025-08-22
Legacy Studio maintenance

Fishbowler:
## Proposed changes

- Removes Storybook, which was almost never used anyway
- Bumps to Node v20, which matches the build environment
- Updates caniuse, just to stop the build-time nag. No idea if this actually helps in production though.
- Runs a big npm upgrade to update a bunch of dependencies

## Testing

Ran `./maestro studio` after each commit - still works on my machine 🤷 

## Issues fixed

A bunch of security vulns


maestro-studio/web/.nvmrc
```diff
@@ -1 +1 @@
-v16.14.0
+v20.19.4
```
amanjeetsingh150, line 1: Umm, I’m not fully sure about updating all these npm packages and the Node version. My main concern is what happens if someone runs into issues on the web studio. I don’t think we should be solving those, which is why I feel we should just leave it as is.

Fishbowler: That bump is only for local developers. GitHub Actions already builds with whatever is local.

amanjeetsingh150: I don't think I understood. Wouldn't this update reflect in `maestro studio` command for users?

Fishbowler: No node required by users. They get the built react webapp.

Fishbowler: And they're already getting an app built in Node v20 (or maybe newer - unsure what's on a GitHub runner currently)

proksh (approved):

PR #2667, opened 2025-08-25
Fixing iOS connection issues during workspace execution

amanjeetsingh150:
## Proposed changes

During execution of the workspace deviceInfo endpoint, calls fail when the view hierarchy is very large:

```
let screenSize = app.firstMatch.frame.size
```

The firstMatch.frame forces XCTest to resolve the full hierarchy and blows up at depth. We already call `app.snapshot()` in this function, so instead of querying firstMatch, we can take the frame directly from the snapshot’s application object which doesn't break XCTest framework.

This PR updates the logic to use the snapshot object for frame size, avoiding hierarchy resolution errors.

## Testing

- [x] Locally


Fishbowler (approved):

Fishbowler:
This solves the issue I was having locally. I've got from being able to run 1-3 consecutive tests to >90.

proksh (approved):

PR #2671, opened 2025-08-26
feat: Add minimum JVM version check

mdedetrich:
## Proposed changes

This PR adds a minimum JVM version check, it is a reimplementation of https://github.com/mobile-dev-inc/Maestro/pull/2635 however instead of trying to parse the output of the `java -version` command we instead run java against a `JvmVersion.class` file which is guaranteed to work on all JDK's from 1.1 (this `JvmVersion.class` was compiled with JDK 1.1).

This is what the source input of `JvmVersion.class` is

```java
import java.util.StringTokenizer;

class JvmVersion {
    public static void main(String[] args) {
        try {
            String javaVersion = System.getProperty("java.version");
            StringTokenizer tokenizer = new StringTokenizer(javaVersion, ".");
            String[] split = new String[tokenizer.countTokens()];
            int count = 0;
            while (tokenizer.hasMoreTokens()) {
                split[count] = tokenizer.nextToken();
                count++;
            }
            if (javaVersion.startsWith("1.")) {
                String version = split[1];
                if (Integer.parseInt(version) >= 1 && Integer.parseInt(version) <= 8) {
                    System.out.println(version);
                    System.exit(0);
                } else {
                    String base = "Expected a JVM version of 1.0 through to 1.8 for legacy JVM versioning. Instead got ";
                    String output = base.concat(version);
                    System.out.println(output);
                    System.exit(1);
                }
            } else {
                String version = split[0];
                if (Integer.parseInt(version) >= 9) {
                    System.out.println(version);
                    System.exit(0);
                } else {
                    String base = "Expected a JVM version of 9 or greater for new JVM versioning. Instead got ";
                    String output = base.concat(version);
                    System.out.println(output);
                    System.exit(1);
                }
            }
        } catch (Exception e) {
            System.err.println(e.getMessage());
            System.exit(1);
        }
    }
}
```

Since the output of `System.getProperty("java.version")` is guaranteed to be consistent for every JDK build (i.e. oracle vs eclipse vs amazon coretto etc etc) and `java -classpath <CLASSPATH> <FILE>` works on every JVM/java since 1.1 JDK, this check should work in any conceivable scenario. Furthermore unlike the previous implementation, all of the business logic for calculating the version is contained within `JvmVersion.class` which means that the necessary POSIX shell/Windows bat modifications are so trivial that it will work with every conceivable POSIX/Windows system (at least since 1997)

You can use a decompiler on `JvmVersion.class` to confirm the JVM bytecode contained within it, i.e. the output of mine is

image

Notice how it says bytecode version 45.3 (Java 1.1)

The `JvmVersion.class` file is manually copied into the `bin` folder of `startupScripts`. This decision is deliberate, as we don't want this file to be accidentally packaged as a resource where it could be included in the main applications classpath and somehow interact with the main application. In other words, this setup means that `JvmVersion.class` is completely isolated from the rest of the app, it only gets manually included right at the start for the JVM check and isn't part of the main app.

This PR was manually tested on both Windows and Linux (see screenshots). Unlike the previous of implementation of this, testing between different JDK builds of the same JDK version is not necessary.

Screenshot 2025-08-26 153806
Screenshot 2025-08-26 153814

```
<@Maestro/m/b/distributions>-<⎇ main>-<±>-<*>-> Maestro/bin/maestro --version
2.0.0
<@Maestro/m/b/distributions>-<⎇ main>-<±>-<*>-> jenv global 1.8
<@Maestro/m/b/distributions>-<⎇ main>-<±>-<*>-> Maestro/bin/maestro --version

ERROR: Java 17 or higher is required.

Please update Java, then try again.
To check your Java version, run: java -version

See https://maestro.dev/blog/what-s-new-inmaestro-2-0-0 for more details.

<@Maestro/m/b/distributions>-<⎇ add-minimum-jvm-version-check>-<±>-<*>-1->
```

### DIFF
--- a/maestro-cli/build.gradle.kts
+++ b/maestro-cli/build.gradle.kts
@@ -68,7 +68,7 @@ if [ "${'$'}JAVA_VERSION" -lt $minimumJavaVersion ]; then
 Please update Java, then try again.
 To check your Java version, run: java -version
 
-See https://maestro.dev/blog/what-s-new-inmaestro-2-0-0 for more details."
+See https://maestro.dev/blog/introducing-maestro-2-0-0 for more details."
 fi
 """.trimIndent()
 


maestro-cli/build.gradle.kts
```diff
@@ -42,8 +42,60 @@ tasks.named<JavaExec>("run") {
...
+)
+
+if %JAVA_VERSION% LSS $minimumJavaVersion (
+  echo.
+  echo ERROR: Java $minimumJavaVersion or higher is required.
+  echo.
+  echo Please update Java, then try again.
+  echo To check your Java version, run: java -version
+  echo.
+  echo See https://maestro.dev/blog/what-s-new-inmaestro-2-0-0 for more details.
```
proksh:
```suggestion
  echo See https://maestro.dev/blog/introducing-maestro-2-0-0 for more details.
```

maestro-cli/build.gradle.kts
```diff
@@ -42,8 +42,60 @@ tasks.named<JavaExec>("run") {
...
+
+fun unixMinimumJavaText(minimumJavaVersion: String): String = """
+JAVA_VERSION=$( "${'$'}JAVACMD" -classpath "${'$'}APP_HOME"/bin/ JvmVersion )
+if [ "${'$'}JAVA_VERSION" -lt $minimumJavaVersion ]; then
+  die "ERROR: Java $minimumJavaVersion or higher is required.
+
+Please update Java, then try again.
+To check your Java version, run: java -version
+
+See https://maestro.dev/blog/what-s-new-inmaestro-2-0-0 for more details."
```
proksh:
```suggestion
See https://maestro.dev/blog/introducing-maestro-2-0-0 for more details."
```

Issue #2677, opened 2025-08-29
In Maestro 2.0.0 did not find the Java path


pravin-debug:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

while running the command ->maestro --version

I found the error 
In the Windows system ->Suppose that your Java installation Directory is the default Java path, i.e "C:/Program Files/Java/JDK-17/bin"

Then the  Meastro could not recognise this path, causing the error as "'C:\Program' is not recognised as an internal or external command, operable program or batch file. ERROR: Java 17 or higher is required."

### Actual results

I found the error 
In the Windows system ->Suppose that your Java installation Directory is the default Java path, i.e "C:/Program Files/Java/JDK-17/bin"

Then the  Meastro could not recognise this path, causing the error as "'C:\Program' is not recognised as an internal or external command, operable program or batch file. ERROR: Java 17 or higher is required."

Solution :
I moved the JDK path to the new Path like -> "C:/JAVA/JDK-17"

and run the maestro successfully

### Expected results

maestro should find the Java installation default path

### About app

I'm just exploring maestro 2.0.0

### About environment

Using Windows System

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

2.0.0

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

mdedetrich:
@pravin-debug Which JDK did you have installed on your Windows machine? Also what version of Windows?

pravin-debug:
Initially, I have 2 JDKs installed in My System, i.e JDK 17 and JDK 24
Windows 11
After that, I moved the JDK to the newly created directory and updated the path in the Windows environment variables, and it works for me

mdedetrich:
> Initially, I have 2 JDKs installed in My System, i.e JDK 17 and JDK 24 Windows 11 After that, I moved the JDK to the newly created directory and updated the path in the Windows environment variables, and it works for me

Which distribution of JDK, Oracle or Eclipse or another one?

pravin-debug:
> > Initially, I have 2 JDKs installed in My System, i.e JDK 17 and JDK 24 Windows 11 After that, I moved the JDK to the newly created directory and updated the path in the Windows environment variables, and it works for me
> 
> Which distribution of JDK, Oracle or Eclipse or another one?

Using Oracle JDK

Issue #2673, opened 2025-08-28
[2.0.0] Maestro cannot find java due to being in a path with spaces in Windows


oriol-serrabassa:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Be on a Windows machine:

1. Have a Java SDK installed on a path with spaces like (C:\Program Files\jdk-17.0.2\bin)
2. Make your java commands available via system environment variables
3. Run any maestro command

<img width="876" height="475" alt="Image" src="https://github.com/user-attachments/assets/17441ee9-6d54-45c1-aa23-fdf5a7e86a11" />

### Actual results

Any maestro command fails because the program cannot find the Java SDK executable, which is actually installed on a path with spaces.

### Expected results

Maestro should be able to properly parse an installation path with spaces.
There was no issues about this problem in version 1.x.

### About app

This is not a testing issue.

### About environment

OS
Microsoft Windows 11 Home
10.0.26100 Build 26100

JAVA
openjdk 17.0.2 2022-01-18
OpenJDK Runtime Environment (build 17.0.2+8-86)
OpenJDK 64-Bit Server VM (build 17.0.2+8-86, mixed mode, sharing)

PROCESSOR
x64-based processor

### Maestro version

2.0.0

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

PR #2678, opened 2025-08-29
fix: Fix windows quoting issue and use jar

mdedetrich:
## Proposed changes

Aside from fixing spaces in paths issue with Windows, it turns out that the java executable on windows cannot handle setting the classpath if the path has spaces in it if you are using `.class` files directly.

```
C:\Users\mdedetrich\Desktop\maestro stuff\bin>"java.exe" -classpath "C:\Users\mdedetrich\Desktop\maestro stuff\bin\JvmVersion.class" JvmVersion
Error: Could not find or load main class JvmVersion

C:\Users\mdedetrich\Desktop\maestro stuff\bin>"java.exe" -classpath C:\Users\mdedetrich\Desktop\maestro stuff\bin\ JvmVersion
Error: Could not find or load main class stuff\bin\

C:\Users\mdedetrich\Desktop\maestro stuff\bin>"java.exe" -classpath "C:\Users\mdedetrich\Desktop\maestro stuff\bin\*" JvmVersion
8
```

With java the only way to use `.class` files with the `-classpath` flag is to not specify `/*` at the end (the `/*` only picks up `jar` files) however, and only on Windows, putting double quotes (so you can handle spaces) on a classpath dir without the `/*` at the end doesn't work.

So in short, in order to properly solve this problem I had to make a jar out of `JvmVersion.class` along with a `MANIFEST` that points to Main so that I can double quote the classpath path.

## Testing

Tested on my Windows machine

## Issues fixed

Resolves: https://github.com/mobile-dev-inc/Maestro/issues/2677
Resolves: https://github.com/mobile-dev-inc/Maestro/issues/2673


herval:
Where is the source for that .class?

mdedetrich:
> Where is the source for that .class?

It was in the description of the previous PR, should I include it as a comment in the gradle build file

Fishbowler (changes requested):
I can reproduce the problem on my Windows laptop, and this doesn't resolve it for me.

I've got a Java path with spaces, and a classpath without.

Local experimentation shows that quoting Java but _not_ quoting classpath solves it (jar vs class seems unrelated for my situation).

Bleugh. Windows is hard.

Fishbowler:
Also fixes #2673?

PR #2681, opened 2025-08-29
login even if user already logged in

herval:
totally unnecessary check.


Fishbowler (approved):
Looks sensible, good intent. Not tested it though (although unsure how many possible states this can need)

PR #2684, opened 2025-09-01
fix: added back runs to cli cloud upload output

proksh:
### Problem

All the Runs output was removed from the cli cloud upload after this [Pull Request](https://github.com/mobile-dev-inc/Maestro/pull/2652)

### Fix

Added back the Runs output logs and used terminalStatus array to detect the end state of run

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -255,7 +255,7 @@ class CloudInteractor(
 
         var pollingInterval = minPollIntervalMs
         var retryCounter = 0
-        val printedFlow = mutableSetOf<UploadStatus.FlowResult>()
+        val printedFlows = mutableSetOf<UploadStatus.FlowResult>()
 
         do {
             val upload = try {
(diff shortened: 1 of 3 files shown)


maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -275,6 +279,16 @@ class CloudInteractor(
...
             }
 
+            for (uploadFlowResult in upload.flows) {
+                if(printedFlows.contains(uploadFlowResult)) { continue }
+                if(!terminalStatuses.contains(uploadFlowResult.status)) { continue }
+
+                printedFlows.add(uploadFlowResult);
+                TestSuiteStatusView.showFlowCompletion(
+                  uploadFlowResult.toViewModel()
+                )
```
amanjeetsingh150, line 289: Can we add a test for cloud interactor that checks its output especially for local.

amanjeetsingh150: You may need to tweak the structure a bit to inject a small class responsible for displaying output/results, and then assert against that during any function call. This will decouple side effects and make the behavior testable.

maestro-cli/src/main/java/maestro/cli/view/TestSuiteStatusView.kt
```diff
@@ -168,16 +169,13 @@ object TestSuiteStatusView {
                 }
             )
 
-            fun UploadStatus.FlowResult.toViewModel(
-                duration: Duration? = null
-            ) = FlowResult(
+            fun UploadStatus.FlowResult.toViewModel() = FlowResult(
```
mdedetrich, line 172: Nice pickup!

mdedetrich (approved):
For me its lgtm but I agree with @amanjeetsingh150 that it would be good to add tests. I accidentally removed this functionality in my PR but due to lack of tests I had no idea I created a regression in functionality

Issue #2687, opened 2025-09-02
HTML report summary wrong when sharding


Fl0p:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Just run

```maestro test --output=output/report.html --format=HTML  --shard-split=N flows/```

### Actual results

Just look at screenshots

<img width="1143" height="336" alt="Image" src="https://github.com/user-attachments/assets/880c4a01-4ffd-4886-a40a-210761562d8b" />

<img width="1034" height="325" alt="Image" src="https://github.com/user-attachments/assets/dc583b96-5609-4b80-ab79-08c93d5f4046" />

Numbers are weird.

### Expected results

Expect correct numbers
Totals flows was 20. And i suppose 10 was failed and 10 was sucessed

### About app

masetro 1.41.0

### About environment

Mac os 15.6 

### Logs

<details>
<summary>Logs</summary>

```

[maestro.log](https://github.com/user-attachments/files/22103645/maestro.log)

```

</details>

### Maestro version

1.41.0

### How did you install Maestro?

Homebrew

Fishbowler:
Can you edit to provide some logs?
Does this report work when you're not sharding?

cc @mdedetrich

Fl0p:
With 1 shard it is okay
<img width="1038" height="325" alt="Image" src="https://github.com/user-attachments/assets/cc985e1e-dd27-4542-9278-943e0542c7d3" />

Fl0p:
attached fresh logs

only report is wrong

terminal output is okay

<img width="727" height="141" alt="Image" src="https://github.com/user-attachments/assets/ef0fc2ec-3344-4ba8-9af4-8fae975b16c6" />

Fishbowler:
Can you give this a whirl in v2.0? There were a few reporting changes there. I'm not expecting this to be fixed, but any change we make is gonna happen there, so getting a grip on what that change needs to be would be useful.

mdedetrich:
Wanted to mention that I have found the root cause of the issue and its solved in this PR https://github.com/mobile-dev-inc/Maestro/pull/2691 . When that gets merged and a new cli release gets made you shouldn't experience this issue

PR #2691, opened 2025-09-03
fix: Failed test reporting in HtmlTestSuiteReporter

mdedetrich:
## Proposed changes

This PR fixes incorrect reporting of failures in `HtmlTestSuiteReporter`. The error appeared to be a simple one, due to the logic of calculating failures there was double counting of test failures. The logic of `getFailedTests` was simplified to just use a basic filter.

`failures` was also moved to a common function so that we don't get divergence in what constitutes a failure. Note that there are some other data structures which also represent test reports but that should be fixed in a future PR as its out of scope.

## Testing

Common testing logic was moved out of `JUnitTestSuiteReporterTest` into a `TestSuiteReporterTest` so that we could add newly added tests to `HtmlTestSuiteReporterTest` in order to verify that the failure count is working (amongst other things).

## Issues fixed

Resolves: https://github.com/mobile-dev-inc/Maestro/issues/2687#issuecomment-3246749465

### DIFF
--- a/maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
@@ -1,0 +1,166 @@
+package maestro.cli.report
+
+import com.google.common.truth.Truth.assertThat
+import okio.Buffer
+import org.junit.jupiter.api.Test
+
+class HtmlTestSuiteReporterTest : TestSuiteReporterTest() {
+
+    @Test
+    fun `HTML - Test passed`() {
+        // Given
+        val testee = HtmlTestSuiteReporter()
+        val sink = Buffer()
+
+        // When
+        testee.report(
+            summary = testSuccessWithWarning,
+            out = sink
+        )
+        val resultStr = sink.readUtf8()
+
+        // Then
+        assertThat(resultStr).isEqualTo(
+            """
+            <html>
+              <head>
+                <title>Maestro Test Report</title>
+                <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet">
+              </head>
+              <body>
+                <div class="card mb-4">
+                  <div class="card-body">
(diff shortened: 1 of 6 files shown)


maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        testee.report(
+            summary = testSuccessWithWarning,
+            out = sink
+        )
+        val resultStr = sink.readUtf8()
+
+        println(resultStr)
+
+        // Then
+        assertThat(resultStr).isEqualTo(
```
proksh, line 23: Unsure we should assert the entire html here. With each change there we might need to change every test html.

maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        val sink = Buffer()
+
+        // When
+        testee.report(
+            summary = testSuccessWithWarning,
+            out = sink
+        )
+        val resultStr = sink.readUtf8()
+
+        println(resultStr)
```
proksh: Was this for debugging?

amanjeetsingh150 (approved):

maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
mdedetrich: Yes ill remove it!

maestro-cli/src/test/kotlin/maestro/cli/report/HtmlTestSuiteReporterTest.kt
mdedetrich: That is a fair point, but trying to grab the data from the produced HTML is a bit of work (you would need to use something like tagsoup) so I did the same thing we did for JUnit tests.

Doing something more comprehensive definitely works as a future improvement

PR #2693, opened 2025-09-03
fix: rotate screenshots according to device orientation

amanjeetsingh150:
## Problem

When capturing screenshots from the iOS, the returned image orientation does not always match the device orientation. For example, if the app is locked in portrait but the simulator is rotated to landscape, the screenshot still comes back as portrait. This creates mismatches between the actual device state and the images we record from driver impacting:

1. takeScreenshot command
2. studio showing the device stream when rotated

## Testing

- [x] takeScreenshot
- [ ] CLI 
- [ ] Cloud local
- [ ] studio


.github/scripts/boot_simulator.sh
```diff
@@ -1,32 +1,37 @@
 #!/bin/bash
 
-# Specify the device type and runtime as per your requirements
-DEVICE_TYPE="iPhone 15 Pro"
-RUNTIME="iOS17.5"
+echo "Looking for the first available and installed iOS simulator..."
```
Fishbowler, line 3: Does this come with a risk of our CI changing over time as GitHub roll out new emulators?
Or is this specifically an attempt to try to roll with those changes?

amanjeetsingh150: > Does this come with a risk of our CI changing over time as GitHub roll out new emulators?

Yeah that's correct, this broke recently because it was hardcoded to 17.5 and it was not installed on macos-latest. 

This PR would change to boot the first runtime available.

Fishbowler: Then for debugability, should we print.... nm, you've already done it :)

maestro-ios-xctest-runner/maestro-driver-lib/ScreenshotPreprocessor.swift
```diff
@@ -0,0 +1,99 @@
...
+// MARK: - Output Format
+
+public enum ScreenshotPreprocessorOutput {
+    case png
+    case jpeg(_ quality: CGFloat) // 0.0 ... 1.0
+}
+
+// MARK: - Implementation
+
+public final class DefaultScreenshotPreprocessor: ScreenshotPreprocessor {
```
steviec, line 31: Is there a good reason we're making this a protocol + class? I usually try keep things as a simpler struct unless there's a good reason to add the abstraction. Do you think we'll have other /different implementations that we'll want to use?

amanjeetsingh150:
> I'm going to approve this, but I have one biggish architectural question: is there a reason why we are pulling this helper utility into its own framework? Managing frameworks brings additional annoying overhead. If this new framework will be solely consumed by the maestro-ios-xctest-runner, then couldn't we just add a new target in the existing XCode project called "Unit tests" or something, and add those to a new folder like maestro-driver-iosUnitTests? I've been down the framework management hell path and it's no fun.

The only reason was to make all the logic code inside the maestro-driver-iosUITests testable. I actually didn't tried making new target like you said. Let me try that. Curious though what type of management issues?

maestro-ios-xctest-runner/maestro-driver-lib/ScreenshotPreprocessor.swift
amanjeetsingh150: This could be a struct yeah let me change it.

steviec:
> > I'm going to approve this, but I have one biggish architectural question: is there a reason why we are pulling this helper utility into its own framework? Managing frameworks brings additional annoying overhead. If this new framework will be solely consumed by the maestro-ios-xctest-runner, then couldn't we just add a new target in the existing XCode project called "Unit tests" or something, and add those to a new folder like maestro-driver-iosUnitTests? I've been down the framework management hell path and it's no fun.
> 
> The only reason was to make all the logic code inside the maestro-driver-iosUITests testable. I actually didn't tried making new target like you said. Let me try that. Curious though what type of management issues?

Yeah, if the only goal was to make the logic code testable, pulling it out into a separate framework feels like overkill and another thing to maintain. My "management issue" comment probably stems from having to use frameworks in conjunction with Cocoapods and React Native native modules, and frameworks can be far more of a headache to deal with in those contexts.

Issue #2694, opened 2025-09-03
Env scoping doesn't work correctly for `runScript`


steviec:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

After the Maestro 2.0 rollout, users are reporting that env variables are not being scoped correctly with `runScript`. Easy repro:

flow.yaml:
```
appId: com.example.app
---
- runScript:
    file: print.js
    env: { VALUE: 123 }

- runScript:
    file: print.js
```

print.js:
```
console.log(`Value: ${VAL}`);
```

### Actual results

```
Value: 123
Value: 123
```

### Expected results

```
Value: 123
Value: null
```

The weird thing is that the behavior is correct when use `jsEngine: rhino`, which is the exact opposite of what I'd expect.

### About app

N/A

### About environment

N/A

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

2.0.0

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

PR #2695, opened 2025-09-03
Using subscope for runScript execution in Graal engine

steviec:
The graal engine implementation ignored the `runInSubScope` boolean flag. So even though it would correctly isolate env scopes set by commands, it didn't do it for `runScript`, which is a "special" command that calls `evaluateScript` directly.

I added integration and graal engine tests, they all failed, then I fixed them to confirm this is working as expected.

Fixes MAE-270 / #2694

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
@@ -76,9 +76,24 @@ class GraalJsEngine(
         sourceName: String,
         runInSubScope: Boolean,
     ): Value {
-        envBinding.putAll(env)
-        val source = Source.newBuilder("js", script, sourceName).build()
-        return createContext().eval(source)
+        if (runInSubScope) {
+            // Save current environment state
+            enterEnvScope()
+            try {
+                // Add the new env vars on top of the current scope
+                envBinding.putAll(env)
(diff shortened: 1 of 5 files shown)


maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -74,11 +74,24 @@ class GraalJsEngine(
...
+                return createContext().eval(source)
+            } finally {
+                // Restore previous environment state
+                leaveEnvScope()
+            }
+        } else {
+            // No env vars - no scoping needed
+            val source = Source.newBuilder("js", script, sourceName).build()
+            return createContext().eval(source)
+        }
```
Fishbowler: Any chance that without scoping, you could affect parent level variables here? Unsure of what the desired behaviour should be here.

```
appId: com.example
env:
  MY_VAR: 0
---
- runScript:
    file: myScript.js
    label: Increment MY_VAR
```

steviec: You are a genius Dan. I added this test and the tests failed. Thanks for saving my behind. I reverted my removal of `runInSubScope` and kept the more direct (and verbose) approach of just implementing `runInSubScope` for the graal engine.

Issue #2651, opened 2025-08-20
Assert visible selected view with text not working if descendant of matching view also has matching text


winghouchan:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

> [!NOTE]
>
> A minimal, reproducible example is still work-in-progress. See **anything else** section for more details.

#### Environment

- iOS 17.x Simulator
- iOS 18.x Simulator

#### Test flow

```yaml
- launchApp

- tapOn:
    text: Text

- assertVisible:
    text: Text
    selected: true
```

#### View hierarchy

```json
// ...

{
  "attributes": {
    "accessibilityText": "Text",
    "bounds": "[101,103][262,140]",
    "enabled": "true",
    "focused": "false",
    "selected": "true",
    "checked": "false"
  },
  "children": [
    {
      "attributes": {
        "accessibilityText": "Text",
        "bounds": "[127,113][236,130]",
        "enabled": "true",
        "focused": "false",
        "selected": "false",
        "checked": "false"
      },
      "children": [
        {
          "attributes": {
            "accessibilityText": "Text",
            "bounds": "[127,113][236,130]",
            "enabled": "true",
            "focused": "false",
            "selected": "false",
            "checked": "false"
          },
          "enabled": true,
          "focused": false,
          "checked": false,
          "selected": false
        }
      ],
      "enabled": true,
      "focused": false,
      "checked": false,
      "selected": false
    }
  ],
  "enabled": true,
  "focused": false,
  "checked": false,
  "selected": true
}

// ...
```

### Actual results

```yaml
- assertVisible:
    text: Text
    selected: true
```

The selector **does not** have a match.

### Expected results

```yaml
- assertVisible:
    text: Text
    selected: true
```

The selector has a match.

### About app

- React Native `0.79.5`
- Expo `53.0.19`

### About environment

Development machine:
 
- OS: macOS 15.5
- Processor: Apple M1, arm64
- Java: 17.0.2
- Xcode: 16.4

CI machine:

- OS:  macOS 14.7.6
- Processor: Apple M1, arm64
- Java: 17
- Xcode: 15.0.1

### Logs

<details>
<summary>Logs</summary>

```
00:00:00.376 [ INFO] maestro.cli.runner.MaestroCommandRunner.invoke: Tap on "${text}" COMPLETED
[...]

PR #2711, opened 2025-09-09
fix: text and id combined selector

### LINKED ISSUE #2533: [ios] using assertVisible with id and text doesn't work

### Steps to reproduce

Create `Counter` component with

```
import React from 'react'
import { Text, View, TouchableOpacity, StyleSheet, useColorScheme } from 'react-native'
import { getFontStyles } from '@/constants/fontStyles'
import { ThemedText } from '@/components/ThemedText'
import { colors } from '@/constants/Colors'

const Counter = ({ 
  count, 
  onIncrease, 
  onDecrease, 
  containerStyle, 
  buttonStyle, 
  textStyle, 
  counterTextStyle,
  id
}) => {
  console.log(`${id}-count`)
    const colorScheme = useColorScheme()
    const fontStyles = getFontStyles(colorScheme)
    const styles = StyleSheet.create({
      container: {
        flexDirection: 'row',
      },
      button: {
        // backgroundColor: colors.mainBlue,
        width: 37,
        // borderRadius: 20,
        alignItems: 'center',
        justifyContent: 'center',
      },
      buttonText: {
        fontSize: 20,
        fontWeight: 'bold',
        color: colors.appGreen,
      },
      counterText: {
        fontWeight: 'bold',
        margin: 5,
        textAlignVertical: 'center'
      },
      inactiveButton: {}
    });
  
    return (
        <View style={[styles.container, containerStyle]} accessible={true}>
          <TouchableOpacity 
            activeOpacity={0.8}
            disabled={count <= 1}
            onPress={onDecrease}
            style={[styles.button, count <= 1 && styles.inactiveButton, buttonStyle]}
            testID={`${id}-decrease`}
          >
              <Text style={[styles.buttonText, textStyle]}>-</Text>
          </TouchableOpacity>
          <Text 
            style={[fontStyles.smallText, counterTextStyle, styles.counterText, textStyle]}
            testID={`${id}-count`}
          >{count}</Text>
          <TouchableOpacity 
            activeOpacity={0.8} 
            onPress={onIncrease} 
            style={[styles.button, buttonStyle]}
            testID={`${id}-increase`}
          >
              <Text style={[styles.buttonText, textStyle]}>+</Text>
          </TouchableOpacity>
        </View>
    );
};

export default Counter;
```

Then use it like this:

```
<Counter
	id="contact-settings-counter"
	containerStyle={containerStyle}
	buttonStyle={buttonStyle}
	textStyle={textStyle}
	counterTextStyle={counterTextStyle}
	onIncrease={() => handleNumberOfMems(Number(countToNotify) + 1)}
	onDecrease={() => handleNumberOfMems(Number(countToNotify) - 1)}
	count={countToNotify}
/>
```

### Actual results

Image

### Expected results

I can target specific text inside of a specific id

### About app

Expo / React Native
Closed source

### About environment

Latest maestro
macOS
iOS 18.2

### Logs

_No response_

### Maestro version

1.40.3

### Anything else?

[hierarchy.log](https://github.com/user-attachments/files/20691921/hierarchy.log)

proksh:
PR Description: Fix deepestMatchingElement Filter Logic for Text and ID Selectors

## Summary
This PR fixes a issue with the deepestMatchingElement filter logic that was causing incorrect element selection when using `text` and `id` selectors. The fix ensures that the filter properly returns correct deepest matching elements in the UI hierarchy.

## Problem
The previous implementation of deepestMatchingElement had a flawed algorithm that would incorrectly select child elements even when matching selectors was pointing at parent. This caused issues with element selection, particularly when multiple nested elements matched any one selector from `text` or `id`.

## Changes Made

1. Improved Filter Composition Logic ([maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt](https://github.com/mobile-dev-inc/Maestro/pull/2711/files#diff-00989d49f0eaa1f9e614580322e3b2ec79d9915a804ffa004e2f53ba36215698))

Before: We were applying filters with `text` and only selecting deepestOne from there. Then again doing the same with `id`, which can give wrong result if `id` & `text` was some parent of first deepest selected node. ([Test 129](https://github.com/mobile-dev-inc/Maestro/pull/2711/files#diff-bb8086499ecc43e4b6eb8c6abca0d48b393ac903c3dd21461ae8686fa639e078))

- Separated filter types: Distinguished between "basic filters" (text, ID, size, traits, etc.) and "relative filters" (below, above, leftOf, rightOf, etc.)
- Applied `deepestMatchingElement` correctly: Now only applies to basic filters, not relative filters
- Better filter intersection: Properly combines basic and relative filters using intersection logic

2. Fixed `deepestMatchingElement` Algorithm ([maestro-client/src/main/java/maestro/Filters.kt](https://github.com/mobile-dev-inc/Maestro/pull/2711/files#diff-2dfb6ca77500e06b2fffc98666d632770fa0ccda87e1dacdc904b6ff8167e3b9))

Before: Used a flawed approach that would return parent elements even when children matched. It used to work because we were sending pre filtered nodes, but since we are not sending a combined basicFilter nodes, this would fail.

After: Implemented a proper recursive algorithm that:
- Recursively checks children first
- Returns children if they match the filter
- Only returns the parent if no children match
- Handles multiple root elements correctly
- Uses .distinct() to avoid duplicates

3. Added Tests for DeepestMatchingElementTest ([maestro-test/src/test/kotlin/maestro/test/DeepestMatchingElementTest.kt](https://github.com/mobile-dev-inc/Maestro/pull/2711/files#diff-76396e1c4f6bd44404de67c0b2f6cc222a53f2fe835ee7cadfa6b440ea5ee401))

4. Added tests for Integration Test for cases that were failing before ([maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt](https://github.com/mobile-dev-inc/Maestro/pull/2711/files#diff-5f00691128e9f122afec58e59902dd0e34941d9c943ee56859390061d0d08060))

Fixes #2533 and #2651

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -3940,7 +3940,49 @@ class IntegrationTest {
         }
 
         // Then
-        // No test failure
+        // No test failure - if we reach this point, the test passed successfully
+    }
+
+    @Test
+    fun `Case 130 - Duplicate elements case for checking deepestHierarchy is working`() {
+        // Given
+        // We're looking for an element with the given text and id, but it has a child element that is only a partial match
+        val commands = readCommands("130_text_and_index")
+        val driver = driver {
+            id = "0"
+            element {
+                id = "1"
+                text = "some_text"
+                bounds = Bounds(0, 0, 200, 200)
+            }
+            element {
+                id = "2"
+                text = "some_text"
+                bounds = Bounds(0, 0, 200, 200)
+                element {
+                    id = "3"
+                    text = "some_text"
+                    bounds = Bounds(0, 0, 200, 200)
+
+                    element {
+                        id = "4"
+                        text = "some_text"
+                        bounds = Bounds(50, 50, 150, 150)
+                    }
+                }
+            }
+
+        }
+
+        // When
+        Maestro(driver).use {
+            runBlocking {
+                orchestra(it).runFlow(commands)
+            }
+        }
+
+        // Then
+        // No test failure - if we reach this point, the test passed successfully
     }
     
     private fun orchestra(
--- a/maestro-client/src/main/java/maestro/Filters.kt
+++ b/maestro-client/src/main/java/maestro/Filters.kt
@@ -269,13 +269,16 @@ object Filters {
 
     fun deepestMatchingElement(filter: ElementFilter): ElementFilter {
         return { nodes ->
-            filter(nodes)
-                .map {
-                    val matchingChildren = deepestMatchingElement(filter)(it.children)
-
-                    matchingChildren.lastOrNull()
-                        ?: it
+            nodes.flatMap { node ->
+                val matchingChildren = deepestMatchingElement(filter)(node.children)
+                if (matchingChildren.isNotEmpty()) {
+                    matchingChildren
+                } else if (filter(listOf(node)).isNotEmpty()) {
+                    listOf(node)
+                } else {
+                    emptyList()
                 }
+            }.distinct()
         }
     }
 
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -56,6 +56,7 @@ import org.slf4j.LoggerFactory
 import java.io.File
 import java.lang.Long.max
 import java.nio.file.Path
+import java.util.logging.Filter
 import kotlin.coroutines.coroutineContext
 
 // TODO(bartkepacia): Use this in onCommandGeneratedOutput.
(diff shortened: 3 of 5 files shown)


Fishbowler:
Why wouldn't we take the first deepest matching element of all filters?

Isn't there a problem where we have:

```
tapOn:
  text: This One
  checked: true
```

But it doesn't match because deepestMatchingElement picks a different "This One"?

i.e. won't this problem apply to all "id + something" and "text + something" filters?

proksh:
> Why wouldn't we take the first deepest matching element of all filters?
> 
> Isn't there a problem where we have:
> 
> ```
> tapOn:
>   text: This One
>   checked: true
> ```
> 
> But it doesn't match because deepestMatchingElement picks a different "This One"?
> 
> i.e. won't this problem apply to all "id + something" and "text + something" filters?

Ohh got it, but what about index? If I do:
```
tapOn:
  text: This One
  index: 2
```
Should it also count the nesting? Wont it lead to multiple count for that single element? Should this deepestMatchingElement be before index only?

Fishbowler:
That makes sense.

The deepest matching element is to prevent getting a container rather than "the actual thing". And you can still end up with multiple matches, so indexes still make sense too.

proksh:
> That makes sense.
> 
> The deepest matching element is to prevent getting a container rather than "the actual thing". And you can still end up with multiple matches, so indexes still make sense too.

Updated PR descriptions for the changes made.

proksh:
I have tested this on some workspace

@Fishbowler It would be great if you can test in some workspace that you know use alot of combined selector to be sure this is not breaking anything.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -3912,6 +3912,78 @@ class IntegrationTest {
             }
         }
     }
+
+    @Test
+    fun `Case 129 - Text and ID with child elements`() {
```
amanjeetsingh150, line 3917: Nice!

PR #2712, opened 2025-09-09
add install app command

aprokofjev-mwb:
## Proposed changes

copilot:summary


Fishbowler:
Interesting change @aprokofjev-mwb - I can see how this would be useful when you need companion apps, or wanted to have an accessibility tool to run alongside.

PR #2714, opened 2025-09-10
Suggests initial biometry implementation

sidferreira:
## Proposed changes

copilot:summary

### DIFF
--- a/maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
+++ b/maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
@@ -145,7 +145,7 @@ class FakeDriver : Driver {
 
     override fun setBiometry(result: Boolean) {
         ensureOpen()
-        // record event for testing if needed in the future
+        events.add(Event.Biometry(result))
     }
 
     override fun tap(point: Point) {
@@ -507,6 +507,10 @@ class FakeDriver : Driver {
         object TakeScreenshot : Event()
 
         object ClearKeychain : Event()
+
+        data class Biometry(
+            val result: Boolean
+        ) : Event()
 
         data class SetLocation(
             val latitude: Double,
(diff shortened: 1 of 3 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1008,6 +1021,18 @@ class Orchestra(
             throw MaestroException.UnableToLaunchApp("Unable to launch app ${command.appId}: ${e.message}", e)
         }
 
+        // If requested, enroll biometry on iOS simulators by sending enrollment changed notifications.
+        if (command.enrollBiometry == true && maestro.cachedDeviceInfo.platform == Platform.IOS) {
+            try {
+                val parts1 = listOf("xcrun", "simctl", "spawn", "booted", "notifyutil", "-s", "com.apple.BiometricKit.enrollmentChanged", "1")
+                val parts2 = listOf("xcrun", "simctl", "spawn", "booted", "notifyutil", "-p", "com.apple.BiometricKit.enrollmentChanged")
```
Fishbowler, line 1028: This should be in the `launch` method of the device too?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/MaestroFlowParser.kt
```diff
@@ -142,6 +142,10 @@ private val stringCommands = mapOf<String, (JsonLocation) -> YamlFluentCommand>(
         _location = location,
         assertNoDefectsWithAI = YamlAssertNoDefectsWithAI()
     )},
+    "biometry" to { location -> YamlFluentCommand(
+        _location = location,
+        biometry = YamlBiometry.parse("match")
+    )},
```
Fishbowler, line 148: I don't think this should be passing in a value to parse

Issue #2701, opened 2025-09-04
Maestro does not work in iPad landscape mode


yolpsoftware:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

- Clone the following repo (was used to demonstrate #2480): https://github.com/yolpsoftware/maestro-bugtest
- cd a terminal into the repo's folder
- run `npm install`
- Start an iPad simulator (I used iPad mini A17 Pro with iOS 18.3).
- run `npx expo start`
- Press the key `i` in the terminal. This will install the app Expo Go on the simulator and then run the app.
- Flip the iPad simulator to landscape mode.
- Run `maestro studio`

You will see this:

<img width="480" height="734" alt="Image" src="https://github.com/user-attachments/assets/8ba8c530-9dc4-4fe1-b9e8-3b80ca1ac073" />

<img width="475" height="716" alt="Image" src="https://github.com/user-attachments/assets/2a5de3d4-5f27-4128-b1b0-bf694349ca18" />

🤔 Maestro interpreting `height x width` as `width x height`?

- If that's not enough, run `- tapOn: Click me!`, either in Maestro Studio, or in a test.

Note that this is a regression. Those tests worked fine before the v2 update.

### Actual results

Test does not work, "Click Me!" will not be clicked (no alert shown). Maestro Studio showing strange image.

### Expected results

Test should work. Maestro Studio should show iPad simulator as it actually appears.

### About app

React Native app with Expo 52. However, the Maestro bug does not seem to be connected to React Native. When I run Maestro Studio on an iPad in Landscape mode, it looks strange even on the home screen.

### About environment

openjdk 21.0.7
Sequoia 15.5
Macbook Pro M4
Maestro 2.0.2

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

2.0.2

### How did you install Maestro?

Homebrew

### Anything else?

Am I the only one writing tests in Landscape?

yolpsoftware:
Reverting to v1, as Maestro is unusable like this.

amanjeetsingh150:
Hey @yolpsoftware there is a fix for this in the works right now.

yolpsoftware:
@amanjeetsingh150 any estimate on when this will be completed?

[...]

PR #2720, opened 2025-09-12
[MA 3559] Hierarchy fix to support orientation changes

amanjeetsingh150:
## Problem 

1. Element selectors broken: All element-selector commands (e.g., tap, swipe) are currently not working in Maestro, impacting CLI, Studio, and Cloud.
2. On iOS, screens that respect rotation return a hierarchy that matches the current orientation and can be used as-is. For screens that don’t respect rotation, iOS returns the hierarchy in the locked orientation (often portrait), which breaks Studio bounding boxes.
3. Because many apps have rotation-respecting and rotation-locked screens, this inconsistency makes it hard to write and validate flows locally for apps with orientation changes specially with studio.
4. Taps are impacted: Since the returned hierarchy is wrong on rotation-locked screens, computed target coordinates are wrong and taps land incorrectly. (A fix for taps will be addressed in a separate PR.)
5. We need both the visual hierarchy (what users see, orientation-normalized) and the AX hierarchy (how iOS/XCTest reports it) so Studio can draw boxes correctly while element selectors remain accurate.

## Proposed changes

Scope of this PR: add support for both visual and AX hierarchies so clients can choose the appropriate mode. Visual is used for Studio overlays; AX is used for element-selector commands.

1. This PR introduces both **visualElement** and **axElement** in the hierarchy response, allowing clients to opt into the desired mode. 
2. When the app orientation differs from the device orientation on rotation-locked screens, rotate/normalize element coordinates for the visual hierarchy only. AX remains unchanged 
3. Allow Maestro to select Visual (for Studio) or AX (default, for element interactions in CLI/Cloud).

## Testing

- [ ] Locally: Wikipedia, Customer

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -55,6 +55,11 @@ import java.io.File
 import java.net.SocketTimeoutException
 import kotlin.collections.set
 
+enum class HierarchyMode {
+    AX,
+    VISUAL
+}
+
 class IOSDriver(
     private val iosDevice: IOSDevice,
     private val insights: Insights = NoopInsights,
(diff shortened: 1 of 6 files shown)


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeProvider.swift
```diff
@@ -1,24 +1,22 @@
...
-struct ScreenSizeHelper {
+struct ScreenSizeProvider {
 
     private static var cachedSize: (Float, Float)?
     private static var lastAppBundleId: String?
     private static var lastOrientation: UIDeviceOrientation?
+    
+    private static let springboardBundleId = "com.apple.springboard"
 
     static func physicalScreenSize() -> (Float, Float) {
```
amanjeetsingh150, line 12: This is quiet confusing abstraction we have both physicalScreenSize and actualSize. 

I would be deleting this in subsequent PR since taps need to take place using ax coordinates.

Fishbowler:
Fixes #2701 ?

Fishbowler:
@amanjeetsingh150 I don't claim to have the skill to appropriately review implementation, but the tests look solid.

What do you want to happen with this PR now?

mahmoudalaa97:
Why was this PR closed? Was there another PR that fixed the main issue with the iPad landscape orientation? #2701

PR #2732, opened 2025-09-17
remove dadb forwarder map

herval:
this is not only completely unnecessary, it actually leads to connections being kept around forever.


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -183,17 +180,9 @@ class AndroidDriver(
         }
 
         LOGGER.info("[Start] close port forwarder")
-        PORT_TO_FORWARDER[hostPort]?.close()
+        postForwarder?.close()
```
amanjeetsingh150: nit: tcpPortForwarder or portForwarder. 

post confuses the context.

herval: agree, that was a typo

Fishbowler:
@herval Any chance this will break sharding?

PR #2733, opened 2025-09-17
Fix iOS permission setting when including 'all' with overrides

Fishbowler:
## Proposed changes

The following should work, are in our docs, but on iOS, the 'all' overrode the individual permission, rather than the other way around.

```
- launchApp:
    permissions:
      all: allow
      location: inuse


- launchApp:
    permissions:
      all: allow
      contacts deny
```

This change fixes that behaviour. I suspect the change was only needed in simctl, and there's a chance that the Apple SimUtils change might even be a no-op, but it feels like it now has more explicit intent.

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
@@ -419,19 +419,28 @@ object LocalSimulatorUtils {
 
     fun setAppleSimutilsPermissions(deviceId: String, bundleId: String, permissions: Map<String, String>) {
         val permissionsMap = permissions.toMutableMap()
+        val effectivePermissionsMap = mutableMapOf<String, String>()
+
         if (permissionsMap.containsKey("all")) {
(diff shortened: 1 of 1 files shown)


maestro-ios-driver/src/main/kotlin/util/LocalSimulatorUtils.kt
```diff
@@ -484,11 +494,17 @@ object LocalSimulatorUtils {
...
 
+        // Write the explicit permissions, potentially overriding the 'all' permissions
+        permissionsMap.forEach {
+            if (simctlPermissions.contains(it.key)) {
+                effectivePermissionsMap[it.key] = it.value
+            }
+        }
 
-        permissionsMap
+        effectivePermissionsMap
```
amanjeetsingh150, line 507: Man this reminds me of cleanup on permissions as well, there is lot of duplication there and applesimutils is deprecated 😢

Fishbowler: This is one area where we'd need to do some work for real device support too 😬

PR #2737, opened 2025-09-21
Test case properties added for JUnit reports

kprakash2:
## Proposed changes

In some cases, it maybe required to add certain meta data to our JUnit test reports. My use-case is, we need a way to associate each test with an arbitrary ID that is generated by our quality management system. So we can easily track test for our requirements from our validation reports.

Changes in this MR, provides a new configuration in the Flow files, where you can add optional properties to their test cases if needed.

Example my-test.yaml

```
appId: com.example.app
properties:
  testid: my-test
  external-test-id: abc-deg-f-123
---
- launchApp:
    clearState: true
- assertVisible: "My Button"
```

Example report.xml generated by above

```
<?xml version='1.0' encoding='UTF-8'?>
<testsuites>
  <testsuite name="Test Suite" device="Pixel_8" tests="1" failures="0" time="1.0">
    <testcase id="my-test" name="my-test" classname="my-test" time="1.0" status="SUCCESS">
      <properties>
        <property name="testid" value="my-test"/>
        <property name="external-test-id" value="abc-deg-f-123"/>
      </properties>
    </testcase>
  </testsuite>
</testsuites>

```

copilot:summary

## Testing

- Added properties to one of the flow yaml files under resources, to make sure no other test or anything else is broken by this change
- Manually tested the flow using example provided above


.gitignore
```diff
@@ -15,4 +15,7 @@ bin
 maestro-orchestra/src/test/resources/media/assets/*
 
 # Local files
-local/
\ No newline at end of file
+local/
+
+# IntelliJ IDEA files
+.idea
```
Fishbowler: Overkill and potentially undesirable

maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
```diff
@@ -28,6 +28,7 @@ data class TestExecutionSummary(
         val failure: Failure? = null,
         val duration: Duration? = null,
         val startTime: Long? = null,
+        val properties: Map<String, String>? = null,
```
Fishbowler, line 31: Is there a more descriptive name than properties?

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -172,7 +172,8 @@ class TestSuiteInteractor(
             .readCommands(flowFile.toPath())
             .withEnv(env)
 
-        var flowName: String = YamlCommandReader.getConfig(commands)?.name ?: flowFile.nameWithoutExtension
+        val yaml = YamlCommandReader.getConfig(commands)
```
Fishbowler: Should this be "flowConfig" rather than yaml?

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroConfig.kt
```diff
@@ -12,6 +12,7 @@ data class MaestroConfig(
     val ext: Map<String, Any?> = emptyMap(),
     val onFlowStart: MaestroOnFlowStart? = null,
     val onFlowComplete: MaestroOnFlowComplete? = null,
+    val properties: Map<String, String> = emptyMap(),
```
Fishbowler, line 15: What happens when the user provides a number or Boolean in the YAML?

.gitignore
kprakash2: Reverted.

maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
kprakash2: So, I did some research but seems like there is no JUnit report xml standard, there are some definitions available online and they have `properties`. Most places I found `<properties>` declaration to represent environment or optional properties in the report.

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
kprakash2: It will be `MaestroConfig`. Updated it to `maestrConfig`, also made it `val`. There was no need for it to be `var`

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroConfig.kt
kprakash2: They will be treated as string, so it will produce something like following in the report:

```
<property name="bool-property" value="true"/>
<property name="num-property" value="1234"/>
```

kprakash2:
@Fishbowler can you look at the latest changes, when you get a chance. Thanks.

Fishbowler:
That was me doing the assigning and requesting reviewing just above 😉

Fishbowler (approved):

Fishbowler:
I've had a good test of this locally, and this looks ace!

Could I ask for one more thing? Could you update the docs in https://github.com/mobile-dev-inc/maestro-docs to add instructions on properties in `/cli/test-suites-and-reports.md`? When this gets released, we can merge the docs at the same time, and let everyone know how to take advantage of this :heart:

Fishbowler:
Rebased.

@kprakash2 Are you up for some small docs changes to match? (see above)

Fishbowler:
Rebased

Fishbowler:
Docs: done.

PR #2743, opened 2025-09-23
Automate driver building for Maestro

amanjeetsingh150:
## Proposed changes

1. Remove not used github workflows.
2. Automates iOS driver building on latest macOS so that it builds automatically fro iOS 26
3. Automates android driver building as well 

## Testing

Idea is to test this with workflow dispatch.

### DIFF
--- a/.github/workflows/build_maestro_android_driver.yaml
+++ b/.github/workflows/build_maestro_android_driver.yaml
@@ -1,0 +1,32 @@
+name: Build Maestro Android driver
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'maestro-android/**'
+
+permissions:
+  contents: write  # allow committing back to the repo
+
+jobs:
+  build_android_driver:
+    name: Build & Commit Android Driver
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build Android driver
+        run: |
+          ./gradlew clean :maestro-android:assemble
+          ./gradlew clean :maestro-android:assembleAndroidTest
+
+      - name: Commit updated driver
+        uses: stefanzweifel/git-auto-commit-action@v6
+        with:
+          commit_message: "chore(android-driver): update built Maestro android runner"
+          file_pattern: |
+            maestro-android/**


.github/workflows/build_maestro_android_driver.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build Android driver
+        run: |
+          ./gradlew clean :maestro-android:assemble
+          ./gradlew clean :maestro-android:assembleAndroidTest
```
Fishbowler, line 25: Will this require setup-java?

amanjeetsingh150:
> Do we care if someone already contributed the binaries? (My guess: no)

I don't think people would do that. If this automation works we can actually remove from readme if its needed.

But if someone still does it I think worse is that it can create conflict on the bot's PR.


> Should we deal with community contributions where someone unchecked "Maintainers are allowed to edit this pull request"? I guess it's fine (and correct) for this to error, but do we want to error early?

Not sure if I got this one are you saying should it be allowed for non maintainers to edit this update driver PR?

> Do we want to include changes to these workflows as triggers for these workflows, as changes occur?

Yeah that'd good idea let me see if there is a way not sure

.github/workflows/build_maestro_android_driver.yaml
amanjeetsingh150: Ohh yeah I forgot let me update it

Fishbowler:
> > Should we deal with community contributions where someone unchecked "Maintainers are allowed to edit this pull request"? I guess it's fine (and correct) for this to error, but do we want to error early?
>
> Not sure if I got this one are you saying should it be allowed for non maintainers to edit this update driver PR?

Non-maintainers will submit other PRs.
Those PRs might have changes to relevant files that would trigger this workflow.
Those PRs might _not_ have "Maintainers are allowed to edit this pull request" enabled (although that's usual)
In that case, the commit step would fail.

Fishbowler:
> > Do we care if someone already contributed the binaries? (My guess: no)
>
> I don't think people would do that. If this automation works we can actually remove from readme if its needed.
> But if someone still does it I think worse is that it can create conflict on the bot's PR.

I think this workflow would just overwrite them with known good versions. That might be good for security as well as other things. I reckon let's ignore it.

PR #2746, opened 2025-09-24
Fix e2e test flake

Fishbowler:
## Proposed changes

The test:
```
- inputText: 'testing'
- assertVisible: 'testing'
- eraseText: 3
- assertNotVisible: 'testing'
```

The last line would flake due to keyboard suggestions

image

This introduces 'above' in the visibility selectors to avoid keyboard suggestions.

It also removes the optional keyword for the last assertion - we should find out if that last assertion is flaky.


amanjeetsingh150 (approved):

Fishbowler:
Needs #2779

Issue #2059, opened 2024-09-16
Support tapping on a point within an element


tokou:
### Use case

Tapping on an element will automatically tap on the center of the element
This can lead to issues in cases where the element id refers to a large area that could have specific tap zones

Example: This whole area is tappable and has a single element id. However it contains links in the spannable text.
![Capture d’écran 2024-09-16 à 14 22 19](https://github.com/user-attachments/assets/f2dde509-1e70-47f0-851b-2c7299baefac)

Other examples include needing to tap on specific areas of large images:
![Capture d’écran 2024-09-16 à 14 26 51](https://github.com/user-attachments/assets/16724084-1161-4c12-a355-be72e634b729)

### Proposal

Make it so that 
```
- tapOn:
    id: 'large_button'
    point: '10%, 50%'
```
Would tap on the left side of the selected element instead of the current behavior of simply ignoring the id

### Anything else?

This is the related code

https://github.com/mobile-dev-inc/maestro/blob/21edc1442aa96d6c435fb88b57bd5e72a65af0c8/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt#L472-L497

ben-pomelo:
Is this still an issue? the docs imply that you can do this: https://docs.maestro.dev/api-reference/commands/tapon#tapping-on-a-specific-point-within-another-element but it doesn't seem to work when I have used it in my tests.

ben-pomelo:
This should be a P0 or P1: it's a very important feature that makes it very hard to test certain workflows, can we get some eyes on this from the maestro.dev team @felipeduartea ?

howwellapp:
I ran into this also. Can someone delete https://docs.maestro.dev/api-reference/commands/tapon#tapping-on-a-specific-point-within-another-element  since it doesn't work?

Of course, fixing it would be better, my test could really use this. :)

Fishbowler:
@howwellapp Hard no on the delete. Instead, let's fix it, so that the docs are correct again. If you've run into it, clearly it's useful.

Fishbowler:
In fact, there's already an open PR that needs reviewing and testing.

Anyone up for contributing some feedback there?

simon-gilmurray:
This has been fixed as of v2.0.4 🎉

PR #2752, opened 2025-10-01
feat: added the missing feature of point-within-another-element

proksh:
# Element-Relative Point Tapping

### What This Adds
- [This PR adds the missing functionality that was already documented but not available in the codebase.
](https://docs.maestro.dev/api-reference/commands/tapon#tapping-on-a-specific-point-within-another-element) Ability to tap at specific coordinates within an element's bounds, not just at the element's center.
- Update command description to include this info

### YAML Syntax
```
# Tap at 50% from left, 90% from top of the Submit button
- tapOn:
    text: Submit
    point: 50%, 90%

# Tap at 25px from left, 75px from top of element
- tapOn:
    id: submit-btn
    point: 25, 75
```

### Why This Matters
- **Fills documentation gap:** Feature was documented but not implemented
- **Testing:** Tap specific areas of buttons/elements

### Added Tests
Added integration test to check the feature as well as backward compatibility


maestro-orchestra/src/main/java/maestro/orchestra/CoordinateUtils.kt
```diff
@@ -0,0 +1,53 @@
...
+import maestro.Bounds
+import maestro.MaestroException
+import maestro.Point
+import maestro.UiElement
+
+/**
+ * Utility class for coordinate calculations.
+ * This class contains pure functions that can be easily tested and reused.
+ */
+internal object CoordinateUtils {
```
amanjeetsingh150: Why object here? This could just be function in file

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1053,16 +1055,34 @@ class Orchestra(
...
-            appId = config?.appId,
-            tapRepeat = command.repeat,
-            waitToSettleTimeoutMs = command.waitToSettleTimeoutMs,
-        )
+
+        // Handle element-relative tap if specified
+        val elementRelativePoint = command.elementRelativePoint
+        if (elementRelativePoint != null) {
+            val tapPoint = calculateElementRelativePoint(result.element, elementRelativePoint)
+            println("Element-relative tap: calculated point $tapPoint within element bounds ${result.element.bounds}")
```
amanjeetsingh150: Why? println here?

maestro-orchestra/src/test/java/maestro/orchestra/ElementRelativeTapRealIntegrationTest.kt
```diff
@@ -0,0 +1,223 @@
...
+            elementRelativePoint = "25, 75"
+        )
+
+        // When: We test the real coordinate calculation
+        val testElement = createTestUiElement()
+        val calculatedPoint = CoordinateUtils.calculateElementRelativePoint(testElement, command.elementRelativePoint!!)
+
+        // Then: Verify the real calculation works correctly
+        assertThat(calculatedPoint.x).isEqualTo(125) // 100 + 25 = 125
+        assertThat(calculatedPoint.y).isEqualTo(175) // 100 + 75 = 175
```
amanjeetsingh150: this could be with other IntegrationTests?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -287,14 +287,16 @@ data class TapOnElementCommand(
     val longPress: Boolean? = null,
     val repeat: TapRepeat? = null,
     val waitToSettleTimeoutMs: Int? = null,
+    val elementRelativePoint: String? = null, // New parameter for element-relative coordinates
```
amanjeetsingh150: Why not just relativePoint? This is already part of tapOnElementCommand

amanjeetsingh150 (commented):
Few question, I'll suggest writing adding test for this:

1. Does this work with command doubleTapOn?
2. Does this also work on tapOn with repeat?

Fishbowler:
Fixes #2059

maestro-orchestra/src/main/java/maestro/orchestra/CoordinateUtils.kt
proksh: The main issue with the function inside the `Orchestra` class is that it cannot be easily tested. Mocking the entire `Orchestra` class for testing would be complex for no reason.

Separating the function out made sense as the function can now be included in unit tests without needing to mock the entire `Orchestra` class.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
proksh: Removed. It was by mistake.

maestro-orchestra/src/test/java/maestro/orchestra/ElementRelativeTapRealIntegrationTest.kt
proksh: Yeah, made sense. Included these into `YamlCommandReaderTest` and one that were related to Coordinates into `CoordinateUtilsTest`

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
proksh: Updated

proksh:
> Few question, I'll suggest writing adding test for this:
> 
> 1. Does this work with command doubleTapOn?
> 2. Does this also work on tapOn with repeat?

Yes, works for both case. Added tests for both.
- https://github.com/mobile-dev-inc/Maestro/pull/2752/commits/62bb6f5c873ad64de4ce63674044ca122c5643ff
- https://github.com/mobile-dev-inc/Maestro/pull/2752/commits/c7345c1bb3e0a54e9cc935b0d243afa0304ac8ef

maestro-orchestra/src/main/java/maestro/orchestra/CoordinateUtils.kt
proksh: Removed it from the object and created a separate function

PR #2755, opened 2025-10-01
Feat/posthog setup

proksh:
Changed posthog setup

Added tracking for:

- cli command run (any command run on cli, gives all user using cli)
- test command
- record command
- hierarchy command
- clout upload command

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/App.kt
+++ b/maestro-cli/src/main/java/maestro/cli/App.kt
@@ -21,6 +21,7 @@ package maestro.cli
 
 import maestro.MaestroException
 import maestro.cli.analytics.Analytics
+import maestro.cli.analytics.AnalyticsEvents
 import maestro.cli.command.BugReportCommand
 import maestro.cli.command.ChatCommand
 import maestro.cli.command.CheckSyntaxCommand
@@ -114,8 +115,7 @@ fun main(args: Array<String>) {
     // https://stackoverflow.com/a/17544259
     try {
         System.setProperty("apple.awt.UIElement", "true")
-
-        Analytics.maybeAskToEnableAnalytics()
+        Analytics.warnAndEnableAnalyticsIfNotDisable()
 
         Dependencies.install()
         Updates.fetchUpdatesAsync()
@@ -149,7 +149,8 @@ fun main(args: Array<String>) {
                 1
             }
 
-        Analytics.maybeUploadAnalyticsAsync(args[0])
+        // Track CLI run
+        AnalyticsEvents.trackCliCommandRun(args[0])
 
         val generateCompletionCommand = commandLine.subcommands["generate-completion"]
         generateCompletionCommand?.commandSpec?.usageMessage()?.hidden(true)
--- a/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
@@ -1,194 +1,154 @@
 package maestro.cli.analytics
 
-import com.fasterxml.jackson.annotation.JsonFormat
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties
-import com.fasterxml.jackson.databind.SerializationFeature
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
+import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
-import com.fasterxml.jackson.module.kotlin.readValue
 import com.posthog.java.PostHog
-import maestro.cli.util.CiUtils
+import maestro.auth.ApiKey
+import maestro.cli.api.ApiClient
 import maestro.cli.util.EnvUtils
-import maestro.cli.util.IOSEnvUtils
-import maestro.device.util.AndroidEnvUtils
 import org.slf4j.LoggerFactory
-import java.net.ConnectException
 import java.nio.file.Path
-import java.time.Duration
-import java.time.Instant
-import java.util.*
-import kotlin.io.path.createDirectories
-import kotlin.io.path.exists
-import kotlin.io.path.readText
-import kotlin.io.path.writeText
+import kotlin.String
 
-/**
- * The new analytics system for Maestro CLI.
- *  - Sends data to /maestro/analytics endpoint.
- *  - Uses configuration from $XDG_CONFIG_HOME/maestro/analytics.json.
- */
 object Analytics : AutoCloseable {
     private const val POSTHOG_API_KEY: String = "phc_XKhdIS7opUZiS58vpOqbjzgRLFpi0I6HU2g00hR7CVg"
     private const val POSTHOG_HOST: String = "https://us.i.posthog.com"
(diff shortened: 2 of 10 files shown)


maestro-cli/src/main/java/maestro/cli/analytics/AnalyticsEvents.kt
```diff
@@ -0,0 +1,120 @@
...
+import maestro.cli.util.IOSEnvUtils
+import maestro.device.util.AndroidEnvUtils
+import kotlin.time.Duration
+
+/**
+ * Event-specific analytics functions for Maestro CLI.
+ * Contains functions that track specific business events.
+ * Uses Analytics internally for core functionality.
+ */
+object AnalyticsEvents {
```
steviec: I love the typed events. But if these helper methods are simple passthrough wrappers, do they provide much value above just calling trackEvent directly? e.g.
```
Analytics.trackEvent(CliCommandRunEvent(command = "blah"))
 ```

File: maestro-cli/src/main/java/maestro/cli/analytics/PostHogEvents.kt [hunk 2]
```diff
@@ -0,0 +1,209 @@
+package maestro.cli.analytics
```
steviec, line 1: Have you ensured that we're sending the same payloads as in Studio for events and superproperties? If not, we might want to cross reference and ensure we have a single schema.

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -1,194 +1,154 @@
...
 
-            if (!analyticsState.enabled) {
-                logger.trace("Analytics disabled with config file, not uploading")
-                return
+    /**
+     * Track events with optional PostHog-level deduplication
+     * Use this for important events like authentication, errors, test results, etc.
+     * For CLI runs and other repeatable events, set enableDeduplication = true
+     */
+    fun trackEvent(event: PostHogEvent) {
```
steviec, line 101: The posthog.capture() call will queue up calls to be sent async, but everything else in this method is synchronous. I think we should move everything into a background thread so that it will never block. This would be especially bad if our backend server isn't available for some reason or hangs and it causes the CLI to stop functioning. trackEvent calls should always be "fire and forget".

maestro-cli/src/main/java/maestro/cli/analytics/AnalyticsEvents.kt
proksh: Yeah make sense, let me try this.

amanjeetsingh150 (approved):
Couple of questions:

1. The CLI analytics that we earlier had on BigQuery now go to PostHog? Have we removed BigQuery analytics completely?

2. The behavior around analytics is the same right — i.e. we don’t block the user and analytics are enabled by default?

proksh:
> 1. The CLI analytics that we earlier had on BigQuery now go to PostHog? Have we removed BigQuery analytics completely?

Yes, that was already removed, and we were using Poshog for a while, but it was not properly setup.

> 2. The behavior around analytics is the same right — i.e. we don’t block the user and analytics are enabled by default?

Yes, the behaviour is same.

maestro-cli/src/main/java/maestro/cli/analytics/PostHogEvents.kt
proksh: Updated to match Studio schema.

Note some property like "studioUserAcquisitionSource", "isFirstAuth", is not here.
And some additional parameters related to version and env being added to superproperties.

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
proksh: capture was already an async event. But Now I have updated the entire trackEvent to be async.

maestro-cli/src/main/java/maestro/cli/App.kt
```diff
@@ -180,11 +181,14 @@ fun main(args: Array<String>) {
...
             printVersion()
+            Analytics.close()
             exitProcess(0)
         }
 
+        Analytics.close()
         exitProcess(exitCode)
-    } finally {
+    } catch (e: Throwable) {
         Analytics.close()
```
proksh, line 191: `Analytics.close()` was never being called because we were calling `exitProcess(exitCode)` before it. 🤦🏻‍♂️

gradle/libs.versions.toml
```diff
@@ -124,7 +124,7 @@ mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
 mozilla-rhino = { module = "org.mozilla:rhino", version.ref = "mozillaRhino" }
 picocli = { module = "info.picocli:picocli", version.ref = "picocli" }
 picocli-codegen = { module = "info.picocli:picocli-codegen", version.ref = "picocli" }
-posthog = { module = "com.posthog.java:posthog", version.ref = "posthog" }
+posthog = { module = "com.posthog:posthog-server", version.ref = "posthog" }
```
proksh, line 127: Had to change the package and version to the one in posthog doc - https://posthog.com/docs/libraries/java

To get the `flush()`

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
```diff
@@ -1,194 +1,182 @@
...
+   /**
+    * Close and cleanup resources
+    * Ensures pending analytics events are sent before shutdown
+    */
     override fun close() {
-        posthog.shutdown()
+        // First, flush any pending PostHog events before shutting down threads
+        try {
+            posthog.flush()
+            Thread.sleep(500) // No sync flush is available in the PostHog Java client
```
proksh: Flush is being called to send all the pending events. 

Thread.sleep is because flush is async function. Let me know if there is a better solution.

maestro-cli/src/main/java/maestro/cli/analytics/AnalyticsEvents.kt
proksh: Much cleaner now.

maestro-cli/src/main/java/maestro/cli/App.kt
steviec: Good catch!

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
steviec: Yeah, let's not do that :).

Based on the documentation, I think we can completely get rid of this; posthog.close() should call flush internally and close all connections. And while posthog.capture queues events to send async, posthog.close is a sync method that will close up everything before moving on (I think). https://posthog.com/docs/libraries/java#shutdown

proksh: The problem here is that it is not working. Server will just close without flushing the cloud upload finished event.

Tried using 
- only `close`, 
- `flush` & then `close`. 

Both didn't work

It is working only with `flush` and Thread.sleep. :(

I will try to figure out a better way.

proksh: Removed `thread.sleep` and changed the structure a bit. Can confirm that it is working.

PR #2759, opened 2025-10-06
feat: cli project selector

proksh:
## Proposed changes

While doing cloud upload:
1. Fixed authentication by fixing the url
2. In case user belongs to multiple project and none are specified in the command: added prompt so that user can see and select the project he want the upload to happen.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -7,6 +7,7 @@ import maestro.cli.analytics.CloudUploadStartedEvent
 import maestro.cli.analytics.CloudUploadTriggeredEvent
 import maestro.cli.api.ApiClient
 import maestro.cli.api.DeviceConfiguration
+import maestro.cli.api.ProjectResponse
 import maestro.cli.api.UploadStatus
 import maestro.cli.auth.Auth
 import maestro.device.Platform
@@ -80,7 +81,6 @@ class CloudInteractor(
         deviceModel: String? = null,
         deviceOs: String? = null,
     ): Int {
-        if (projectId == null) throw CliError("Missing required parameter '--project-id'")
         if (appBinaryId == null && appFile == null && !flowFile.isWebFlow()) throw CliError("Missing required parameter for option '--app-file' or '--app-binary-id'")
         if (!flowFile.exists()) throw CliError("File does not exist: ${flowFile.absolutePath}")
         if (mapping?.exists() == false) throw CliError("File does not exist: ${mapping.absolutePath}")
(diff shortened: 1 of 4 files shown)


maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -173,24 +176,68 @@ class CloudInteractor(
...
             ))
             
-            // Flush analytics events immediately after tracking the upload finished event
-            Analytics.flush()
-            
             return uploadResponse
         }
     }
 
+    private fun selectProject(authToken: String): String {
```
trindadetiago, line 195: what happens if user is not logged in?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -173,24 +176,68 @@ class CloudInteractor(
...
+        val projects = try {
+            client.getProjects(authToken)
+        } catch (e: ApiClient.ApiException) {
+            throw CliError("Failed to fetch projects. Status code: ${e.statusCode}")
+        } catch (e: Exception) {
+            throw CliError("Failed to fetch projects: ${e.message}")
+        }
+
+        if (projects.isEmpty()) {
+            throw CliError("No projects found. Please create a project first at https://console.mobile.dev")
```
trindadetiago, line 207: should we create a default one or make it possible for the user to create it with the CLI?

What I am thinking is: If we are going to allow the user to create his account / org in CLI, we could as well make it create the project here also. Though I think this wouldn't be a blocker for this PR

Maybe I am miss understood, let me know

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
proksh: We get authtoken before this:
```       
 val authToken = auth.getAuthToken(apiKey)
```

Which will trigger the signin flow if user is not logged in
```
   fun getAuthToken(apiKey: String?, triggerSignIn: Boolean = true): String? {
        if (triggerSignIn) {
            return apiKey // Check for API key
                ?: ApiKey.getToken()
                ?: triggerSignInFlow() // Otherwise, trigger the sign-in flow
        }
        return apiKey // Check for API key
            ?: ApiKey.getToken()
    }
```

If he is still not logged in, an error will be thrown:
```
if (authToken == null) throw CliError("Failed to get authentication token")
```

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
proksh: Yeah, backend already creates a new project whenever a new org is created. So its already handled.

PR #2761, opened 2025-10-06
Consolidate Single enum for `Platform`

handstandsam:
## What?
I finally realized `maestro.Platform` and `maestro.device.Platform` were the same enums (except `maestro.device.Platform` was a superset).  So, now there is one canonical `Platform` enum.

## Why?
In my own codebase I was attempting to add an extension function on the `Platform` type and was getting weird compiler errors that didn't make sense.


Fishbowler:
This later got done by staff in another PR. You were ahead of the curve! ❤️

PR #2769, opened 2025-10-08
Add path aliases

sidferreira:
## Proposed changes

This PR is result of a discussion between me and @Fishbowler about ways to have TS Config path aliases in yaml files.

The initial idea was about using JS variables, but it would make test files more complicated and less readable. The proposed solution lets java handle the replacements and validation as well.

AI was used to develop this code.

Follow up ideas:
- maestro.paths.yaml (or similar) file or something for cases where the paths are automatically generated
- use a script to generate the paths dynamically (for example reading fro tsconfig)


Fishbowler:
Rebased to include a test fix on `main`

sidferreira:
> Rebased to include a test fix on `main`

ty!

maestro-cli/src/test/java/maestro/cli/util/WorkingDirectoryTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    data class ResolveTestCase(
+        val inputPath: String,
+        val expectedPath: String
+    )
+
+    companion object {
+        private val testPathAliases = mapOf(
+            "@proj" to "libs/screens",
+            "@libs" to "src/components"
```
Fishbowler: Are these weird mappings?

sidferreira: common uses in my experience... some use `~` or `!`

Fishbowler: No, I mean that "proj" maps to a directory in libs, but "libs" maps elsewhere.

sidferreira: oh, yah, that's AI slob

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
```diff
@@ -649,7 +649,11 @@ data class YamlFluentCommand(
     }
 
     private fun resolvePath(flowPath: Path, requestedPath: String): Path {
-        val path = flowPath.fileSystem.getPath(requestedPath)
+        val resolvedAliasPath = maestro.orchestra.workspace.WorkspaceConfigProvider.workspaceConfig?.let { 
+            maestro.orchestra.workspace.PathResolver.resolve(requestedPath, it)
```
Fishbowler: This function is used in a few places - runFlow, runScript, retry. 
`runScript` has me worried - will this "just work" with the new placeholders?

The only other place we've got paths is in AddMedia, but that already has JS interpolation, and files are read later. Given you've got a good grip on the use cases here, should that also work with placeholders?

sidferreira: So, the way I envision it is allowing devs to write like this:

```
# config.yaml
- pathAliases:
    @monorepo/login-screen: src/screens/login/src

#test.yaml
- runFlow: @monorepo/login-screen/e2e/login-fail.yaml
```

Let me check about addMedia

Fishbowler: Would be good to add a test for runScript. I _think_ it'll just work.

sidferreira: @Fishbowler just in case, I added `interpolation` to runScript/runFlow and `pathAlias` to addMedia... and tests

Fishbowler:
I'm unsure about the tests.

Take a look at IntegrationTest.kt where you see stuff happening against a FakeDriver.

I'm not sure that your PathAliasResolutionEndToEndTest is quite as EndToEnd as I'd consider that term to mean. It's pretty similar to the integration test you've written, but with some files already on disk.

More problematic is that I can't actually get this to work.

[path-substitution.zip](https://github.com/user-attachments/files/22874410/path-substitution.zip)

Here's a trivial workspace where the flows folder contains tests and other files are elsewhere. The path substitution appears to be doing string substitution and not finding paths relative to the config, but relative to the file in question, which makes this hard to use for a project with varied depth. Am I misunderstanding? Or were you expecting absolute paths there?

sidferreira:
@Fishbowler I'll rebase to main and slowly adding the commits to find out any issues. I must admit I'm relying a lot in AI cause my main dev env has some limitations to access NPM/YARN packages

sidferreira:
@Fishbowler I rewrote it cause it was too complex... new one seems to be ligher and cleaner

maestro-orchestra/src/test/resources/workspaces/016_path-aliases_config-in-folder/workspace/.maestro/config.yaml
Fishbowler, line 1: The established convention here is that `resources/workspaces` contains isolated workspaces.

Maybe refactor to 016a and 016b?

maestro-orchestra/src/test/java/maestro/orchestra/PathAliasesIntegrationTest.kt
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Integration tests for path aliases and JavaScript interpolation in runFlow and runScript commands.
+ * 
+ * These tests verify that:
+ * 1. Path aliases (like ~js, ~sf) are correctly resolved
+ * 2. JavaScript interpolation in file paths (like ${output.var}) works
+ * 3. Both features can be combined
+ */
+class PathAliasesIntegrationTest {
```
Fishbowler: What's the purpose of separating this from PathAliasesRuntimeTest? They don't use different approaches to their validation.

PR #2770, opened 2025-10-09
chore: added trial start prompt event

proksh:
Added `maestro_cli_trial_start_prompted` event for posthog

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -660,6 +660,68 @@ class ApiClient(
         }
     }
 
+    fun getOrgs(authToken: String): List<OrgResponse> {
+        val url = "$baseUrl/v2/maestro-studio/orgs"
+      
+        val request = Request.Builder()
+            .header("Authorization", "Bearer $authToken")
+            .url(url)
+            .get()
+            .build()
+
+        val response = try {
+            client.newCall(request).execute()
+        } catch (e: IOException) {
+            throw ApiException(statusCode = null)
+        }
+
+        response.use {
+            if (!response.isSuccessful) {
+                throw ApiException(
+                    statusCode = response.code
+                )
+            }
+            val responseBody = response.body?.string()
+            try {
+                val orgs = JSON.readValue(responseBody, object : TypeReference<List<OrgResponse>>() {})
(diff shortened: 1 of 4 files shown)


maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlCommandReaderTest.kt
```diff
@@ -657,7 +657,7 @@ internal class YamlCommandReaderTest {
         assertThat(tapCommand.selector.textRegex).isEqualTo("Submit")
         assertThat(tapCommand.selector.enabled).isTrue()
         assertThat(tapCommand.relativePoint).isEqualTo("25%, 75%")
-        assertThat(tapCommand.originalDescription).isEqualTo("Tap on \"Submit\" at 25%, 75%")
+        assertThat(tapCommand.originalDescription).isEqualTo("Tap on \"Submit\", enabled at 25%, 75%")
```
steviec, line 660: Why did this change?

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -426,7 +427,7 @@ class ApiClient(
         println("Starting your trial...")
         val url = "$baseUrl/v2/start-trial"
 
-        val jsonBody = """{ "companyName": "$companyName", "referralSource": "maestro-cli" }""".toRequestBody("application/json".toMediaType())
+        val jsonBody = """{ "companyName": "$companyName" }""".toRequestBody("application/json".toMediaType())
```
steviec: Anytime we're handwriting json I get nervous. Can we use a better serialization approach here?

maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlCommandReaderTest.kt
proksh: It was just a change that broke the test. Should already be merged now on main: https://github.com/mobile-dev-inc/Maestro/pull/2772

So you can ignore this.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
proksh: True, updated. :)

PR #2771, opened 2025-10-09
feat: added org and project_id selectors

proksh:
Added org and project selectors in case they are not specified.

1. Case where none are specified
https://www.loom.com/share/fedd497a00a34307b5b4213a4b6e0e08

2. Case where either of api key is specified and when both api key and project id is specified.
https://www.loom.com/share/419a0d633c0b4f31872e56373ea7b743

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -252,7 +252,6 @@ class CloudInteractor(
         }
 
         return when (orgs.size) {
-            0 -> { authToken }
             1 -> {
                 val org = orgs.first()
                 PrintUtils.message("Using organization: ${org.name} (${org.id})")


Fishbowler:
Docs upon release: https://app.gitbook.com/o/zCVYm3M93B0sOcjR1Oj4/s/n5KVIOjVkVjYRyVWZ0yT/~/diff/~/changes/596 (sorry casual readers - this one's behind auth)

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -210,35 +213,87 @@ class CloudInteractor(
...
+            client.getOrgs(authToken)
+        } catch (e: ApiClient.ApiException) {
+            throw CliError("Failed to fetch organizations. Status code: ${e.statusCode}")
+        } catch (e: Exception) {
+            throw CliError("Failed to fetch organizations: ${e.message}")
+        }
+
+        if (orgs.isEmpty()) {
+            throw CliError("No organizations found. Please create an organization first at https://console.mobile.dev")
+        }
```
steviec: If they must be logged in to get to this point, don't they always have an org? If it's possible for them not to have an org, can we just use our existing prompt to let them create one?

proksh: True, removed this error.

We already trigger sign in which adds the org and project if there is none on backend.

PR #2775, opened 2025-10-11
Add SetPermissions command

### LINKED ISSUE #2702: Separate permission config from launchApp
### Use case

Right now I'm doing this, which wastes a bit of time:
```yaml
- clearState
- launchApp:
    permissions: { all: allow }
- killApp

# We re-open with deep link to control the env.
# If we just launch the app, we can't set the env.
# If we just open the link, Maestro won't accept permission dialogs.
- openLink: 'myscheme://my/link/has/?data=to&speed=up&test=setup'
# iOS might ask the first time you open a deep link 'Open in “.*”\?'
- tapOn:
    text: Open
    optional: true
```

If I just launch the app using `openLink`, I don't get all of the neat automatic permission handling that `launchApp` offers. This dance of `launchApp`, `killApp` and finally `openLink` gives us the permission handling that `launchApp` offers, but it takes a while to run.

### Proposal

Some ways of addressing this:
1. Lift out the permission handling from `launchApp` to the flow level, like so:
```yaml
appId: com.android.contacts
permissions:
  all: allow
---
- launchApp
```
2. Alternatively, add a permissions option to `openLink`:
```yaml
appId: com.google.maps
---
- clearState
- openLink: comgooglemaps://?center=40.765819,-73.975866&zoom=14&views=traffic
  permissions:
    location: allow
```

### Anything else?

Love the tool, keep up the good work! ❤️

Fishbowler:
## Proposed changes

Adds a `setPermissions` command. Like `clearState` and `clearKeychain`, you can now set permissions independently of launching an app. This is useful for deeplink testing,

## Testing

Unit test, integration tests, and local testing with Android

## Issues fixed

Fixes #2702

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -926,9 +926,11 @@ class AndroidDriver(
 
     private fun setPermissionInternal(appId: String, permission: String, permissionValue: String) {
         try {
-            dadb.shell("pm $permissionValue $appId $permission")
+            shell("pm $permissionValue $appId $permission")
         } catch (exception: Exception) {
-            /* no-op */
+            // We don't need to be loud about this. IOExceptions were caught in shell. Remaining issues are likely due
+            // to "all" containing permissions that the app doesn't support.
+            logger.debug("Failed to set permission $permission for app $appId: ${exception.message}")
         }
     }
 


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -520,6 +520,24 @@ data class LaunchAppCommand(
...
+    override val originalDescription: String
+        get() = "Set permissions"
+
+    override fun evaluateScripts(jsEngine: JsEngine): SetPermissionsCommand {
+        return copy(
+            appId = appId.evaluateScripts(jsEngine),
+            label = label?.evaluateScripts(jsEngine)
+        )
+    }
+}
```
amanjeetsingh150, line 539: Why have this as a command I'd prefer to have this somewhere like platform config either upload wide or flow specific?

We already have platform config for this type of stuff.

I'd even say we should remove the coupling of this from launch app. We should start doing this internally in driver as a setup.

Fishbowler: A flow can contain different permutations of permissions during its run. I wouldn't mind inheriting defaults from flow and workspace config, but I'd absolutely want to retain the ability to change my permissions mid-flow.

This exists for folks doing openLink tests, where they don't necessarily want to have to jump thought launchApp then stopApp to know that their links will launch the app correctly. Imagine a maps link or something, where location is absolutely something you'd want to sort before the link opens.

Fishbowler:
Rebased.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1017,6 +1022,16 @@ class Orchestra(
         return true
     }
 
+    private fun setPermissionsCommand(command: SetPermissionsCommand): Boolean {
+        try {
+            maestro.setPermissions(command.appId, command.permissions)
+        } catch (e: Exception) {
+            throw MaestroException.UnableToSetPermissions("Unable to set permissions for app ${command.appId}: ${e.message}", e)
```
amanjeetsingh150, line 1029: This has a risk of breaking on cloud, because applesimutils is deprecated and I remember it fails everytime. I'll dev test this on cloud first.

We should then address the problems in setPermissions first IMO.

Fishbowler: This is running the same code as launchApp, split out as a convenience. I don't expect this to be any more problematic in Cloud than the launchApp. I think fixing how we do permissions is worthy, but isn't this change.

amanjeetsingh150: So, the risk here is the newly added exception. The behaviour changes by throwing failure explicitly here that was not happening before.

I would either suggest to keep the behaviour same like before and just doing setPermissions call. Otherwise we can pair and test on a cloud host before.

Fishbowler: Just above this code is the launch app code which would have explicitly thrown `UnableToClearState` when `maestro.setPermissions` threw. I've updated that to be more accurate, then duplicated it here. 

I don't think there's any new trapping around applesimutils failing that wasn't there before. Just more readable versions of what was already there.

amanjeetsingh150: > I don't think there's any new trapping around applesimutils failing that wasn't there before.

This is something that happens on cloud due to old os versions and not updating our strategy here. Anyways I missed the UnableClearState one you pointed. Cool yeah! lets get this in.

Upsilon-Iridani:
Nice! 🥳

PR #2780, opened 2025-10-13
Store random output

Fishbowler:
## Proposed changes

When we use random commands, store what was generated for use later.

```
- inputRandomPersonName
- evalScript: ${console.log(maestro.randomText)}
```

## Testing

Tested locally, and added IntegrationTests

## Issues fixed

None. This was something I wanted :)


Fishbowler:
Draft until I do a docs PR

PR #2794, opened 2025-10-16
chore: match posthog events and properties

proksh:
Match posthog events and properties to Maestro studio

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -431,7 +431,7 @@ class ApiClient(
         println("Starting your trial...")
         val url = "$baseUrl/v2/start-trial"
 
-        val request = StartTrialRequest(companyName)
+        val request = StartTrialRequest(companyName, referralSource = "cli")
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
steviec, line 129: @proksh : now that we're adding some more complex logic to our Posthog tracking, I'm wondering if we need a way to determine if our tracking system is _failing_. From what I can tell, we just silently ignore failures (we obviously don't want to break CLI), but the flip side is that we won't know if a certain percentage of event calls are dying. Any thoughts? Could we use Sentry here?

steviec (approved):
Looks great, I added one comment that we might want to save for a different PR.

maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
proksh: Yeah make sense.
But we don't have Sentry setup on CLI side, and setting it up then configuring it to reduce noise will take time. 😕 What do you think about that? I have added a linear ticket - https://linear.app/mobile-dev/issue/MA-3746/track-posthog-tracking-errors-on-sentry. We can discuss this on standup/weekly.

PR #2801, opened 2025-10-23
feat: add timeout property for conditions

HannesArni:
## Summary
This PR adds support for a timeout property on conditions used in `while` and `when` conditions for loops or conditional commands

## Motivation
Currently, conditions inherit timeout values from global configuration or command-level settings. I've been writing tests with Maestro, and tend to add these optional handling flows that can correct the app if it's in an unexpected state. For example:
```yaml
- runFlow:
    label: Login, if not already logged in
    when:
      visible: 'Login'
    commands:
      - tapOn: 'Login'
```
This is treated as an optional condition, which defaults to 7 seconds, so the test would wait for 7 seconds before continuing when the login screen is not presented. 

With these changes, we'll be able to specify a short timeout, because we already expect the app to be in a settled state when this condition is triggered:
```yaml
- runFlow:
    label: Login, if not already logged in
    when:
      visible: 'Login'
      timeout: 500
    commands:
      - tapOn: 'Login'
```

## Changes
- Added timeout: Long? property to Condition data class
- Updated YamlCondition to accept timeout parameter in YAML
- Modified Orchestra.evaluateCondition() to respect condition-level timeout with proper fallback chain
- Added integration test (Case 131) validating timeout behavior

## Tests
Added `131_repeat_while_timeout.yaml` test flow, as well as tests in our tests suite

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -4015,6 +4015,51 @@ class IntegrationTest {
         // No test failure - if we reach this point, the test passed successfully
     }
 
+    @Test
+    fun `Case 131 - Repeat while with custom timeout`() {
+        // Given
+        val commands = readCommands("131_repeat_while_timeout")
+        val driver = driver {
+            var counter = 0
+
+            val counterView = element {
+                text = "Value 0"
+                bounds = Bounds(0, 100, 100, 100)
+            }
+
+            element {
+                text = "Always Visible Element"
+                bounds = Bounds(0, 100, 100, 100)
+            }
+
+            element {
+                text = "Button"
+                bounds = Bounds(0, 100, 100, 100)
+                onClick = {
+                    counter++
+                    counterView.text = "Value $counter"
+                }
+            }
+        }
+
+        // When
+        Maestro(driver).use {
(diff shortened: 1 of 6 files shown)


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4015,6 +4015,51 @@ class IntegrationTest {
...
+
+        // Then
+        // The "Always Visible Element" is always visible, so the while condition
+        // (notVisible: "Always Visible Element") should be false.
+        // With a 100ms timeout, it should check quickly and not run the loop.
+        // Button should never be tapped.
+        driver.assertEventCount(
+            Event.Tap(Point(50, 50)),
+            expectedCount = 0
+        )
```
Fishbowler, line 4060: Given the bounds (100x100, starting at 0,100), should this be 50,150 for the middle?

Fishbowler (commented):
I've added one comment on the test.

More generally - does this make `assertVisible` and `assertNotVisible` now directly equivalent to `extendedWaitUntil`?

Fishbowler:
@HannesArni Thoughts on the above?

Fishbowler:
I've assumed this is abandoned. Do please reopen if you've got some energy for it!

PR #2816, opened 2025-11-04
Fixing edge case for DependencyResolver when uploading to cloud

trindadetiago:
## Proposed changes

This PR handles a problem users were facing when uploading a flow to cloud that have dependencies. Specifically, It errored when multiple flows referenced the same file with different relative paths.

Example: a flow and a subflow in a subfolder both reference the same sub flow, so the relative path used are different.

In this case, the error that happened was of a duplicated file, where Workspace utils tried to add the same file more than once.

The fix for this was adding a normalization for the path during the dependency discovery.

## Testing

Included a test case scenario in Test file that covers this problem.

## Issues fixed

Cloud upload error

### DIFF
--- a/maestro-cli/src/test/kotlin/maestro/cli/util/DependencyResolverTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/util/DependencyResolverTest.kt
@@ -644,4 +644,51 @@ class DependencyResolverTest {
         assertThat(dependencies.any { it.fileName.toString() == "subflow.yaml" }).isTrue()
     }
 
+    @Test
+    fun `treats files with same name but different real paths as different dependencies`(@TempDir tempDir: Path) {
+        // Directory layout:
+        // tempDir/
+        //   main_flow.yaml
+        //   scripts/
+        //     createUser.js
+        //     usa/
+        //       createUser.js
+
+        // Create directories
+        tempDir.resolve("scripts").toFile().mkdirs()
+        tempDir.resolve("scripts/usa").toFile().mkdirs()
+
+        // Create two different scripts with the same filename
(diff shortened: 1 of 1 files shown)


maestro-cli/src/test/kotlin/maestro/cli/util/DependencyResolverTest.kt
```diff
@@ -594,4 +594,54 @@ class DependencyResolverTest {
...
+            ---
+            - runFlow: sub/subflow.yaml
+            - runScript: ./sub/../scripts/createAccount.js
+            """.trimIndent()
+        )
+
+        val dependencies = DependencyResolver.discoverAllDependencies(mainFlow)
+
+        // Expect exactly one instance of the script, despite two different references
+        val scriptCount = dependencies.count { it.fileName.toString() == "createAccount.js" }
```
steviec, line 639: Let's add another test (or augment this test) to guarantee that if a dependency has the same name _but a different real path_ that it's treated as a different dependency. For example, what if a user had:
```
scripts/createUser.js
scripts/usa/createUser.js
```

And they had a script that referenced both, for whatever reason, we should ensure that they are treated as different dependencies. I think your code already does that, but your tests don't confirm it.

steviec: And first make sure that the workflow tests pass too :).

trindadetiago: > And first make sure that the workflow tests pass too :).

The tests that are failing are unrelated. I see that the last PR's error out in the same. 
Example in the last PR merged: https://github.com/mobile-dev-inc/Maestro/actions/runs/19067178519/job/54461177510

Issue #2703, opened 2025-09-05
Maestro should detect ports already in use, rather than give obscure errors


Fishbowler:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Run something to listen on port 7001. Then do something like `maestro test` or `maestro studio`.

### Actual results

You get errors like described in #1125

### Expected results

Do the thing, or explain to the user neatly why you can't.

### About app

n/a

### About environment

n/a

### Logs

n/a

### Maestro version

2.0.2

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

### Anything else?

This feels like a nice QoL improvement, but moreover, it's normally a first-time user hitting this.

Issue #2556, opened 2025-06-26
Running 2 different Tests on 2 Devices on the Same Machine


Revono:
### Use case

Hi all,

I'm trying to run two different tests simultaneously on the same machine, each targeting a different locally connected ADB device.

Yes know there is the "shards" feature but with this i can't specify a device and a specific test.

However, it seems like the issue is due to the connection already being in use. Our use case involves starting different tests on multiple ADB-connected devices at the same time—not duplicating the same test.

Unfortunately, Maestro seems unable to handle this. I consistently get the following error:

Maestro can't handle this i get always an error like this:
`Exception in thread "pool-4-thread-1" java.io.IOException: Command failed (tcp:7001): closed
	at dadb.adbserver.AdbServer.send$dadb(AdbServer.kt:103)
	at dadb.adbserver.AdbServerDadb.open(AdbServer.kt:149)
	at dadb.forwarding.TcpForwarder.handleForwarding$lambda-1(TcpForwarder.kt:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
`

i know on the screenshot is the same test but it doesn't matter, mostly I don't use the same test.

<img width="1466" alt="Image" src="https://github.com/user-attachments/assets/4fdb2363-3aa2-4903-a448-97cb33927146" />

Has anyone successfully run multiple parallel tests like this? Or is there a known workaround for handling multiple ADB sessions concurrently with Maestro?

I tried fancy stuff like this but without success (run in different terminals):
maestro --device ROF4P3BXQ test --shard-all 1 validate_all_settings_available.yaml
maestro --device ROSCIXF16 test --shard-all 1 validate_all_settings_available.yaml

Another Error i see sometimes:

`io.grpc.StatusRuntimeException: UNAVAILABLE: io exception                                                    
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:275)                           
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:256)                                       
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:169)                                  
[...]

PR #2339, opened 2025-02-28
Fixes Root cause of sharding race condition and allows driver to run on a custom port

avinash-bharti:
**This PR has two parts**
1. **Parallel session execution**
    1. Fixes The root cause of
        1. #1853
        2. #1867
    2. Add support for
        1. #1485
        2. #1818 
3. **Adds --driver-host-port support for Driver server similar to --port for dadb server.**
--------------------------
#### Steps to reproduce
1. Run parallel sessions on different devices using multiple instance of maestro cli.
    1. Test will run in one instance but will fail in another due to grpc io exception.
Screenshot 2025-02-28 at 9 23 57 AM
4. This same issue was faced for sharding flow in this issue #1853 and was fixed by a hack in this PR #1867

#### Root cause of the issue
1. Maestro test execution logs of the errored session shows that Issue happens when maestro makes first call to the driver server on mobile device.
```
09:23:13.850 [ INFO] maestro.Maestro.invoke: Getting device info
09:23:23.295 [ERROR] maestro.cli.runner.TestRunner.runCatching: Failed to run flow
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:275)
```
2. On checking logcat log of the device, found that for this errored session no driver app was installed and no instrumentation command was triggered and the device driver didn't receive the ``device info`` request.
3. This is happening because of `connectToExistingSession` feature which reconnects to existing session in case of some interruption or retries of session execution occurs.
    1. How `connectToExistingSession` feature  works?
        1. This feature was implemented before introducing sharding feature  #1732 
        2. On every new session creation it checks in keystore (~/.maestro/sessions) if any session other than current one exists that were active in last 21 second. If yes, **then skip setting up the driver on the device** (Assuming driver was set already in last session) #575
    2. Why this feature doesn't goes well with sharding and create issues?
        1. Sharding runs parallel sessions on **different devices**
        2. Due to `connectToExistingSession` it skips setting up driver on devices of session which were created after the first session initialisation of the shard (race condition).
            1. Heartbeat creates entry of session_platform key and timestamp as value on session creation and keeps updating it every 5 sec during whole session lifecycle.
        3. Since the device of the second session doesn't have the android server setup, grpc call couldn't find any server on that particular port and gives  io exception.
    3. Why PR #1867 (Adding a delay of 1 sec while recording first heartbeat) solves the issue?
[...]

Issue #2082, opened 2024-10-08
 [Question] [1.37.9] -  Running tests on two different devices


dannmf:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

- Connect two devices to the machine
- Open two terminals
- Run test routine on both terminals

### Actual results

My doubt is related to running tests on different devices, in this case I tried opening different consoles and specifying the device ID, but an io exception is returned.
Do you have a future resolution or resource to perform this type of execution simultaneously on different devices?

### Expected results

The expected result would be to be able to run different test scripts on two or more devices simultaneously

### About app

Closed source application, using Flutter in its architecture

### About environment

Ubuntu 22.04.2 LTS

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

1.37.9

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

reinvent-heucles:
I am having the same issue here !!!

reinvent-heucles:
@dannmf I was able to workarround it using Docker and changing adb to TCP mode, I intend to polish the solution and publish it, but if you'd like, let me know and I share it with you

dannmf:
"That's great, I'd love to know how you solved it. If you’re able to share, I’d really appreciate it. I'm actually no longer at the company that uses Maestro for testing, but I'll pass it along to my colleagues who are still on the team so they can implement the solution.

omnarayan:
I've built on @avinash-bharti's work from #2339 and created an unofficial release with `--driver-host-port` support as a stopgap while [PR #2821](https://github.com/mobile-dev-inc/maestro/pull/2821) is under review.

  **Upgrade:**
```bash
  curl -fsSL https://raw.githubusercontent.com/omnarayan/Maestro/feature/driver-host-port-distribution/upgrade-maestro-ports.sh | sh
```

  **Usage:**
```bash
  maestro --driver-host-port 7005 --device device1 test flow.yaml
  maestro --driver-host-port 7006 --device device2 test flow.yaml
```
  **Restore:**  `bash ~/.maestro-backups/backup/restore.sh`

  📦 https://github.com/omnarayan/Maestro/releases/tag/fork-2.0.9-ports

PR #2821, opened 2025-11-08
Enable concurrent testing on multiple devices/simulators from same machine

### LINKED ISSUE #1485: [Feature Request] Run maestro on multiple devices at the same time on a single computer
A computer can only run one maestro even if there are multiple devices connected, and if i want to run multiple maestro, i need multiple computers, which is a waste of resources.

Run maestro on multiple devices at the same time on a single computer. If multiple devices are connected to the PC, run the ‘-device’ command to select a specific device to run the maestro on.

omnarayan:
## Proposed changes

<copilot:summary>
Introduces a new global `--driver-host-port` CLI option that allows users to specify custom driver ports for both Android and iOS, enabling parallel test execution on multiple devices/simulators without port conflicts.

### Problem
Currently, running tests concurrently on different devices fails because all Maestro instances attempt to use the default port 7001, resulting in:
- `Connection refused: localhost:7001`
- `Command failed (tcp:7001): closed`

Users cannot run different tests on different devices simultaneously.

### Solution
- Added `--driver-host-port` global CLI option to `App.kt`
- Modified `TestCommand.kt` to respect user-specified port in `selectPort()` method
- Updated all commands (`RecordCommand`, `StudioCommand`, `QueryCommand`, `PrintHierarchyCommand`, `StartDeviceCommand`) to pass custom port to session manager
- Maintains backward compatibility - defaults to 7001 if not specified
- Works for both Android devices and iOS simulators

### Usage Example
```bash
# Terminal 1: Run auth tests on device 1 with port 7005
maestro --driver-host-port 7005 --device device1 test auth-suite.yaml

# Terminal 2: Run product tests on device 2 with port 7006
maestro --driver-host-port 7006 --device device2 test product-suite.yaml
```

<details>

<summary>Try it now (before merge)</summary>

To use this feature, you have three options:

1. **Wait for PR merge** - This PR needs to be reviewed and merged
2. **Build from source** - Clone this branch and build Maestro yourself
3. **Use pre-built version** - Install the updated JAR directly:

**Upgrade:**
```bash
curl -fsSL https://raw.githubusercontent.com/omnarayan/Maestro/feature/driver-host-port-distribution/upgrade-maestro-ports.sh | sh
```

This replaces your Maestro JAR with the updated version. Usage:
```bash
maestro --driver-host-port 7005 --device device1 test flow.yaml
maestro --driver-host-port 7006 --device device2 test flow.yaml
```

**Restore Original:**
```bash
bash ~/.maestro-backups/backup/restore.sh
```

</details>

### Credit
This implementation builds upon the approach proposed by @avinash-bharti in #2339, implementing the port configuration feature without the additional session management changes.

### Testing
- ✅ Verified concurrent execution on physical Android device (port 7005) + emulator (port 7006) running different test suites simultaneously
- ✅ Confirmed no session mixing between devices - each test ran on its intended device with correct app
- ✅ Backward compatibility verified: default port 7001 works without the flag
- ✅ Tested on both Android devices and iOS simulators
- ✅ All test steps completed successfully with no port conflicts

Test logs show both devices executing tests in parallel without interference.

### Issues fixed
Resolves #1485, #1853, #2082, #2556, #2703
Related to #1818

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -51,6 +51,7 @@ import maestro.cli.util.CiUtils
 import maestro.cli.util.EnvUtils
 import maestro.cli.util.FileUtils.isWebFlow
 import maestro.cli.util.PrintUtils
+import maestro.cli.util.isPortAvailable
 import maestro.cli.insights.TestAnalysisManager
 import maestro.cli.view.greenBox
 import maestro.cli.view.box
@@ -505,12 +506,23 @@ class TestCommand : Callable<Int> {
 
     private fun selectPort(effectiveShards: Int): Int {
         // If user specified driver host port via CLI, use it
-        parent?.driverHostPort?.let { return it }
+        parent?.driverHostPort?.let { port ->
+            if (!isPortAvailable(port)) {
+                throw CliError("Port $port is already in use. Please specify a different port with --driver-host-port")
+            }
+            return port
+        }
 
         // Otherwise use default behavior
-        return if (effectiveShards == 1) 7001
-        else (7001..7128).shuffled().find { port ->
-            usedPorts.putIfAbsent(port, true) == null
+        if (effectiveShards == 1) {
+            if (!isPortAvailable(7001)) {
+                throw CliError("Default port 7001 is already in use. Use --driver-host-port to specify a different port")
+            }
+            return 7001
+        }
+
+        return (7001..7128).shuffled().find { port ->
+            usedPorts.putIfAbsent(port, true) == null && isPortAvailable(port)
         } ?: error("No available ports found")
     }
 
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -503,11 +503,16 @@ class TestCommand : Callable<Int> {
         }
     }
 
-    private fun selectPort(effectiveShards: Int): Int =
-        if (effectiveShards == 1) 7001
+    private fun selectPort(effectiveShards: Int): Int {
+        // If user specified driver host port via CLI, use it
+        parent?.driverHostPort?.let { return it }
+
+        // Otherwise use default behavior
+        return if (effectiveShards == 1) 7001
         else (7001..7128).shuffled().find { port ->
             usedPorts.putIfAbsent(port, true) == null
         } ?: error("No available ports found")
+    }
 
     private fun runSingleFlow(
         maestro: Maestro,
--- a/maestro-cli/src/main/java/maestro/cli/App.kt
+++ b/maestro-cli/src/main/java/maestro/cli/App.kt
@@ -91,6 +91,12 @@ class App {
     var port: Int? = null
 
     @Option(
+        names = ["--driver-host-port"],
+        description = ["AndroidDriver host port for instrumentation communication (default: 7001)"]
+    )
+    var driverHostPort: Int? = null
+
+    @Option(
         names = ["--device", "--udid"],
         description = ["(Optional) Device ID to run on explicitly, can be a comma separated list of IDs: --device \"Emulator_1,Emulator_2\" "],
     )
--- a/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
@@ -109,7 +109,7 @@ class PrintHierarchyCommand : Runnable {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = null,
+            driverHostPort = parent?.driverHostPort,
             teamId = appleTeamId,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
--- a/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
@@ -73,7 +73,7 @@ class QueryCommand : Runnable {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = null,
+            driverHostPort = parent?.driverHostPort,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
             teamId = appleTeamId,
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -130,7 +130,7 @@ class RecordCommand : Callable<Int> {
         return MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = null,
+            driverHostPort = parent?.driverHostPort,
             deviceId = deviceId,
             teamId = appleTeamId,
             platform = parent?.platform,
--- a/maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
@@ -99,7 +99,7 @@ class StartDeviceCommand : Callable<Int> {
                 PrintUtils.message(if (p == Platform.IOS) "Launching simulator..." else "Launching emulator...")
                 DeviceService.startDevice(
                     device = device,
-                    driverHostPort = parent?.port
+                    driverHostPort = parent?.driverHostPort
                 )
             }
         } catch (e: LocaleValidationIosException) {
--- a/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/StudioCommand.kt
@@ -75,7 +75,7 @@ class StudioCommand : Callable<Int> {
         MaestroSessionManager.newSession(
             host = parent?.host,
             port = parent?.port,
-            driverHostPort = null,
+            driverHostPort = parent?.driverHostPort,
             teamId = appleTeamId,
             deviceId = parent?.deviceId,
             platform = parent?.platform,
--- a/maestro-cli/src/main/java/maestro/cli/util/SocketUtils.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/SocketUtils.kt
@@ -10,3 +10,11 @@ fun getFreePort(): Int {
     }
     ServerSocket(0).use { return it.localPort }
 }
+
+fun isPortAvailable(port: Int): Boolean {
+    return try {
+        ServerSocket(port).use { true }
+    } catch (e: Exception) {
+        false
+    }
+}


Fishbowler:
I noted that you've said this resolves #2703 - if you manually specify a port that's already in use, how does this behave?

omnarayan:
> I noted that you've said this resolves #2703 - if you manually specify a port that's already in use, how does this behave?

  Good catch on the earlier gap! I've just pushed a fix for this.

  Now if you specify a port that's already in use:

  maestro --driver-host-port 7005 test flow.yaml

  You'll get a clear error:

  Port 7005 is already in use. Please specify a different port with --driver-host-port

  Same for the default port - if 7001 is occupied:

  Default port 7001 is already in use. Use --driver-host-port to specify a different port

  The port check happens right before the driver starts, so it catches conflicts at the point of use.

  We missed this earlier because our internal tooling checks port availability before passing --driver-host-port to Maestro, so we weren't hitting this scenario directly. Apologies for the oversight.

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -476,11 +477,27 @@ class TestCommand : Callable<Int> {
...
+        if (effectiveShards == 1) {
+            if (!isPortAvailable(7001)) {
+                throw CliError("Default port 7001 is already in use. Use --driver-host-port to specify a different port")
+            }
+            return 7001
+        }
+
+        return (7001..7128).shuffled().find { port ->
+            usedPorts.putIfAbsent(port, true) == null && isPortAvailable(port)
         } ?: error("No available ports found")
```
Fishbowler, line 526: Should we do the same thing that you've done to the other messages, and make the feedback to the user really actionable here?

(Not a hill I'll die on - I suspect if you can't find any of these ports, there's something else amiss)

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -476,11 +477,27 @@ class TestCommand : Callable<Int> {
...
+        // Otherwise use default behavior
+        if (effectiveShards == 1) {
+            if (!isPortAvailable(7001)) {
+                throw CliError("Default port 7001 is already in use. Use --driver-host-port to specify a different port")
+            }
+            return 7001
+        }
+
+        return (7001..7128).shuffled().find { port ->
+            usedPorts.putIfAbsent(port, true) == null && isPortAvailable(port)
```
Fishbowler, line 525: Will this evaluate consecutively or concurrently?
Does it matter if the left succeeds then the right fails? 🤔

Fishbowler: (GitHub hasn't rendered this well - I'm referring to the two tests on L498)

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
omnarayan: Extending this thought — with parallel/sharding, there's an interesting failure mode: what if we need 2 ports but only get 1? Do we rollback the first allocation and fail entirely, or run with fewer shards than requested?

Either way, expanding the range beyond 7001-7128 would reduce the probability significantly. It's a custom internal server, nothing stopping us from using 10000-49000.

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
omnarayan: It evaluates left-to-right (short-circuit). So if `putIfAbsent` succeeds but `isPortAvailable` fails, the port is marked as "used" in the map but never actually used — it's leaked from the pool.

But even fixing the order doesn't help — there's still a race window between checking and actually binding.

The atomic solution is to attempt the bind directly:
```kotlin
fun findAvailablePort(): Int {
    return (7001..7128).shuffled().firstNotNullOfOrNull { port ->
        try {
            ServerBuilder.forPort(port)
                .addService(maestroService)
                .build()
                .start()  // Binds atomically — if it works, we own it
            port
        } catch (e: IOException) {
            null  // Port unavailable, try next
        }
    } ?: error("No available ports in range 7001-7128")
}
```

No check-then-use, the bind IS the check.

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
Fishbowler: I think if 128 ports are in use (or at least, can't be bound) maybe you've got other problems?

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
Fishbowler: Maybe if the port can't be bound, we _should_ put into the `usedPorts` map, so as not to retry it?
If so, this is good as-is?

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
Fishbowler: What are the chances that someone sees this message and actually might've been able to use Maestro for sharding if we'd used a wider port range? I'm concerned that someone who can't bind might be waiting a long time for feedback if we test 39000 ports on some locked down machine.

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
omnarayan: Or start the next cycle from the last found port. There are more ways to optimize, the question is, what defines over-optimization?

maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
omnarayan: Fair point — 128 ports is a decent range.

The real optimization is avoiding sequential checks. `netstat -tuln `grabs all busy ports in one shot, so you instantly know what's free.

Fishbowler: I think we're agreed here that I'm talking nonsense, and this bit is fine as-is :)

Fishbowler:
Rebased

Fishbowler:
Spotted an e2e failure, but that's unrelated - opened #2920 to solve that

Fishbowler (changes requested):
I had a little test of this.

This works:
1. `maestro --platform android test ...`
2. `maestro --driver-host-port 7010 --platform ios test ....`

This doesn't work:
1. `maestro --driver-host-port 7010 --platform ios test ....`
2. `maestro --platform android test ...`

Maybe first command locks 7001 as well as the custom port in use?

Inverting the platforms works, so perhaps specific to Android implementation?

omnarayan:
> I had a little test of this.
> 
> This works:
> 
> 1. `maestro --platform android test ...`
> 2. `maestro --driver-host-port 7010 --platform ios test ....`
> 
> This doesn't work:
> 
> 1. `maestro --driver-host-port 7010 --platform ios test ....`
> 2. `maestro --platform android test ...`
> 
> Maybe first command locks 7001 as well as the custom port in use?
> 
> Inverting the platforms works, so perhaps specific to Android implementation?

If you need my help, I'm happy to contribute — just need some time since it's not just a code change, it requires proper testing. But if you could take it forward, I'd be more than happy.

**Fix location:** `maestro-cli/src/main/java/maestro/cli/session/MaestroSessionManager.kt:106-111`

```kotlin
connectToExistingSession = if (isStudio) {
    false
} else if (driverHostPort != null) {
    // Custom port specified → skip session store check for all devices
    false  // Changed from 'true' to 'false'
} else {
    SessionStore.hasActiveSessions(
        sessionId,
        // ...
    )
}
```

varun-jupiter:
Hey @Fishbowler @omnarayan, I am a completely new to Maestro and I too have this use case. Stumbled upon this PR as a fix for it. I see you are just changing port selection in `selectPort`. I just randomised the port in just the test command and tried running 2 parallel sessions after having done `installDist`. 
` ./maestro --device d1 test /path/to/test --include-tags t1 --debug-output /path/to/output -e APP_ID=my-app`
`./maestro --device d2 test /path/to/test --include-tags t1 --debug-output /path/to/output -e APP_ID=my-app`
Both are android devices. 
However, I see this in the debug log. 
```
12:42:59.316 [ INFO] maestro.cli.runner.TestSuiteInteractor.runFlow:  Running flow flow1
12:42:59.322 [ INFO] maestro.Maestro.cachedDeviceInfo_delegate$lambda$0: Getting device info
12:42:59.416 [ERROR] maestro.drivers.AndroidDriver.runDeviceCall: Not able to reach the gRPC server while doing android device call
12:42:59.416 [ERROR] maestro.cli.runner.TestSuiteInteractor.runFlow: Failed to complete flow
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:275) ~[grpc-stub-1.57.2.jar:1.57.2]
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:256) ~[grpc-stub-1.57.2.jar:1.57.2]
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:169) ~[grpc-stub-1.57.2.jar:1.57.2]
	at maestro_android.MaestroDriverGrpc$MaestroDriverBlockingStub.deviceInfo(MaestroDriverGrpc.java:740) ~[maestro-client.jar:?]
	at maestro.drivers.AndroidDriver.deviceInfo$lambda$4(AndroidDriver.kt:212) ~[maestro-client.jar:?]
	at maestro.drivers.AndroidDriver.runDeviceCall(AndroidDriver.kt:1263) ~[maestro-client.jar:?]
	at maestro.drivers.AndroidDriver.deviceInfo(AndroidDriver.kt:211) ~[maestro-client.jar:?]
	at maestro.Maestro.cachedDeviceInfo_delegate$lambda$0(Maestro.kt:49) ~[maestro-client.jar:?]
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:86) ~[kotlin-stdlib-2.2.20.jar:2.2.20-release-333]
	at maestro.Maestro.getCachedDeviceInfo(Maestro.kt:47) ~[maestro-client.jar:?]
	at maestro.orchestra.Orchestra.initJsEngine(Orchestra.kt:301) ~[maestro-orchestra.jar:?]
	at maestro.orchestra.Orchestra.runFlow(Orchestra.kt:160) ~[maestro-orchestra.jar:?]
	at maestro.cli.runner.TestSuiteInteractor.runFlow(TestSuiteInteractor.kt:243) ~[maestro-cli-2.
```
Am I missing something here? I have been trying to figure out what exactly is happening here but its taking a while. 
Would be really helpful if one of you can point me in the right direction!

My change in `selectPort`:
```
    private fun selectPort(effectiveShards: Int): Int =
        if (effectiveShards == 1) {
            val portToUse = (7001..7128).random()
            logger.info("Port picked for usage: $portToUse")
            portToUse
        } else (7001..7128).shuffled().find { port ->
            usedPorts.putIfAbsent(port, true) == null
        } ?: error("No available ports found")
```

Jonatthu:
Why no one wants to merge this? or make this continue doing progress? :( chat apps would be amazing to test with this

varun-jupiter:
Ok so I dug a little more and as per @omnarayan's suggestions, looks like the android driver does not open if there are active sessions detected. 
I have made a change to accommodate this and its working for me now. 
Tested with 3 concurrent executions across 3 separate devices.
Will try and raise the change soon.

Fishbowler:
Thanks!

pamartineza:
I would like to add another usecase where this feature is very valuable:

We have a dedicated powerful Mac Studio Machine to run up to 6 jobs concurrently of a self-hosted Giltab CI runner (shell executor mode).

If two or more different branches are pushed and maestro tests are run, despite starting 2 or more different android emulators with different device ids, the maestro tests fail due to port collision.

Right now we have to run different jobs that execute maestro test in series rather than in parallel which is not an optimal experience.

d2callpod:
Is there a reason why this hasn’t been merged? This is a great feature/bug. We have a lot of cross device features that we would like to test.

Fishbowler:
We're not ignoring it. We're considering it alongside other changes. We're also working on Studio improvements, and we've been at a bunch of MCP changes recently.

How do you think this compares to #3138 ?

d2callpod:
> We're not ignoring it. We're considering it alongside other changes. We're also working on Studio improvements, and we've been at a bunch of MCP changes recently.
> 
> How do you think this compares to #3138 ?

I’ve done some testing on Android devices and it works well and is a lot simpler to use. There is no need for any configuration for #3138, so that is definitely my preference.

orennu:
when is this expected to be on offiical release?
our usecase is running maestro as ui test step from pytest since our e2e flow includes device management via adb/ios equivilant which is not supported by maestro.
we would like to run tests in parallel from the same machine for runtime optimization.

Fishbowler:
As above.

Fishbowler:
Superseded by #3138

PR #2822, opened 2025-11-10
chore: added maestro cloud promotional messages

proksh:
# "Get faster result" promotion

Show this message on top if:
- In case there are more than 5 test running
- This promotional message was not shown on the same day before
- Cloud command wasn't used with last 3 days

Screenshot 2025-11-07 at 5 12 49 PM

# "Easy Debug" promotion

Show this message on bottom if:
- Any flow was failed
- This promotional message was not shown on the same day before
- Cloud command wasn't used with last 3 days

Screenshot 2025-11-07 at 5 13 18 PM

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -47,6 +47,7 @@ import maestro.cli.runner.TestSuiteInteractor
 import maestro.cli.runner.resultview.AnsiResultView
 import maestro.cli.runner.resultview.PlainTextResultView
 import maestro.cli.session.MaestroSessionManager
+import maestro.cli.util.CiUtils
 import maestro.cli.util.EnvUtils
 import maestro.cli.util.FileUtils.isWebFlow
 import maestro.cli.util.PrintUtils
@@ -659,6 +660,11 @@ class TestCommand : Callable<Int> {
     }
 
     private fun showCloudFasterResultsPromotionMessageIfNeeded() {
+        // Don't show in CI environments
+        if (CiUtils.getCiProvider() != null) {
+            return
+        }
+        
         val promotionStateManager = PromotionStateManager()
         val today = LocalDate.now().toString()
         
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -642,4 +657,44 @@ class TestCommand : Callable<Int> {
...
+        val promotionStateManager = PromotionStateManager()
+        val today = LocalDate.now().toString()
+        
+        // Don't show if already shown today
+        if (promotionStateManager.getLastShownDate("fasterResults") == today) {
+            return
+        }
+        
+        // Don't show if user has used cloud command within last 3 days
+        if (promotionStateManager.wasCloudCommandUsedWithinDays(3)) {
```
amanjeetsingh150, line 677: Why we should even show if user is using maestro cloud command. I think we should not show this if they are cloud user.

Reason: Upselling of cloud should be part of cloud and not CLI for paid users.

proksh: I thought about this:

My point of view is:
- Using `maestro cloud` should be a regular part of workflow if someone is using cloud
- Even if user has used `maestro cloud`, if they are running large amount of test locally and haven't used cloud for some time, a prompt for that would be good

amanjeetsingh150: > Even if user has used maestro cloud, if they are running large amount of test locally and haven't used cloud for some time, a prompt for that would be good

I assumed the large amount might actually be the reason for choosing a cloud subscription in the first place. 

If we circle back to what we’re trying to solve here, it might get clearer — from my understanding, the goal was awareness to drive activation. But if the user is already activated, I’m not sure what problem this solves? 🤔

proksh: Discussed this on call. Conclusion is, This would be useful if:
- new user ran maestro cloud command just to try it out and then forgot about it
- old user who are running test locally frequently and not using cloud command, get a ping to do so once a day. As they are paying for it, it will be a good thing to convert them to run maestro cloud more often than maestro test.

Change needed:
- This shouldn't show up for CI, which is the case. Working on updating this.

PR #2827, opened 2025-11-12
feat: add sleep command

rasyid7:
## Proposed changes

add sleep command.
currently, there is no exact sleep command.
literally sleep or wait.

## Testing

the sleep itself is just simple `Thread.sleep(time)`
added serialization test.


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -573,6 +573,12 @@ class Maestro(
         ScreenshotUtils.waitUntilScreenIsStatic(timeout, SCREENSHOT_DIFF_THRESHOLD, driver)
     }
 
+    fun sleep(time: Long?) {
+        val time = time ?: ANIMATION_TIMEOUT_MS
+        LOGGER.info("Sleep for $time ms")
+        Thread.sleep(time)
```
Fishbowler: I think this should be using `MaestroTimer.sleep(...)`?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSleepCommand.kt
```diff
@@ -0,0 +1,14 @@
+package maestro.orchestra.yaml
+import com.fasterxml.jackson.annotation.JsonCreator
+data class YamlSleepCommand(
+    val time: Long?,
```
Fishbowler: Is there a better label for this?
Should this be in milliseconds? Is there a good use case for something more granular than seconds?

Would this be more readable?
```
- sleep
    seconds: 5
```

(that's intended as a real question, not a leading question)

maestro-client/src/main/java/maestro/Maestro.kt
rasyid7: @Fishbowler ah yes, i was following other implementations like here https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-client/src/main/java/maestro/Maestro.kt#L538 

but sure i will change it to `MaestroTimer`

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSleepCommand.kt
rasyid7: how about this?
```
-sleep: 5s
```
so we can put 5s or 5000ms
i prefer default to ms.

Fishbowler (commented):
This is looking awesome! Thanks so much!

Are there any other unit or integration tests we can add? Mostly thinking about the variations of the command syntax possible via YamlSleepCommand.

Let me know if this is something I can help with.

Fishbowler:
Oooh, we're gonna need a matching PR in https://github.com/mobile-dev-inc/maestro-docs too

rasyid7:
> This is looking awesome! Thanks so much!
> 
> Are there any other unit or integration tests we can add? Mostly thinking about the variations of the command syntax possible via YamlSleepCommand.
> 
> Let me know if this is something I can help with.

i have added more tests here https://github.com/mobile-dev-inc/Maestro/pull/2827/commits/a121f823162aa3a5eb05205a3fa2d3a8a7e7cae4

> Oooh, we're gonna need a matching PR in https://github.com/mobile-dev-inc/maestro-docs too

and here is the PR for the docs https://github.com/mobile-dev-inc/maestro-docs/pull/157

thanks!

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -573,6 +573,12 @@ class Maestro(
         ScreenshotUtils.waitUntilScreenIsStatic(timeout, SCREENSHOT_DIFF_THRESHOLD, driver)
     }
 
+    fun sleep(time: Long?) {
+        val time = time ?: ANIMATION_TIMEOUT_MS
```
Fishbowler: Just spotted this default. That's 15s. That's fine when we're waiting for elements to appear. Do you think it's a good default for this command?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -943,6 +943,20 @@ data class WaitForAnimationToEndCommand(
...
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+
+    override val originalDescription: String
+        get() = "Sleep for $time ms"
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return this
+    }
```
Fishbowler, line 1006: I wonder if we should make the timeout possible via expression too?
I've just done the same activity in #2864 - WDYT?

Fishbowler: Would probably mean that the parsing logic in YamlSleepCommand would need to move to a private function here too, I think?

maestro-orchestra/src/test/resources/YamlCommandReaderTest/031_sleep_commands.yaml
```diff
@@ -0,0 +1,5 @@
+appId: com.example.app
+---
+- sleep: 100
+- sleep: 1000ms
+- sleep: 1s
```
Fishbowler, line 5: We should test the default too

Fishbowler (commented):
Had another pass over this, and conjured some more comments. This is an often requested feature, and I want to be sure we get it right.

Do let me know if you can't finish this - I'm happy to help!

Fishbowler:
@rasyid7 Lemme know if you're stuck for time to finish this off, and I'll look to find some time

maestro-client/src/main/java/maestro/Maestro.kt
rasyid7: good question. i think it's better to remove the default timeout. thanks!

rasyid7:
> @rasyid7 Lemme know if you're stuck for time to finish this off, and I'll look to find some time

sure thing @Fishbowler !
if you believe there is a better approach to this, it would be great!
specifically this one https://github.com/mobile-dev-inc/Maestro/pull/2827#discussion_r2571202466

thank you in advance

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSleepCommand.kt
```diff
@@ -0,0 +1,42 @@
...
+        @JsonCreator(mode = JsonCreator.Mode.PROPERTIES)
+        fun fromObject(duration: String?, ms: Long?, seconds: Long?, label: String?): YamlSleepCommand {
+            val time = when {
+                duration != null -> parseDuration(duration)
+                ms != null -> ms
+                seconds != null -> seconds * 1000
+                else -> null
+            }
+            return YamlSleepCommand(time, label)
+        }
```
Fishbowler: What's the intention of this bit? To support a broader object?

e.g.

```
- sleep:
    duration: 4s
- sleep:
    ms: 5000
- sleep:
    seconds: 6
```

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSleepCommand.kt
```diff
@@ -1,42 +1,15 @@
...
-        private fun parseDuration(value: String): Long {
-            val trimmed = value.trim().lowercase()
-            return when {
-                trimmed.endsWith("ms") -> trimmed.removeSuffix("ms").toLong()
-                trimmed.endsWith("s") -> trimmed.removeSuffix("s").toLong() * 1000
-                else -> trimmed.toLong() // Assume ms for backward compatibility
-            }
+        fun parse(value: String): YamlSleepCommand {
+            return YamlSleepCommand(value, null)
         }
```
Fishbowler: This moved to Commands, and was replaced with a single regex. I've commented on that elsewhere...

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -988,7 +988,21 @@ data class SleepCommand(
...
+    private fun String.parseDuration(): String? {
+        return Regex("""^(\d+)(ms|s)?$""").matchEntire(this.trim().lowercase())
+            ?.let {
+                val value = it.groupValues[1].toLong() // Get the digits
+                when (it.groupValues[2]) { // If there's a duration unit
+                    "s"  -> (value * 1_000).toString()
+                    "ms", "" -> value.toString()
+                    else -> null // Unknown unit, ignore the digits and return null
+                }
+            } // When no match, we'll return null
```
Fishbowler, line 1005: I've moved the parseDuration to a single regex - it's faster and Kotlin idiomatic, but unsure if it's drastically reduced readability - WDYT @rasyid7 ?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlSleepCommand.kt
```diff
@@ -1,42 +1,15 @@
...
-
-        private fun parseDuration(value: String): Long {
-            val trimmed = value.trim().lowercase()
-            return when {
-                trimmed.endsWith("ms") -> trimmed.removeSuffix("ms").toLong()
-                trimmed.endsWith("s") -> trimmed.removeSuffix("s").toLong() * 1000
-                else -> trimmed.toLong() // Assume ms for backward compatibility
-            }
+        fun parse(value: String): YamlSleepCommand {
+            return YamlSleepCommand(value, null)
```
Fishbowler: The companion object has been drastically simplified now that parsing logic happens later on

maestro-orchestra/src/test/java/maestro/orchestra/yaml/YamlSleepCommandTest.kt
```diff
@@ -1,39 +0,0 @@
...
-import org.junit.jupiter.api.Assertions.assertThrows
-import org.junit.jupiter.api.Test
-
-class YamlSleepCommandTest {
-
-    @Test
-    fun `fromValue with Int`() {
-        val command = YamlSleepCommand.fromValue(100)
-        assertThat(command.time).isEqualTo(100L)
-    }
```
Fishbowler: Now that the parsing is later, these tests are gone, replaced with integration tests

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4148,6 +4148,77 @@ class IntegrationTest {
...
+            runBlocking {
+                orchestra(it).runFlow(commands)
+            }
+        }
+
+        // Then
+        // No test failure
+        fakeTimer.assertEvent(MaestroTimer.Reason.COMMAND_SLEEP, 1000L) // Env var
+        fakeTimer.assertEvent(MaestroTimer.Reason.COMMAND_SLEEP, 2000L) // Env var with duration unit
+        fakeTimer.assertEvent(MaestroTimer.Reason.COMMAND_SLEEP, 3000L) // JS calculation expression
```
Fishbowler, line 4259: Checking that the JS interpolated values actually turn into real-world values

Fishbowler (commented):
I've added the stuff for JS interpolation. I'll try to self-review my changes to show what I've done.

rasyid7:
hi @Fishbowler !
sorry for my late response, just came back from holiday 🙇 

awesome, thanks!
i'll fix the conflict.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
rasyid7: > I've moved the parseDuration to a single regex - it's faster and Kotlin idiomatic, but unsure if it's drastically reduced readability - WDYT @rasyid7 ?

awesome Dan!
not drastically reduced the readability, i'm fine with it.

PR #2845, opened 2025-11-24
Reduce disk size for Android integration tests

Fishbowler:
This fixes CI failures due to Android Emulators not launching - failing due to disk size limitations on the ubuntu GH runner.

https://github.com/ReactiveCircus/android-emulator-runner/issues/455 led me onto API 33 requiring more disk space than others, so bumped to API 34. (I tried reducing the space needed using `disk-size` but it either didn't work, or it wouldn't go smaller than the minimum required).

Whilst I was there, I fixed up some test artifact storage.

## Proposed changes

copilot:summary

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -60,7 +60,7 @@ jobs:
         if: failure()
         with:
           name: maestro-unit-test-report
-          path: ./maestro-test/build/reports
+          path: ./**/build/reports/tests/test
           retention-days: 1
           include-hidden-files: true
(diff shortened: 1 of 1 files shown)


Fishbowler:
It seems that Android API 34 is slightly flakier than API 33 for the integration tests. [Example](https://github.com/mobile-dev-inc/Maestro/actions/runs/19633960838/job/56220076783). Any suggestions?

amanjeetsingh150 (approved):

PR #2855, opened 2025-11-26
Fix/accessibility support

PATAKAMURIVENKATAGANESH:
## Proposed changes

copilot:summary


Fishbowler:
Thanks @PATAKAMURIVENKATAGANESH!

What's the real world effect of this? Or, put another way, how do I test it?

PR #2858, opened 2025-11-26
Legacy Studio - Dependency bumps

Fishbowler:
## Proposed changes

Bumps a few dependencies to solve this little badge:
image

## Testing

Ran `maestro studio` locally and had a little play. LGTM.

### DIFF
--- a/maestro-studio/web/package-lock.json
+++ b/maestro-studio/web/package-lock.json
@@ -55,7 +55,7 @@
         "ajv": "^6.12.6",
         "ajv-keywords": "^3.5.2",
         "autoprefixer": "^10.4.13",
-        "postcss": "^8.4.19",
+        "postcss": "^8.4.31",
         "tailwindcss": "^3.2.4",
         "webpack": "^5.75.0"
       }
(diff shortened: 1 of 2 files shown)


Fishbowler:
@amanjeetsingh150 I want to keep the security patches going whilst it's still operational and in use by folks. I'm not advocating for any first-party  maintenance of the functionality. If it stops actually working, I'd advocate for dropping it from the codebase.

amanjeetsingh150:
> @amanjeetsingh150 I want to keep the security patches going whilst it's still operational and in use by folks.

Exactly what I'm thinking, why is this still used when we have new studio? Are there any blockers we are not aware of? Let me know, if you know of any and we should priortize those and plan studio web deprecation.


Otherwise I'd say we can announce deprecation maybe someday in January?

oscar7-gh:
Hello,

This section of the code brings up alerts on Checkmarx for npm inflight vulnerability, if it were dropped the repo scan would return clean 100% on this tool.

Fishbowler:
@oscar7-gh Does checkmarx support the gradle version catalog we use?

oscar7-gh:
@Fishbowler  It should, yes

Fishbowler:
Ooh, maybe not.

https://docs.checkmarx.com/en/34965-19111-files-used-for-manifest-resolution.html doesn't include [libs.versions.toml](https://github.com/mobile-dev-inc/Maestro/blob/main/gradle/libs.versions.toml)

Fishbowler:
This code won't get removed immediately. If you can open an issue about the scan result, there might be something we can do in the interim.

PR #2864, opened 2025-11-28
Allow evaluation of JS expressions in waitForAnimationToEnd timeout

### LINKED ISSUE #2734: waitForAnimationToEnd does not accept JS parameter as a timeout

### Steps to reproduce

JS:
```
output.utils = {
    timeout: 12000
}
```

YAML:
```
appId: com.example
---
- launchApp
- waitForAnimationToEnd:
    timeout: ${output.utils.timeout}
```

### Actual results

> Incorrect Format: timeout

```
╭──────────────────────────────────────────╮
│ 4 | ---                                  │
│ 5 | - waitForAnimationToEnd:             │
│ 6 |     timeout: ${output.utils.timeout} │
│                  ^                       │
│ ╭─────────────────────────────────────╮  │
│ │ The format for timeout is incorrect │  │
│ ╰─────────────────────────────────────╯  │
│ 7 | - runFlow:                           │
│ 8 |     when:                            │
╰──────────────────────────────────────────╯
```

### Expected results

waitForAnimationToEnd to accept an integer or a JS expression

`scrollUntilVisible` and `extendedWaitUntil` will both accept a JS parameter as a timeout.

### About app

Native iOS and Android app

### About environment

macOS 15.6.1

### Logs

n/a

### Maestro version

2.0.2

### Anything else?

_No response_

Fishbowler:
## Proposed changes

Allows JS expressions in the `timeout` field of `waitForAnimationToEnd`

## Testing

```
appId: com.nonexistent.app
---
- evalScript: ${output.customTimeout = 1234}
- waitForAnimationToEnd:
    timeout: ${output.customTimeout}
```

image

Unsure if there's a sensible place in the existing tests (or if there's value) for yet another JS evaluation test.

## Issues fixed

Fixes #2734

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
@@ -1023,16 +1023,29 @@ data class RunScriptCommand(
 }
 
 data class WaitForAnimationToEndCommand(
-    val timeout: Long?,
+    val timeout: String?,
     override val label: String? = null,
     override val optional: Boolean = false,
 ) : Command {
 
     override val originalDescription: String
-        get() = "Wait for animation to end"
+        get() {
+            var description = "Wait for animation to end"
+            timeout?.let {
+                description += " within $it ms"
+            }
+            return description
+        }
 
+    private fun String.timeoutToMillis(): String? {
+        return if (this.toLong() < 0) {
+            null
+        } else this
+    }
     override fun evaluateScripts(jsEngine: JsEngine): Command {
-        return this
+        return copy(
+            timeout = timeout?.evaluateScripts(jsEngine)?.timeoutToMillis()
+        )
     }
 }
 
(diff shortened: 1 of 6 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -930,16 +931,29 @@ data class RunScriptCommand(
...
+            var description = "Wait for animation to end"
+            timeout?.let {
+                description += " within $it ms"
+            }
+            return description
+        }
 
+    private fun String.timeoutToMillis(): String? {
+        return if (this.toLong() < 0) {
+            null
```
amanjeetsingh150, line 1042: Ummmm why this validation here shouldn't this be part of yaml validation?

Changing the data type might also break parsing for the commands that are working with long timeout.

IMO, this should be long and this validation should be responsibility of yaml layer.

Fishbowler: It can't be `long` if it's gonna be a JS parseable string. 

This mirrors the approach taken by ScrollUntilVisible. I'd rather be consistent in the imperfections for easy refactoring later.

PR #2872, opened 2025-12-04
Fix WebView element access on iOS 26 by including SafariViewService hierarchy

### LINKED ISSUE #2735: Can't access elements in browser which is opened from an app on iOS v26.0

### Steps to reproduce

1. Install iOS 26
2. Launch a Simulator which iOS 26 e.g iPhone 17
3. Build & install your App
4. Run a Maestro flow

### Actual results

In case the app opens a browser during the flow for example to enter login credentials to have a secure sign-in process with oauth, Maestro isn't able to detect any elements in the browser which was opened by the app.

### Expected results

As in previous versions of iOS Maestro should be able to access the elements in the browser which was opened by the app.

### About app

It is a native app built with expo and react native. Here are the versions:
- Expo: 52.0.39
- React Native: 0.77.2

The app isn't open source and therefore i cannot share a build or a reproducible sample. However i assume that the behaviour will be the same in other apps

### About environment

- Java: OpenJDK 17
- MacBook M3 with MacOS 26
- Architecture: Arm Apple Silicon

### Logs

<details>
<summary>Logs</summary>

```
Assertion is false: "Customer login" is visible

Assertion '"Customer login" is visible' failed. Check the UI hierarchy in debug artifacts to verify the element state and properties.

Possible causes:
```

</details>

### Maestro version

2.0.3

### Anything else?

_No response_

leovarmak:
## Summary

Fixes #2735 

On iOS 26, `SFSafariViewController` and similar WebViews run in a separate application container (`com.apple.SafariViewService`) rather than within the host app's process. This causes Maestro to miss all WebView elements when capturing the view hierarchy.

## Changes

Added `getSafariWebViewHierarchy()` helper in `ViewHierarchyHandler.swift` that:
- Detects iOS 26+ and checks if SafariViewService is running with active webviews
- Fetches the Safari WebView hierarchy and merges it into the main view hierarchy response
- Applies the existing `snapshotMaxDepth` limit to prevent memory issues with complex web pages

## Safeguards

Based on feedback from earlier attempts (crashes on M1 CI machines), this implementation includes:
- iOS version gating (only runs on iOS 26+)
- Pre-checks for service state and webview presence before attempting snapshot
- Depth limiting via `AXClientSwizzler` 
- Graceful error handling - returns nil instead of crashing if snapshot fails

## Testing

- Tested on iOS 26 simulator with apps using `SFSafariViewController`. WebView elements are now visible in Maestro Studio and accessible for automation.
- Tested on iOS 18 simulator. WebView elements are now visible in Maestro Studio and accessible for automation.

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -254,10 +254,9 @@ jobs:
 
   test-ios:
     name: Test on iOS
-    runs-on: macos-latest
+    runs-on: macos-26
     needs: build
     timeout-minutes: 120
-    #if: ${{ false }}
 
     env:
       MAESTRO_DRIVER_STARTUP_TIMEOUT: 240000 # 240s


Fishbowler:
Rebased.

Fishbowler (commented):
I've tested this locally using https://github.com/dornelas-atx/sample-app-with-webview (thanks @dornelas-atx!) and appears to solve the problem on iOS 26.0 and 26.1

I'd like a review from someone who understands our swift a little better before merging

Fishbowler:
Sorry for all of the noise 😬

amanjeetsingh150 (approved):

Fishbowler:
Thanks for the contribution @leovarmak and for your patience whilst we got this reviewed and tested. It'll be in a Maestro release over the next few days! 🎉

PR #2874, opened 2025-12-05
ceeate a singleton for mcp web browser

adrum:
## Proposed changes

I was trying to use Maestro MCP for the web, and it was a painful experience as every new command would spawn a new browser instance. I was expecting the agent to have a consistent browser it can use for repeated commands. For example, the take screenshot command would just be a fresh browser with a white screen.

This proposed solution simply keeps a reference around in the MCP areas of the code base so the agent can keep one instance around for all its commands.

A perhaps more robust approach would have been to properly implement IDs similar to how iOS/Android devices are assigned. That would require a fix in the core parts of the code, and thought it might be easier to focus in improving just the MCP experience first and start some discussion around a more robust solution.

copilot:summary

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/McpServer.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/McpServer.kt
@@ -26,13 +26,19 @@ import maestro.cli.mcp.tools.InspectViewHierarchyTool
 import maestro.cli.mcp.tools.CheatSheetTool
 import maestro.cli.mcp.tools.QueryDocsTool
 import maestro.cli.util.WorkingDirectory
+import kotlin.concurrent.thread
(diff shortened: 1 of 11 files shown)


Fishbowler (approved):
LGTM, but would love some eyes from @steviec - he knows this code much better than me :)

DevGeorge:
Is this looking to be committed anytime soon?

Fishbowler:
You're in the right place to see any and all updates to this.

Ideally we'd merge everything as soon ~instantly and get releases out. In reality, we've got a single digit number of engineers, and we're not just maintaining the open source project, but also the software that pays the bills.

We'll get to this just as soon as we can, but I'd expect it to be January at this point.

HTH

Issue #1087, opened 2023-05-18
Very high memory usage with Maestro Studio on Linux


meatnordrink:
**Describe the bug**
Memory use on Linux increases dramatically when using Maestro Studio - it will slowly and steadily increase from a baseline of about 44% (of 32G) until the computer crashes. Closing Maestro Studio immediately reduces memory usage back to ~44%. (The tests finishing does not affect memory usage.)

It appears that this happens when Maestro Studio is running while a Maestro test on the device it is observing is running, even if Maestro Studio itself is not being interacted with. 

**To Reproduce**

(On Linux)
Launch an emulator with an app built. 
Launch Maestro Studio - `maestro studio`. 
Run a few long Maestro tests. Watch your memory use increase.

**Expected behavior**

Memory use to increase slightly, due to Maestro Studio running, but then remain steady.

**Environment information (please complete the following information):**

- Maestro version: 1.27.0
- Platform: Android
- Framework: React Native
- Device model and OS version: any android (usually Pixel 6 or Nexus S)
- Simulator: Android Studio Emulator
- Host: Linux (Ubuntu)

**Workspace (if applicable)**
I can't share the application code, nor is the issue really about a Maestro test, but if there's specific information you need let me know and I'll do my best.

meatnordrink:
A further note on this - on Linux, killing Maestro Studio in one terminal with `ctrl + c` _also_ kills any simultaneously-running Maestro test. 

I thought this was known, but I just now realized this isn't the case on iOS - and so wondered if it might be connected to the memory leak bug. (Which is still very much happening - Maestro Studio is only usabel for about ~7min on Linux.)

synapticvoid:
I have the same issue with the latest version of Maestro.

scriptas:
I run maestro tests from a raspberry, so this high memory usage is a massive issue. It completely freezes a (32gb sd card) raspberry pi in around 20 hours.

Restarting RPi works, but just for around 24 hours.

After investigating, I found that ./tmp folder is getting jammed up by hundreds of maestro-app*.apk & maestro-server*.apk files. Deleting those will clear up the memory.

As a temporary fix, I will just involve a cleanup of those files as a solution.

bartekpacia:
Hey @scriptas – do you mean that you use Maestro Studio, or just run tests with `maestro test` on your Raspberry Pi? Also, thanks for sharing the findings with us.

scriptas:
[...]

PR #2881, opened 2025-12-10
Only persist context when maestro or output binding modified

steviec:
## Proposed changes

Graal JS contexts take almost a full 1MB of memory, and if we have a repeated loop of commands, we can easily hit 1000+ commands (with associated contexts). The contexts are only cleared when the full maestro flow is done executing, which means we can bloat our memory considerably.

However, if we close a context, then values that have been stored on our `output` or `maestro` objects will get wiped. So we need to keep those values around.

This PR immediately closes contexts _only_ when we detect that the the bindings have not been updated. Otherwise we keep the context around until the engine is closed.

## Testing

See test suite to understand how this was tested.

## NOTE

This fix relies on the fact that all callers of evaluateScript cast the result to a string. It would probably be safer to force the return type of this to `String` instead of `Any`; we have to do that because we need to copy the `Value` returned by the js engine to a kotlin type. There might be a better solution, but I didn't want to touch too much code around this.

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
@@ -39,7 +39,7 @@ class GraalJsEngine(
     private val outputBinding = HashMap<String, Any>()
     private val maestroBinding = HashMap<String, Any?>()
     private val envBinding = HashMap<String, String>()
-    
+
     // Stack to track environment variable scopes for proper isolation
     private val envScopeStack = mutableListOf<HashMap<String, String>>()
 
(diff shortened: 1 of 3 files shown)


Fishbowler:
Oooh, I wonder how this affects runs inside Studio? I wonder if this helps with #1087

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -75,25 +78,50 @@ class GraalJsEngine(
...
+
+        // Only close context if shared bindings weren't modified.
+        // If modified, the stored values may be Value objects tied to this context.
+        val bindingsModified = outputBinding != outputBefore || maestroBinding != maestroBefore
+        if (!bindingsModified) {
+            // Convert to string before closing - Value objects become invalid after context close.
+            // Callers (see Env.kt) call .toString() on the result anyway.
+            val stringResult = result.toString()
+            context.close()
+            openContexts.remove(context)
```
amanjeetsingh150, line 120: There might be this edge case still where bindings change in every js scope and they'd never close and still lead to this problem?

steviec:
Closing in favor of https://github.com/mobile-dev-inc/Maestro/pull/2886

PR #2884, opened 2025-12-11
fix: respect cancellations on repeatWhile

amanjeetsingh150:
## Proposed changes

Repeat while commands can have conditions that are always true and can have infinite loop, these infinite loops should only execute when the coroutine is active.

This ensures that the loops are cooperative to cancellations.

## Testing

Integration test to reproduce this problem and fixing it.

### DIFF
--- a/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
+++ b/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
@@ -4073,8 +4073,6 @@ class IntegrationTest {
                     val maxReasonableSkips = 1_00
 
                     withTimeout(2000) {
-                        val flowJob = coroutineContext[Job]
-
                         val orchestra = Orchestra(
                             maestro = maestro,
                             lookupTimeoutMs = 0L,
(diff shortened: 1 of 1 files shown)


Fishbowler:
Nice - Studio suffers from this too ❤️

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -701,7 +697,7 @@ class Orchestra(
                 ?.let { evaluateCondition(it, commandOptional = command.optional) } != false
         }
 
-        while (checkCondition() && counter < maxRuns) {
+        while (checkCondition() && counter < maxRuns && currentCoroutineContext().isActive) {
```
steviec, line 700: NICE. I was trying to think of a good way to do this, and this is a simple, elegant solve.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4036,6 +4038,100 @@ class IntegrationTest {
         driver.assertHasEvent(Event.SetPermissions(appId = "com.example.app", permissions = mapOf("all" to "deny", "notifications" to "unset")))
     }
 
+
+    @Test
+    fun `Case 132 - repeatWhile respects coroutine timeout and gets cancelled`() {
```
steviec, line 4043: Unless you really need them, I think the callbacks should be simplified to do nothing, or just have onCommandComplete wait 10ms or something. If a test has a lot more going on in it unrelated to the specific thing we're testing it's hard to read.

amanjeetsingh150: Removed unnecessary callbacks and vars.

PR #2886, opened 2025-12-12
Testing a single context, IIFE-based approach to isolating vars set i…

steviec:
## Proposed changes

Graal JS contexts take almost a full 1MB of memory, and if we have a repeated loop of commands, we can easily hit 1000+ commands (with associated contexts). The contexts are only cleared when the full maestro flow is done executing, which means we can bloat our memory considerably.

However, if we close a context, then values that have been stored on our `output` or `maestro` objects will get wiped. So we need to keep those values around.

This PR keeps a single, shared context, which guarantees that `output` and `maestro` objects persist across invocations. It uses IIFE to isolate vars defined in the user script, and has to do additional management of the dynamic env vars to maintain our sub scope logic.

## Testing

See test suite to understand how this was tested. The tests are identical to the other PR: https://github.com/mobile-dev-inc/Maestro/pull/2881.

### DIFF
--- a/maestro-client/src/test/java/maestro/device/DeviceCatalogTest.kt
+++ b/maestro-client/src/test/java/maestro/device/DeviceCatalogTest.kt
@@ -12,11 +12,10 @@ internal class DeviceCatalogTest {
 
         assertThat(spec.platform).isEqualTo(Platform.ANDROID)
         assertThat(spec.model).isEqualTo("pixel_6")
-        assertThat(spec.os).isEqualTo("android-34")
+        assertThat(spec.os).isEqualTo("android-33")
         assertThat(spec.locale.code).isEqualTo("en_US")
         assertThat(spec.orientation).isEqualTo(DeviceOrientation.PORTRAIT)
         assertThat(spec.disableAnimations).isEqualTo(false)
-        assertThat(spec.snapshotKeyHonorModalViews).isEqualTo(false)
     }
 
     @Test
@@ -29,6 +28,7 @@ internal class DeviceCatalogTest {
         assertThat(spec.locale.code).isEqualTo("en_US")
         assertThat(spec.orientation).isEqualTo(DeviceOrientation.PORTRAIT)
         assertThat(spec.disableAnimations).isEqualTo(false)
+        assertThat(spec.snapshotKeyHonorModalViews).isEqualTo(false)
     }
 
     @Test


maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -82,26 +88,82 @@ class GraalJsEngine(
...
+            bindings.putMember("faker", faker)
+            bindings.putMember("output", ProxyObject.fromMap(outputBinding))
+            bindings.putMember("maestro", ProxyObject.fromMap(maestroBinding))
         }
+
+        // Clear non-permanent (env) bindings, then set current env vars
+        bindings.memberKeys
+            .filter { it !in permanentBindingKeys }
+            .forEach { bindings.removeMember(it) }
+        envBinding.forEach { (k, v) -> bindings.putMember(k, v) }
```
steviec: Note that this would be much cleaner if we just made ENV variables available on a named binding for it, e.g. `env` instead of putting them directly on the context root.

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -82,26 +88,82 @@ class GraalJsEngine(
...
+        if (!bindings.hasMember("http")) {
+            maestroBinding["platform"] = platform
+            bindings.putMember("http", httpBinding)
+            bindings.putMember("faker", faker)
+            bindings.putMember("output", ProxyObject.fromMap(outputBinding))
+            bindings.putMember("maestro", ProxyObject.fromMap(maestroBinding))
         }
+
+        // Clear non-permanent (env) bindings, then set current env vars
+        bindings.memberKeys
```
Leland-Takamine, line 153: Does this allow env vars to override static bindings? I don't think that was possible before

maestro-test/src/test/kotlin/maestro/test/GraalJsEngineTest.kt
```diff
@@ -91,16 +91,118 @@ class GraalJsEngineTest : JsEngineTest() {
...
     }
 
+    @Test
+    fun `memory should remain bounded after many evaluations without binding modifications`() {
+        // This test simulates a repeat loop with a condition like:
+        //   - repeat:
+        //       while:
+        //         notVisible: '${MARKET}'
+        //       commands:
+        //         - swipe: ...
```
amanjeetsingh150, line 113: I'm looking at this problem now from another perspective, wasn't the real problem flows going beyond the limit defined for orchestra in our case 15?

If we fix that should we have this guard? I think if a timeout is enough to be that guard. I would consider a timeout that ensures JS code doesn't end up taking very high memory.

So, one solution can be having this timeout defined on our CLI as well. Right now we just have that on cloud.

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
steviec: Fixed

PR #2887, opened 2025-12-12
Add shard index option to StartDeviceCommand

cem2ran:
This update introduces a new command line option `--shard-index` to the StartDeviceCommand, allowing users to specify a shard index for device creation, enabling the user to start multiple emulators via the maestro cli. 

The cli appeared to already support this internally so this PR simply exposes this to the end user.

## Proposed changes

copilot:summary

## Testing

Usage example:

```bash
for ((i=0; i<$MAX_EMULATORS; i++)); do
  echo "Starting emulator $((i+1))/$MAX_EMULATORS with shard-index=$i"
  maestro start-device --platform=android --shard-index="$i"
done
```

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
@@ -64,6 +64,13 @@ class StartDeviceCommand : Callable<Int> {
     )
     private var forceCreate: Boolean = false
 
+    @CommandLine.Option(
+        order = 5,
+        names = ["--shard-index"],
+        description = ["Shard index to use for the device"],
+    )
+    private var shardIndex: Int? = null
+
     override fun call(): Int {
         TestDebugReporter.install(null, printToConsole = parent?.verbose == true)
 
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -95,7 +102,7 @@ class StartDeviceCommand : Callable<Int> {
         try {
             val (deviceLanguage, deviceCountry) = LocaleUtils.parseLocaleParams(locale, maestroPlatform)
 
-            DeviceCreateUtil.getOrCreateDevice(p, o, deviceLanguage, deviceCountry, forceCreate).let { device ->
+            DeviceCreateUtil.getOrCreateDevice(p, o, deviceLanguage, deviceCountry, forceCreate, shardIndex).let { device ->
```
Fishbowler, line 105: You've seen that the impl will add 1 to the input value?

Fishbowler: I've seen the usage example you've added - I guess it doesn't matter in this instance?

Fishbowler: If not, should the description be updated for other folks who might perceive this as unusual?

cem2ran: Yeah, I think it makes sense that the emulators have the shard-index in the avd-name?

cem2ran: Oh I see what you mean. Probably should not be adding `_1` if shard-index is not present. Is that what you mean? Or do you mean that it should be there at all?

Fishbowler: No, I mean that if you input 3 it's gonna put 4 on the name.

Fishbowler: Presumably a decision based on counting from 0, etc. You could cater for this with explicit logic, I imagine, if that were desirable.

Fishbowler:
How does this compare to the approach in #2909 ?

Fishbowler:

PR #2897, opened 2025-12-17
feat: added a strongly typed device locale

proksh:
## Proposed changes

- Changed the device locale:
      - Strongly typed
      - Combined for IOS, Android & Web
      - Added functions that can be utilised throughout: Maestro CLI, Cloud & Studio 
      - Added Web supported configs
      - Added check for correct combination by using java valid combinations
- Used the new type in maestro codebase
- Updated tests

## Fix

Rather than starting device and then get the error, we get an error before even starting the device

Screenshot 2025-12-17 at 6 53 09 PM

We can use the same code for detection in Cloud apis as well


steviec (commented):
Hey @proksh , I love strongly typing our locales, but I'm wondering if we could make this way simpler. Instead of giving line-by-line feedback, here's a zooming out set of suggestions/questions.

### Name it like types, not utils
"LocaleUtils" suggests helper functions (which they previously were), but these are actually types that represent valid locales. We should stop using "Utils" and helper methods like "getBlahBlah". It seems far more straightfoward to be interacting with types like `AndroidLocale.code`, etc.

### Remove sealed class hierarchy
The sealed class LocaleUtils with Android and Ios subclasses adds complexity without benefits that I can see. These are separate types for separate platforms - they don't need a common parent. Code that handles locales already knows which platform it's dealing with. Just have:
```
data class AndroidLocale(val language: AndroidLanguage, val country: AndroidCountry)
enum class IosLocale(val code: String) (flat, because iOS has special formats like zh-Hans)
```

And clients can directly use the locale for whatever platform they are handling. If we need to guarantee that these have some common helpers, we can create a sealed class JUST for those, e.g.
```
sealed interface DeviceLocale {
    val code: String
}

data class AndroidLocale(...) : DeviceLocale {
    override val code get() = "${language.code}_${country.code}"
}
...
```

### Validation
I'm seeing a lot of validation that I'm wondering if it's necessary. e.g. in the IosSupportedLocale we have an init() method that does validation. But doesn't the enum itself already define what we accept? Is there a good reason to double that validation?

proksh:
> Hey @proksh , I love strongly typing our locales, but I'm wondering if we could make this way simpler. Instead of giving line-by-line feedback, here's a zooming out set of suggestions/questions.
> 
> ### Name it like types, not utils
> "LocaleUtils" suggests helper functions (which they previously were), but these are actually types that represent valid locales. We should stop using "Utils" and helper methods like "getBlahBlah". It seems far more straightfoward to be interacting with types like `AndroidLocale.code`, etc.
> 
> ### Remove sealed class hierarchy
> The sealed class LocaleUtils with Android and Ios subclasses adds complexity without benefits that I can see. These are separate types for separate platforms - they don't need a common parent. Code that handles locales already knows which platform it's dealing with. Just have:
> 
> ```
> data class AndroidLocale(val language: AndroidLanguage, val country: AndroidCountry)
> enum class IosLocale(val code: String) (flat, because iOS has special formats like zh-Hans)
> ```
> 
> And clients can directly use the locale for whatever platform they are handling. If we need to guarantee that these have some common helpers, we can create a sealed class JUST for those, e.g.
> 
> ```
> sealed interface DeviceLocale {
>     val code: String
> }
> 
> data class AndroidLocale(...) : DeviceLocale {
>     override val code get() = "${language.code}_${country.code}"
> }
> ...
> ```
> 
> ### Validation
> I'm seeing a lot of validation that I'm wondering if it's necessary. e.g. in the IosSupportedLocale we have an init() method that does validation. But doesn't the enum itself already define what we accept? Is there a good reason to double that validation?

Updated the code, it's much cleaner now. :)

Also organised the Android, Ios, Web locale separately so its easier to read and manage.

maestro-client/src/test/java/maestro/locale/DeviceLocaleTest.kt
```diff
@@ -0,0 +1,154 @@
+package maestro.locale
+
+import maestro.Platform
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import com.google.common.truth.Truth.assertThat
+
+internal class DeviceLocaleTest {
```
amanjeetsingh150, line 8: Can you add a test here for locale: he-IL? This is valid and I remember we had issues while parsing those since testing got missed.

amanjeetsingh150 (commented):
I like where this PR is heading. Given we want all the device capabilities like OS, locale, model to be strongly typed. Different requirements here would be:

1. We can list the supported types
2. We have validation from just one layer instead of studio, CLI and cloud their own thing.

Right now the problem is all of this is quiet spread out and we should be consolidating all of this under one layer.

# Making DeviceCatalog

Device catalog is where we know what are different specs on device like os, locale, model etc. For now you can create locale here.

# Idea of DeviceCatalog

- DeviceCatalog's returns deviceSpec that can be used directly by studio/cloud or CLI.
- These specs like locale are strongly type instead of using primitives directly

Something like:

```
data class DeviceSpec(
  val os: DeviceOsId,
  val model: DeviceModelId,
  val locale: LocaleId
)
```
- Single validation layer for one spec like:

```
DeviceCatalog.parseLocale(platform: Platform, raw: String): LocaleId
```

- If needed for any clients CLI, studio or cloud. This can also return supported matrices:

```
DeviceCatalog.supported(platform): SupportedDeviceMatrix
```

maestro-client/src/test/java/maestro/locale/DeviceLocaleTest.kt
proksh: Added

PR #2899, opened 2025-12-17
Quieter permissions logging when setting permissions

Fishbowler:
## Proposed changes

In #2775, some debug logging was added for failures to set permissions. That's actually still really noisy when, by default, we set 'all'.

Example from our e2e tests:

```
==> [demo_app] [passing] => disableAnsi=false([ INFO]) Launch app "com.example.example" RUNNING
==> [demo_app] [passing] => disableAnsi=false([DEBUG]) Failed to set permission android.permission.INTERNET for app com.example.example: pm grant com.example.example android.permission.INTERNET: 
==> [demo_app] [passing] => Exception occurred while executing 'grant':
==> [demo_app] [passing] => java.lang.SecurityException: Permission android.permission.INTERNET requested by com.example.example is not a changeable permission type
==> [demo_app] [passing] => 	at com.android.server.pm.permission.PermissionManagerService.grantRuntimePermissionInternal(PermissionManagerService.java:1508)
```

This filters out the permissions that error because they're not settable.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -934,9 +934,13 @@ class AndroidDriver(
         try {
             shell("pm $permissionValue $appId $permission")
         } catch (exception: Exception) {
-            // We don't need to be loud about this. IOExceptions were caught in shell. Remaining issues are likely due
-            // to "all" containing permissions that the app doesn't support.
-            logger.debug("Failed to set permission $permission for app $appId: ${exception.message}")
+            // Ignore if it's something that the user doesn't have control over (e.g. you can't grant / deny INTERNET)
+            if (exception.message?.contains("is not a changeable permission type") == false) {
+                // Debug level is fine.
+                // We don't need to be loud about this. IOExceptions were already caught in shell(..)
+                // Remaining issues are likely due to "all" containing permissions that the app doesn't support.
+                logger.debug("Failed to set permission $permission for app $appId: ${exception.message}")
+            }
         }
     }
 


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -934,9 +934,13 @@ class AndroidDriver(
         try {
             shell("pm $permissionValue $appId $permission")
         } catch (exception: Exception) {
-            // We don't need to be loud about this. IOExceptions were caught in shell. Remaining issues are likely due
-            // to "all" containing permissions that the app doesn't support.
-            logger.debug("Failed to set permission $permission for app $appId: ${exception.message}")
+            // Ignore if it's something that the user doesn't have control over (e.g. you can't grant / deny INTERNET)
+            if (exception.message?.contains("is not a changeable permission type") == false) {
```
proksh, line 938: Wouldn't this become a problem if user is trying to change permission for a value that is not controllable but also doesn't see any log?

For example: If I am trying to grant permission for INTERNET and its not happening. And when I see the logs, I don't see anything that will help me explain the issue.

Fishbowler: I thought about that, and figured that it's far outweighed by the number of folks just doing `launchApp` with no explicit permissions and getting lots of unnecessary noise. I don't think I've ever seen anyone explicitly trying to control the internet permission - it's an Android legacy permission.

This log line is new to this version (#2775) so we wouldn't be removing observability that anyone already has.

proksh: Got it. Since no one complaint about the missing log before, I wouldn't expect this to be a big deal.

PR #2900, opened 2025-12-17
Add "setClipboard"command  (issue - #2029)

### LINKED ISSUE #2029: Add new `copyText` command to set clipboard contents
### Use case

There is a lot of issues with input flakiness, but there is no need to test input each time
We can just paste text in input, and I hope it will be more stable

https://github.com/mobile-dev-inc/maestro/issues/1225

### Proposal

There is the `pasteText` command in maestro, so adding `copyText` with string argument will give us more complete api.

Instead of 

```
- tapOn:
    id: email-feild

- inputText: 'custom@example.com'

```

will be something like this

```
- tapOn:
    id: email-feild

- copyText: 'custom@example.com'
- pasteText 

```

Or even simplier - it could be just string argument to pasteText, so test will be

```
- tapOn:
    id: email-feild
- pasteText  'custom@example.com'

```

### Anything else?

_No response_

ryuuhei0729:
## Summary

Adds a new `setClipboard` command to set clipboard contents directly, complementing the existing `copyTextFrom` command.

Fixes #2029

## Usage

```yaml
- setClipboard: "text to copy"
- pasteText
```

## Changes

- Add `SetClipboardCommand` class in `Commands.kt`
- Add `setClipboardCommand` field to `MaestroCommand`
- Add `YamlSetClipboard` class for YAML parsing (string syntax only)
- Add command handler in `Orchestra.kt`
- Add integration test (Case 133)

## Testing

Added integration test `Case 133 - Set clipboard` that verifies:
- `setClipboard` command sets text to clipboard
- `pasteText` command pastes the copied text
- Both commands work together correctly

Run tests with:
```bash
./gradlew :maestro-test:test --tests "*Case 133*"
```

## Notes

- This command uses the existing clipboard infrastructure (`copiedText` variable and `jsEngine.setCopiedText()`)
- No changes to `Maestro` or `Driver` APIs were needed
- The `text` parameter supports JavaScript expression evaluation via `evaluateScripts`
- The command name `setClipboard` was chosen to accurately reflect that it sets Maestro's internal clipboard, not the device's system clipboard

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
@@ -377,8 +377,8 @@ data class YamlFluentCommand(
             setClipboard != null -> listOf(
                 MaestroCommand(
                     SetClipboardCommand(
-                        text = setClipboard.text.
-                        label = setClipboard.label.
+                        text = setClipboard.text,
+                        label = setClipboard.label,
                         optional = setClipboard.optional
                     )
                 )


Fishbowler:
This doesn't appear to actually interact with the clipboard at all... Am I missing something?

Fishbowler:
This is just trying to set the fake clipboard contents with direct text, rather than via an element?

If so, should we name it something more like `setClipboard`?

Also, don't forget a matching docs PR for the new command - https://github.com/mobile-dev-inc/maestro-docs - there's past and current PRs you can use as prior art to see how it's done.

ryuuhei0729:
@Fishbowler , thanks for the response!

I've implemented it based on [this comment](https://github.com/mobile-dev-inc/Maestro/issues/2029#issuecomment-2337819429). 
Should I rename the command to `setClipboard`?


Also, I’ll make sure to submit a PR to https://github.com/mobile-dev-inc/maestro-docs after the fix. Thanks!

Fishbowler:
I think the rename makes sense. I'll take a closer look at the implementation now.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -267,6 +267,23 @@ data class CopyTextFromCommand(
...
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+
+    override val originalDescription: String
+        get() = "Copy text to clipboard"
+
+    override fun evaluateScripts(jsEngine: JsEngine): CopyTextCommand {
+        return copy(
+            text = text.evaluateScripts(jsEngine),
```
Fishbowler, line 281: ❤️  This gets forgotten too often

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -267,6 +267,23 @@ data class CopyTextFromCommand(
...
 }
 
+data class CopyTextCommand(
+    val text: String,
+    override val label: String? = null,
+    override val optional: Boolean = false,
+) : Command {
+
+    override val originalDescription: String
+        get() = "Copy text to clipboard"
```
Fishbowler: Should this include the value?

Something like:

```suggestion
        get() = "Set Maestro clipboard to $text"
```

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlCopyText.kt
```diff
@@ -0,0 +1,31 @@
...
+                    val optional = text.getOrDefault("optional", false) as? Boolean ?: false
+                    return YamlCopyText(input, label, optional)
+                }
+                is Int, is Long, is Char, is Boolean, is Float, is Double -> text.toString()
+                else -> throw UnsupportedOperationException("Cannot deserialize copy text with data type ${text.javaClass}")
+            }
+            return YamlCopyText(text = copyText)
+        }
+    }
+}
```
Fishbowler: Can't this be as simple as a text parser, and leave parsing or the user to do any implicit/explicit conversions?


```suggestion
companion object {
        @JvmStatic
        @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
        fun parse(text: String) = YamlCopyText(
            text = text,
        )
    }
```

ryuuhei0729:
@Fishbowler 
Thanks for the review. 
I've made the updates, could you take another look?

ryuuhei0729:
I also added documentation in [this PR](https://github.com/mobile-dev-inc/maestro-docs/pull/165),  and would love your review on it.

maestro-orchestra-models/src/main/java/maestro/orchestra/MaestroCommand.kt
```diff
@@ -57,8 +57,8 @@ data class MaestroCommand(
     val setLocationCommand: SetLocationCommand? = null,
     var setOrientationCommand: SetOrientationCommand? = null,
     val repeatCommand: RepeatCommand? = null,
-    val copyTextFromCommand: CopyTextFromCommand? = null,
-    val copyTextCommand: CopyTextCommand? = null,
+    val copyTextCommand: CopyTextFromCommand? = null,
```
Fishbowler, line 60: Unintentional rename?

Fishbowler: Ah no, that's naming back.

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
```diff
@@ -372,6 +374,13 @@ data class YamlFluentCommand(
             )
 
             copyTextFrom != null -> listOf(copyTextFromCommand(copyTextFrom))
+            setClipboard != null -> listOf(
+                MaestroCommand(
+                    SetClipboardCommand(
+                        text = setClipboard.text
+                    )
```
Fishbowler, line 383: I think label and optional still need plumbing through here

```suggestion
                    SetClipboardCommand(
                        text = setClipboard.text.
                        label = setClipboard.label.
                        optional = setClipboard.optional
                    )
```

ryuuhei0729:
@Fishbowler 
My bad, I accidentally deleted it!
I fixed it now—could you re-review it when you have a chance?

Fishbowler (approved):

PR #2904, opened 2025-12-23
Show hierarchy, query, and check-syntax commands in help

Fl0p:
Remove the hidden flag from hierarchy, query, and check-syntax commands to make them visible in the maestro --help output. These are useful commands that users should be aware of.

Changes:
- PrintHierarchyCommand: Remove hidden = true
- QueryCommand: Remove hidden = true
- CheckSyntaxCommand: Remove hidden = true

## Proposed changes

Aligned with 

https://docs.maestro.dev/cli/view-hierarchy#hierarchy

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/QueryCommand.kt
@@ -39,7 +39,8 @@ import picocli.CommandLine.Spec
     name = "query",
     description = [
         "Find elements in the view hierarchy of the connected device"
-    ]
+    ],
+    hidden = true
 )
 class QueryCommand : Runnable {
 


Fishbowler:
I'm not sure I agree that `query` is useful to expose. It doesn't support complex or interpreted selectors, and is prone to expansions and interpretations on the CLI.

Fl0p:
Meanwhile It is already exposed
"It doesn't support complex or interpreted selectors" - thats is sad and have to be addressed

We (and AI agents) need some kind feedback loop to perform well. 
To have the ability touch and feel the screen to create robust and precise UI element selectors.

And right now there is no good way to do so

Only two workarounds:
- run studio and make UI selectors attempts inside studio
- create temp yaml file with assertVisible and run it with maestro test

First is not usable for tools (Clause code as example)
Second one is weird and messy. and give only true/false result not good for multiple items

So... i think it is better to  keep `query` and evolve it to support all possible selectors

Fishbowler:
It's exposed, but it's not easily discoverable, and it's rarely used.

I'd rather make it useful before exposing it to additional users. I'd want it to take the full range of selectors to address the stuff you're talking about.

Fishbowler:
Query isn't useful to most folks in its current form. Lets make it useful before it's shown.

Fl0p:
@Fishbowler  query hidden

PR #2905, opened 2025-12-23
Add maestro.flowName to JavaScript context

Fl0p:
Enable flow scripts to access the current flow name via maestro.flowName. This allows users to make flow-specific decisions and log which flow is currently executing.

Implementation:
- Add flowName field to maestro object (Js.kt)
- Pass flowName through Orchestra to JS engines (GraalJsEngine, RhinoJsEngine)
- Implement fallback chain: CLI parameter → config.name → "unknown"
- Use sanitizeJs() to handle special characters in flow names

Usage example:
```yaml
appId: com.example.app
name: MyFlow
---
- runScript: console.log("Running: " + maestro.flowName)
- takeScreenshot: ${maestro.flowName}/start
```

## Proposed changes

- Added maestro.flowName to JavaScript context
Useful wen multiple flows running and all of them write same screenshots 

## Testing

Runned locally

Screenshot 2025-12-23 at 13 05 34

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
@@ -57,13 +57,15 @@ object TestRunner {
             flowFile = flowFile,
         )
 
+        val commands = YamlCommandReader.readCommands(flowFile.toPath())
+        val flowName = YamlCommandReader.getFlowName(commands, flowFile.nameWithoutExtension)
+        aiOutput = aiOutput.copy(flowName = flowName)
+
         val updatedEnv = env
             .withInjectedShellEnvVars()
-            .withDefaultEnvVars(flowFile, deviceId)
+            .withDefaultEnvVars(flowFile, flowName, deviceId)
+        val commandsWithEnv = commands.withEnv(updatedEnv)
 
-        val commands = YamlCommandReader.readCommands(flowFile.toPath()).withEnv(updatedEnv)
-        val flowName = YamlCommandReader.getConfig(commands)?.name ?: flowFile.nameWithoutExtension
-        aiOutput = aiOutput.copy(flowName = flowName)
         logger.info("Running flow ${flowFile.name}...")
 
         // Per-flow folder ArtifactsGenerator writes the bundle into (see BundleLayout).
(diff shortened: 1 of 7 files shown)


Fishbowler:
I'm not sure this is the right approach. Take a look at how `MAESTRO_FILENAME` is rendered, by adding it to the env of a runner, without specific need for JSEngine changes. Right now, it's evaluated just _before_ a flow is read, but I suspect that it could suffer a small refactor to set the parsed name at the same time.

Fl0p:
@Fishbowler  Done

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
```diff
@@ -58,17 +58,17 @@ object TestRunner {
...
-            .withInjectedShellEnvVars()
-            .withDefaultEnvVars(flowFile)
-
         val result = runCatching(resultView, maestro) {
-            val commands = YamlCommandReader.readCommands(flowFile.toPath())
-                .withEnv(updatedEnv)
-
-            val flowName = YamlCommandReader.getConfig(commands)?.name ?: flowFile.nameWithoutExtension
+            val rawCommands = YamlCommandReader.readCommands(flowFile.toPath())
+            val flowName = YamlCommandReader.getConfig(rawCommands)?.name ?: flowFile.nameWithoutExtension
```
Fishbowler: Is this inconsitent with implementation in other places?

In other places, the flowName is a nullable flow.name. 
Here, it defaults to the filename if there's no name.

Is this the most user friendly, to always have _some_ name on that var?

Fishbowler (commented):
This is looking great! One comment on consistency. Do you think this should have its own IntegrationTest?

maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt
Fl0p: Done !

Fl0p: сс @Fishbowler

Fl0p:
@Fishbowler  still actual ?

Fishbowler:
Right, the rebase on this was tricky, given the work that's happened on artifact generation and storage, but I think it's there.

I've also added 1 abstraction to cover the read commands / read name / updatedEnv / commands.withEnv thing that we kept needing to repeat.

Give this a read and another test and be sure I haven't corrupted it, since it'll end up in your name 😂

PR #2908, opened 2025-12-24
Add equality conditions and commands assertEqual and assertNotEqual

ryanpato:
## Proposed changes

This PR introduces two new assertion commands, `assertEqual` and `assertNotEqual`, along with support for equality conditions in `runFlow`, `runScript`, and `repeat`.

## Motivation

Currently, equality checks in Maestro are typically expressed via `assertTrue`. While functional, this approach obscures intent, produces suboptimal labels and error messages. Equality is a common and semantically distinct operation in test flows, and treating it as a first-class assertion will improve readability, diagnostics, and clarity.

## Usage

**Assert Equal** 

```
- assertEqual:
    value1: "maestro"
    value2: "maestro"
```

```
 ║  > Flow: 133_assert_equal                                     
 ║                                                               
 ║    ✅   Assert that maestro is equal to maestro               
 ║    ✅   Assert that 4 is equal to 4   
```

```
 ║                                                               
 ║  > Flow: 133_assert_equal                                     
 ║                                                               
 ║    ❌   Assert that maestro is equal to maestros              
 ║    🔲   Assert that ${2 + 2} is equal to ${8 - 4}             
 ║                                                               
                                                                 
Assertion failed: expected 'maestros', but got 'maestro'

Assertion 'maestro is equal to maestros' failed. Check the UI hierarchy in debug artifacts to verify the element state and properties.

Possible causes:
```

**Assert Not Equal** 

```
- assertNotEqual:
    value1: "maestro"
    value2: "maestro"
```

**Conditional Equality Check**

```
- runFlow: 
    file: subflow.yaml
    when:
      equal:
        value1: "maestro"
        value2: "maestro"
```

## Testing

- Manually testing commands and conditions (CLI/Web Studio)
  - Commands & Conditions
  - Happy paths
  - Negative paths
  - Error messages
- Integration tests

## Documentation

https://github.com/mobile-dev-inc/maestro-docs/pull/174%5D


maestro-orchestra-models/src/main/java/maestro/orchestra/AssertEqual.kt
```diff
@@ -0,0 +1,16 @@
+package maestro.orchestra
+
+import maestro.js.JsEngine
+import maestro.orchestra.util.Env.evaluateScripts
+
+data class AssertEqual(
+    val value1: String,
+    val value2: String,
```
ryanpato: Not attached to naming here. Would actual/expected, expected/actual be better?

ryanpato:
> Test on Java 17 failed on CI

I don't think this was related to my changes. I did initially hit a similar looking UI assert error locally on an emulator without google apis on `main`, but it passed on an emulator with google apis. Not super confident with Android, though from the logs it looks like the device _did_ boot eventually, and the os image looks fine? Is this perhaps a known flaky CI issue?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -810,6 +813,18 @@ class Orchestra(
             }
         }
 
+        condition.equal?.let {
+            if (it.value1 != it.value2) { 
+                return false
+            }
+        }
```
Fishbowler, line 811: Are there any other equality operators that need consideration here?
(Not a leading question - I'm unsure)

maestro-orchestra-models/src/main/java/maestro/orchestra/AssertEqual.kt
```diff
@@ -0,0 +1,20 @@
...
+) {
+    fun evaluateScripts(jsEngine: JsEngine): AssertEqual {
+        return copy(
+            value1 = value1.evaluateScripts(jsEngine),
+            value2 = value2.evaluateScripts(jsEngine),
+        )
+    }
+
+    fun failureMessage(): String {
+        return "Assertion failed: expected '${value2}', but got '${value1}'"
```
Fishbowler: I'm not sure that this wording will fit all use cases. This assumes value1 as a source of truth, and value2 as the item under test. What if both elements come from different sources, e.g. the same balance must be displayed on the both the profile and transactions page

Fishbowler: Something simple like this?
```suggestion
        return "Assertion failed: expected '${value2}' to equal '${value1}'"
```

ryanpato: Thank you @Fishbowler , that makes total sense and i've gone with your suggestion

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
ryanpato: I'm thinking now these changes might be a bit naive and not doing what someone would expect. It's essentially a string comparison where I noticed `assertTrue` is a JS handled condition and would support the following:

```
- assertTrue: ${true == 1}
- assertTrue: ${1 == 1.0}
- assertTrue: ${JSON.stringify(["abc"]) == JSON.stringify(["abc"])}
- assertTrue: ${ ["abc"][0] == "abc" }
```

I think it might be too inconsistent as is. What do you think? I'll perhaps close this and raise it again in a better state.

ryanpato: Do you mean other checks like less/greater than, or contains etc sorry? I think i may have misunderstood your question initially and got in my head about whether the string equality was good enough or whether we'd want to use the js engine to evaluate the comparison itself, rather than separately to also include type coercion

Fishbowler: I was considering other kinds of equality. Do we need other operators for string v number v object?

But then if they're going to be coerced into strings before the final evaluation anyway, the test developer can maybe deal with that constraint?

ryanpato: Yeah I think my initial implementation is naive in that I assume most comparisons will be strings from UI elements (our desired use case) or stringified expressions.

I was considering today whether to evaluate the comparison in JS at least (like `assertTrue`). This might handle type coercion a _little_ better, but I think still has limitations like objects and arrays would fail due to reference equality in JS? 

I think it could be documented and worked around by the test developer, but i'm also happy to try iterate further here, or alternatively I can just close this PR and open an issue to discuss the design more thoroughly? (probably should've done this anyway sorry). Let me know your preference!

Fishbowler: Why not both?

If there's something that works for a lot of cases (string equality) but we're aware of edge cases (objects, floats, yaml parsing) then we could merge something that delivers most of the value and open an issue to discuss further improvements.

ryanpato: Why not both! If you're happy with that and think it's ok, i'm happy to proceed as such. Though i'm not sure how much more is required to get something release ready? Not sure if there's any documentation requirements with something like this or how to contribute?

Fishbowler: The next step for a new command would be to add a PR in the docs repo - https://github.com/mobile-dev-inc/maestro-docs. You'll see a couple of open PRs there, waiting for the next release before they're merged.

ryanpato: PR: raised: [maestro-docs/pull/174](https://github.com/mobile-dev-inc/maestro-docs/pull/174)
Also linked in PR body

PR #2915, opened 2026-01-05
Make CLI's integrationTest task part of the verification category

Fishbowler:
## Proposed changes

Entirely a developer experience change.

When I want to run the integrationTest task locally, I look in `verification`, see it's not there, then dig for it in `other`.

This fixes that, saving me _seconds_ every single week.

## Testing

Still runs.

## Issues fixed

n/a

### DIFF
--- a/maestro-cli/build.gradle.kts
+++ b/maestro-cli/build.gradle.kts
@@ -340,6 +340,7 @@ jreleaser {
 }
 
 tasks.register<Test>("integrationTest") {
+    group = "verification"
     useJUnitPlatform {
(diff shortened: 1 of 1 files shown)


proksh:
> I look in `verification`

Not sure what you mean by looking in `verification`, maybe I am not having the same workflow. Can you share steps or screen recording?

Fishbowler:
@proksh You're right, that was really terse!

I was talking about working inside IntelliJ.

Before:

image

After:

image

proksh (approved):
Ohh nice, this is new for me. 😄

Issue #2885, opened 2025-12-11
Referencing images


iMostfa:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Thanks for the great project. we are facing a blocking issue in our project where referencing images from JS scripts. it works locally, but doesn't work on maestro cloud. 

steps to reproduce: 
1. Fork `maestro-examples`
```bash
git clone https://github.com/Fishbowler/maestro-examples.git
```
2. run the following on test on cloud
```
https://github.com/Fishbowler/maestro-examples/blob/main/multipart-uploads/flow.yaml
```

### Actual results

it fails in the JS phase, with unable to find the file. 

### Expected results

it should work as expected and uploads the image in multiform request

### About app

-

### About environment

it doesn't matter  because it's the cloud itself, not local maestro

### Maestro version

latest

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

### Anything else?

full context:
https://maestro--dev.slack.com/archives/C041FU72T54/p1765293212452679

PR #2917, opened 2026-01-07
fix: Resolve multipart form file paths relative to JS script location

proksh:
## Problem

When using runScript with HTTP multipart form file uploads, file paths in the script were resolved relative to the JVM's current working directory instead of relative to the script file's location, causing all file references to fail.

### Example scenario:

```
workspace/
  flows/
    flow.yaml          # runScript: "./script/upload.js"
  script/
    upload.js          # filePath: "../media/image.png"
  media/
    image.png          # ❌ File not found!
```

The script would fail with `FileNotFoundException` even though the relative path `../media/image.png` is correct relative to the script's location.

## Root Cause

The codebase had no mechanism to track or use the script file's location when resolving file paths in multipart forms.

Before the fix:
```
// HttpUtils.kt
fun Map<*, *>.toMultipartBody(): MultipartBody {
    // ...
    val file = File(filePath.toString())  // ❌ Always resolves from JVM working directory
}
```

File paths were always resolved using `File(filePath)`, which interprets relative paths from the JVM's current working directory. Since the JVM could be running from anywhere (project root, build directory, temp folder), file resolution was unreliable and typically failed.

## Solution

Added a complete mechanism to track and use the script's directory when resolving file paths:

1. HttpUtils.kt - Added script directory parameter and smart resolution
```
fun Map<*, *>.toMultipartBody(scriptDir: File? = null): MultipartBody

private fun resolveFilePath(filePath: String, scriptDir: File?): File {
    val file = File(filePath)
    
    // If absolute path exists, use it
    if (file.isAbsolute && file.exists()) return file
    
    // If we have script directory, resolve relative to it
    if (scriptDir != null) {
        val resolvedFile = File(scriptDir, filePath)
        if (resolvedFile.exists()) return resolvedFile
    }
    
    // Fallback to JVM working directory (backward compatibility)
    return file
}
```

2. JsHttp.kt & GraalJsHttp.kt - Track current script directory
```
@Volatile
private var currentScriptDir: java.io.File? = null

fun setCurrentScriptDir(scriptPath: String?) {
    currentScriptDir = scriptPath?.let { java.io.File(it).parentFile }
}

// Pass to HttpUtils
multipartForm.toMultipartBody(currentScriptDir)
```

3. RhinoJsEngine.kt & GraalJsEngine.kt - Set script directory before execution
```
override fun evaluateScript(..., sourceName: String, ...): Any? {
    jsHttp.setCurrentScriptDir(if (sourceName != "inline-script") sourceName else null)
    // ...
}
```

4. YamlFluentCommand.kt - Provide absolute script path
```
runScript != null -> {
    val scriptPath = resolvePath(flowPath, runScript.file)  // Resolve to absolute path
    listOf(
        MaestroCommand(
            RunScriptCommand(
                script = scriptPath.readText(),
                sourceDescription = scriptPath.toString(),  // ✅ Absolute path
                // ...
            )
        )
    )
}
```

This ensures the script directory can be correctly extracted via File(sourceName).parentFile.

## Testing

✅ All existing tests pass (backward compatible)
✅ File paths now resolve correctly relative to script location
✅ Both relative and absolute file paths work as expected
✅ Graceful fallback to JVM CWD if script directory unavailable

## Side Effect (Minor UI Change)

Commands without custom labels now show absolute paths in CLI output:

- Before: Run `../utils/upload.js`
- After: Run `/full/path/to/utils/upload.js`

Users with custom labels see no change. The absolute path will be better for debugging and provides clarity.


Fishbowler:
Fixes #2885

iMostfa (approved):
Thanks a lot for taking a look and fixing it

maestro-test/src/test/kotlin/maestro/test/JsEngineTest.kt
```diff
@@ -253,4 +254,104 @@ abstract class JsEngineTest {
         // Then
         assertThat(result.toString()).isEqualTo("POST endpoint")
     }
+
+    @Test
```
steviec: Great test, but we probably only need one of them. The multiple files one is a superset of this, so kill this one.

PR #2922, opened 2026-01-10
feature/custom commands

ryanpato:
## Proposed changes

POC to allow for custom commands by mapping them to run flow commands. 

## Benefits

Custom commands are just a convenience. They can allow for shorter, cleaner and easier to maintain tests and reusable  flows by removing the verbosity of `runFlow` and `runScript`.

**Before**

```
appId: com.app.example
---
- launchApp
- runFlow:
    file: ../common/login.yaml
    env:
        email: user@email.com
        password: pass123
```

**After**

```
appId: com.app.example
---
- launchApp
- login:
    email: user@email.com
    password: pass123
```

----

## Example

**workspace**
```
my-workspace/
├── config.yaml
├── commands/
│   ├── login.yaml
└── flows/
    └── test.yaml
```

**commands/login.yaml**
```
appId: com.app.example
params:
  email: { required: true }
  password: { required: true }
---
- tapOn: "email"
- inputText: ${email}
- tapOn: "password"
- inputText: ${password}
- tapOn: "Submit"
```

**flows/test.yaml**
```
appId: com.app.example
---
- launchApp
- login:
    email: user@email.com
    password: pass123
```

## How to use
- add a config file in the root of the workspace (currently used to find the root of the workspace)
- add a commands directory in the root of the workspace (to read custom commands from)
- add a custom command in the commands dir e.g. `login.yaml` (supports both string and object types)
- add a test case using that command

## Considerations
- Adding `params` will allow for parse time failure. It will still work if you've not specified the params, but provided valid params. However, if you do not explicitly detail params, and do not correctly pass it, it will fail later at run-time.
- The filename is the name of the command, camel case recommended.
- **Test Maintenance**. If a test fails in CI, you wanna run just that test and see it fail locally so that you can make changes. With flow syntax tied to the config, you'd need to do something fun with tags to run just one test, I think?

## TODO
- How could this work with Maestro Studio?
- Is 'default' in params needed? i can't think of any examples but felt like it made sense to include
- FIx unit tests (currently i've changed the error output)
- Write new unit tests and integration tests

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/yaml/CustomCommandLoader.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/yaml/CustomCommandLoader.kt
@@ -1,0 +1,27 @@
+package maestro.orchestra.yaml
+
+import java.nio.file.Path
+import kotlin.io.path.exists
+import kotlin.io.path.isDirectory
+
+object CustomCommandLoader {
+    fun findWorkspaceRoot(flowPath: Path): Path? {
+        var current = flowPath.toAbsolutePath().parent
+        while (current != null) {
+            if (current.resolve("config.yaml").exists()) return current
+            current = current.parent
+        }
+        return null
+    }
+
+    fun getCommandsDirectory(workspaceRoot: Path): Path? {
+        val commandsDir = workspaceRoot.resolve("commands")
+        return if (commandsDir.exists() && commandsDir.isDirectory()) commandsDir else null
+    }
+
+    fun findCustomCommand(flowPath: Path, commandName: String): Path? {
+        val workspaceRoot = findWorkspaceRoot(flowPath) ?: return null
+        val commandsDir = getCommandsDirectory(workspaceRoot) ?: return null
+        return commandsDir.resolve("$commandName.yaml").takeIf { it.exists() }
+    }
+}
--- a/e2e/workspaces/custom_commands/commands/login.yaml
+++ b/e2e/workspaces/custom_commands/commands/login.yaml
@@ -1,0 +1,10 @@
+appId: com.ryan.CustomCommands
+params:
+  email: { required: true }
+  password: { required: true }
+---
+- tapOn: "email"
+- inputText: ${email}
+- tapOn: "password"
+- inputText: ${password}
+- tapOn: "Submit"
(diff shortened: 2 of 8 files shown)


Fishbowler:
Another consideration: test maintenance.

If a test fails in CI, you wanna run just that test and see it fail locally so that you can make changes. With flow syntax tied to the config, you'd need to do something fun with tags to run just one test, I think?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/CustomCommandLoader.kt
```diff
@@ -0,0 +1,27 @@
...
+        var current = flowPath.toAbsolutePath().parent
+        while (current != null) {
+            if (current.resolve("config.yaml").exists()) return current
+            current = current.parent
+        }
+        return null
+    }
+
+    fun getCommandsDirectory(workspaceRoot: Path): Path? {
+        val commandsDir = workspaceRoot.resolve("commands")
```
Fishbowler, line 18: If you're already using the config to determine the workspace root, maybe it makes sense to have a config key to configure it, so that folks can have control of their directory structure, especially since they already have control for custom config locations that aren't necessarily in the root.

If this is the route followed, it might also make sense to make this a list, mirroring the `flows` key.

We'd need to document that when doing this you'd have to specify the config when running an individual flow, since it wouldn't be detected.

ryanpato:
> Another consideration: test maintenance.
> 
> If a test fails in CI, you wanna run just that test and see it fail locally so that you can make changes. With flow syntax tied to the config, you'd need to do something fun with tags to run just one test, I think?

I think i'm misunderstanding you, or the way people run tests sorry. Ignoring studio, the happy path would be `maestro test` from root with config determining flows? Then that fails on the CI and the user wants to run the test locally. Would `maestro test flows/test1.yaml` from the workspace not work?

maestro-orchestra/src/main/java/maestro/orchestra/yaml/CustomCommandLoader.kt
ryanpato: > especially since they already have control for custom config locations that aren't necessarily in the root.

🤯 i think i have some research to do, thank you

Please do let me know if this is a fools errand though, or if there's too many gotchas for a UX. I'm not attached to any code or time spent, so won't be offended if it's not worth pursing!

Fishbowler:
Whilst we're not set on implementing this, I'm gonna close this in favour of #3339 that's got some more recent activity.

Issue #1142, opened 2023-06-09
[Feature Request] broadcast an intent during a flow


andre-artus:
**Is your feature request related to a problem? Please describe.**
Some of my use cases are initiated from a broadcast  (i.e. implements a BroadcastReceiver). I would like to automate the tests for these receivers.  

**Describe the solution you'd like**
Some mechanism to initiate or simulate a broadcast (Local or System), or call adb to do it and check the result.
I would like to test when the following are sent:
 - correct broadcast (data/params)
 - incorrect broadcast ,
 - multiple concurrent broadcast (attempt to overload)
 Also
 - no ANR 
 - What happens when app is in non-interrupt state (e.g. user filling in a form) and intent is broadcast
- Behaviour under Doze, standby or process death.
- Test order of broadcast receipt (if dependent on order).

I'm sure I'm leaving something out, but that should cover a lot of it.

**Describe alternatives you've considered**
I'm currently just using scripts calling into ADB with a variety of known values and logging the interactions.

amanjeetsingh150:
Hey @andre-artus can you explain the detailed use case you are looking at? What type of behavior do you want to verify in the maestro test?

MaaxGr:
We use broadcast Intents to simulate the behaviour of the physical scan engine on enterprise devices (as Zebra TC58) on emulators:
![image](https://github.com/mobile-dev-inc/maestro/assets/43404580/c53f725c-753c-4bde-915e-c47e44b1ccbd)

Example: Send Barcode "test" to our application:
adb shell am broadcast -a de.guettlerlogistik.hallenapp.scan --es barcode "test" de.guettlerlogistik.hallenapp

That should also be supported with maestro. :)

Have a great week.

Max

andre-artus:
Sorry, @amanjeetsingh150 , I failed to respond to you (I don;t know if I missed the notification) but Max's examples is very similar to ours (scanner or other accessories on enterprise devices).

MaaxGr:
@andre-artus How did you solved it with scripts calling into ADB. Can you share an example?

bartekpacia:
Hey all,

This seems to a bit of a more niche use case, and we're a small team busy with work, so I cannot share any timeline on when we will implement it. That said if anyone decided to contribute this feature, we're always happy to review PRs.

LukStankovic:
Yes, this feature would be extremely useful for my workflow too, as I need to test an application on a Zebra scanner 🙏🏻

PR #2925, opened 2026-01-13
Add sendBroadcast feature

markushi:
## Proposed changes

Implements a new `sendBroadcast` command for sending out signals which can be retrieved by an Android `BroadcastReceiver` component. 
First time contributing, if you have any feedback please let me know!

## Testing

Added tests by following existing patterns + some manual testing on top.

## Issues fixed

[Support for `sendBroadcast`.](https://github.com/mobile-dev-inc/Maestro/issues/1142)

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/ListCloudDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListCloudDevicesCommand.kt
@@ -35,9 +35,7 @@ class ListCloudDevicesCommand : Callable<Int> {
         TestDebugReporter.install(null, printToConsole = parent?.verbose == true)
 
         val platformFilter = platform?.let { input ->
-            Platform.fromString(input) ?: throw CliError(
-                "Unknown platform: '$input'. Supported values: ${Platform.entries.joinToString { it.name.lowercase() }}"
-            )
+            Platform.fromString(input)
         }
 
         val apiClient = ApiClient(EnvUtils.BASE_API_URL)
--- a/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
@@ -36,9 +36,7 @@ class ListDevicesCommand : Callable<Int> {
         TestDebugReporter.install(null, printToConsole = parent?.verbose == true)
 
         val platformFilter = platform?.let { input ->
-            Platform.fromString(input) ?: throw CliError(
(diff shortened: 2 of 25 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -678,6 +678,40 @@ data class KillAppCommand(
...
+    override val optional: Boolean = true
+
+    override val originalDescription: String
+        get() {
+            var result = "Broadcast $action"
+            if (receiver != null) {
+                result += " to $receiver"
+            }
+            if (extras != null && extras.isNotEmpty()) {
+                result += " (extras: $extras)"
```
Fishbowler: Will this always coerce to string cleanly?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -678,6 +678,40 @@ data class KillAppCommand(
...
+        }
+
+    override fun evaluateScripts(jsEngine: JsEngine): Command {
+        return copy(
+            action = action.evaluateScripts(jsEngine),
+            receiver = receiver?.evaluateScripts(jsEngine),
+            extras = extras?.entries?.associate {
+                val value = it.value
+                it.key.evaluateScripts(jsEngine) to if (value is String) value.evaluateScripts(jsEngine) else it.value
+            },
```
Fishbowler, line 739: Where both key and value could be expressions, what would this look like in YAML?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
markushi: Yes, Kotlin's `Map` produces some pretty output, but in theory the extras could contain sensitive data. Let me remove it instead to avoid any future hassle.

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
markushi: Good catch, IMHO it actually makes little sense to evaluate the key anyway. Let me fix that.

markushi:
@Fishbowler any chance you could have another look at this?

markushi:
@Fishbowler in case there's still interest in getting this merged, please let me know and I can take care of the merge conflicts.

markushi:
@Fishbowler  I've updated the PR once more, but will refrain from investing more time into this, as there seems to be little interest in getting this merged.

PR #2927, opened 2026-01-14
Fix: Preserve original errors and display onFlowComplete output on failures

erik-agens:
## Summary

Fixes two related issues with onFlowComplete error handling:

1. **onFlowComplete output was hidden when JS errors occurred** - When a test failed with a JavaScript error, the polyglot stacktrace would be displayed but the onFlowComplete commands and their console output would be completely swallowed
2. **Original test failures were overwritten** - When both the test AND onFlowComplete failed, only the onFlowComplete error was shown, making it impossible to debug the actual cause of the test failure

This PR ensures onFlowComplete commands are always displayed (even on failure) and both errors are shown when dual failures occur.

### Background

From [Slack discussion](https://maestro--dev.slack.com/archives/C041FU72T54/p1768301314388559):

**Problem 1: onFlowComplete output was hidden on JS errors**

When a test like this failed:
```yaml
appId: ${APP_ID}
onFlowComplete:
  - runScript: debugFlowComplete.js  # Logs test results
---
- runScript: config.js  # Throws error
```

The entire polyglot stacktrace was printed, but the onFlowComplete section and its console logs were completely missing from the output. Users couldn't see their debug/cleanup output because the stacktrace swallowed it.

**Problem 2: Dual failures overwrote original errors**

When testing a local fix for Problem 1, a new edge case was discovered: if the test fails AND onFlowComplete also throws an error, the onFlowComplete error would completely replace the original test failure, hiding the root cause.

### Key Changes

**1. Always Display onFlowComplete Commands**
- Modified `AnsiResultView.kt` to track and display executed onFlowComplete commands even when errors occur
- Shows console logs and command status from onFlowComplete regardless of test outcome
- Preserves the valuable debug information that was previously hidden

**2. Error Preservation for Dual Failures**
- Created `OnFlowCompleteFailedException` to wrap both the original test failure and the onFlowComplete failure
- Modified `Orchestra.kt` to catch onFlowComplete exceptions and wrap them with the original error
- Updated `MaestroCommandRunner.kt` to detect dual failures (assertion + onFlowComplete)

**3. Enhanced Error Display**
- Primary error (original test failure) displayed in red
- Yellow separator line (━━━) to visually distinguish errors
- Secondary error (onFlowComplete failure) displayed in yellow with "Note: onFlowComplete also failed:"
- Both errors are preserved and accessible for debugging

**4. Improved Error Messages**
- Added "Script failed:" prefix for JavaScript/PolyglotException errors to clarify error type
- Show first line of stacktrace with file location: `at config.js:1`
- Added reference to maestro.log for full polyglot stacktraces
- Cleaner, more readable error output vs. raw stacktraces

**5. Comprehensive Test Coverage**
- Added 3 new integration test cases covering all error scenarios:
  - Case 137: Flow fails with assertion AND runScript throws JS error
  - Case 138: Flow fails with assertion AND onFlowComplete fails  
  - Case 139: Flow fails with JS error AND onFlowComplete fails

#### Test 1: Assertion failure only

Standard assertion failure output remains unchanged:

test1_assertion_only

#### Test 2: Assertion + onFlowComplete both fail ⭐️

This demonstrates the **key fix** - both errors are shown with clear visual separation:

test2_assertion_and_onflowcomplete_fail

- ✅ onFlowComplete commands are executed and shown
- ✅ Primary error (assertion) appears first in red
- ✅ Yellow separator line clearly divides the errors
- ✅ Secondary error (onFlowComplete) appears below with "Note: onFlowComplete also failed:"

#### Test 3: JS error + onFlowComplete both fail

Similar to Test 2, but with a JS error as the primary failure:

test3_js_error_and_onflowcomplete_fail

- ✅ Shows "Script failed:" prefix for JS errors
- ✅ Both errors preserved with clear hierarchy

#### Test 4: Only onFlowComplete fails

When the test passes but onFlowComplete fails:

test4_onflowcomplete_only

- ✅ onFlowComplete error shown clearly
- ✅ Concise error message with file location

### Test Plan

- ✅ All existing tests pass (143+ integration tests)
- ✅ 3 new integration tests added and passing
- ✅ Manual testing with various error scenarios (screenshots above)
- ✅ Verified backward compatibility (optional fields, null-safe operations)
- ✅ No breaking changes to report generation or serialization

### Benefits

- **Better debugging**: See onFlowComplete debug logs even when tests fail
- **Root cause analysis**: Original errors are never lost, even if cleanup fails
- **Clear visual hierarchy**: Color coding and separators make it obvious which error is primary
- **Cleaner output**: Concise error messages with references to full logs instead of huge stacktraces

### Development Notes

This implementation was developed collaboratively with Claude Code (Sonnet 4.5) through an iterative process:

1. **Initial exploration** - Analyzed codebase patterns, error handling flow, and existing test structure
2. **Planning phase** - Created comprehensive implementation plan addressing all edge cases
3. **Implementation** - Developed `OnFlowCompleteFailedException`, updated error handling in Orchestra/TestRunner/MaestroCommandRunner, enhanced UI rendering
4. **Testing** - Added integration tests covering all error scenarios and performed manual verification
5. **Code review** - Audited comment patterns, checked for breaking changes, verified backward compatibility

The entire development process, including code exploration, implementation, testing, and documentation, was conducted through pair programming with Claude Code.

### DIFF
--- a/maestro-test/src/test/resources/138_assertion_and_onflowcomplete_fail.yaml
+++ b/maestro-test/src/test/resources/138_assertion_and_onflowcomplete_fail.yaml
@@ -1,5 +1,5 @@
 appId: com.example.app
 onFlowComplete:
-  - runScript: 137_failing_cleanup.js
+  - runScript: 138_failing_cleanup.js
 ---
 - assertVisible: "NonExistentElement"


maestro-test/src/test/resources/137_assertion_and_runscript_js_error.yaml
```diff
@@ -0,0 +1,4 @@
+appId: com.example.app
+---
+- runScript: 137_failing_script.js
+- assertVisible: "NonExistentElement"
```
Fishbowler: The assertion here won't ever run, so this won't behave any differently if you omit the second line.

maestro-cli/src/main/java/maestro/cli/view/ErrorViewUtils.kt
```diff
@@ -3,20 +3,37 @@ package maestro.cli.view
...
+            is PolyglotException -> {
+                // Get the first stack trace element which should contain the JS location
+                val firstTrace = e.stackTrace.firstOrNull()?.toString() ?: ""
+                // Extract path from "at <js>.:program(/path/to/file.js:line)"
+                val location = if (firstTrace.contains("(") && firstTrace.contains(")")) {
+                    firstTrace.substringAfter("(").substringBefore(")")
+                } else {
+                    firstTrace.ifEmpty { "unknown location" }
+                }
+                "Script failed: ${e.message}\n    at $location\n\nCheck maestro.log for full polyglot stacktrace."
```
Fishbowler:
```suggestion
                "Script failed: ${e.message}\n    at $location\n\nCheck maestro.log for full stacktrace."
```

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -927,13 +945,34 @@ class Orchestra(
...
+
+                // If no exceptions, return the combined success result
+                if (exception == null && onCompleteException == null) {
+                    return onCompleteSuccess && flowSuccess
+                }
             }
-            onCompleteSuccess && flowSuccess
+
+            // This should never be reached due to throws above, but compiler needs it
+            return false
```
Fishbowler: Given the throws...

* Is the if block necessary? Won't that always evaluate to true at that point?
* Can't this be simplified to returning `return onCompleteSuccess && flowSuccess` outside the finally block? Static analysis isn't a fan of returns inside finally. (Would require rescoping those vals to vars further up.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4215,6 +4217,59 @@ class IntegrationTest {
...
+                    orchestra(it).runFlow(commands)
+                }
+            }
+        }
+
+        // Verify both exceptions are preserved
+        assertThat(exception.originalException).isInstanceOf(PolyglotException::class.java)
+        assertThat(exception.onFlowCompleteException).isInstanceOf(PolyglotException::class.java)
+    }
+
```
Fishbowler, line 4332: One more test for a failing assertion but passing onFlowComplete?

Fishbowler (changes requested):
A few little bits.

Also a finding from local testing: If both the flow fails and the onFlowComplete fails, but the onFlowComplete failure _isn't_ JS, then the old behaviour still persists. Should these be consistent?

```
appId: com.example.app
onFlowComplete:
  - assertTrue: ${false}
---
- assertTrue: ${false}
```

maestro-test/src/test/resources/137_assertion_and_runscript_js_error.yaml
erik-agens: Good catch! This test does not really make sense at all to be honest. Because a test assert and test runScript failure will never happen together. Whichever comes first will stop the test. I'll remove this test all together.

maestro-cli/src/main/java/maestro/cli/view/ErrorViewUtils.kt
erik-agens: I agree, I'm not even sure what polyglot is - better to make it more general. I'll adjust!

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
erik-agens: yeah, good catch on both points. I'll do the following refactor:                                                                                                                                                                            
- Move `onCompleteSuccess` declaration outside the finally block
- Remove the redundant if check
- Return outside finally to avoid static analysis issues

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
erik-agens: yes, makes sense! I'll add it!

erik-agens:
> A few little bits.
> 
> Also a finding from local testing: If both the flow fails and the onFlowComplete fails, but the onFlowComplete failure _isn't_ JS, then the old behaviour still persists. Should these be consistent?
> 
> ```
> appId: com.example.app
> onFlowComplete:
>   - assertTrue: ${false}
> ---
> - assertTrue: ${false}
> ```

@Fishbowler  Thanks for the thorough review!                                                                                                                     
                                                                                                                                                                                                                 
Great catch on the double assertion when none was js - missed this completely!                                                                      
I added a new test case that verifies dual assertion failures - should show both errors with proper visual hierarchy.

Let me know if you find something else and I'll do my best to address it!

PR #2086, opened 2024-10-09
feat: add partial screenshots capability

TheKohan:
## Proposed changes
This PR extends current capabilities of `takeScreenshot` command with `targetComponentId` prop.

## Motivation
Visual regression testing is something that has been on the horizon for some time #1222. This PR is aiming to fill the gaps and prepare the ground for further development in this area (will fit #2078 great). Component level screenshots are something that will greatly decrease screenshot size and make component-level tests possible.

### Example usage

`test-screenshot.yml`
```yml
appId: com.example.app
---
- launchApp
- takeScreenshot: 
      path: "TestScreen"
```
```terminal
 ║                                                             
 ║  > Flow: test-flow                                          
 ║                                                             
 ║    ✅   Launch app "com.exapmle.com"                                                        
 ║    ✅   Tap on "Test Button"Screen                          
 ║    ✅   Take screenshot TestScreen                          
 ║            
```
`test-component-screenshot.yml`
```yml
appId: com.example.app
---
- launchApp
- takeScreenshot: 
      path: "TestScreen"
      targetComponentId: "test-btn"
```
```terminal
 ║                                                             
 ║  > Flow: test-flow                                          
 ║                                                             
 ║    ✅   Launch app "com.exapmle.com"      
 ║    ✅   Tap on "Test Button"
 ║    ✅   Take screenshot TestScreen, cropped on component with id test-btn
```

| Screenshot | Component Screenshot |
|----------|---------|
| Login screen | TestScreen |

## Testing
* would gladly accept any ideas of how to test it with confidence

PR #2933, opened 2026-01-16
feat: add cropOn screenshots capability

SosenWiosen:
## Proposed changes
This PR extends current capabilities of takeScreenshot command with cropOn prop which is of an ElementSelector type.

## Motivation
The motivation has been explained in #2086. This PR aims to get @TheKohan 's feature merged.
### Example usage

`takeScreenshot.yaml`

```yaml
appId: com.example.example
---
- launchApp
- tapOn: "Nesting Test"

- takeScreenshot:
    path: cropped_screenshot
    cropOn:
      id: "level-1"
```

```
 ║                                                       
 ║  > Flow: takeScreenshot                               
 ║                                                       
 ║    ✅   Launch app "com.example.example"              
 ║    ✅   Tap on "Nesting Test"                         
 ║    ✅   Take screenshot cropped_screenshot, cropped to id: level-1                                             
 ║              
```

| Nesting Test    | Cropped |
| -------- | ------- |
| Simulator Screenshot - iPhone 16 Pro - 2026-01-16 at 20 35 42  |  cropped_screenshot  |

## Testing

A simple test flow has been added. My idea for better e2e testing this would be to save a cropped screenshot of some view that doesn't change between platforms in the test apps - with fixed dimensions, etc.  Subsequent test runs would then compare the new screenshots to this one.  This would require adding some additional script and some special handling in `run_tests`.

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -506,62 +506,55 @@ class Maestro(
         }
     }
 
-    fun takeScreenshot(sink: Sink, compressed: Boolean) {
-        LOGGER.info("Taking screenshot")
-
-        sink
-            .buffer()
-            .use {
-                ScreenshotUtils.takeScreenshot(it, compressed, driver)
-            }
-    }
-
-    /**
-     * Takes a screenshot cropped to the given bounds.
-     * Bounds are in "grid" units: on Android they match pixels (grid == pixels), on iOS they are points (logical).
-     * We convert to screenshot pixel coordinates using scale = heightPixels/heightGrid, which is 1.0 on Android
-     * and the device scale factor on iOS, so one implementation works correctly on both platforms.
-     */
-    fun takePartialScreenshot(sink: Sink, bounds: Bounds, compressed: Boolean) {
-        LOGGER.info("Taking partial screenshot")
-        val (x, y, width, height) = bounds
-
-        val originalImage = Buffer().apply {
-            ScreenshotUtils.takeScreenshot(this, compressed, driver)
-        }.let { buffer ->
-            buffer.inputStream().use { ImageIO.read(it) }
-        }
-
-        val info = cachedDeviceInfo
-        val scale = if (info.heightGrid > 0) {
-            info.heightPixels.toDouble() / info.heightGrid
+    fun takeScreenshot(sink: Sink, compressed: Boolean, bounds: Bounds? = null) {
+        if (bounds == null) {
+            LOGGER.info("Taking screenshot")
+            sink
+                .buffer()
+                .use {
+                    ScreenshotUtils.takeScreenshot(it, compressed, driver)
+                }
         } else {
-            1.0
-        }
-        val startX = (x * scale).toInt().coerceIn(0, originalImage.width)
-        val startY = (y * scale).toInt().coerceIn(0, originalImage.height)
-        val cropWidthPx = (width * scale).toInt()
-            .coerceIn(0, originalImage.width - startX)
-        val cropHeightPx = (height * scale).toInt()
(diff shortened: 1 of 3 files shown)


SosenWiosen:
Hi, @Fishbowler, @Leland-Takamine could you take a look at this? Thanks!

Leland-Takamine:
Hey @SosenWiosen - thanks for reviving this PR! Copying my comment from #2086 as it still applies:

> The main thing missing here is ensuring that our test-e2e.yaml workflow validates that this command is indeed working as intended. Ie: When this command is run, does the cropOn option end up producing the expected image?
> 
> One way do to this:
> 
> * Add a flow to our end to end test suite that uses this command
> * After the test suite runs, validate that the screenshot produced matches an image committed to the codebase (perhaps using the compare command from imagemagick - open to other ideas on what to use for this)
> 
> If you're able to add end to end test coverage and rebase then we can get this merged.

e2e/run_tests
```diff
@@ -105,6 +105,15 @@ for workspace_dir in ./workspaces/*; do
     fi
 done
 
+###
+### Validate screenshots
+###
+if [ "$ALL_PASS" = "true" ]; then
+	_h2 "screenshot validation" "comparing screenshots"
+	chmod +x ./validate_screenshots.sh
```
Leland-Takamine: Any reason not commit the file as executable?

SosenWiosen:
This turned out a little more complicated than I would like. I've had to add a separate script for image validation, and also a separate PR to the demo app - the other available screens didn't have a view which would be of consistent size between devices.
https://github.com/mobile-dev-inc/demo_app/pull/9

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4253,6 +4254,45 @@ class IntegrationTest {
...
+        // No test failure
+        driver.assertEvents(
+            listOf(
+                Event.TakeScreenshot,
+            )
+        )
+        val file = File("138_take_cropped_screenshot_with_filename.png")
+        val image = ImageIO.read(file)
+
+        assert(file.exists())
```
Fishbowler: Do these two lines need to be swapped?

Leland-Takamine:
Spoke with @SosenWiosen on slack and we've decided to land the assertScreenshot changes first in order to use assertScreenshot to simplify the end to end testing for this PR!

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
SosenWiosen: I think it shouldn't matter.

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -514,6 +516,54 @@ class Maestro(
             }
     }
 
+    /**
+     * Takes a screenshot cropped to the given bounds.
+     * Bounds are in "grid" units: on Android they match pixels (grid == pixels), on iOS they are points (logical).
+     * We convert to screenshot pixel coordinates using scale = heightPixels/heightGrid, which is 1.0 on Android
+     * and the device scale factor on iOS, so one implementation works correctly on both platforms.
+     */
+    fun takePartialScreenshot(sink: Sink, bounds: Bounds, compressed: Boolean) {
```
Leland-Takamine: Instead of a separate takePartialScreenshot method, can we just add an optional bounds parameter to the existing takeScreenshot method?

SosenWiosen: That's sensible, will do it.

Leland-Takamine:
This looks good to go. Waiting on internal green light before merging

PR #2934, opened 2026-01-17
fix(android): Add gRPC retry logic and channel recreation for UNAVAILABLE errors

olivier-motium:
## Summary

This PR fixes a long-standing issue where Maestro intermittently fails to communicate with Android emulators/devices, presenting the error:

```
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
Connection refused: localhost/[0:0:0:0:0:0:0:1]:7001
```

The current workaround (`rm -f ~/.maestro/session`) works but requires manual intervention and degrades user experience significantly.

## Root Cause Analysis

After investigating the gRPC connection lifecycle in `AndroidDriver.kt`, I identified three contributing factors:

| Issue | Impact | Current Behavior |
|-------|--------|------------------|
| **Immutable channel** | Connection cannot recover after termination | `channel` declared as `val`, never recreated |
| **Inconsistent retry logic** | Random failures depending on which call hits a dead connection | Only `callViewHierarchy()` retries; `runDeviceCall()` fails immediately |
| **Keep-alive mismatch** | Premature connection termination | Client: 2 min interval, Server: 30s minimum |

The gRPC channel, once terminated due to network issues, device disconnection, or idle timeout, cannot be recovered. Subsequent calls fail immediately instead of attempting reconnection.

### 1. Channel Recreation on Connection Loss

Made `channel`, `blockingStub`, and `asyncStub` mutable to support recreation:

```kotlin
private var channel = ManagedChannelBuilder.forAddress("localhost", hostPort)
    .usePlaintext()
    .keepAliveTime(45, TimeUnit.SECONDS)
    .keepAliveTimeout(20, TimeUnit.SECONDS)
    .keepAliveWithoutCalls(true)
    .build()
```

Added `recreateChannelIfNeeded()` that checks channel state and rebuilds if shutdown/terminated:

```kotlin
private fun recreateChannelIfNeeded(): Boolean {
    if (channel.isShutdown || channel.isTerminated) {
        // Recreate channel and stubs
        return true
    }
    return false
}
```

### 2. Unified Retry Logic

Added consistent 3-attempt retry with 1-second backoff to both `runDeviceCall()` and `callViewHierarchy()`:

```kotlin
private fun <T> runDeviceCallWithRetry(call: () -> T, remainingAttempts: Int): T {
    return try {
        call()
    } catch (throwable: StatusRuntimeException) {
        when (Status.fromThrowable(throwable).code) {
            Status.Code.UNAVAILABLE -> {
[...]

PR #2935, opened 2026-01-17
fix(mcp): Auto-install Maestro driver APKs when starting Android devices via MCP

olivier-motium:
### Problem

When using the MCP `start_device` tool to boot an Android emulator, the tool successfully starts the device but **does not install the Maestro driver APKs**. This causes all subsequent MCP tool calls (like `inspect_view_hierarchy`, `tap_on`, `run_flow`) to fail with:

```
io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 119.981718041s.
[closed=[], open=[[buffered_nanos=..., waiting_for_connection]]]
```

### Root Cause Analysis

The `StartDeviceTool` was the only MCP tool that didn't use `MaestroSessionManager` for driver initialization:

| MCP Tool | Uses SessionManager | Driver Initialized |
|----------|--------------------|--------------------|
| `start_device` | ❌ No | ❌ No |
| `inspect_view_hierarchy` | ✅ Yes | ✅ Yes |
| `run_flow` | ✅ Yes | ✅ Yes |
| `tap_on` | ✅ Yes | ✅ Yes |
| `launch_app` | ✅ Yes | ✅ Yes |
| (all other tools) | ✅ Yes | ✅ Yes |

The `StartDeviceTool` only called `DeviceService.startDevice()` which boots the emulator but does **not**:
1. Install `maestro-app.apk` and `maestro-server.apk`
2. Start instrumentation via `am instrument`
3. Set up gRPC port forwarding (`adb forward tcp:7001 tcp:7001`)

### Solution

After `DeviceService.startDevice()` successfully boots an Android device, we now call `sessionManager.newSession()` to initialize the Maestro driver. This is the same pattern used by all other MCP tools.

```kotlin
// For Android devices, initialize driver to install APKs and start instrumentation
if (connectedDevice.platform == Platform.ANDROID) {
    sessionManager.newSession(
        host = null,
        port = null,
        driverHostPort = null,
        deviceId = connectedDevice.instanceId,
        platform = "android"
    ) { _ ->
        // Session initialization installs driver APKs and starts instrumentation
        // The session is then closed but the driver remains running
    }
}
```

### Why This Works

When `MaestroSessionManager.newSession()` is called:
1. `selectDevice()` identifies the Android device
2. `createMaestro()` calls `createAndroid()` which:
   - Creates an `AndroidDriver` with `reinstallDriver = true` (default)
   - The driver constructor installs APKs if not present
   - Starts instrumentation and waits for gRPC connection
3. The session is closed via the block, but the driver remains running on the device
4. Subsequent MCP tool calls find an active driver and connect immediately

### Changes

- **`StartDeviceTool.kt`**: 
  - Accept `MaestroSessionManager` parameter in `create()` function
  - Call `sessionManager.newSession()` after booting Android devices (both `device_id` and `platform` code paths)
  
- **`McpServer.kt`**: 
  - Pass `sessionManager` to `StartDeviceTool.create()`

### Manual Testing Steps

1. **Clean state**: Ensure driver APKs are not installed
   ```bash
   adb uninstall dev.mobile.maestro
   adb uninstall dev.mobile.maestro.test
   adb kill-server  # Optional: ensure clean ADB state
   ```

2. **Build and install**:
   ```bash
   ./gradlew :maestro-cli:installDist
   ```

3. **Test via MCP** (using any MCP client):
   ```
   # Start the device
   mcp__maestro__start_device(platform: "android")
   
   # Verify driver works immediately (should NOT timeout)
   mcp__maestro__inspect_view_hierarchy(device_id: "emulator-5554")
   ```

### Expected Results

| Scenario | Before | After |
|----------|--------|-------|
| Fresh emulator, `start_device` then `inspect_view_hierarchy` | ❌ DEADLINE_EXCEEDED after 120s | ✅ Returns hierarchy immediately |
| Fresh emulator, `start_device` then `run_flow` | ❌ DEADLINE_EXCEEDED | ✅ Flow executes |
| Already-running emulator with driver | ✅ Works (reconnects) | ✅ Works (no change) |
| iOS simulator | ✅ Works (no driver needed) | ✅ Works (no change) |

### Edge Cases Verified

- ✅ Device already running with driver installed - no duplicate installation
- ✅ iOS simulators - no Android-specific initialization attempted
- ✅ `device_id` parameter path - driver initialized
- ✅ `platform` parameter path - driver initialized

## Issues fixed

This PR addresses the UX issue where MCP users must manually install Maestro driver APKs before using MCP tools, which defeats the purpose of automated device management through MCP.

---

**Note**: This is separate from the gRPC channel recovery fix in PR #2934. That fix handles reconnection after connection loss; this fix ensures the connection is established in the first place.


Fishbowler:
Why would the lack of drivers cause the deadline exceeded? Suggests that there's something else wrong - why would installing the driver and taking a screenshot take more than 120s?

olivier-motium:
Great question! Let me explain the flow:

## Why DEADLINE_EXCEEDED (not a better error)?

The 120s timeout comes from the gRPC client configuration in `AndroidDriver.kt:83`:
```kotlin
private val blockingStubWithTimeout get() = blockingStub.withDeadlineAfter(120, TimeUnit.SECONDS)
```

### What happens without this fix:

1. **User calls `start_device`** → Device boots successfully ✅
2. **User calls `inspect_view_hierarchy`** → This creates a new session via `sessionManager.newSession()`
3. **Session creation** triggers `AndroidDriver` initialization which:
   - Installs APKs (if `reinstallDriver=true`, which it is by default)
   - Starts instrumentation (`am instrument`)
   - **Waits for gRPC connection** on port 7001

4. **The wait loop** in `AndroidDriver.kt:126-168` polls for the gRPC server:
```kotlin
while (System.currentTimeMillis() - startTime < getStartupTimeout()) {
    // Try to connect...
    Thread.sleep(500)
}
throw AndroidDriverTimeoutException("Maestro Android driver did not start up in time")
```

The startup timeout is 15 seconds (`SERVER_LAUNCH_TIMEOUT_MS`). But here's the issue: **the APK installation and instrumentation start can take significant time**, and if something fails silently (APK already partially installed, instrumentation crashes, etc.), the gRPC server never comes up.

### The real timing breakdown:

When calling a tool like `inspect_view_hierarchy` on a fresh device:
- APK installation: 5-15s (depending on device speed)
- Instrumentation start: 2-5s
- gRPC channel establishment: 1-2s
- **If instrumentation fails**: waits full 15s startup timeout, then the gRPC call itself has a 120s deadline

The error you see (`DEADLINE_EXCEEDED`) is from the **gRPC call deadline**, not the driver startup timeout. By the time the gRPC call is made, the driver thinks it started (no exception thrown), but the server isn't actually responding.

### Why this fix helps:

By calling `sessionManager.newSession()` in `start_device` for Android:
1. Driver initialization happens **during device startup** (user expects this to take time)
2. Any failures are surfaced **before** the user tries to use other tools
3. Subsequent tool calls find a working driver and connect immediately

---

**Side note**: The error message could definitely be improved. The gRPC layer doesn't know *why* the connection failed - it just knows it exceeded the deadline. A follow-up improvement could add better diagnostics (check if driver is installed, check if instrumentation is running, etc.) before attempting gRPC calls.

Fishbowler:
I'm not sure I like polluting start_device with something unrelated. The same problem you describe would occur if I'd already started a device by other means. 

From your description above, perhaps a better method would be to complete driver installation before the 15s timer begins?

olivier-motium:
## Summary of Revised Changes (per reviewer feedback)

This update addresses @fishbowler's feedback to avoid polluting `StartDeviceTool` with session management.

### Changes Made

#### 1. `DeviceService.kt` (maestro-client)
- Added `installDriver: Boolean = false` parameter to `startDevice()` method
- After boot completes for Android, if `installDriver` is true, installs the Maestro driver APKs:
```kotlin
if (installDriver) {
    PrintUtils.message("Installing Maestro driver...")
    val driver = AndroidDriver(dadb, driverHostPort)
    driver.installMaestroDriverApp()
}
```

#### 2. `StartDeviceTool.kt` (maestro-cli)
- **Removed** the `sessionManager` dependency completely
- **Removed** the workaround that used `sessionManager.newSession()` to install drivers
- Now passes `installDriver = available.platform == Platform.ANDROID` to `DeviceService.startDevice()`

#### 3. `McpServer.kt` (maestro-cli)
- Updated call site from `StartDeviceTool.create(sessionManager)` to `StartDeviceTool.create()`

### Why This Approach Is Better

| Aspect | Benefit |
|--------|---------|
| **Correct abstraction** | Driver installation is device preparation, belongs in `DeviceService` |
| **Opt-in behavior** | Default `false` preserves existing behavior for CLI and other callers |
| **Works for any startup method** | Benefits anyone using `DeviceService.startDevice()` |
| **Simpler code** | No session manager dependency in `StartDeviceTool` |
| **Follows existing pattern** | Same approach already used for locale setting in `DeviceService` |

### Build Status
✅ `./gradlew :maestro-cli:installDist` completed successfully

olivier-motium:
## CI Re-run Triggered

The previous CI run failed due to a flaky test:

```
AndroidIntegrationTest > test travel command works as expected() FAILED
    Assertion is false: ".*Eiffel.*" is visible
```

This test simulates GPS travel to Eiffel Tower coordinates and checks for "Eiffel" visibility - it's inherently flaky due to UI/network dependencies.

**Evidence this is a flaky test (not related to these changes):**
- The previous CI run on this exact same branch **passed**
- The current CI run on the same branch **failed**
- The failing test is unrelated to the MCP `start_device` functionality changed in this PR

I've pushed an empty commit to trigger a fresh CI run.

Fishbowler:
Should we be limiting this to Android? Will the same issue also occur on iOS? They've got a driver that needs installing too.

PR #2941, opened 2026-01-20
fix(recording): disable ANSI output when --no-ansi is set

### LINKED ISSUE #1009: --no-ansi does not work when using "maestro record"
While using `maestro record`, I noticed that it seems to ignore the `--no-ansi` option. When using `maestro test`, it works just fine. Currently using Maestro version 1.26.1.

Command that doesn't work;
```
maestro --no-ansi record {flow path}
```

Command that does work;
```
maestro --no-ansi test {flow path}
```
https://user-images.githubusercontent.com/68027680/233732432-d2217c7e-5f0f-40e8-9b71-d58386cb9f43.mp4

mohswell:
## Proposed changes
The record command was always creating an AnsiResultView() regardless of whether --no-ansi was passed, while the test command properly checked DisableAnsiMixin.ansiEnabled to decide between AnsiResultView and PlainTextResultView.

> While using maestro record, I noticed that it seems to ignore the --no-ansi option. When using maestro test, it works just fine

Added missing imports:
1. PlainTextResultView - for non-ANSI output
2. EnvUtils - to detect Windows (which disables emojis)

So in my changes, I have just replicated the exact same pattern used in TestCommand (line 550-553), ensuring consistent behavior.
```kotlin
val resultView = if (DisableAnsiMixin.ansiEnabled) {
    AnsiResultView(useEmojis = !EnvUtils.isWindows())
} else {
    PlainTextResultView()
}
```

What This Fixes
- maestro --no-ansi record {flow path} (strips ANSI codes)
- maestro --no-color record {flow path} (alias works too)

## Testing
I have tested my changes locally 

## Issues fixed
Resolved #1009

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -147,7 +147,7 @@ class RecordCommand : Callable<Int> {
                     )
                 }
 
-                val resultView = AnsiResultView(useEmojis = DisableAnsiMixin.ansiEnabled && !EnvUtils.isWindows())
+                val resultView = AnsiResultView(useEmojis = !EnvUtils.isWindows())
                 val screenRecording = kotlin.io.path.createTempFile(suffix = ".mp4").toFile()
                 val exitCode = screenRecording.sink().use { out ->
                     maestro.startScreenRecording(out).use {
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -32,7 +32,6 @@ import maestro.cli.graphics.SkiaFrameRenderer
 import maestro.cli.report.TestDebugReporter
 import maestro.cli.runner.TestRunner
 import maestro.cli.runner.resultview.AnsiResultView
-import maestro.cli.runner.resultview.PlainTextResultView
 import maestro.cli.session.MaestroSessionManager
 import maestro.cli.util.EnvUtils
 import maestro.cli.util.FileUtils.isWebFlow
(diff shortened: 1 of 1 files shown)


Fishbowler:
@mohswell Thanks! This will help folks looking to generate test evidence in CI!

Have you got some additional local changes that you used in your local testing? This doesn't compile because the PlainTextResultView doesn't have any of the Frame stuff.

Maybe that belongs on the ResultView interface, if it's needed for recordings?

mohswell:
hi @Fishbowler encountered a local build environment issue, I need guidance

- Local PC environment: Java 25, Windows 10
- Project requires: Java 17 toolchain
- Error: `No matching toolchains found for requested specification: {languageVersion=17}`

### What I've verified 
- Kotlin compilation passes: `./gradlew :maestro-cli:compileKotlin` ✓
- Changes follow the existing pattern used in TestCommand
- No breaking changes

Ready for ci workflow since GitHub Actions has Java 17
- Would appreciate guidance on:
  1. Is there a workaround for local Java 17 toolchain setup on Windows?

Thanks!

Fishbowler:
Maestro intentionally requires a Java 17 toolchain, since that's the minimum version required to run Maestro so we can't have bytecode that's ahead of that. You could change it to 25 for some local testing - just be sure **not** to commit it to the repo.

There's a `maestro` helper in the root of the repo that runs the local code as if it were installed Maestro. e.g. If you run `./maestro record flow.yaml` in the repo, does that succeed?

mohswell:
@Fishbowler updated my logic a bit. you can review now.

i tried with version 21 seems to build just fine. i was testing the google messages app. 

 .\maestro-cli\build\install\maestro\bin\maestro --no-ansi record test_messages.yaml

i have removed some of the output, becuase it was malformed by my terminal but it managed to start the recording. 
Running on R9PW40HKQBY
 > Flow: test_messagesli/build/instal ║  > Flow: test_messages.jar)
 ║    >   Launch app ║    X   Launch app "com.google.android.apps.messaging"
 ║        Assert that "Google Messages" is visibleuture release unless nati ║ access is enabled ║

Unable to launch app com.google.android.apps.messaging

Fishbowler:
Won't this still be using ansi though?

Fishbowler:
Right now, I think this is a good change, but I don't think it fully resolves the original issue.

But does it just need [this bit](https://github.com/mobile-dev-inc/Maestro/blob/f952bc951413835507b66fadc2ef6e5876f5bec4/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt#L550-L554) from the `maestro test` command?

Fishbowler:
@mohswell Got any energy for continuing with this one?

Fishbowler:
Feel free to reopen this if you get the impetus. I'd love to lend you a hand getting it over the finish line.

PR #2078, opened 2024-10-07
feat: add `assertVisual` command draft

TheKohan:
## Proposed changes
This PR is  built on top of #1898 and implements a basic `assertVisual` flow which is much awaited in the community (#1222). 

Given a simple React Native app

`app.tsx`
```ts
export default function App() {
  const [toggle, setToggle] = React.useState(false);
  return (
    <View style={styles.container}>
      <View style={[styles.dot, {backgroundColor: toggle ? 'red' : 'blue'}]}>
        <Button title="test" onPress={() => setToggle(!toggle)}></Button>
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    marginTop: 200,
  },
  dot: {
    backgroundColor: 'red',
    padding: 30,
    borderRadius: '50%',
  },
});
```

and a flow

`test-flow.yml`
```yml
appId: com.example.app
---
- launchApp
- tapOn: 'test'
- assertVisual: Login screen # compares the current screen against `.maestro/visual_regression/Login screen.png`
```

`maestro` will compare screenshots, producing result image with differences masking.

```terminal
 ║  > Flow: test-flow                                             
 ║                                                                
 ║    ✅   Launch app "com.example.app"                                                              
 ║    ✅   Tap on "test"                                          
 ║    ✅   Tap on "test"                                          
 ║    ❌   Assert visual difference with baseline Login screen (threshold: 95%)                                                     
 ║   
```
| Baseline | Current | Result |
|----------|---------|--------|
| Login screen | image | Login screen |

`Result` for now is being saved in `.maestro/failed_visual_regression/` directory.

If there's no `baseline`, `current` will be saved as one.

### What is missing:
* changes accepting flow
* cloud integration
* testing
* component-scoped tests (partial screenshots for component of given `test-id`)

## Testing

* `in progress` - test plan has not yet been laid out due to potentially big changes in the proposed flow.

## Issues fixed
* #1222

PR #2949, opened 2026-01-22
feat: add new assertScreenshot command

### LINKED ISSUE #1222: [Feature Request] Visual Regression Testing
 It always takes a lot of time to check if certain components looks different/wrong. Tools like Chromatics visual regression testing help a lot with that. They take a screenshot of a component and with every build a new screenshots is taken and being compared against the old one.

- Maybe there should be a command like `- assertVisual: xyz` - which creates a screenshot the first time it is used. the xyz is a keyword so you can compare multiple screenshots and flows
- When the command is used again maestro will check if a screenshot is already created and check if there are visual differences.
  - if there is a difference the flow fails. 
  - you would have to manually check if the visual difference is ok or not
    - if it is ok and wanted you can accept the new screenshot as the news "master"-screenshot. flows will compare against that screenshot in the future

SosenWiosen:
## Proposed changes
This PR implements the assertScreenshot command for visual regression testing. It's based on #2078 by @TheKohan.

Given a simple React Native App

```tsx
import React from 'react';
import { Button, StyleSheet, View } from 'react-native';

export default function App() {
  const [toggle, setToggle] = React.useState(false);
  return (
    <View style={styles.container}>
      <View
        testID="dot"
        style={[styles.dot, { backgroundColor: toggle ? 'red' : 'blue' }]}
      >
        <Button title="test" onPress={() => setToggle(!toggle)} />
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    marginTop: 200,
  },
  dot: {
    backgroundColor: 'red',
    padding: 30,
    borderRadius: '50%',
  },
});

```

And with a sample flow:

```yaml
appId: com.maestrotestapp
---
- launchApp
- tapOn: "test"
- assertScreenshot:
    path: screen.png
    cropOn:
      id: dot
    thresholdPercentage: 95.3
    
```

Maestro will compare the screenshots:

```
 ║                                                                                        
 ║  > Flow: test-flow
 ║                                                                                        
 ║    ✅   Launch app "com.maestrotestapp"
 ║    ✅   Tap on "test"                                                                  
 ║    ❌   Assert screenshot matches screen.png (threshold: 95.3%) (cropped on id: dot)
 ║                                                                                        
  
Comparison error: Assert screenshot matches screen.png (threshold: 95.3%) (cropped on id: dot) - threshold not met, current: 58.970116%
                                                                                          
Screenshot comparison failed. Check the diff image at /Users/sosen/WorkProjects/Maestro/screen_diff.png to see the differences. Adjust the thresholdPercentage if the differences are acceptable.
```

| App | Correct Screenshot | Screenshot with error and diff |
|-----|-------------------|----------------------|
|   Screenshot_1769120930  |    screen |      screen_diff   |

## Testing
Two e2e tests have been added - one which should pass which compares to an already existing screenshot, and one which should fail which compares to a modified screenshot.

## Issues fixed
#1222

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -607,7 +607,7 @@ class Orchestra(
         comparison.apply {
             allowingPercentOfDifferentPixels = thresholdDifferencePercentage
             rectangleLineWidth = 10
-            pixelToleranceLevel = 50.00
+            pixelToleranceLevel = 0.1 
             minimalRectangleSize = 40
         }
 


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -523,6 +534,79 @@ class Orchestra(
...
+        val cropOn = command.cropOn
+        if (cropOn != null) {
+            val elementResult = findElement(cropOn, optional = command.optional)
+            val bounds = elementResult.element.bounds
+            val x = bounds.x.coerceAtLeast(0)
+            val y = bounds.y.coerceAtLeast(0)
+            val width = bounds.width.coerceAtMost(photoNow.width - x)
+            val height = bounds.height.coerceAtMost(photoNow.height - y)
+
+            if (width > 0 && height > 0) {
```
Leland-Takamine: If the element has width zero it seems like this will default to comparing against the full screenshot which I don't think is the expected behavior. I'd expect this case to fail with a specific error message

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -523,6 +534,79 @@ class Orchestra(
         return false
     }
 
+    private suspend fun assertScreenshotCommand(command: AssertScreenshotCommand): Boolean {
+        val path = if (command.path.endsWith(".png")) command.path else command.path + ".png"
```
Leland-Takamine: Seems we're assuming png format in a few places here. Any way we can make this command work regardless of the image format?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
SosenWiosen: Noted!

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
SosenWiosen: I changed it to work based on the suffix in the filename.

SosenWiosen:
Hi, @Fishbowler  what do you think?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -523,6 +534,97 @@ class Orchestra(
...
+        }
+        val diffFileName = "${baseName}_diff.$extension"
+        val diffFile = if (screenshotsDir != null) {
+            screenshotsDir.resolve(diffFileName).toFile()
+        } else {
+            File(diffFileName)
+        }
+
+        val comparison =
+            ImageComparison(expectedImage, actualImage, diffFile)
```
Leland-Takamine, line 605: Looks like we are setting the extension of the diff file to match the extension of the provided path. What's the reasoning for that?

Leland-Takamine:
I believe these should be the last updates needed before merging. Thanks again for the work on this!

* Comment here on diff image format / extension: https://github.com/mobile-dev-inc/Maestro/pull/2949#discussion_r2737588539
* Add a test for jpg input
* Add a test for unsupported image format (and expected error message)

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
SosenWiosen: I thought that's what the users would expect. We could always produce PNG if you think that's right.

Leland-Takamine: My main concern is whether the outputted diff image actually respects the format specified by the extension of the provided path? If we input "foo.jpg" here, will ImageComparison actually output a jpeg image for the diff?

SosenWiosen: I've totally missed that, that's a good catch - It's hardcoded in the imagecomparison library. I'll always save it as png, unless we want to change the library or fork. (I would prefer not)

https://github.com/romankh3/image-comparison/blob/0e9c63792af31dac63b439554dcc0d35c34e0b78/src/main/java/com/github/romankh3/image/comparison/ImageComparisonUtil.java#L81

Leland-Takamine: Yeah we shouldn't fork - let's just make sure we have the correct extension for the diff file

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -523,6 +534,96 @@ class Orchestra(
...
+
+        comparison.apply {
+            allowingPercentOfDifferentPixels = thresholdDifferencePercentage
+            rectangleLineWidth = 10
```
Leland-Takamine, line 609: What do these values do and how were these values chosen?

SosenWiosen: I've used the values from the previous PR without giving them that much thought (which was sloppy - I should have looked at it, sorry!)

allowingPercentOfDifferentPixels - how many pixels can be different (in %) for the images to still be considered a match - can be specified by user,
rectangleLineWidth - how wide is the rectangle showcasing the diff - 10 seems to work fine and is quite visible,
pixelToleranceLevel - how much can the RGB value of pixels differ - this should be 0 to 1, it's a % of euclidean distance. I will fix it in a next commit and set it to 0.1 (default of the library)
minimalRectangleSize - minimal size  (width*height) of the rectangles which show the diff, 40 seems to make sens so that there aren't too many tiny rectangles.

Image from #2078 
Image

PR #2849, opened 2025-11-24
Add --format html-detailed for enhanced HTML reports with step details

sidferreira:
## Proposed changes

This PR adds a new `--format html-detailed` option for HTML test reports that displays detailed step information with timing and status indicators. This enhancement provides better visibility into test execution flow and serves as the foundation for upcoming features like console log capture (PR #2850).

### 🎯 What this adds

**New Format:**
- `--format html-detailed`: Enables enhanced HTML reports with detailed step breakdown

### 📋 Features

**Enhanced Step Display:**
- Visual status indicators for each step (✅ ⚠️ ❌ ⏭️)
- Step-by-step breakdown of test execution
- Individual timing for each command
- Clean, organized layout

**Status Icons:**
- ✅ `COMPLETED`: Step executed successfully
- ⚠️ `WARNED`: Step completed with warnings (e.g., optional assertions)
- ❌ `FAILED`: Step failed during execution
- ⏭️ `SKIPPED`: Step was skipped

**External CSS File:**
- Introduced `html-detailed.css` for clean separation of styles
- Consistent styling across all detailed reports
- Easy to maintain and extend

### 📸 Visual Comparison

<table>
<tr>
<td>Normal</td>
<td>Detailed</td>
<tr>
<tr>
<td></td>
<td></td>
<tr>
<tr>
<td></td>
<td></td>
<tr>
</table>

### 💻 Usage

```bash
# Generate detailed HTML report
maestro test --format html-detailed myFlow.yaml

# Custom output location
maestro test --format html-detailed --output report.html myFlow.yaml
```

### 🔧 Technical Details

**Data Model Changes:**
- Added `StepResult` data class to `TestExecutionSummary`
- Captures: description, status, and duration for each command
- Stored in `FlowResult.steps` list

**HTML Reporter:**
- New constructor parameter: `detailed: Boolean = false`
- Conditional rendering: shows steps only when `detailed = true`
- External CSS loaded from `html-detailed.css` resource

**CSS Styling:**
- `.step-item`: Visual separator between steps
- `.step-header`: Monospace font for command descriptions
- `.step-name`: Bold text for step identification
- Clean, minimal design with border accents

**Test Suite Interactor:**
- New parameter: `captureSteps: Boolean = false`
- Captures command execution details during test run
- Formats durations (< 1000ms as "Xms", >= 1000ms as "X.Xs")

**Command Ordering:**
- Uses sequence numbers to preserve exact execution order
- Prevents timestamp-based ordering issues with fast-executing commands

[...]

PR #2951, opened 2026-01-23
fix: orchestra command order

proksh:
## Problem

The env variables were not showing up on cloud. The metadata was not saving correctly while executing the run.

## Solution

There was an execution order change in this [PR](https://github.com/mobile-dev-inc/Maestro/pull/2849), which made `updateMetadata` run before `onCommandStart` which caused the issue. Reverting that change.

Added test so this don't happen again.


maestro-orchestra/src/test/java/maestro/orchestra/OrchestraCommandOrderTest.kt
```diff
@@ -0,0 +1,164 @@
...
+        val type: String,
+        val commandIndex: Int,
+        val sequence: Int,
+    )
+
+    @Test
+    fun `callback order should be correct for successful command in subflow`() {
+        // Given
+        val events = mutableListOf<CallbackEvent>()
+        var sequence = 0
```
amanjeetsingh150: Should we use integration test for this instead of creating another class here?

maestro-orchestra/src/test/java/maestro/orchestra/OrchestraCommandOrderTest.kt
```diff
@@ -0,0 +1,164 @@
...
+            val metadataEvent = cmdEvents.find { it.type == "onCommandMetadataUpdate" }
+            val completeEvent = cmdEvents.find { it.type == "onCommandComplete" }
+            
+            // Verify onCommandStart comes before onCommandMetadataUpdate
+            if (startEvent != null && metadataEvent != null) {
+                assertThat(startEvent.sequence).isLessThan(metadataEvent.sequence)
+            }
+            // Verify onCommandMetadataUpdate comes before onCommandComplete
+            if (metadataEvent != null && completeEvent != null) {
+                assertThat(metadataEvent.sequence).isLessThan(completeEvent.sequence)
```
amanjeetsingh150: Instead of something like less than. Can we use something like containsExactly, which ensures the order should also be same? Just a stricter check.

maestro-orchestra/src/test/java/maestro/orchestra/OrchestraCommandOrderTest.kt
proksh: Done

maestro-orchestra/src/test/java/maestro/orchestra/OrchestraCommandOrderTest.kt
proksh: Done

PR #2955, opened 2026-01-28
Fail test if hide keyboard fails

proksh:
## Changes
- **Verify Keyboard is hidden**: After hiding keyboard verifies the keyboard is hidden and throws `MaestroException.HideKeyboardFailure` if it's still visible (iOS and Android).

Right now hideKeyboard just passes everytime irresepective if it's successful in hiding keyboard.

## Motivation
- iOS hidekeyboard default behavior (taps/swipes) can be flaky depending on screen elements.
- Failures should be explicit: if hideKeyboard doesn't hide the keyboard, the test should fail.

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -778,7 +778,7 @@ class Orchestra(
 
         // Throw error in case keyboard is still visible
         if (maestro.isKeyboardVisible()) {
-            throw MaestroException.HideKeyboardFailure("Couldn't hide the keyboard. This can happen if the app uses a custom input or doesn't expose a standard dismiss action.")
+            throw MaestroException.HideKeyboardFailure("Couldn't hide the keyboard. This can happen if the app uses a custom input or doesn't expose a standard dismiss action. Instead of hideKeyboard, try tapping on interactive element to hide keyboard.")
         }
 
         return true


maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
```diff
@@ -286,6 +286,7 @@ data class YamlFluentCommand(
             hideKeyboard != null -> listOf(
                 MaestroCommand(
                     HideKeyboardCommand(
+                        commands = hideKeyboard.commands?.flatMap { it.toCommands(flowPath, appId) },
```
amanjeetsingh150: This is missing the environment variable injection, check the runFlow command for the correct implementation.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -672,8 +672,26 @@ class Orchestra(
         )
     }
 
-    private fun hideKeyboardCommand(): Boolean {
-        maestro.hideKeyboard()
+    private suspend fun hideKeyboardCommand(command: HideKeyboardCommand, config: MaestroConfig?): Boolean {
+        val customCommands = command.commands
```
amanjeetsingh150: nit: change the variable name for readability hideKeyboardCommand.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -672,8 +672,26 @@ class Orchestra(
...
+            if (maestro.isKeyboardVisible()) {
+                throw MaestroException.HideKeyboardFailure("Failed to hide keyboard, custom commands didn't hide keyboard")
+            }
+        } else {
+            // Use default behavior (platform-specific: swipes for iOS, back press for Android)
+            maestro.hideKeyboard()
+
+            // Verify that the keyboard was actually hidden (works for both iOS and Android)
+            if (maestro.isKeyboardVisible()) {
+                throw MaestroException.HideKeyboardFailure("Failed to hide keyboard, you can use 'commands' to provide custom commands to hide keyboard")
```
amanjeetsingh150: Can we give example here?

```
Couldn’t hide the keyboard automatically. This can happen if the app uses a custom input or doesn’t expose a standard dismiss action.

You can provide your own steps to hide the keyboard using commands, and Maestro will run those instead:

- hideKeyboard:
    commands:
      - tapOn:
          text: "Static label"
```

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
proksh: Do we need to inject env here like runflow??

I think it make sense to keep it like repeat command. Where there can be subcommand but no direct env injection from command itself.

proksh:
> Can we add an integration test for logic that it executes commands if default hideKeyboard doesn't work and vice versa?

Added test. Logic is a little different than what you described in this line. Let me know if you wanna discuss.


> Another thing here:
> Since we are only consuming commands on iOS, an argument could be why its not used on android. We should maybe add this in android as well. Although I am not aware of many cases on android.

Yes, that wont make sense. So same logic is being used for android as well. If `commands` not provided default flow is used. If command is provided, that will overwrite the default hideKeyboard functionality and check for keyboard visibility at the end.

> The parsing of hideKeyboard is not getting parsed as expected, its still creating with null commands even when I supply non empty commands. I think mostly because we miss passing commands here:

Not sure what you mean here.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -380,6 +380,11 @@ class IOSDriver(
                 durationMs = 50,
             )
 
+
+            if (isKeyboardHidden()) {
```
steviec: Shouldn't this be `!isKeyboardHidden()` ? Can we add a test to make sure we're doing the right thing :).

proksh: Ohh nice catch. Got confused with `isKeyboardHidden` and `isKeyboardVisible`. Updating and adding test.

proksh: Updated

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlFluentCommand.kt
amanjeetsingh150: Cool ya makes sense, let's not add unless someone really asks for it.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -775,6 +775,12 @@ class Orchestra(
 
     private fun hideKeyboardCommand(): Boolean {
         maestro.hideKeyboard()
+
+        // Throw error in case keyboard is still visible
+        if (maestro.isKeyboardVisible()) {
+            throw MaestroException.HideKeyboardFailure("Couldn't hide the keyboard. This can happen if the app uses a custom input or doesn't expose a standard dismiss action.")
```
amanjeetsingh150: nit: Can we help user here what should they do instead?

For example, tap non interactive element to hide keyboard like static texts.

PR #2956, opened 2026-01-28
Fix screenshot error handling with shouldFailOnError flag

proksh:
## Summary
Fixes screenshot error handling to prevent commands from getting stuck on failures. Adds `shouldFailOnError` parameter to distinguish between test commands and internal screenshot usage. Test commands fail fast on screenshot errors, while internal screenshots fail silently.

**_Review commit-by-commit: Check maestro-android changes first (commit 1), then proto (commit 2), then maestro-client integration (commit 3)._**

### Commit 1: Add ScreenshotService in maestro-android
- Created `ScreenshotService` class with testable design (accepts screenshot provider function)
- Refactored `MaestroDriverService.screenshot()` to delegate to `ScreenshotService`
- Added tests covering both `shouldFailOnError=true` and `shouldFailOnError=false` 
- Added `GrpcErrorExtensions` helper for consistent error handling

### Commit 2: Add shouldFailOnError to proto
- Added `shouldFailOnError` field to `ScreenshotRequest` message

### Commit 3: Integrate shouldFailOnError in maestro-client
- Added `shouldFailOnError` parameter to `Driver.takeScreenshot()` interface (defaults to false)
- Updated `AndroidDriver` to pass `shouldFailOnError` via gRPC
- Updated `Orchestra` to pass `shouldFailOnError=true` for `- takeScreenshot` commands
- Updated all internal screenshot calls to use `shouldFailOnError=false`

## Benefits
- Fixes issue where commands would get stuck indefinitely on screenshot failures
- Test commands properly fail when screenshots cannot be taken
- Internal screenshot failures no longer cause unnecessary errors
- Screenshot service is testable via dependency injection pattern


maestro-test/src/main/kotlin/maestro/test/drivers/FakeDriver.kt
```diff
@@ -227,7 +227,7 @@ class FakeDriver : Driver {
         events += Event.HideKeyboard
     }
 
-    override fun takeScreenshot(out: Sink, compressed: Boolean) {
+    override fun takeScreenshot(out: Sink, compressed: Boolean, shouldFailOnError: Boolean) {
```
amanjeetsingh150, line 230: This feels like weird design, shouldn't it be consumer who decides if I want to ignore exception?

As a function takeScreenshot should just throw an expected exception if it doesn't work. In known cases for example null bitmap, it could screenshot not found and in other cases we ensure that throwables are properly getting streamed from GRPC server to maestro client.

Booleans, can be hard to maintain specially because it touches with state. So, I would only add them when needed.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -942,7 +942,22 @@ class Orchestra(
...
         val pathStr = command.path + ".png"
         val fileSink = getFileSink(screenshotsDir, pathStr)
-        maestro.takeScreenshot(fileSink, false)
+        val resolvedFile = screenshotsDir?.resolve(pathStr)?.toFile() ?: File(pathStr)
+        val absoluteFile = resolvedFile.absoluteFile
+
+        try {
+            // Use shouldFailOnError = true so screenshot failures propagate as exceptions
+            maestro.takeScreenshot(fileSink, false, shouldFailOnError = true)
+        } catch (e: Exception) {
```
amanjeetsingh150, line 951: We should not be catching all the exceptions here, we should just call takeScreenshot and it should blow up with expected exceptions.

maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotService.kt
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Takes a screenshot and sends the response to the observer.
+     *
+     * Behavior depends on [MaestroAndroid.ScreenshotRequest.getShouldFailOnError]:
+     * - `true`: Failures are reported as gRPC INTERNAL errors
+     * - `false`: Failures return an empty response silently
+     */
+    fun screenshot(
+        request: MaestroAndroid.ScreenshotRequest,
+        responseObserver: StreamObserver<MaestroAndroid.ScreenshotResponse>
```
amanjeetsingh150, line 30: It seems like class is aware more than what it needs to in this case its all the GRPC domain. The responsibility of class it to just take bitmap and doing all the encoded options and handling failures, something like:



```kotlin
class ScreenshotService(
    private val loggerTag: String = "Maestro",
    private val maxEncodedBytes: Int = 25 * 1024 * 1024, // 25MB safety limit (tune)
) {

    /**
     * Encodes a screenshot bitmap to PNG bytes.
     *
     * @throws ScreenshotException if compression fails or output is invalid/too large.
     */
    fun encodePng(bitmap: Bitmap, quality: Int = 100): ByteString =
        encode(bitmap, Bitmap.CompressFormat.PNG, quality)

    /**
     * Generic encoder for any [Bitmap.CompressFormat].
     *
     * @throws ScreenshotException if compression fails or output is invalid/too large.
     */
    fun encode(
        bitmap: Bitmap,
        format: Bitmap.CompressFormat,
        quality: Int = 100,
    ): ByteString {
        validateBitmap(bitmap)
        validateQuality(format, quality)

        val ok = try {
            bitmap.compress(format, quality, outputStream)
        } catch (t: Throwable) {
            throw ScreenshotException.CompressCrashed(
                message = "Bitmap compression crashed (${format.name})",
                cause = t
            )
        }

        if (!ok) {
            // compress() returns false if it failed to encode
            throw ScreenshotException.CompressFailed(
                message = "Failed to compress bitmap (${format.name})"
            )
        }

        val bytes = outputStream.toByteArray()
        if (bytes.isEmpty()) {
            throw ScreenshotException.EmptyOutput(
                message = "Bitmap compressed but produced empty output (${format.name})"
            )
        }
        return ByteString.copyFrom(bytes)
    }
```

Writing test for this would ensure that we can easily isolate it with bitmap object and mock compress function to fail, and throw errors.

maestro-android/src/androidTest/java/dev/mobile/maestro/GrpcErrorExtensions.kt
```diff
@@ -0,0 +1,15 @@
+package dev.mobile.maestro
+
+import io.grpc.Status
+import io.grpc.StatusException
+
+/**
+ * Converts a Throwable to a gRPC INTERNAL StatusException.
+ */
+internal fun Throwable.internalError(): StatusException {
```
amanjeetsingh150, line 9: We already have this, in MaestroDriverService:


```
internal fun Throwable.internalError() = Status.INTERNAL.withDescription(message).asException()
```

I think what's missing is the cause so, we can add that but need to pass cause and not `this` exception which is the exception itself:

```
internal fun Throwable.internalError() = Status.INTERNAL.withDescription(message).withCause(cause).asException()
```

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -323,17 +320,9 @@ class Service(
...
-        if (bitmap.compress(Bitmap.CompressFormat.PNG, 100, outputStream)) {
-            responseObserver.onNext(screenshotResponse {
-                bytes = outputStream.toByteString()
-            })
-            responseObserver.onCompleted()
-        } else {
-            Log.e("Maestro", "Failed to compress bitmap")
-            responseObserver.onError(Throwable("Failed to compress bitmap").internalError())
-        }
+        ScreenshotService {
```
amanjeetsingh150, line 323: This service should just call Service for bitmap operations and have try catch here to properly handle exception in GRPCy way:


```
try {
  ScreenshotService.encodeBitmap(uiAutomation.takeScreenshot)
     responseObserver.onNext({})
     responseObserver.onCompleted({})
} catch(exception: Exception) {
     responseObserver.onError(t.internalError())
}
```

maestro-cli/src/main/java/maestro/cli/util/ScreenshotUtils.kt
```diff
@@ -40,7 +40,7 @@ object ScreenshotUtils {
             val out = File
                 .createTempFile("screenshot-${status}-${System.currentTimeMillis()}", ".png")
                 .also { it.deleteOnExit() } // save to another dir before exiting
-            maestro.takeScreenshot(out.sink(), false)
+            maestro.takeScreenshot(out.sink(), false, shouldFailOnError = false)
```
amanjeetsingh150, line 43: This should not be here, it's the responsibility of cloud to actually catch it. By default client should fail when maestro didn't work to take and encode screenshots.

proksh:
Moved the PR here: https://github.com/mobile-dev-inc/Maestro/pull/2966

Issue #2811, opened 2025-10-30
Cannot run example flow on Android


haizhiguang:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

I am follow the doc to run sample :https://docs.maestro.dev/getting-started/run-a-sample-flow

cd ./samples
adb install sample.apk
maestro test android-flow.yaml

I can see the app was installed successfully, but with exception when I run 'maestro test android-flow.yaml
' , I have try this command by many maestro version.

Tested JDK version  : openlogic-openjdk-17.jdk ,  jdk 17.0.12 , jdk 21.0.9 , jdk 25.0.1 , jdk 17.0.17, jdk 17.0.0
All with this issue.

### Actual results

WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

No content to map due to end-of-input
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1]

The stack trace was:
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1]
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4996)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4898)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3848)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3831)
	at util.LocalIOSDevice.listDeviceViaDeviceCtl(LocalIOSDevice.kt:76)
	at maestro.device.DeviceService.listIOSConnectedDevices(DeviceService.kt:266)
	at maestro.device.DeviceService.listIOSDevices(DeviceService.kt:262)
	at maestro.device.DeviceService.listDevices(DeviceService.kt:144)
	at maestro.device.DeviceService.listConnectedDevices(DeviceService.kt:130)
	at maestro.cli.command.TestCommand.getDeviceIds(TestCommand.kt:304)
[...]

PR #2960, opened 2026-01-29
Add error handling for devicectl execution

NobodyNo1:
## Proposed changes

Fix device detection crash
- Added error handling for devicectl call
- Improve logging for process execution
- Refactor device list parsing 
- Remove unused temp file

copilot:summary

## Description
Issue happens when `xcrun devicectl` command fails
For me it was failing because of `.zshenv` caused by cargo
```
. "$HOME/.cargo/env"
# ... other things
```

which causes `xcrun devicectl` command to fail with 
```
Error: 2 unexpected arguments: 'PATH ...
```

That leads `devicectl_response.json` to be empty which afterwards crashes during the deserialization.

```
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
...
```

## Testing

Tested with:

`./maestro-cli/build/install/maestro/bin/maestro test android-test.yaml`

With `xcrun devicectl` is failing

image
 
With normal `xcrun devicectl` execution

image

Run test using `./gradlew :maestro-cli:test`

image

## Reasoning

The command-line output should give the user enough information to understand and fix this issue, since it is caused by a local environment problem. If the user is only running an Android flow, a failure in xcrun devicectl should not block them from proceeding.
 

## Issues fixed

linked issue https://github.com/mobile-dev-inc/maestro/issues/2811

### DIFF
--- a/maestro-ios-driver/src/main/kotlin/util/LocalIOSDevice.kt
+++ b/maestro-ios-driver/src/main/kotlin/util/LocalIOSDevice.kt
@@ -5,11 +5,7 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import java.io.File
 
-object DeviceCtlProcess {
-
-    val devicectlDevicesOutput: File? by lazy {
-        devicectlDevicesOutput()
-    }
+class DeviceCtlProcess {
 
     /**
      * Executes `xcrun devicectl list devices` (lists Apple devices visible to Xcode)
@@ -17,7 +13,7 @@ object DeviceCtlProcess {
      *
      * @return temp JSON file on success, or null if the command fails.
      */
-    private fun devicectlDevicesOutput(): File? {
+    fun devicectlDevicesOutput(): File? {
         val tempOutput = File.createTempFile("devicectl_response", ".json")
 
         val process = ProcessBuilder(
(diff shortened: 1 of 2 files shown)


maestro-ios-driver/src/main/kotlin/util/LocalIOSDevice.kt
```diff
@@ -38,12 +55,9 @@ class LocalIOSDevice(private val deviceCtlProcess: DeviceCtlProcess = DeviceCtlP
     }
 
     fun listDeviceViaDeviceCtl(deviceId: String): DeviceCtlResponse.Device {
-        val tempOutput = File.createTempFile("devicectl_response", ".json")
+        val tempOutput = deviceCtlProcess.devicectlDevicesOutput()
+            ?: throw java.lang.IllegalArgumentException("Unable retrieve device list")
```
Fishbowler, line 59: Why not a plain Kotlin IllegalArgumentException?
For that matter, shouldn't this be a RuntimeException?

maestro-ios-driver/src/main/kotlin/util/LocalIOSDevice.kt
```diff
@@ -1,20 +1,37 @@
...
+            .redirectError(ProcessBuilder.Redirect.PIPE)
+            .start()
+
+        val error = process.errorStream.bufferedReader().readText()
+        val exit = process.waitFor()
+
+        if (exit != 0 || error.isNotBlank()) {
+            println("[WARNING] unable to retrieve list of Apple device (devicectl failed)")
+            println("   > Make sure that `xcrun devicectl --help` command runs without an issue")
+            println("Proceeding without Apple devices...\n")
```
Fishbowler, line 31: Is `println` definitely what we want here?
This is used pretty sparingly in Maestro, and wouldn't make it to logging.

PR #2962, opened 2026-01-30
Shard queue system & automatic retry

tylerqr:
## Proposed changes
I've been using this queue and retry system for a week or so now, and it's been great. I thought I'd share.

Instead of dividing the number of tests among the shards and assigning them at the beginning, we have a queue system where shards pick up the next available test in the queue when it finishes a test. 
When a test fails, it will be re-added to the top of the queue to be tested again for 3 retries, a total of 4 attempts. 

My reasons for changing this: 
- My test lengths vary greatly. Some are almost 3 minutes log and more brittle. but the use case requires it. Most other tests are a lot shorter. The 'split and divide evenly' approach would leave 2/4 simulators having completed their allotted tests while the remaining shards still had several tests to churn through. The queue system keeps all shards working until no tests are left in the queue, completing the run faster.
- Maestro tests have always failed for silly non-bug reasons. Not often enough to be a bug with Maestro, but often enough that I couldn't really just walk away from my pc while tests ran, and realized I was wasting a lot of time trying to reproduce a bug from a failed test and constantly questioning if its' a real bug and just re-running the test to watch where it fails and adding more defensive 'wait.yaml' and 'waitForAnimationToEnd', etc. 
- I need to run more than 2 simulators at a time. When I do this I risk increasing random crashes and failures because of limited PC ram but the tests can complete faster.
- I'm comfortable knowing if a test fails after 4 attempts, that's a real issue I need to go fix.

image

## Testing

I've run shards for ios and android simulators, up to 4 simultaneously (maxing out my 32GB ram) with running backend, node, stripe cli listeners, etc. 

## Issues fixed
Android simulator can now freely start and stop running tests even while an ios test is ongoing. Previously, you had to stop the ios test and start android first. Twas quite annoying.


Fishbowler:
Maestro has always intentionally avoided automatic retries in order to maintain trust. 

If folks ship an app where an important button only worked half the time, and automatic retries hid that, some would lose trust in the tool. Instead, we've got a `retry` command to enable folks to wrap just the bits that they expect to be flaky (e.g. a dev API that runs on underspec'd hardware)

PR #2966, opened 2026-01-30
Fix screenshot error handling and improve gRPC error metadata propagation

proksh:
## Summary                                                                                                                                                                                                        
                                                                                                                                                                                                                 
  - Added proper error handling for `takeScreenshot` with detailed exception information                                                                                                                           
  - Fixed gRPC internal error handling to include error type, message, and cause in metadata trailers                                                                                                            
  - Enables better analysis of infrastructure errors from the Android driver                                                                                                                                     
                                                                                                                                                                                                                 
## Changes                                                                                                                                                                                                        
                                                                                                                                                                                                                 
### Screenshot Error Handling (`ScreenshotService.kt` - new file)                                                                                                                                                    
  - Created ScreenshotException with nullable cause for proper error propagation                                                                                                                                 
  - Added ScreenshotService with `encodePng()` that wraps bitmap compression in try-catch                                                                                                                          
  - Error messages include diagnostic info: format, quality, width, height, and config                                                                                                                           

### gRPC Error Metadata (`MaestroDriverService.kt`)                                                                                                                                                                  
  - Fixed `internalError()` extension function to return `StatusRuntimeException` with metadata trailers                                                                                                             
  - Added metadata keys: `error-type`, `error-message`, `error-cause`                                                                                                                                      
  - Errors now propagate exception class name, message, and cause through gRPC                                                                                                                                   

### Client-Side Error Extraction (`AndroidDriver.kt`)                                                                                                                                                                
  - Added handling for `Status.Code.INTERNAL` in `runDeviceCall()`                                                                                                                                                
  - Extracts and logs error metadata from trailers: type, message, cause                                                                                                                                         
  - Enables debugging of Android-side exceptions from the client                                                                                                                                                 
                                                                                                                                                                                                                 
## Test                                                                                                                                                                                                                                                                                                                                                                                 

- [X] Run existing Maestro flows that take screenshots                                                                                                                                                             
- [X] Verify error logs contain detailed metadata when screenshot fails                                                                                                                                            
- [ ] Monitor for ScreenshotException in production logs post-merge


maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotService.kt
```diff
@@ -0,0 +1,113 @@
...
+        val limitBytes: Int
+    ) : ScreenshotException(message)
+}
+
+/**
+ * Encodes screenshots to PNG (or other formats) with validation and size limits.
+ */
+class ScreenshotService(
+    private val loggerTag: String = "Maestro",
+    private val maxEncodedBytes: Int = 25 * 1024 * 1024, // 25MB safety limit
```
amanjeetsingh150: Let's not add this yet. We don't know why the screenshots are breaking. We would first want ensure that server propogates errors successfully.

maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotService.kt
```diff
@@ -0,0 +1,113 @@
...
+                message = "Screenshot size ${bytes.size} exceeds limit $maxEncodedBytes (${format.name})",
+                sizeBytes = bytes.size,
+                limitBytes = maxEncodedBytes
+            )
+        }
+        return ByteString.copyFrom(bytes)
+    }
+
+    private fun validateBitmap(bitmap: Bitmap) {
+        if (bitmap.isRecycled) {
```
amanjeetsingh150: What does this isRecycled mean here? Do we really need to do this validation?

maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotServiceTest.kt
```diff
@@ -0,0 +1,125 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Assert.fail
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class ScreenshotServiceTest {
```
amanjeetsingh150, line 14: Ok the structure here is weird, ideally we should not have this test here. This is place for android ui tests. And this is a unit test.

One way to solve this is having the screenshot service in a module which would have its own unit test and that module can be implemented in maestro-android.

I would say let's disable this for now ensuring it passes locally. We might just restructure the maestro modularization first and then automate this as well.

maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotService.kt
```diff
@@ -0,0 +1,113 @@
...
+        override val message: String
+    ) : ScreenshotException(message)
+
+    /** Compression produced no bytes. */
+    data class EmptyOutput(
+        override val message: String
+    ) : ScreenshotException(message)
+
+    /** Encoded size exceeded [ScreenshotService.maxEncodedBytes]. */
+    data class OutputTooLarge(
```
amanjeetsingh150: Let's not add this yet. The only ones we care about right now is: CompressCrashed. We can actually even get rid of this sealed class, CompressFailed and Crash sounds like the same thing.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -323,16 +324,14 @@ class Service(
...
+        try {
+            val bitmap = uiAutomation.takeScreenshot()
+            val bytes = screenshotService.encodePng(bitmap)
+            responseObserver.onNext(screenshotResponse { this.bytes = bytes })
             responseObserver.onCompleted()
-        } else {
-            Log.e("Maestro", "Failed to compress bitmap")
-            responseObserver.onError(Throwable("Failed to compress bitmap").internalError())
+        } catch (e: ScreenshotException) {
+            Log.e(TAG, e.message ?: "Screenshot encoding failed")
```
amanjeetsingh150: Like any other logger android logger has the throwable as well. Let's use that:

```
Log.e(TAG, "Screenshot encoding failed with ${e.message}", e)
```

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -1276,8 +1276,12 @@ class AndroidDriver(
                         throw throwable
                     }
                 }
+                Status.Code.INTERNAL -> {
+                    LOGGER.error("Device call failed: ${status.description ?: throwable.message}", throwable)
```
amanjeetsingh150: Why this special handling for internal type of GRPC erorrs?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1057,7 +1058,22 @@ class Orchestra(
     private fun takeScreenshotCommand(command: TakeScreenshotCommand): Boolean {
         val pathStr = command.path + ".png"
         val fileSink = getFileSink(screenshotsDir, pathStr)
-        maestro.takeScreenshot(fileSink, false)
+        val resolvedFile = screenshotsDir?.resolve(pathStr)?.toFile() ?: File(pathStr)
+        val absoluteFile = resolvedFile.absoluteFile
+
+        try {
+            maestro.takeScreenshot(fileSink, false)
```
amanjeetsingh150: We would want to avoid swallowing exceptions here. This should actually be untouched.

maestro-android/src/androidTest/java/dev/mobile/maestro/screenshot/ScreenshotService.kt
proksh: It is used to check if a bitmap's memory has been freed to prevent crashes. https://learn.microsoft.com/fil-ph/dotnet/api/android.graphics.bitmap.isrecycled?view=net-android-35.0

Should we not check this??

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -323,16 +333,14 @@ class Service(
...
-            })
+        try {
+            val bitmap = uiAutomation.takeScreenshot()
+            val bytes = screenshotService.encodePng(bitmap)
+            responseObserver.onNext(screenshotResponse { this.bytes = bytes })
             responseObserver.onCompleted()
-        } else {
-            Log.e("Maestro", "Failed to compress bitmap")
-            responseObserver.onError(Throwable("Failed to compress bitmap").internalError())
+        } catch (e: ScreenshotException) {
```
amanjeetsingh150, line 341: This would still hang if there is some other exception, other than Screenshot Exception. We want to still log that and use responseObserver to propagate unknown exceptions as well.

amanjeetsingh150 (approved):

PR #2970, opened 2026-02-02
Fix --no-ansi breaks and stops printing nested sub flows

eldare:
## Proposed changes

This fix addresses a bug in the `--no-ansi` option.

When using nested complex runFlow inside runFlow, and even deeper, the `--no-ansi` option fails to keep track and stops printing. It gets out of sync. 

NOTE: Maestro testing is still running as expected.

Here is an example of the bug:

main.yml file:
```
appId: ${APP_ID}
---
- runFlow: open_app.yml  <- printed fine
- runFlow: login_to_app.yml  <- printed fine
- runFlow: tests.yml  <-- this flow and everything in it are not being printed.
- runFlow: logout_from_app.yml
- runFlow: login_to_app.yml
- runFlow: tests.yml
```

open_app.yml file:
```
appId: ${APP_ID}
---
- launchApp
```

login_to_app.yml file:
```
appId: ${APP_ID}
---
- runFlow:
    when:
      visible: "name@example.com"
    commands:
      - tapOn: "name@example.com"
      - inputText: ${USERNAME}
      - tapOn: "Continue"
```

tests.yml file:
```
appId: ${APP_ID}
---
- runFlow: test1.yml
- runFlow: test2.yml
```

test1.yml file:
```
appId: ${APP_ID}
---
- assertVisible: "hello"
```

## Fix explained:

The new implementation uses **set-based tracking with unique keys**:
- `printedStartItems` - a `mutableSetOf<String>()` to track which commands have started printing
- `printedCompleteItems` - a `mutableSetOf<String>()` to track which commands have completed printing
- `devicePrinted`, `flowNamePrinted`, `onFlowStartPrinted`, `onFlowCompletePrinted` - boolean/state flags for top-level items

Each command is now identified by a unique **composite key** (e.g., `"$prefix:$index:$description:start"`), which ensures:
1. Commands are only printed once (prevents duplicates)
2. Nested sub-flows are correctly tracked independently
3. The status of each command (`PENDING`, `RUNNING`, `COMPLETED`, etc.) is properly handled

#### **Key Behavioral Improvements**
- The `renderCommandsPlainText()` function now accepts a `prefix` parameter to generate unique keys for nested commands
- Sub-flow sections like `subOnStartCommands`, `subCommands`, and `subOnCompleteCommands` each get their own key namespace
- The fix properly handles deeply nested command structures that were previously not printing correctly

## Testing

I've built and tested the fix locally with complex nested runFlows.

### DIFF
--- a/maestro-cli/src/test/kotlin/maestro/cli/runner/resultview/PlainTextResultViewTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/runner/resultview/PlainTextResultViewTest.kt
@@ -1,0 +1,337 @@
+package maestro.cli.runner.resultview
+
+import com.google.common.truth.Truth.assertThat
+import maestro.cli.runner.CommandState
+import maestro.cli.runner.CommandStatus
+import maestro.orchestra.AssertConditionCommand
+import maestro.orchestra.Condition
+import maestro.orchestra.ElementSelector
+import maestro.orchestra.MaestroCommand
+import maestro.orchestra.RunFlowCommand
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+
+class PlainTextResultViewTest {
+
+    private lateinit var outputStream: ByteArrayOutputStream
+    private lateinit var originalOut: PrintStream
+
+    @BeforeEach
+    fun setUp() {
+        outputStream = ByteArrayOutputStream()
+        originalOut = System.out
+        System.setOut(PrintStream(outputStream))
+    }
+
+    private fun tearDown() {
+        System.setOut(originalOut)
+    }
+
+    private fun getOutput(): String {
+        return outputStream.toString()
+    }
+
+    /**
+     * This test verifies that deeply nested runFlow commands are all printed correctly.
+     *
+     * Bug description: When using nested complex runFlow inside runFlow (and even deeper),
+     * the --no-ansi option fails to keep track and stops printing. It gets out of sync.
+     *
+     * Example structure that was failing:
+     * main.yml:
+     *   - runFlow: open_app.yml        <- printed fine
+     *   - runFlow: login_to_app.yml    <- printed fine
+     *   - runFlow: tests.yml           <- this flow and everything in it were NOT being printed
+     *
+     * Where tests.yml contains:
+     *   - runFlow: test1.yml
+     *   - runFlow: test2.yml
+     *
+     * And login_to_app.yml contains a conditional runFlow.
+     *
+     * The fix ensures unique keys are generated for each nested command by using
+     * hierarchical prefixes (e.g., "main:0:sub:0:sub:0") instead of flat indices.
+     */
+    @Test
+    fun `nested runFlow commands should all be printed correctly`() {
+        // Given
+        val resultView = PlainTextResultView()
+
+        // Create a deeply nested structure similar to the bug scenario:
+        // main flow -> runFlow (tests.yml) -> runFlow (test1.yml) -> assertVisible
+
+        val deepestCommand = MaestroCommand(
+            assertConditionCommand = AssertConditionCommand(
+                condition = Condition(visible = ElementSelector(textRegex = "hello"))
+            )
+        )
+
+        val deepestCommandState = CommandState(
+            status = CommandStatus.COMPLETED,
+            command = deepestCommand,
+            subOnStartCommands = null,
+            subOnCompleteCommands = null,
+            subCommands = null
+        )
+
+        // test1.yml - contains assertVisible
+        val test1RunFlow = MaestroCommand(
+            runFlowCommand = RunFlowCommand(
+                commands = listOf(deepestCommand),
+                sourceDescription = "test1.yml",
+                config = null
+            )
+        )
+
+        val test1State = CommandState(
+            status = CommandStatus.COMPLETED,
+            command = test1RunFlow,
+            subOnStartCommands = null,
+            subOnCompleteCommands = null,
+            subCommands = listOf(deepestCommandState)
+        )
+
+        // test2.yml - another nested flow
+        val test2RunFlow = MaestroCommand(
+            runFlowCommand = RunFlowCommand(
+                commands = listOf(deepestCommand),
+                sourceDescription = "test2.yml",
+                config = null
+            )
+        )
+
+        val test2State = CommandState(
+            status = CommandStatus.COMPLETED,
+            command = test2RunFlow,
+            subOnStartCommands = null,
+            subOnCompleteCommands = null,
+            subCommands = listOf(deepestCommandState.copy())
+        )
+
+        // tests.yml - contains test1 and test2
+        val testsRunFlow = MaestroCommand(
+            runFlowCommand = RunFlowCommand(
+                commands = listOf(test1RunFlow, test2RunFlow),
+                sourceDescription = "tests.yml",
+                config = null
+            )
+        )
+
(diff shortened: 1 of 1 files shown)


Fishbowler:
Can you describe the problem this solves in more detail?

eldare:
@Fishbowler I've updated the description.

eldare:
@Fishbowler Hi, do you think this fix can be reviewed? This bug is really problematic when we use Maestro in CI.
Thank you 🙏

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
```diff
@@ -7,8 +7,12 @@ import maestro.utils.chunkStringByWordCount
...
 class PlainTextResultView: ResultView {
 
-    private var currentStep = 0
-    private var renderStepCount = 0
+    private val printedStartItems = mutableSetOf<String>()
+    private val printedCompleteItems = mutableSetOf<String>()
+    private var devicePrinted = false
+    private var flowNamePrinted: String? = null
+    private var onFlowStartPrinted = false
+    private var onFlowCompletePrinted = false
```
amanjeetsingh150: This feels like too much of state maintenance. This would be harder to maintain over time when things are changed and added on MaestroCommands semantics.

Can't we just have one set:

```
private val printed = mutableSetOf<String>()
```
And print once function that adds in set:

```
private inline fun printOnce(key: String, block: () -> Unit) {
    if (printed.add(key)) block()
}
```

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
```diff
@@ -22,106 +26,109 @@ class PlainTextResultView: ResultView {
...
+                    println("  ".repeat(indent + 1) + "> On Flow Complete")
+                    printedStartItems.add(onCompleteKey)
                 }
-                renderCommandsPlainText(command.subOnCompleteCommands, indent = indent + 1)
+                renderCommandsPlainText(command.subOnCompleteCommands, indent = indent + 1, prefix = "$commandKey:subOnComplete")
             }
 
-            if (shouldPrintStep()) {
-                println("  ".repeat(indent) + "${c?.description()}... " + status(command.status))
+            if (completeKey !in printedCompleteItems && command.status in setOf(CommandStatus.COMPLETED, CommandStatus.FAILED, CommandStatus.SKIPPED, CommandStatus.WARNED)) {
```
amanjeetsingh150: This completeKey !in printedCompleteItems is repetitive pattern that can be replaced with function as suggested in printOnce function in other comment.

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
```diff
@@ -22,106 +26,109 @@ class PlainTextResultView: ResultView {
...
     }
 
-    private fun renderCommandPlainText(command: CommandState, indent: Int) {
+    private fun renderCommandPlainText(command: CommandState, indent: Int, commandKey: String) {
         val c = command.command.asCommand()
         if (c?.visible() == false) { return }
 
-        if (command.subCommands != null) {
+        val description = c?.description() ?: "Unknown command"
+        val startKey = "$commandKey:$description:start"
```
amanjeetsingh150: I think the key can be just combination of position like:

Position-only keys ("$key:start")

1. main:0:start, main:0:complete
2. main:0:sub:0:start
3. main:0:onComplete

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
```diff
@@ -22,106 +26,109 @@ class PlainTextResultView: ResultView {
...
+        val completeKey = "$commandKey:$description:complete"
 
-            if (shouldPrintStep()) {
-                println("  ".repeat(indent) + "${c?.description()}...")
+        if (command.subCommands != null) {
```
amanjeetsingh150: I think the cases here can be simplified like:

```
 private fun renderCommandPlainText(command: CommandState, indent: Int, key: String) {                                                                                                                    
      val c = command.command.asCommand()                                                                                                                                                                  
      if (c?.visible() == false) return                                                                                                                                                                    
                                                                                                                                                                                                           
      val desc = c?.description() ?: "Unknown command"                                                                                                                                                     
      val pad = "  ".repeat(indent)                                                                                                                                                                        
                                                                                                                                                                                                           
      when {                                                                                                                                                                                               
          c is CompositeCommand -> {
              // --- Composite (RunFlow, Repeat, Retry) ---

              // The actual sub-commands of the composite
              command.subCommands?.let { cmds ->
                  renderCommandsPlainText(cmds, indent + 1, "$key:sub")
              }

              // onFlowComplete hooks
              command.subOnCompleteCommands?.let { cmds ->
                  renderCommandsPlainText(cmds, indent + 1, "$key:onComplete")
              }

              // Print completion line once when it reaches a terminal status
              if (command.status in TERMINAL_STATUSES) {
                  printOnce("$key:complete") { println("$pad$desc... ${status(command.status)}") }
              }
          }

          else -> {
              // --- Simple command (tapOn, assertVisible, etc.) ---
              when (command.status) {

                  CommandStatus.RUNNING -> {
                      printOnce("$key:start") { print("$pad$desc...") }
                  }

                  in TERMINAL_STATUSES -> {

                      }
                  }

                  else -> {}
              }
          }
      }
  }
```

amanjeetsingh150 (commented):
You've done really good job in explaining the problem on PR description now. It would ensure we don't break this again if you add a test in Integration test here:

https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
eldare: Fixed.

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
eldare: Fixed.

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
eldare: Fixed.

maestro-cli/src/main/java/maestro/cli/runner/resultview/PlainTextResultView.kt
eldare: Fixed.

eldare:
@amanjeetsingh150 Thank you for the review. I've fixed your notes.

As for adding tests to the integration tests, I couldn't figure out how to apply cli related code there, so instead I've created dedicated unit tests for the fixed area.
Please let me know if I should add more.

Thank you 🙏

amanjeetsingh150:
> @amanjeetsingh150 Thank you for the review. I've fixed your notes.
> 
> As for adding tests to the integration tests, I couldn't figure out how to apply cli related code there, so instead I've created dedicated unit tests for the fixed area.
> Please let me know if I should add more.
> 
> Thank you 🙏

Ahh, makes sense this is the rendering logic. I thought for some reason we had those test. But the one that you add are also good. I think we should be ready to get this in once tests pass.

eldare:
🎉 Thank you @amanjeetsingh150

PR #2972, opened 2026-02-03
upgrade gradle

proksh:


Fishbowler:
Is this regular maintenance, or is it solving a problem?

proksh:
> Is this regular maintenance, or is it solving a problem?

Its regular maintenance. I have updated the gradle on `maestro-android` for easy compiling and testing. So have to update in maestro and cloud as well - else build will fail.

PR #2974, opened 2026-02-04
Fix connection failures due to failing driver tests while auto granting permissions

amanjeetsingh150:
## Problem

Auto granting permissions via the XCUITest driver can cause tests to fail when the expected permission buttons are not present on screen. The XCUITest tap API retries and also scrolls automatically while attempting to locate the button, which eventually results in a test failure if we don't find it and, in some cases, connection errors. These retries also add unnecessary slowness, for overall test execution.

This PR fixes by tapping on coordinates we already know from hierarchy snapshot from caller class.

### Bug Fix

Currently we use:

```swift
let predicate = NSPredicate(format: "label CONTAINS[c] %@", notificationsPermissionLabel)
let alert = foregroundApp.alerts.matching(predicate).element
let allowButton = alert.buttons.element(boundBy: 1)
allowButton.tap()

```
This is the main culprit, to tackle this we would just inject AXElement from caller that already has access to hierarchy.
This would just find the buttons and tap on them using coordinates:

Check this here: [SystemPermissionsHelper](https://github.com/mobile-dev-inc/Maestro/blob/MA-3925/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/SystemPermissionHelper.swift#L23).

### Structural Changes

1. Writing tests for this layer was getting increasingly difficult, especially for validating filtering logic. To address this, we extracted the Swift-only domain logic into an isolated, testable module: [MaestroDriverLib](https://github.com/mobile-dev-inc/Maestro/tree/MA-3925/maestro-ios-xctest-runner/MaestroDriverLib/Sources/MaestroDriverLib).

2. We’ve also moved integration tests into a separate unit-test scheme that only requires a device. This decouples them from the driver scheme and significantly speeds up execution.
 
### CI changes:

With the structural changes above we can simplify CI for iOS further:

- Removing the confusing warning of building iOS driver. Its flaky and doesn't add value. 
- Runs the MaestroDriverLib unit test, and maestro-driverTests unit test 
- These tests are fast because either they don't require any device or they can be run just with simulator.

## Testing

- [x] Sample Demo App: Check PR here: https://github.com/mobile-dev-inc/demo_app/pull/11
- [x] User flow
- [x] Local environment

https://github.com/user-attachments/assets/b4127a6a-3e08-4bca-8485-783ec66dbfcc

## Issues fixed

- Slowness on test execution, this is big one because this was part of view hierarchy which maestro does frequently
- Connection failures due to failing tests.


.github/workflows/warn_build_xctestrunner.yaml
Fishbowler, line 1: This isn't useful for external contributors?

maestro-ios-xctest-runner/maestro-driver-iosTests/SnapshotParametersTests.swift
```diff
@@ -0,0 +1,82 @@
+import XCTest
+
+/// Tests to verify XCTest's accessibility snapshot parameters and swizzling behavior.
+/// These tests run as unit tests hosted in the simulator, giving access to XCTest runtime.
+/// The test scheme sets `snapshotKeyHonorModalViews=false` to trigger swizzling.
```
Fishbowler, line 5: swizzling is a great word 😂

steviec (approved):
I want to approve this so you are free to merge if you want.

PR #2979, opened 2026-02-05
Add tags & properties to test reports

proksh:
- Display flow tags and custom properties in HTML reports (tags as badges, properties as a table)
  - Include tags and properties in JUnit XML reports for CI integration


maestro-cli/src/main/java/maestro/cli/report/HtmlTestSuiteReporter.kt
Fishbowler: When testing this, found that this line causes the collapse to fail when the name has spaces. Is this something you can fix?

proksh: Updated

Fishbowler: Approved, but there's a unit test fix needed... sorry 😬

proksh: Fixed that as well, sorry about that. :)

PR #2981, opened 2026-02-05
fix: screenshot exceptions

proksh:
## Problem

- `Nullpoint` exception while takeScreenshot
- Logs hard to connect to the call made

## Reason

The issue is that `uiAutomation.takeScreenshot()` can return null in transient situations (window not laid out, invalid SurfaceControl, SurfaceFlinger errors). These are often temporary [[UiAutomation](https://developer.android.com/reference/android/app/UiAutomation#takeScreenshot(android.view.Window))].
   
## Solution

1. Added 3 time exponential backoff retry before failing. 

Why retry is appropriate here??    
  1. Window not laid out - A brief wait allows rendering to complete
  2. Invalid SurfaceControl - Often transient during activity transitions
  3. SurfaceFlinger errors - Can be momentary system hiccups
    
The retry approach with 3 attempts (200ms, 400ms, 800ms) adds at most ~1.4s on failure cases while handling the common transient scenarios.

2. Added call name to logs inside runDeviceCall.


maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -334,15 +342,32 @@ class Service(
...
     ) {
         try {
-            val bitmap = uiAutomation.takeScreenshot()
+            // Retry on IO failures and when screenshot returns null (window not ready / SurfaceControl invalid).
+            val bitmap = runBlocking {
+                val retryOnIOExceptionOrNull = continueIf<Throwable> {
+                    it.failure is IOException || it.failure is ScreenshotException
+                }
+                retry(
+                    stopAtAttempts<Throwable>(3) + retryOnIOExceptionOrNull + binaryExponentialBackoff(min = 500L, max = 5000L)
```
amanjeetsingh150: This is business logic and should be tested and moved to ScreenshotService.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
```diff
@@ -334,15 +342,32 @@ class Service(
...
+                ) {
+                    val result = uiAutomation.takeScreenshot()
+                    if (result == null) {
```
amanjeetsingh150: Why is this added check here? Can't we just keep the lines exactly where we have NullPointerException in retry block.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
proksh: Make sense, updated.

maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt
proksh: Done

PR #2988, opened 2026-02-10
Fix: only dismiss notifications dialog automatically

amanjeetsingh150:
## Proposed changes

Refactoring the auto dismissal, started automatically dismissing other dialogs. This makes sure we only dismiss notification dialogs.

## Testing

- [x] Adds tests for make sure we only dismiss buttons inside the notifications dialogs aligns with how it was before


maestro-ios-xctest-runner/MaestroDriverLib/Sources/MaestroDriverLib/Helpers/PermissionButtonFinder.swift
```diff
@@ -6,14 +6,16 @@ public enum PermissionButtonResult: Equatable {
...
-    /// Permission value doesn't require action (unset/unknown)
+    /// Permission value doesn't require action (unset/unknown/not a permission dialog)
     case noActionRequired
 }
 
 /// Pure logic for finding permission dialog buttons in the view hierarchy.
 /// This class has no XCUITest dependencies and can be unit tested.
 public final class PermissionButtonFinder {
 
+    static let notificationsPermissionLabel = "Would Like to Send You Notifications"
```
proksh, line 17: Is it possible that notification permission message is different between os or updates??

Fishbowler: We've not seen this on iOS 16/17/18
We do get complaints that it only happens in English, but that's a different issue altogether 😂

amanjeetsingh150: Not that I'm aware of. But definitely this can be sensitive to OS and locales. This right now aligns with our past logic. We can make sure to add cases for that. Let me actually write this on permissions rewrite we'll be doing.

PR #2990, opened 2026-02-10
fix: iOS call timeout being 60 seconds

amanjeetsingh150:
## Proposed changes

iOS driver HTTP client doesn't configure a call timeout right now. This is a huge problem because this can leave the calls hanging on our XCUITest server. 

Read and write timeout are just small part of HTTP transaction. With this we have a complete timeout defined for each call to server.

## Testing

- [x] Locally.
- [x] Unit test that verifies thread doesn't block forever like before.


Fishbowler (approved):
👀  https://www.baeldung.com/okhttp-timeouts

Yep, this look sensible.

PR #2997, opened 2026-02-15
Feat/support allure reporter

vfadc:
## Proposed changes

[Allure](https://allurereport.org/) is a widely used test reporting format, and this change adds support for generating Allure results in maestro:
- supports `--format allure` output
- exports artifacts as Allure attachments (`image/png`, `video/mp4`)
- includes screenshot diff artifact on visual mismatch
- maps Maestro metadata into Allure [labels](https://allurereport.org/docs/how-it-works-test-result-file/#labels) (`framework`, `suite`, `host`, `tag`, flow, properties)

## Testing

Tested locally on `e2e/workspaces/demo_app` by running `e2e/run_tests` with Allure output enabled (added Allure reporter flag in `e2e/run_tests` (`--format allure`))

  Repro steps:
  ```bash
  cd e2e
  ./run_tests android
  allure serve allure-results
```

Verified:
  - Allure result files are generated
  - attachments are present in report
  - image attachments are correct (regular screenshots + screenshot diff)
  - video attachment is present and playable (`video/mp4`)
  - labels are populated correctly in Allure
  - parallel execution is handled correctly, and test timelines are formed/rendered correctly in Allure
Screenshot 2026-02-15 at 18 39 10
Screenshot 2026-02-15 at 18 32 22
Screenshot 2026-02-15 at 18 32 59
Screenshot 2026-02-15 at 18 33 15

  ## Allure docs

  - https://allurereport.org/docs/install/
  - https://allurereport.org/docs/v2/generate-report/
  - https://allurereport.org/docs/how-it-works-test-result-file/#labels

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/report/AllureTestSuiteReporter.kt
+++ b/maestro-cli/src/main/java/maestro/cli/report/AllureTestSuiteReporter.kt
@@ -63,7 +63,7 @@ class AllureTestSuiteReporter(
                 val testResult = TestResult()
                     .setUuid(uuid)
                     .setName(flow.name)
-                    .setFullName(buildFullName(suite, flow))
+                    .setFullName(buildFullName(flow))
                     .setHistoryId(buildHistoryId(flow))
                     .setStatus(mapFlowStatus(flow.status))
                     .setLabels(buildLabels(suite, flow))
@@ -120,11 +120,8 @@ class AllureTestSuiteReporter(
         }
     }
 
-    private fun buildFullName(
-        suite: TestExecutionSummary.SuiteResult,
-        flow: TestExecutionSummary.FlowResult,
-    ): String {
-        val suitePart = testSuiteName ?: suite.deviceName ?: "maestro"
+    private fun buildFullName(flow: TestExecutionSummary.FlowResult): String {
+        val suitePart = testSuiteName ?: "maestro"
         return "$suitePart#${flow.name}"
     }
 
@@ -141,7 +138,6 @@ class AllureTestSuiteReporter(
         val labels = mutableListOf<Label>()
 
         labels.add(Label().setName("framework").setValue("maestro"))
-
         if (testSuiteName != null) {
             labels.add(Label().setName("suite").setValue(testSuiteName))
         }
--- a/maestro-cli/src/main/java/maestro/cli/report/AllureTestSuiteReporter.kt
+++ b/maestro-cli/src/main/java/maestro/cli/report/AllureTestSuiteReporter.kt
@@ -1,0 +1,188 @@
+package maestro.cli.report
+
+import io.qameta.allure.AllureLifecycle
+import io.qameta.allure.FileSystemResultsWriter
+import io.qameta.allure.model.Attachment
+import io.qameta.allure.model.Label
+import io.qameta.allure.model.Parameter
+import io.qameta.allure.model.Status
+import io.qameta.allure.model.StatusDetails
+import io.qameta.allure.model.StepResult
+import io.qameta.allure.model.TestResult
+import io.qameta.allure.util.ResultsUtils
+import maestro.cli.CliError
+import maestro.cli.model.FlowStatus
+import maestro.cli.model.TestExecutionSummary
+import java.io.File
+import java.nio.file.Files
+import java.util.UUID
+
+class AllureTestSuiteReporter(
+    private val outputDir: File,
+    private val testSuiteName: String?,
+) : TestSuiteReporter {
+
+    companion object {
+        fun from(output: File?, testSuiteName: String?): AllureTestSuiteReporter {
+            return AllureTestSuiteReporter(
+                outputDir = resolveOutputDir(output),
+                testSuiteName = testSuiteName,
+            )
+        }
+
+        private fun resolveOutputDir(output: File?): File {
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

**Comment by vfadc** [comment]
@Fishbowler Can I get a review on this?)

**Comment by Fishbowler** [comment]
Have you tested that the non-Allure outputs are still operational, given the changes?

File: maestro-cli/src/test/kotlin/maestro/cli/report/TestSuiteReporterTest.kt [hunk 1]
```diff
@@ -130,22 +133,26 @@ abstract class TestSuiteReporterTest {
...
-                                duration = "2.0s"
+                                durationMs = 2000,
+                                startTime = nowPlus1.toInstant().toEpochMilli() + 3500,
                             ),
                             TestExecutionSummary.StepResult(
                                 description = "4. Assert visible",
                                 status = "SKIPPED",
-                                duration = "0ms"
+                                durationMs = 0,
+                                startTime = nowPlus1.toInstant().toEpochMilli() + 5500,
```
Fishbowler: For maintenance purposes, should it be more obvious where these numbers are coming from?

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
```diff
@@ -30,6 +28,7 @@ import kotlin.time.Duration.Companion.seconds
 import maestro.cli.util.ScreenshotUtils
 import maestro.orchestra.util.Env.withDefaultEnvVars
 import maestro.orchestra.util.Env.withInjectedShellEnvVars
+import java.util.Locale
```
Fishbowler: Is this used?

maestro-cli/src/main/java/maestro/cli/report/ReporterFactory.kt
```diff
@@ -1,16 +1,39 @@
...
-import okio.BufferedSink
+import okio.sink
+import java.io.File
 
 object ReporterFactory {
 
-    fun buildReporter(format: ReportFormat, testSuiteName: String?): TestSuiteReporter {
+    private const val DEFAULT_JUNIT_OUTPUT_FILE = "report.xml"
+    private const val DEFAULT_HTML_OUTPUT_FILE = "report.html"
+    private const val DEFAULT_ALLURE_OUTPUT_DIR = "allure-results"
```
Fishbowler, line 10: Have these moved from somewhere I've not spotted, or are they declared elsewhere in the codebase too?

vfadc:
Tested `JUNIT`, `HTML`, and `HTML_DETAILED` manually, and the existing test suite passes as well. `NOOP` has no output.

maestro-cli/src/main/java/maestro/cli/report/ReporterFactory.kt
vfadc: New constants, not moved from anywhere. The old code built the filename dynamically, extensions were in `ReportFormat.fileExtension (.xml, .html)` and combined as `File("report$extension")` in `TestCommand`. I removed fileExtension from the enum when adding Allure (which outputs a directory, not a file) and made the defaults explicit constants in `ReporterFactory` instead.

maestro-cli/src/test/kotlin/maestro/cli/report/TestSuiteReporterTest.kt
vfadc: Reworked into named variables

maestro-cli/src/main/java/maestro/cli/runner/TestSuiteInteractor.kt
vfadc: removed

maestro-cli/src/main/java/maestro/cli/report/AllureTestSuiteReporter.kt
```diff
@@ -0,0 +1,184 @@
...
+        flow.properties?.forEach { (key, value) ->
+            labels.add(Label().setName(key).setValue(value))
+        }
+
+        return labels
+    }
+
+    private fun buildParameters(env: Map<String, String>): List<Parameter> {
+        return env.map { (key, value) ->
+            Parameter().setName(key).setValue(value)
```
aprokofjev-mwb: I would prefer to mask values with sensitive keys.
```
private fun maskSensitiveValue(key: String, value: String): String {
        val sensitiveKeywords = listOf(
            "password", "pwd", "secret", "token", "key", "auth",
            "credential", "apikey", "api_key", "private"
        )

        val isSensitive = sensitiveKeywords.any { keyword ->
            key.lowercase().contains(keyword)
        }

        return if (isSensitive) "***MASKED***" else value
    }
```

vfadc:
@Fishbowler  re-review, please

sdillon1224:
commenting to follow progress 👀

dmitriishabalin:
Is there anything we can do to help make this happen?

Fishbowler:
I don't think additional report formats is the way we're gonna choose to take Maestro. Instead, what we'd like to do is create a intermediate format (JUnit or JSON) that enables anyone to create any report format that they need.

I think that Allure would make a great test case to ensure that we've got the right details in something like that. Would you be open to collaborating on that?

PR #3001, opened 2026-02-19
Strongly Typed Device Configuration

proksh:
## Summary

Introduces a strongly typed device configuration system that simplifies device management across the codebase by replacing scattered parameter passing with a unified `MaestroDeviceConfiguration` data class.

## Key Changes

- Added `DeviceCatalog` for handling creating a strongly typed config `MaestroDeviceConfig`, which will be used as main config
- Updated start-device command & Pickdevice to use new config and simplified code
- Get defaults from the API and apply defaults for things not provided
- Added `--device-model` & `--device-os` flag to start-device and depreciated `--os-version` flag. These new flags would take entire os instead of just numbers. Added hint for how these can be listed locally
- Locale validations added
- Removed duplicate platform type
- Fix web device launch

--- 
Note: I haven't deleted LocaleUtils right now, as it is being used by worker. So that this doesn't become a blocker for bumping maestro, I am keeping it for now.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
+++ b/maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
@@ -30,11 +30,11 @@ object PickDeviceView {
                 Platform.fromString(it)
             } ?: throw CliError("Please specify a platform"))
 
-        val maestroDeviceConfiguration = DeviceCatalog.resolve(
+        val spec = DeviceCatalog.resolve(
             platform = selectedPlatform.toString(),
         )
 
-        return maestroDeviceConfiguration
+        return spec
     }
 
     fun pickRunningDevice(devices: List<Device>): Device {
--- a/maestro-client/src/test/java/maestro/device/DeviceCatalogTest.kt
+++ b/maestro-client/src/test/java/maestro/device/DeviceCatalogTest.kt
@@ -75,15 +75,6 @@ internal class DeviceCatalogTest {
     }
 
     @Test
-    fun `resolve succeeds for valid cloud-compatible configs`() {
-        DeviceCatalog.resolve("android", model = "pixel_6",  os = "android-34")
-        DeviceCatalog.resolve("android", model = "pixel_xl", os = "android-33")
-        DeviceCatalog.resolve("ios",     model = "iPhone-11", os = "iOS-18-2")
-        DeviceCatalog.resolve("ios",     model = "iPhone-16", os = "iOS-18-2")
-        DeviceCatalog.resolve("web",     model = "chromium")
-    }
-
-    @Test
     fun `resolve throws on unsupported platform`() {
         assertThrows<IllegalArgumentException> {
             DeviceCatalog.resolve("windows")


maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -0,0 +1,144 @@
...
+  private const val DEFAULT_LOCALE = "en_US"
+  private val DEFAULT_ORIENTATION = DeviceOrientation.PORTRAIT
+  private const val DEFAULT_IOS_MODEL = "iPhone-11"
+  private const val DEFAULT_IOS_OS = "iOS-17-0"
+  private const val DEFAULT_ANDROID_MODEL = "pixel_6"
+  private const val DEFAULT_ANDROID_OS = "30"
+  private const val DEFAULT_WEB_MODEL = "chromium"
+  private const val DEFAULT_WEB_OS = "latest"
+
+  private val CLOUD_CATALOG: Set<MaestroDeviceConfiguration> by lazy { setOf(
```
proksh: These defaults and device list will be moved to cloud side and would come here through APIs.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
             throw CliError("This command is not supported in Windows WSL. You can launch your emulator manually.")
         }
 
-        val p = Platform.fromString(platform)
-            ?: throw CliError("Unsupported platform $platform. Please specify one of: android, ios")
+        val p = Platform.fromString(platform) ?: throw CliError("Unsupported platform $platform. Please specify one of: android, ios")
```
amanjeetsingh150: Can we move this to resolve as well? Seems like something that can fit in resolve abstraction.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
...
         }
 
+        // Get/Create the device
+        val device = DeviceCreateUtil.getOrCreateDevice(
+            maestroDeviceConfiguration,
+            forceCreate
+        )
+
+        // Launch the device
+        PrintUtils.message(if (p == Platform.IOS) "Launching simulator..." else "Launching emulator...")
```
amanjeetsingh150: Do we have start-device on web? How does that work today?

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
...
-                    "\n" +
-                    countries
-            )
-        } catch(e: LocaleValidationWrongLocaleFormatException) {
-            throw CliError("Wrong device locale format was provided $deviceLocale. A combination of lowercase ISO-639-1 code and uppercase ISO-3166-1 code should be used, i.e. \"de_DE\" for Germany. More info can be found here https://maestro.mobile.dev/")
-        } catch (e: LocaleValidationNotSupportedPlatformException) {
-            throw CliError("The feature to set a device locale is not supported by the platform ${p.description}")
+        } catch (e: CloudCompatibilityException) {
+            // Local execution — warn but proceed with the resolved config
+            PrintUtils.warn(e.message.toString())
```
amanjeetsingh150: Does this gets printed everytime I start device? Should user be warned everytime if they are starting a device not compatible with cloud?

Start device should only fail in cases if user is entering invalid input that's not applicable or not available on local envrionment.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
...
+            e.config
+        } catch (e: Exception) {
+            throw CliError(e.message.toString())
```
amanjeetsingh150: Let's log the error and exception so that it enters maestro.log file. Is it possible to have this exception printed on CLI. If this is failing ideally we would want to know from users: Which exception it fails from from both CLI and log outputs.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
...
         }
 
+        // Get/Create the device
+        val device = DeviceCreateUtil.getOrCreateDevice(
```
amanjeetsingh150: I think we can get rid of try/catch above and not swallow exceptions.

All the unknown errors should be printed on CLI. CLIError treat them known and we would not even know the cause behind real exception.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,61 +65,36 @@ class StartDeviceCommand : Callable<Int> {
...
-            val countries = LocaleUtils.ANDROID_SUPPORTED_COUNTRIES.joinToString("\n")
-            throw CliError("${e.country} country is currently not supported by Maestro, please check that it is a valid ISO-3166-1 code. Here is a full list of supported countries:\n" +
-                    "\n" +
-                    countries
-            )
-        } catch(e: LocaleValidationWrongLocaleFormatException) {
-            throw CliError("Wrong device locale format was provided $deviceLocale. A combination of lowercase ISO-639-1 code and uppercase ISO-3166-1 code should be used, i.e. \"de_DE\" for Germany. More info can be found here https://maestro.mobile.dev/")
-        } catch (e: LocaleValidationNotSupportedPlatformException) {
-            throw CliError("The feature to set a device locale is not supported by the platform ${p.description}")
+        } catch (e: CloudCompatibilityException) {
```
amanjeetsingh150: If resolve function is to check compatibility for cloud we can: leave it a noop here, so that its clear that local CLI works for devices that are not yet launched on cloud like OS 26, API 35. 

StartDeviceCommand possible errors, I'd prefer removing try/catch swallowing all exceptions:

- Using local commands (xcrun, avdmanager) errors here should get printed on cli with output for better debugging
- Unknown errors should directly blow up we want to know where its failing

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -55,29 +47,28 @@ object DeviceCreateUtil {
         if (existingDeviceId != null) PrintUtils.message("Using existing device $deviceName (${existingDeviceId}).")
         else PrintUtils.message("Attempting to create iOS simulator: $deviceName ")
 
-
         val deviceUUID = try {
-            existingDeviceId ?: DeviceService.createIosDevice(deviceName, device, runtime).toString()
+            existingDeviceId ?: DeviceService.createIosDevice(deviceName, config.deviceModel, config.deviceOs).toString()
         } catch (e: IllegalStateException) {
```
amanjeetsingh150, line 49: This is the confusing part it seems like this is meant for handling errors from xcrun and avdmanager. It would be useful to catch them as a known exception instead.

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -89,33 +80,23 @@ object DeviceCreateUtil {
...
 
     fun getOrCreateAndroidDevice(
-        version: Int?, language: String?, country: String?, forceCreate: Boolean, shardIndex: Int? = null
+        config: MaestroDeviceConfiguration.Android, forceCreate: Boolean, shardIndex: Int? = null
     ): Device.AvailableForLaunch {
-        @Suppress("NAME_SHADOWING") val version = version ?: DeviceConfigAndroid.defaultVersion
-        if (version !in DeviceConfigAndroid.versions) {
-            throw CliError("Provided Android version is not supported. Please use one of ${DeviceConfigAndroid.versions}")
+        val abi = when (val architecture = EnvUtils.getMacOSArchitecture()) {
+            CPU_ARCHITECTURE.x86_64 -> "x86_64"
```
amanjeetsingh150: Why is this not in device configuration? Or I miss something.

maestro-cli/src/main/java/maestro/cli/device/PickDeviceView.kt
```diff
@@ -37,27 +37,26 @@ object PickDeviceView {
...
-                Platform.IOS -> {
-                    PrintUtils.message("Please specify iOS version ${DeviceConfigIos.versions}: Press ENTER for default (${DeviceConfigIos.defaultVersion})")
-                    readlnOrNull()?.toIntOrNull() ?: DeviceConfigIos.defaultVersion
-                }
-
-                Platform.ANDROID -> {
-                    PrintUtils.message("Please specify Android version ${DeviceConfigAndroid.versions}: Press ENTER for default (${DeviceConfigAndroid.defaultVersion})")
-                    readlnOrNull()?.toIntOrNull() ?: DeviceConfigAndroid.defaultVersion
-                }
+        val available = DeviceCatalog.allCloudAvailableOs(selectedPlatform)
```
amanjeetsingh150: Wait is this filtering other local only options even now? Given the current behaviour of CLI it feels like we should just let users start whatever is available on local and have this validation only on cloud.

It's more or less starting device that's all.

maestro-cli/src/main/java/maestro/cli/model/DeviceStartOptions.kt
```diff
@@ -1,9 +0,0 @@
-package maestro.cli.model
-
-import maestro.device.Platform
-
-data class DeviceStartOptions(
```
amanjeetsingh150, line 5: Feels good to finally get rid of these!

maestro-cli/src/main/java/maestro/cli/util/DeviceConfig.kt
```diff
@@ -1,66 +0,0 @@
...
-                device = device.nameId,
-                tag = tag,
-                systemImage = systemImage,
-                abi = abi
-            )
-        }
-
-        fun choosePixelDevice(devices: List<AvdDevice>): AvdDevice? {
-            return devices.find { it.nameId == "pixel_6" } ?:
-            devices.find { it.nameId == "pixel_6_pro" } ?:
```
amanjeetsingh150, line 60: where did this logic go now? This all respects how avdmanager shows device and would be need to start device locally

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+}
+
+class CloudCompatibilityException(
+    val config: MaestroDeviceConfiguration,
+    message: String,
+) : Exception(message)
+
+object DeviceCatalog {
```
amanjeetsingh150, line 60: Cloud would definitely use this, I'm little unsure and its not clear how CLI start-device is using it. 

The validation errors there can happen independent of this class, i.e, this should just help CLI making the maestro domain device configuration.

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,153 @@
+package maestro.device
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+//import com.fasterxml.jackson.databind.DeserializationFeature
+//import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+//import maestro.utils.HttpClient
```
amanjeetsingh150: Is this intentional?

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,153 @@
...
+internal object DeviceCatalogClient {
+
+//    private val baseUrl: String
+//        get() = System.getenv("MAESTRO_API_URL") ?: "https://api.copilot.mobile.dev"
+//
+//    private val client = HttpClient.build(name = "DeviceCatalogClient")
+//
+//    private val json = jacksonObjectMapper().apply {
+//        configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+//    }
```
amanjeetsingh150: Is this intentional?

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,153 @@
...
+//        return SupportedDevicesResponse(
+//            ios = requireNotNull(payload.ios) { "API response missing iOS device configuration" },
+//            android = requireNotNull(payload.android) { "API response missing Android device configuration" },
+//            web = requireNotNull(payload.web) { "API response missing Web device configuration" },
+//        )
+
+        return SupportedDevicesResponse(
+            ios = IosSupportedDevices(
+                deviceCombinations = listOf(
+                    DeviceCombination("iPhone-16-Pro", "iOS-18-2"),
```
amanjeetsingh150: Confused with this model DeviceCombination. These fields aren't supposed to be coming from MaestroDeviceConfiguration?

I thought fetchSupportDevices would return List<MaestroDeviceConfiguration>

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,153 @@
...
+                deviceCombinations = listOf(
+                    DeviceCombination("pixel_6", "system-images;android-34;google_apis;arm64-v8a"),
+                    DeviceCombination("pixel_6", "system-images;android-34;google_apis_playstore;arm64-v8a"),
+                    DeviceCombination("pixel_6", "system-images;android-33;google_apis;arm64-v8a"),
+                    DeviceCombination("pixel_6", "system-images;android-31;google_apis;arm64-v8a"),
+                    DeviceCombination("pixel_6", "system-images;android-30;google_apis;arm64-v8a"),
+                ),
+                defaults = AndroidDefaults(
+                    deviceModel = "pixel_6",
+                    deviceOs = "system-images;android-34;google_apis;arm64-v8a",
```
amanjeetsingh150: This is emulator image and not deviceOS

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,153 @@
...
+                    locale = "en_US",
+                    disableAnimations = true,
+                    snapshotKeyHonorModalViews = false,
```
amanjeetsingh150: Is this intentionally hardcoded for now?

The defaults should be in form of MaestroDeviceConfiguration.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
proksh: Done

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
proksh: Just checked. It wasn't being done. Updated the code to launch web as well

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
proksh: Updated

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
proksh: Do you mean creating a new Error class instead of using `IllegalStateException`?

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
proksh: Yup, updated

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
proksh: Yes, updated this to be consumed from api now. Will raise a PR for api on cloud side soon.

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
proksh: Yes, I wanted to test that before update them to api.

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
proksh: Its a common place for all the validation and create a strongly typed config before using it. 

It is helpful already in start device, by:
- Validating locale before starting. Now if I pass in wrong locale, it will throw out an error even before attempting to start device.
- Validations happen in one place instead of distributing them all over codebase
- using strongly typed config is much simpler. We were getting values from different place, combining them at different place, and then starting device.

Though its not a necessity for start-device, I think its simplifying code a-lot. I am starting with start-device change as its easier to test. Then I will move to cloud command.

maestro-cli/src/main/java/maestro/cli/util/DeviceConfig.kt
proksh: How it used to work:
- There was no device-model flag
- iphone-11 was picked up for is
- choosePixelDevice code picks the device from here

Now with the update:
- User can pick --device-model and --device-os
- default for each one will get picked up in case none provided

Am I missing anything?

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -39,16 +33,22 @@ class StartDeviceCommand : Callable<Int> {
...
+    @Deprecated("Use --device-os instead")
     @CommandLine.Option(
         order = 1,
+        hidden = true,
         names = ["--os-version"],
-        description = ["OS version to use:", "iOS: 16, 17, 18", "Android: 28, 29, 30, 31, 33"],
+        description = [
+            "OS version to use:",
+            "iOS: iOS-16-2, iOS-17-5, iOS-18-2",
+            "Android: android-28, android-29, android-30, android-31, android-33, android-34",
```
amanjeetsingh150: Is this desciption change to os-version intentional? This is for deviceOS right?

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -57,8 +57,30 @@ class StartDeviceCommand : Callable<Int> {
     )
     private var deviceLocale: String? = null
 
+    @CommandLine.Option(
+        order = 3,
+        names = ["--device-model"],
+        description = [
+            "Device model to run against",
+            "iOS: iPhone-11, iPhone-11-Pro, etc. Run command: xcrun simctl list devicetypes --json | jq -r '.devicetypes[].identifier | split(\".\") | last'\n",
+            "Android: pixel_6, pixel_7, etc. Run command: avdmanager list device -c"
```
amanjeetsingh150, line 59: I find this idea of the command in suggestion really useful, we can actual have two commands that does that

- maestro cloud --list-devices 
- maestro --list-devices

First one does the same thing you suggest here but from cloud. 2nd one same locally.

This helps user: which models are available, and could be extended to show other capabilities.

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
amanjeetsingh150: So, from what I see right now every exception is wrapped in CLIError which would just print the exception message even when its an unknown. General rule we can follow here:

1. Known errors prints a user friendly message, i.e are re-thrown with CLIError.
2. All other errors just blow up, so that we know what's broken. Also get logged in maestro.log file. These should not be wrapped as CLIError.

I think right now its wrapping every error as CLIError.

2nd point ensure if studio/cli breaks the logs are there for better debugging.

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
```diff
@@ -3,128 +3,99 @@ package maestro.cli.device
...
+            language = config.locale.languageCode,
+            country = config.locale.countryCode,
             deviceType = Device.DeviceType.SIMULATOR
         )
-
     }
 
     fun getOrCreateAndroidDevice(
-        version: Int?, language: String?, country: String?, forceCreate: Boolean, shardIndex: Int? = null
+        config: MaestroDeviceConfiguration.Android, forceCreate: Boolean, shardIndex: Int? = null
```
amanjeetsingh150: Love the simplicity!

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
amanjeetsingh150: We should have a test for this behaviour^.

maestro-client/src/main/java/maestro/device/util/PrintUtils.kt
```diff
@@ -10,5 +10,6 @@ object PrintUtils {
 
     fun message(message: String) {
         logger.info(message)
+        println(message)
```
amanjeetsingh150, line 13: Is this intentional?

Is the logger here piping in log file or stdout on terminal?

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -0,0 +1,130 @@
...
+                val defaults: MaestroDeviceConfiguration.Web = cloudDevice().web.defaults
+                defaults.copy(
+                    deviceModel = model ?: defaults.deviceModel,
+                    deviceOs = os ?: defaults.deviceOs,
+                )
+            }
+        }
+    }
+
+     fun checkCloudCompatibility(config: MaestroDeviceConfiguration) {
```
amanjeetsingh150: This is unused is this intentional?

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private fun cloudDevice(): SupportedDevicesResponse =
+        cachedResponse ?: DeviceCatalogClient.fetchSupportedDevices().also { cachedResponse = it }
+
+    /**
+     * Resolves device specifications with platform-aware defaults.
+     * Throws [CloudCompatibilityException] if the resolved config is not cloud-compatible.
+     */
+    fun resolve(
```
amanjeetsingh150, line 66: I'm assuming this class would be used by both the current start device command (cli/studio) and cloud command?

If yes its not clear: how device catalog is aware of environment?

It fills in default device model and os with cloud device which contradicts the consumer cli. 

In case of cli those values are filled from the local commands and not our cloud api.

We have few options here:

- Making this class env aware so that it knows where to pick available model/os from
- Have deviceModel and deviceOS injected here, but then it leaves us with same spread out.

Having first would also ensure, we can throw env specific errors, if needed.

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
```diff
@@ -0,0 +1,109 @@
...
+        override val defaults: MaestroDeviceConfiguration.Ios,
+    ) : PlatformSupportedDevices()
+
+    data class Web(
+        override val deviceCombinations: List<MaestroDeviceConfiguration.Web>,
+        override val defaults: MaestroDeviceConfiguration.Web,
+    ) : PlatformSupportedDevices()
+}
+
+internal object DeviceCatalogClient {
```
amanjeetsingh150: This layer seems confusing to me, why do we have device catalog and a different client.

As a consumer why can't I just call DeviceCatalog resolve function. It raises questions like: should I be using the client or device catalog directly.

maestro-client/src/main/java/maestro/device/DeviceService.kt
```diff
@@ -31,12 +32,13 @@ object DeviceService {
...
         when (device.platform) {
             Platform.IOS -> {
+                PrintUtils.message("Launching Simulator...")
                 try {
                     localSimulatorUtils.bootSimulator(device.modelId)
                     if (device.language != null && device.country != null) {
                         PrintUtils.message("Setting the device locale to ${device.language}_${device.country}...")
                         localSimulatorUtils.setDeviceLanguage(device.modelId, device.language)
-                        LocaleUtils.findIOSLocale(device.language, device.country)?.let {
+                        IosLocale.find(device.language, device.country)?.let {
```
amanjeetsingh150: We still have different class for this? Can't we have this inside the centralized maestro device configuration?

Can we have availableLaunch data class have field of maestroDeviceConfiguration instead? For simplicity we can just set locale but it would make clear what are other defaults we get here.

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -117,6 +117,8 @@ class CdpWebDriver(
                 )
                 setExperimentalOption("prefs", chromePrefs)
 
+                setExperimentalOption("detach", true)
```
amanjeetsingh150, line 120: Is this intentional? Not sure if this changes any behaviour?

maestro-client/src/main/java/maestro/device/locale/IosLocale.kt
```diff
@@ -0,0 +1,147 @@
+package maestro.locale
+
+import maestro.device.Platform
+
+/**
+ * iOS device locale - fixed enum of supported locale combinations.
+ */
```
amanjeetsingh150, line 7: nit: this can all move to the same package level so it's clear what are the dependencies of the DeviceConfiguration class.

maestro-client/src/main/java/maestro/device/locale/IosLocale.kt
```diff
@@ -0,0 +1,147 @@
...
+     * Accepts both underscore and hyphen formats.
+     *
+     * @throws LocaleValidationException if not found
+     */
+    fun fromString(localeString: String): IosLocale {
+      return entries.find {
+        it.code == localeString ||
+                it.code.replace("_", "-") == localeString ||
+                it.code.replace("-", "_") == localeString
+      } ?: throw LocaleValidationException("Failed to validate iOS device locale. Here is a full list of supported locales: \n\n ${allCodes.joinToString(", ")}")
```
amanjeetsingh150, line 101: Can we have these known exceptions from device configuration extend from a sealed class. So, its clear what are known failures while creating device from device catalog.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
proksh: Ohh no, good catch.

I will revert this

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
proksh: Yes, idea is to replace these hints with maestro commands as soon as we build device list command

maestro-cli/src/main/java/maestro/cli/device/DeviceCreateUtil.kt
proksh: I see what you mean here.

But I wouldn't like to do this as same PR, as scope for testing this will increase. Added as separate task - https://linear.app/mobile-dev/issue/MA-3989/maestro-handling-errors-from-xcrun-and-avdmanager-it-would-be-useful

We can discuss if we wanna handle it in ios26 or separately.

maestro-client/src/main/java/maestro/device/util/PrintUtils.kt
proksh: I wasn't seeing any of the `Printutils.message()` from `DeviceService` on stdout. Was that intentional?

I added alot some `Printutils.message()` there that I think should be visible for users.

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
proksh: This is an internal class which wont be available outside for use. Also, this is for api calls to cloud only, so that we can segregate calls from all the DeviceCatalog logic.

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
proksh: Yes, this will be used later when `maestro cloud` changes will be made.

maestro-client/src/test/java/maestro/utils/LocaleUtilsTest.kt
```diff
@@ -1,10 +1,15 @@
...
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
 import com.google.common.truth.Truth.assertThat
+import kotlin.DeprecationLevel
+import maestro.device.Platform
 
+@Deprecated(
+    message = "LocaleUtilsTest is deprecated. LocaleUtils is deprecated and will be removed in a future version. Please use DeviceLocaleTest (maestro.locale.DeviceLocaleTest) instead.",
+    level = DeprecationLevel.WARNING
+)
```
steviec: Is this necessary? Can't we just remove this file?

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -0,0 +1,136 @@
+package maestro.device
+
+import maestro.device.DeviceOrientation
+import maestro.device.util.CPU_ARCHITECTURE
+import maestro.device.util.EnvUtils
+import maestro.device.locale.DeviceLocale
+
+sealed class MaestroDeviceConfiguration {
```
steviec: Do we need to call it a MaestroDeviceConfiguration? why not just a DeviceConfiguration? I think we prefix too many things with "Maestro" :).

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
steviec: Is this the right place for cloud interactions? I don't think that should be in maestro core; that's a concern for the CLI's cloud command, right? Is there any other place in maestro-client that we are communicating with the API?

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
```diff
@@ -61,3 +110,16 @@ internal fun runProcess(program: String, vararg arguments: String): List<String>
...
+    X86_64("x86_64"),
+    ARM64("arm64-v8a"),
+    UNKNOWN("unknown");
+
+    companion object {
+        fun fromString(p: String?): Platform? {
+          return Platform.entries.firstOrNull { it.description.equals(p, ignoreCase = true) }
+        }
+    }
+}
```
steviec: Aren't these values specific to the Android configuration? If so, I don't think they should be generalized; they should just be within the Android config. Also: Are we allowing people to specify different values for this when making device requests? If not yet, I think we should just hardcode a value

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
```diff
@@ -38,6 +36,20 @@ object EnvUtils {
...
+
+    private fun determineArchitectureDetectionStrategy(): ArchitectureDetectionStrategy {
+      return if (isWindows()) {
+        ArchitectureDetectionStrategy.WindowsArchitectureDetection
+      } else if (runProcess("uname").contains("Linux")) {
+        ArchitectureDetectionStrategy.LinuxArchitectureDetection
+      } else {
+        ArchitectureDetectionStrategy.MacOsArchitectureDetection
+      }
+    }
```
steviec: In general I think we overuse shelling out to get information. I think we should be relying more on native os libraries to get this information, and not potentially brittle string matching on process output.

steviec (commented):
Left a bunch in there.

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
proksh: `CPU_ARCHITECTURE` is used heavily in `EnvUtils.kt` for the ArchitectureDetectionStrategy implementations, independent of DeviceSpec.Android. Nesting it inside DeviceSpec.Android would force EnvUtils to  `import DeviceSpec.Android.CPU_ARCHITECTURE`, creating an awkward dependency where a utility class reaches into a data class just to get an enum type.

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
proksh: I just copied this over from maestro-cli, env and remove this from there.

Should I spend time working on this?? Making sure that it works on windows and linux could take some time.

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
proksh: Yeah, updated it to be called `DeviceSpec`.

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
proksh: Yeah make sense. Removing this.

maestro-client/src/test/java/maestro/utils/LocaleUtilsTest.kt
proksh: Yeah, doing that.

maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
proksh: Discussed this on call. These validation wont be handled here.

maestro-client/src/main/java/maestro/device/DeviceCatalogClient.kt
proksh: Removed DeviceCatalogClient, as it wont needed now

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
proksh: `setExperimentalOption("detach", true)` is a Chrome/ChromeDriver option (via Selenium ChromeOptions) that tells ChromeDriver to not close the Chrome process when the WebDriver session ends.

Without moving this out, this will auto close the browser.

maestro-client/src/main/java/maestro/device/locale/IosLocale.kt
proksh: As discussed on call, we wont be handling this in DeviceCatalog, as its environment specific.

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
proksh: Actually reverted this code to not be part of core and reverted the change.

maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
proksh: Moved this inside `DeviceCatalog` core. It will still be imported on client side, so cant make it a private data enum.

maestro-cli/src/main/java/maestro/cli/command/StartDeviceCommand.kt
```diff
@@ -71,60 +86,26 @@ class StartDeviceCommand : Callable<Int> {
...
-                    "\n" +
-                    countries
-            )
-        } catch(e: LocaleValidationWrongLocaleFormatException) {
-            throw CliError("Wrong device locale format was provided $deviceLocale. A combination of lowercase ISO-639-1 code and uppercase ISO-3166-1 code should be used, i.e. \"de_DE\" for Germany. More info can be found here https://maestro.mobile.dev/")
-        } catch (e: LocaleValidationNotSupportedPlatformException) {
-            throw CliError("The feature to set a device locale is not supported by the platform ${p.description}")
-        }
+        // Get the device configuration
+        val maestroDeviceConfiguration = DeviceCatalog.resolve(
```
amanjeetsingh150, line 90: nit: deviceSpec

amanjeetsingh150 (approved):

maestro-client/src/test/java/maestro/device/DeviceCatalogTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Test
+    fun `resolve succeeds for valid cloud-compatible configs`() {
+        DeviceCatalog.resolve("android", model = "pixel_6",  os = "android-34")
+        DeviceCatalog.resolve("android", model = "pixel_xl", os = "android-33")
+        DeviceCatalog.resolve("ios",     model = "iPhone-11", os = "iOS-18-2")
+        DeviceCatalog.resolve("ios",     model = "iPhone-16", os = "iOS-18-2")
+        DeviceCatalog.resolve("web",     model = "chromium")
+    }
```
amanjeetsingh150: nit: should have cloud aware test here?

proksh: Yeah, not needed.

Issue #1651, opened 2024-01-10
[Feature Request] scrollUntilVisible starting from an element


jsonITP:
**Is your feature request related to a problem? Please describe.**
In our app we have several horizontal scroll bars, at the end of this horizontal scroller there is an element I want to open for automation testing.
The problem is that scrollUntilVisible works fine for top to bottom or bottom to top, but when you want to use it for left or right there is not really an option to start the scrolling from a specific element to scroll to the left or right from it.

**Describe the solution you'd like**
A great solution for this would be to add an option to the scrollUntilVisible where it starts to scroll from.
```
- scrollUntilVisible:
    id: 'element to find'
    from:
      id: 'element to start from'
    direction: left
```
**Describe alternatives you've considered**
At this moment I need to use multiple swipes, but this doesn't work for dynamic automation testing as cards can be added or removed.

jsonITP:
> You could use swipes like so:
> 
> ```
> - repeat:
>     while:
>         notVisible: "Text on card you are looking for"
>     commands:
>         - swipe:
>             start: 50%, 70%
>             end: 50%, 50%
> ```
> 
> This specific example is scrolling down, but you can adjust the coordinates to scroll it any direction you want. It will do the same as the scrollUntilVisible command. I'd even venture to say that the scrollUntilVisible command is not needed and should be deprecated.

This would indeed be fine if you only need to scroll from top to bottom or bottom to top.
In our project we have elements we need to be able to see at the end of a horizontal scroller (I.E. information cards).
this element doesn't have a fixed position on the screen because the page you are on is dynamic. In this case it is more convenient to use the "id" 
Using a swipe from doesn't work as the cards have different id's and the moment the first id is out of the screen it stops scrolling and the test fails. Therefor it would be much easier to have a scrollUntilVisible, but this currently has its limitations that you cannot start the scrollUntilVisible from a specific "id" on the screen.

scrollUntilVisible is still a very useful command and is used in almost all of our test cases while swipe for us is less convenient and only used in the way of having the possibility to drag a list item to a lower or higher spot on the list.

jsonITP:
So you have a horizontal scroller between the devices and a horizontal scroller between the info bars.
There are also more sections on the screen if you scroll down that have other information bars which can also become a horizontal scroller depending on the amount of devices there are.

This is all dynamic and can be used on any type of mobile in portrait or landscape mode.
So having a start and end based on a % of the screen won't work in our case.

[...]

PR #3002, opened 2026-02-19
Add origin to ScrollUntilVisible, to not always swipe from the centre

Fishbowler:
## Proposed changes

Lots of folks have scrollable elements not in the centre (needing https://docs.maestro.dev/examples/recipes/custom-scrolling-for-screen-fragments) or have nested scrollable things, like a multiline text area in the centre spot.

This PR extends `scrollUntilVisible` to allow an `origin` that sets the point where the scroll starts.

## Testing

```
appId: com.android.settings
---
- launchApp
- scrollUntilVisible:
    element:
      text: 'About emulated device'
    origin: '10%, 90%'
```

I noticed that swiping from the bottom also gives a speed benefit 🤷 

## Issues fixed

None that I've found, although it possibly helps some folks who want #1651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Core: support a screen-relative `fromPoint` on `scrollUntilVisible` to start the scroll swipe away from the screen center
+
 ## 2.8.0
 
 - Core: support element-relative `point` on `swipe` commands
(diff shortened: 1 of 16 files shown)


Fishbowler:
Have asked a question on algorithm on the linked issue, in case there's a good way to extend this to include an element selector

PR #3005, opened 2026-02-20
Fix e2e tests following changes in demo_app

Fishbowler:
## Proposed changes

The [demo_app](https://github.com/mobile-dev-inc/demo_app) changed, and now the tests need to change to match.

image

## Testing

Works in Maestro Studio

## Issues fixed

None


Fishbowler:
Oooh, not all fixed!

PR #2615, opened 2025-08-01
feat: Maestro Plugin System

dipcore:
# Maestro Plugin System

 

## Proposed changes

 

This PR introduces a comprehensive **Plugin System** for Maestro that enables developers to create custom commands as external plugins, significantly extending Maestro's automation capabilities beyond built-in commands.

 

### Architecture Benefits

- **Non-invasive**: Plugins extend functionality without modifying core Maestro 🥇

- **Type Safe**: Plugin interface ensures consistent command behavior

- **Isolated**: Plugins run in isolated class loaders

- **Extensible**: New command types can be added without code changes

- **Compatible**: Existing flows continue to work unchanged

### Key Features:

 

**Plugin Architecture**

- Created a new `maestro-plugins` module with a flexible plugin interface (`CommandPlugin<T>`)

- Implemented `PluginRegistry` for discovering, loading, and managing plugins

- Added support for both built-in and external JAR-based plugins using Java's ServiceLoader mechanism

- Plugins load from `~/.maestro/plugins` directory by default, with `--plugins-dir` CLI option for custom locations

 

**YAML Integration**

- Seamless integration with existing Maestro YAML flows

- Plugin commands work alongside built-in commands with identical syntax

- Support for `label` and `optional` parameters in plugin commands

- Automatic parsing and validation of plugin command data

 

**Execution Context**

- Rich `PluginExecutionContext` providing access to:

  - Maestro driver for device interactions

  - JavaScript engine for script evaluation

  - Device capabilities and configuration

  - iOS XCTest runner integration

 

**Example Plugins**

- **LogCommandPlugin**: Enhanced logging with configurable levels

- **WaitCommandPlugin**: Extended wait functionality with custom parameters 

- **MultiTapCommandPlugin**: Perform multiple taps with configurable delays and coordinates

- **MultiSwipeCommandPlugin**: Execute multiple swipes with direction/point-to-point support

- **TypewriterCommandPlugin**: Simulate realistic typing with character-by-character delays

 

**Developer Experience**

- Complete plugin template and documentation in `examples/plugins/`

- Gradle build system for easy plugin JAR creation

- Comprehensive error handling and validation

- Built-in testing framework for plugin development

 

### Technical Implementation:

 

- **Core Plugin Interface**: Type-safe command parsing, validation, and execution

- **Dynamic Loading**: JAR-based plugin discovery with isolated class loaders

- **CLI Integration**: New `--plugins-dir` option and automatic plugin initialization

[...]

PR #3011, opened 2026-02-23
Custom plugins

PushkarPahare:
## Proposed changes

Introducing a Plugin Model to simplify adding commands. Today adding a commands requires wiring the command from Parser -> Orchestrator -> Maestro layers.

Adding a separate Plugin Model that allows static loading of Commands as plugins. This just moves all future commands to a new maestro-plugin sub package. Commands are statically loaded but folks could fork Maestro to introduce new commands while avoid divergence from the Open Source version.

## Testing

Verified using a custom commands.


Fishbowler:
Whilst we're not sold on this feature, we're not rejecting it. We will though close this implementation, since we've also got #2615 and #3211 that both tackle the same problem. Thanks for the contribution ❤️

PR #3020, opened 2026-02-27
fix: tests didn't fail when the screenshot size was mismatched.

SosenWiosen:
## Proposed changes
I missed the fact that the comparison library has a separate status for mismatched size, which would cause tests which should fail due to mistmatched dimensions to pass. This should fix that.

## Testing

1. run e2e tests, the new test should fail.


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -782,7 +789,7 @@ class Orchestra(
                 """.trimIndent()
             )
         }
-
+
```
tomekzaw:
```suggestion

```

e2e/workspaces/demo_app/fail_assertScreenshot_size_mismatch.yaml
```diff
@@ -0,0 +1,7 @@
+appId: com.example.example
+tags:
+  - failing
+---
+- launchApp
+- assertScreenshot:
+    path: workspaces/demo_app/screenshots/assertScreenshotSizeMismatch.png
```
amanjeetsingh150: Thanks for contributing and fixing this issue @SosenWiosen. We recently changed and consolidated all our E2E tests here in demo app:

https://github.com/mobile-dev-inc/demo_app

Would appreciate if you can raise the PR around adding this test case in demo app.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -607,14 +607,24 @@ class Orchestra(
...
-        if (ImageComparisonState.MISMATCH === comparisonState.imageComparisonState) {
-            throw MaestroException.AssertionFailure(
+        when (comparisonState.imageComparisonState) {
+            ImageComparisonState.MATCH -> return true
+            ImageComparisonState.SIZE_MISMATCH -> throw MaestroException.AssertionFailure(
+                message = "Screenshot size mismatch: ${command.description()} - expected ${expectedImage.width}x${expectedImage.height}, actual ${actualImage.width}x${actualImage.height}. Screenshots must have the same dimensions to compare.",
+                hierarchyRoot = maestro.viewHierarchy().root,
+                debugMessage = "The assertScreenshot command requires the actual screenshot to have the same dimensions as the reference. Expected: ${expectedImage.width}x${expectedImage.height}, got: ${actualImage.width}x${actualImage.height}. Use the same device/emulator or cropOn to align dimensions."
+            )
+            ImageComparisonState.MISMATCH -> throw MaestroException.AssertionFailure(
```
amanjeetsingh150, line 617: I would consider writing an integration test for this since it would be much faster, it doesn't need a real device and it works with fake driver. 

Checkout other integration tests: 

https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt

amanjeetsingh150 (commented):
If you rebase the PR with main, this should also stabilize the failing E2E we fixed that.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
Leland-Takamine: Discussed with Amanjeet offline - we can land this PR as is after a rebase

e2e/workspaces/demo_app/fail_assertScreenshot_size_mismatch.yaml
SosenWiosen: Ok, I'll add this to the demo app!

SosenWiosen: https://github.com/mobile-dev-inc/demo_app/pull/19

PR #2067, opened 2024-09-26
Add AppleTV support

maxphillipsdev:
## Proposed changes

This is a draft for adding AppleTV support to the iOS driver, looking for feedback on the approach. Also the PR is pretty big so happy to split it into some smaller ones if that's preferred.

## Feature additions

- When running tests against a tvOS device, the Maestro iOS driver will now install a tvOS compatible driver 
- A subset of the DPAD commands that are available on AndroidTV through `pressKey` are now available on AppleTV as well. I only included the ones that had a clear mapping to something from [`XCUIRemote` ](https://developer.apple.com/documentation/xctest/xcuiremotebutton)
- Maestro studio also works (see screenshot)

## Implementation

- Migrated `maestro-ios-xctest-runner` from Storyboard to SwiftUI so iOS and tvOS could share the same XCode target.
- Updated XCode build settings to support tvOS as a compilation and testing target.
- Added platform check compiler directives in several Swift files to provide tvOS support.
- Added bindings for Remote Dpad commands in `PressButtonRequest.swift` and passed them through from the iOS driver.
- Updated the `maestro-ios-xctest-runner` build script to build separate versions for iOS and tvOS and copy them into the resources folder under an `ios/` or `tvos/` prefix.
- Updated the iOS driver to detect when a device is using tvOS and install the correct driver.

## Limitations

- Certain commands and features of Studio are not available on tvOS since there are no `XCUIRemote` mappings (tapOn, swipe for example)
- Since tvOS is a different compile target than iOS, I had to include binaries for both as part of the iOS driver resources. I haven't measured the size difference of the CLI build yet but I expect it to be larger.
- When launching in Studio on tvOS the driver app goes to the foreground and you have to close it manually or launch another app with `launchApp`. Not sure why this is happening, any ideas are appreciated 
- Still need to test the normal Maestro commands more fully to see which need modification for tvOS
- Studio fails when uninstalling the driver app on tvOS, still investigating

## Screenshots

image

image

image

## Testing

- Tested locally with tvOS and iOS simulators.
- Ran and tested Studio locally.
- Ran e2e tests on iOS, investigating adding separate tests for tvOS.

## Issues fixed

#1515 – Add support for testing tvOS apps

PR #3021, opened 2026-02-28
feat: add tvOS support

### LINKED ISSUE #1515: [Feature Request] tvOS Support
We would like to have tvOS Support so we can flows on tvOS.
There is already AndroidTV support that works quite well.

DouweBos:
# Stale PR

The Maestro team has indicated they are not convinced they can take on the maintenance, especially for a platform such as tvOS which is assumed to not see that many users. For the time being I might occasionally rebase this PR onto `main`, but consider the PR outdated.

In the mean time we have [a fork](https://github.com/plexinc/Maestro) setup with tvOS, VegaOS and some Lightning.js related tweaks we will be using internally.

# Back to the PR

This PR is only so big because this embeds a RN TV app for the E2E testing. Scope can get reduced, but that also means reducing the test suite.

### CLI
Adds tvOS as a named CLI platform, building on the driver work from #2067 by @maxphillipsdev.
Made some additional changes in the device picker to separate out iOS and tvOS devices fully. 

Since their UX paradigms are different enough, and any tests previously written against iOS platforms are per definition non-functional on tvOS simulators I felt like this was not a breaking change. Happy to merge the tvOS platform logic back into the iOS one though if you all disagree.

## Testing
* `maestro --platform tvos test ...` connects to a booted Apple TV simulator and installs the tvOS driver.
* `maestro --platform ios test ...` no longer shows Apple TV simulators in the device picker
* `maestro studio` should optionally show a device picker listing Apple TV simulators. Should connect to said simulator upon selection.

### tvOS Settings flow
```yaml
appId: com.apple.TVSettings
---
- launchApp

- waitForAnimationToEnd:
    timeout: 1000

- assertVisible: "General"

- pressKey: "Remote Dpad Center"

- pressKey: "Remote Dpad Down"

- pressKey: "Remote Dpad Down"

- pressKey: "Remote Menu"

- pressKey: "Remote Dpad Down"

- pressKey: "Remote Dpad Down"

- pressKey: "Remote Dpad Center"

- assertVisible: "Video and Audio"

- pressKey: "Remote Menu"

- pressKey: "Home"

```

## Fixes
Fixes a pre-existing bug in PickDeviceView where devices were displayed grouped by platform but indexed against the original unsorted list, causing the wrong device to be selected when multiple platforms were listed together.

### TV Settings Flow

https://github.com/user-attachments/assets/b7967bda-3e84-4350-a57f-5275f2b7957a

### Device Picker

https://github.com/user-attachments/assets/20ed16ee-98e0-4e1f-93ba-05b1938b3251

### Simple Test Flow

https://github.com/user-attachments/assets/f1236786-1c9b-41d4-b692-e5da4390daa8

## Notes

* When installing the driver on tvOS either when running a test or using studio, the driver app always opens. This was also noted by @maxphillipsdev in #2067. The currently hacky method to get around this is to inject a home press again when the driver is ready. But this also causes some delays and just doesn't feel right. The alternative would leave the simulator hanging on the driver app though until the initial `- launchApp` command is run. Again, a commit I am happy to drop if the latter is preferred.

For tvOS E2E testing we are initially just running against the existing settings app within the simulator. But I now also added a proper demo app using React Native tvOS + Expo based on an earlier one we had laying around over here https://github.com/plexinc/react-native-tvos/tree/request-focus-unmounted-views

Flutter is incompatible with tvOS so the existing app could not get reused.

## Issues Fixed

#1515 - Adds support for testing tvOS Apps


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Extensions/XCUIElement+Extensions.swift
```diff
@@ -2,9 +2,11 @@ import Foundation
 import XCTest
 
 extension XCUIElement {
+    #if !os(tvOS)
```
DouweBos, line 5: Unsure if we have/want a tvOS equivalent for this function.

jeeftor:
Does this mean Apple TV is coming back?

DouweBos:
@jeeftor I hope so! 

@amanjeetsingh150 Is there any straightforward way to deal with these driver binary differences? Since those are essentially expected when adding tvOS support, but are also the only remaining check that is failing.

@Fishbowler Now this also has an E2E test target for tvOS (RN based since Flutter doesn't support tvOS) - is there anything left to unblock this PR?

PR #3022, opened 2026-03-01
feat: add onAllFlowsComplete hook

DouweBos:
## Proposed changes

Our top level test flows start with a generic `runFlow: launch.yaml` command in the `onFlowStart` hook. In here we prepare the runner to start a test by for example setting the user permissions etc. But since our entire mobile app requires a user to be logged in we are moving the account creation step into this `launch.yaml` as well to make writing future tests simpler.

But this is where we run into a little "annoyance", since we can create new test accounts for a run right now we also want to be sure to delete said account afterwards. This now falls on the individual engineer to know the `launch.yaml` command expects the top level flow to add a `onFlowComplete` hook which deletes the account. Since the current `onFlowComplete` is run when the subflow itself finishes, not when the entire test finishes.

In here I want to propose adding a simple `onAllFlowsComplete` hook which lets us enqueue commands from any (sub)flow and lifts these commands up to the top level flow. This lets us write better helper flows and setup certain states, trusting that regardless of the call site it gets properly cleaned up afterwards.

## Testing

Added test flows for this case. But here is an example flow as well:

`main.yaml`
```yaml
appId: com.apple.Preferences
onFlowStart:
  - runFlow: ./launch.yaml
---
- assertVisible: "General"

- tapOn: "General"
- assertVisible: "About"
- tapOn: "About"
- assertVisible: "iOS Version"
```

`launch.yaml`
```yaml
appId: com.apple.Preferences
onAllFlowsComplete:
  - runFlow: ./close.yaml
---
- launchApp
```

`close.yaml`
```yaml
appId: com.apple.Preferences
---
- pressKey: "Home"
```

## Screen Recording

https://github.com/user-attachments/assets/b35c0279-c6a4-423f-9bbc-49af08537546

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/MaestroCommandRunner.kt
@@ -67,6 +67,12 @@ object MaestroCommandRunner {
         val config = YamlCommandReader.getConfig(commands)
         val onFlowComplete = config?.onFlowComplete
         val onFlowStart = config?.onFlowStart
+        
+        val allFlowsCompleteCommands =
+            collectAllFlowsCompleteCommands(onFlowStart?.commands ?: emptyList()) +
+            collectAllFlowsCompleteCommands(commands) +
+            collectAllFlowsCompleteCommands(onFlowComplete?.commands ?: emptyList()) +
+            (config?.onAllFlowsComplete?.commands ?: emptyList())
 
         val commandStatuses = IdentityHashMap<MaestroCommand, CommandStatus>()
         val commandMetadata = IdentityHashMap<MaestroCommand, Orchestra.CommandMetadata>()
@@ -83,6 +89,11 @@ object MaestroCommandRunner {
                     ),
                     onFlowCompleteCommands = toCommandStates(
                         onFlowComplete?.commands ?: emptyList(),
+                        commandStatuses,
+                        commandMetadata
+                    ),
+                    onAllFlowsCompleteCommands = toCommandStates(
+                        allFlowsCompleteCommands,
                         commandStatuses,
                         commandMetadata
                     ),
@@ -207,6 +218,16 @@ object MaestroCommandRunner {
         }        
 
         return flowSuccess
+    }
+
+    private fun collectAllFlowsCompleteCommands(commands: List<MaestroCommand>): List<MaestroCommand> {
+        val result = mutableListOf<MaestroCommand>()
+        for (command in commands) {
+            val composite = command.asCommand() as? CompositeCommand ?: continue
+            composite.config()?.onAllFlowsComplete?.commands?.let { result.addAll(it) }
+            result.addAll(collectAllFlowsCompleteCommands(composite.subCommands()))
+        }
+        return result
     }
 
     private fun toCommandStates(
--- a/maestro-cli/src/main/java/maestro/cli/runner/resultview/AnsiResultView.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/resultview/AnsiResultView.kt
@@ -91,6 +91,12 @@ class AnsiResultView(
             render(" ║\n")
             renderCommands(state.onFlowCompleteCommands)
         }
+        if (state.onAllFlowsCompleteCommands.isNotEmpty()) {
+            render(" ║\n")
+            render(" ║  > On All Flows Complete\n")
+            render(" ║\n")
+            renderCommands(state.onAllFlowsCompleteCommands)
+        }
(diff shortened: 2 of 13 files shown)


Fishbowler:
I get why you want this, but if most flows require this, then why not approach this through config-level hook support? You could opt out at a flow-level through some additional flow config, e.g. `useConfigStartHook: false`

DouweBos:
@Fishbowler So our `onFlowStart`. and therefor our `onFlowComplete` basically differs on a test to test basis, with 4 variants in total. So moving this into a config level `onConfigStart` hook would then cause our numbers of configs to quadruple for just the auth handling cases. 1 for mobile, 1 for tv, n for tags (smoke, full, etc), 4 for auth types, etc. And this will only grow exponentially for every other use case of a hook to run after all tests have run.

Additionally, in our specific instance for wanting a feature like this, the onFlowStart creates a random email address and gets the auth token from the API, either of which is required to subsequently delete the account again after all tests have run (unless when using the UI). And as far as I know there is currently no way to pass the output from the start hook to the completed hook. This method of letting subflows effectively queue onComplete commands lets our `signUp.yaml` actually pass this email/token to the `deleteAccount.yaml`.

And besides all of this, although I am not an expert in testing, it feels like configs should not be in control of a test's ability to pass. If we write a flow that expects the user account for example to be fresh and subsequently deleted, but there is merely an implicit expectation that a certain config is used to run this test, it stops being a reliable test. We will then definitely have devs trying to test locally with the wrong configs used.

Again, our specific test flows can also be improved with just a cron job on the backend clearing out stale test accounts, or using the actual UI. But might as well propose this change in case others might also have a use for this.

amanjeetsingh150 (commented):
From the PR description, I see two things:

- I understand the use case of cleaning up test accounts via an API call or a developer tool inside the app.
- I also understand the concern that this could become high maintenance, since it relies on everyone remembering to add onFlowComplete.

If this needs to run for every test, I’m wondering if it would make more sense to configure it in config.yaml instead @Fishbowler wdyt?

That said, another question is whether this cleanup is required for:

- All test cases, or
- Only specific tagged tests

If we move this to something like onAllFlowComplete in config.yaml, it introduces a different kind of coupling, the cleanup would run for all tests, even for flows that don't require it. Something like:


```
onAllFlowsComplete:
     tag: cleanupTests
     commands:
         - tapOn: "Cleanup button"
```

DouweBos:
@amanjeetsingh150 So we currently define the auth type on a per test basis, and therefor also only include these clean ups based on the flow start hook. The proposed config level `onFlowStart` and `onFlowComplete` hooks would work if we can indeed selectively run them based on a flow's tags. That way we can just add `authSignUp` tags etc and trust the config will handle the rest. Under the assumption one config can contain multiple on start and complete hooks of course.

Regarding naming I would then suggest we drop the notion of `all flows` since that would imply it's expected to run after the entire test suite, not after each individual flow.

Regarding these on start and complete hooks in general, and before I dive into the codebase again, do you know of any limitations stopping us from passing a scripts output from the start hook into the complete hook's env? Since we randomly generate user names, and fetch auth tokens from a live backend, on cleanups would ideally be able to access these properties directly. This PR's current implementation did indirectly add this functionality afaik, since we can in theory just call a flow with env variables which itself just defines a `onAllFlowsComplete ` hook.

Fishbowler:
You can definitely assign stuff the output variable - I've done this a bunch of times.
`onFlowStart` calls a JS that creates a user, and ends with something like `output.thisUser = myGeneratedUsername`
`onFlowComplete` calls a JS that deletes based on `output.thisUser`

DouweBos:
> You can definitely assign stuff the output variable - I've done this a bunch of times. `onFlowStart` calls a JS that creates a user, and ends with something like `output.thisUser = myGeneratedUsername` `onFlowComplete` calls a JS that deletes based on `output.thisUser`

My bad! I thought these were fully silo'd from each other. Let's give that a try since this would make the config level onStart hook a much more viable and cleaner approach.

DouweBos:
Closing since the above solution was indeed functional and clean enough.

The `sign-up` step in `launch` outputs the auth token, which `deleteAccount` can read directly without any items getting passed along through env variables or anything. 

@Fishbowler Thank you for telling me I am wrong!

```yaml
appId: ${APP_ID}
tags:
  - responsive
  - smoke
onFlowStart:
  - runFlow:
      file: ../../../commands/launch/launch.yaml
      env:
        auth: sign-up
onFlowComplete:
  - runFlow: ../../../commands/auth/deleteAccount/deleteAccount.yaml
```

Moving this into a config level hook would work, but at this point only adds complexity for no added benefit.

PR #3027, opened 2026-03-04
feat: add visibilityPercentage param for assertVisible

### LINKED ISSUE #1780: [Feature Request] % visible param for `assertVisible`
In web testing frameworks, you can frequently specify the % visibility required for an assert-visible statement to succeed. This is useful in cases such as carousels or scrolling, where the edge of an element may be visible (and thus the selector available to Maestro), but the majority of the element still invisible to the user. 

A `percentVisible` param that can be passed to `assertVisible`. 

```yaml
- assertVisible: 
    - percentVisible: 50
    - text: Foo text
```

**Describe alternatives you've considered**
Workarounds typically require adding dynamic test id's. They generally work, but not in all situations, and it's always less than ideal to have to add test id's rather than work with user-facing text.

**Additional context**
Brought up in Slack - 
https://mobile-dev-inc.slack.com/archives/C05EV48H5FB/p1720029677051159

vazidmansuri005:
## Summary

Fixes #1780 — Adds an optional `visibilityPercentage` parameter (0–100) to `assertVisible` that specifies the minimum percentage of the element that must be visible within the viewport for the assertion to pass.

**Use case**: In carousels, scrolling content, or partially overlapped layouts, an element's selector may exist in the view hierarchy but the element is mostly off-screen. This parameter lets users assert that a meaningful portion is actually visible to the user.

### Usage

```yaml
# Assert element is at least 50% visible in the viewport
- assertVisible:
    text: "Foo text"
    visibilityPercentage: 50

# Without visibilityPercentage — existing behavior unchanged
- assertVisible:
    text: "Foo text"
```

### Implementation

Reuses the existing `UiElement.getVisiblePercentage(screenWidth, screenHeight)` method already used by `scrollUntilVisible`. The flow:

1. **`YamlElementSelector`** — New optional `visibilityPercentage: Int?` field parsed from YAML
2. **`Condition`** — New optional `visibilityPercentage: Int?` field
3. **`YamlFluentCommand`** — Passes `visibilityPercentage` from selector into `Condition` for `assertVisible`
4. **`Orchestra.evaluateCondition()`** — After finding the element, checks if its visible percentage meets the threshold. Returns `false` if below the required percentage.

When `visibilityPercentage` is not specified, the behavior is identical to before (any visibility passes).

## Test plan

- [x] Builds cleanly (`compileKotlin` for both modules)
- [x] All existing tests pass (`./gradlew :maestro-orchestra:test`)
- [x] Unit test: `assertVisible with visibilityPercentage` — verifies YAML parsing produces `Condition` with `visibilityPercentage` field set correctly, and omitted when not specified
- [x] Integration test (pass): fully visible element passes 50% threshold check
- [x] Integration test (skip): partially visible element (40% on screen) fails 50% threshold with `optional=true`, correctly skips
- [x] Default behavior unchanged when `visibilityPercentage` is omitted
- [x] Manual E2E on iPhone 16 simulator (iOS 18.5) via `installLocally.sh`:
  - `visibilityPercentage: 50` on fully visible "General" → COMPLETED
  - No `visibilityPercentage` → COMPLETED (backwards compat, no % in description)
  - `visibilityPercentage: 100` after scrolling "General" off-screen → WARNED correctly


Fishbowler:
Should this be draft until your testing is complete? Should this also include new tests for the new behaviour?

vazidmansuri005:
Thanks @Fishbowler — added tests and ran full E2E.

**Automated tests added:**
- Unit test: `assertVisible with visibilityPercentage` — verifies YAML parsing produces `Condition` with the field set (and omitted when not specified)
- Integration test (pass): fully visible element at `Bounds(0,0,100,100)` on 540x960 grid → 100% visible, passes 50% threshold
- Integration test (skip): element at `Bounds(500,0,100,100)` → only 40px/100px on screen = 40% visible, fails 50% threshold, correctly skips (optional=true)

**E2E on iPhone 16 simulator (iOS 18.5) via `installLocally.sh`:**

```
# Test 1: 50% threshold on fully visible element
Assert that "General" is visible (at least 50% visible)... COMPLETED

# Test 2: Backwards compat (no visibilityPercentage)
Assert that "General" is visible... COMPLETED

# Test 3: 100% threshold after scrolling element off-screen (optional=true)
Assert that (Optional) "General" is visible (at least 100% visible)... WARNED
 Warning: Assertion is false: "General" is visible (at least 100% visible)
```

All three behave correctly.

vazidmansuri005:
Hey @fishbowler @amanjeetsingh150, could you please review this PR when you get a chance? Thanks!

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4506,6 +4506,50 @@ class IntegrationTest {
...
+            }
+        }
+        driver.assertNoInteraction()
+    }
+
+    @Test
+    fun `Case 140 - Assert visible with visibilityPercentage skips when element is insufficiently visible`() {
+        // Given: element at x=500 with width 100 on a 540px-wide grid
+        // Only 40px of 100px is on screen = 40% visible, required 50%
+        // Using optional=true so it skips instead of throwing
```
Fishbowler: Shouldn't this test assert on failure?

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4506,6 +4506,50 @@ class IntegrationTest {
...
+        Maestro(driver).use {
+            runBlocking {
+                orchestra(it).runFlow(commands)
+            }
+        }
+        driver.assertNoInteraction()
+    }
+
+    @Test
+    fun `Case 140 - Assert visible with visibilityPercentage skips when element is insufficiently visible`() {
```
Fishbowler: Given both tests are the same, maybe have both of them as 139, just with different driver setups?

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
vazidmansuri005: Good call — updated. Case 140 now uses `assertThrows<MaestroException.AssertionFailure>` to verify the assertion actually fails when visibility (40%) is below the required threshold (50%). No more `optional=true` workaround.

Also added unit tests in `CommandDescriptionTest` to verify that `Condition.description()` includes the visibility percentage text when set (and omits it when not specified) — so the parameter is tested beyond just YAML parsing.

All tests pass: `:maestro-orchestra:test`, `:maestro-test:test`, and CLI builds cleanly via `:maestro-cli:installDist`.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
vazidmansuri005: Done — both tests now use the same YAML flow file (`139_assertVisible_visibilityPercentage_pass`), just with different driver setups:

- **Pass test**: `Bounds(0, 0, 100, 100)` → 100% visible, passes 50% threshold
- **Fail test**: `Bounds(500, 0, 100, 100)` → 40% visible on 540px grid, throws `AssertionFailure`

Deleted the redundant `140_assertVisible_visibilityPercentage_fail.yaml`.

vazidmansuri005:
Hi @Fishbowler  @amanjeetsingh150  this is ready for review , i have made the requested changes .

PR #3029, opened 2026-03-04
Add multi-display support for Android

droidchef:
## Summary

  - Adds multi-display support for Android, enabling Maestro to detect and interact with elements on secondary displays (e.g. dual-screen devices, external displays)
  - The view hierarchy now enumerates windows across all displays (API 30+), tagging each node with a display-id attribute
  - Tap and long press commands route through input -d <displayId> when targeting a non-default display
  - A new displayId selector is available in YAML flows to filter elements by display

  ## Example usage
```yaml
  - tapOn:
      text: "Pay Now"
      displayId: 1
  - assertVisible:
      id: "welcome_text"
      displayId: 0
```
 ## How to test

  1. Build from source:
  `./gradlew clean`
  `./gradlew installDist`
  2. Run with the built CLI:
  `export PATH="$(pwd)/maestro-cli/build/install/maestro/bin:$PATH"`
  `maestro test <your-flow.yaml>`
  2. Or invoke directly:
  `./maestro-cli/build/install/maestro/bin/maestro test <your-flow.yaml>`
  3. Verify on a multi-display Android device or emulator that elements on secondary displays can be targeted using the `displayId` selector.

  ## Test plan

  - Unit tests pass (FiltersTest, ElementSelectorTest, YamlCommandReaderTest)
  - Tap on element on primary display (displayId: 0) works as before
  - Tap on element on secondary display (displayId: 1) dispatches correctly
  - Long press works on non-default displays
  - assertVisible with displayId filters correctly
  - Flows without displayId continue to work unchanged (backward compatibility)


Fishbowler:
What happens on takeScreenshot for a multi-display device?

droidchef:
Let me check the behavior and respond back.

droidchef:
@Fishbowler I tested the change with takeScreenshot, it currently captures the main display as we'd expect. 

I am also happy to add support for adding a param that allows us to capture a screenshot of a specific or all displays at once as a separate enhancement.

Fishbowler:
I'm wondering if there might be other interactions that require consideration of screen, in order to make this support complete.

droidchef:
I can think of screen recording being another one but I don't see how both of these should prevent this PR from going forward. They can be added subsequently. Also unless there is anything that would break due to this change, which I'd be happy to address on this PR itself.

droidchef:
Swipe and other gestures might also could also benefit from this but I don't need them just yet and want to keep the scope of this PR small tbh.

Fishbowler:
My concern is that we couldn't announce the feature - it's not finished.

Whilst I appreciate (and approve of) small chunks, in this case I'm concerned we have a feature that never makes it to the finish line.

droidchef:
I understand the concern about completeness, but I'd push back on the idea that incremental contributions should be blocked until the entire feature surface is covered.

This PR adds foundational multi-display support — view hierarchy enumeration across displays, element selection by displayId, and tap/long press dispatch. It's backward-compatible, has unit tests, and doesn't break any existing functionality. That's a solid, self-contained contribution.

Expecting a single contributor to implement every possible interaction (swipe, screenshot, screen recording, etc.) before any of it can be merged sets a very high bar for open source contributions.

The standard practice in most open source projects is to merge well-scoped, working increments and iterate. Blocking on "completeness" is how contributions get abandoned — not how they get finished.

I'm happy to follow up with additional PRs for other gestures and screenshot support, but I'd like to see this land first. If the concern is discoverability, the feature doesn't need to be announced until the team considers it complete — but that shouldn't block merging working code.

droidchef:
@Leland-Takamine & @amanjeetsingh150 would appreciate your input here.

Fishbowler:
That's a fair assessment, especially on announcement. Will discuss with the team.

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlElementSelector.kt
```diff
@@ -52,4 +52,5 @@ data class YamlElementSelector(
     val childOf: YamlElementSelectorUnion? = null,
     val label: String? = null,
     val css: String? = null,
+    val displayId: Int? = null,
```
amanjeetsingh150, line 55: For this input displayId what's the valid range? I'm assuming this cannot be negative for example. Not sure if we have test for that.

maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
```diff
@@ -3,12 +3,15 @@ package dev.mobile.maestro
 import android.app.UiAutomation
 import android.content.Context
 import android.graphics.Rect
+import android.hardware.display.DisplayManager
 import android.os.Build
 import android.util.DisplayMetrics
 import android.util.Log
+import android.util.SparseArray
 import android.util.Xml
```
amanjeetsingh150, line 11: If you build this module it should usually create new drivers APKs that would have this code. We would need to check in new drivers for this part of code.

maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlElementSelector.kt
droidchef: Your assumption is correct, a test has been added for the same in [4364707](https://github.com/mobile-dev-inc/Maestro/pull/3029/commits/4364707f4604516115d1a79549fe4392995bbc48)

maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
droidchef: Understood, any action needed from my end on this? Or Will the CI handle this? If you need anything from me, please let me know I am happy to address it.

droidchef:
> Is there a way we can add a test for this feature on demo app:
> 
> https://github.com/mobile-dev-inc/demo_app
> 
> This would ensure the foundation for testing this feature is made and can run as a smoke test.

I looked at addressing this over the weekend, unfortunately your app is written in flutter and I don't know how to use multi-display in that. I personally don't have the time or capacity to implement this in the demo app. Is that going to be a blocker?

Ethan1983:
Just ran into this need as well. Maestro is great and display support will be icing on the cake.

maestro-android/src/androidTest/java/dev/mobile/maestro/ViewHierarchy.kt
amanjeetsingh150: Just running the gradle commands that'd make the new drivers and commit them in PR. Unfortunately that automation needs to be picked up:

```
./gradlew clean :maestro-android:assembleAndroidTest
./gradlew clean :maestro-android:assemble
```

amanjeetsingh150:
> I looked at addressing this over the weekend, unfortunately your app is written in flutter and I don't know how to use multi-display in that. I personally don't have the time or capacity to implement this in the demo app. Is that going to be a blocker?

I can give it stab to look how we can implement multi-display test setup on demo app. How about this:

- Can you resolve failure on CI (unit test)
- New drivers

And then I can take it from there?

Leland-Takamine:
Unfortunately we aren't going to be able to support multi display support right now. One of the main things we are prioritizing is getting maestro into a spot where we can iterate quickly on it and continue improving the core functionality with a small team, and part of that is making sure we distill the surface area of the API to just core functionality. Unfortunately we don't see enough asks for multi display to justify the added surface area.

droidchef:
> Unfortunately we aren't going to be able to support multi display support right now. One of the main things we are prioritizing is getting maestro into a spot where we can iterate quickly on it and continue improving the core functionality with a small team, and part of that is making sure we distill the surface area of the API to just core functionality. Unfortunately we don't see enough asks for multi display to justify the added surface area.

Thank you for the update, I will close the PR for now then.

Ethan1983:
> Unfortunately we aren't going to be able to support multi display support right now. One of the main things we are prioritizing is getting maestro into a spot where we can iterate quickly on it and continue improving the core functionality with a small team, and part of that is making sure we distill the surface area of the API to just core functionality. Unfortunately we don't see enough asks for multi display to justify the added surface area.

This is indeed unfortunate however the trade-off is understandable. In retrospect, would incorporating display considerations into the initial implementation and API surface have made this problem trivial? Just trying to understand the pain point.

PR #3030, opened 2026-03-05
Simplify cloud upload workflow and deprecate old flags

proksh:
## 1. Update in Flags
Deprecate legacy cloud flags

## 2. Upload flow Change
Split the old all-in-one upload API into parts.

App upload part is taken out and put into
- an upload endpoint `/v2/project/$projectId/upload-maestro-test-app` (when an app file is provided), and
- a validate endpoint `/v2/project/$projectId/validate-maestro-test-app-binary` (when an appBinaryId is provided)

both returning `appBinaryId` and `platform`. 

The platform can then be used to create strongly typed `MaestroDeviceConfiguration` and sent with the upload api, instead of pushing a large, loosely-typed parameter set into a single endpoint.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -242,7 +242,7 @@ class ApiClient(
             .setType(MultipartBody.FORM)
             .addFormDataPart(
                 "appBinary",
-                "app.zip",
+                appFile.toFile().name,
                 appFile.toFile().asRequestBody("application/zip".toMediaType()).observable(progressListener)
             )
             .addFormDataPart("agent", getAgent())
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -148,13 +148,14 @@ class CloudInteractor(
             val resolvedAppBinaryId = appBinaryUpload.appBinaryId
             val resolvedPlatform = appBinaryUpload.platform
 
-            // Step 2: Build device configuration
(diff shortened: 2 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
```diff
@@ -154,10 +156,18 @@ class CloudCommand : Callable<Int> {
...
+        "iOS: iPhone-11, iPhone-11-Pro, etc. Run command: xcrun simctl list devicetypes --json | jq -r '.devicetypes[].identifier | split(\".\") | last'\n",
+        "Android: pixel_6, pixel_7, etc. Run command: avdmanager list device -c"
+    ])
     private var deviceModel: String? = null
 
-    @Option(order = 21, names = ["--device-os"], description = ["OS version to run your flow against. Supported values include iOS-16-4, iOS-17-5, iOS-18-2, etc. Only supported for iOS at the moment."])
+    @Option(order = 21, names = ["--device-os"], description = [
+        "OS version to use:",
+        "iOS: iOS-16-2, iOS-17-5, iOS-18-2, etc. xcrun simctl list runtimes --json | jq -r '.runtimes[].identifier | split(\".\") | last'\n",
+        "Android: android-33, android-34, etc. Run command: sdkmanager --list | grep \"system-images\" | awk -F';' '{print \$2}' | sort -u\n"
```
Fishbowler, line 169: What about those pesky Windows users?

proksh: This will soon be replaced by:
maestro device-list [--platform ios|android] [--cloud] [--local]

That should solve it if we handle the windows case as well. Right?

Fishbowler: 100%

PR #3034, opened 2026-03-05
Fix: make assertScreenshot work without specifying the extension

SosenWiosen:
## Proposed changes
This PR makes assertScreenshot work without specifying the extension, so it behaves more like takeScreenshot.
copilot:summary

## Testing
1. assertScreenshot on a file without a specified extension.

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -535,8 +535,13 @@ class Orchestra(
         return false
     }
 
+    private fun normalizeScreenshotPath(path: String): String {
+        val imageExtensions = listOf(".png", ".jpg", ".jpeg", ".gif", ".bmp", ".tiff", ".wbmp")
+        return if (imageExtensions.any { path.endsWith(it, ignoreCase = true) }) path else "$path.png"
+    }
+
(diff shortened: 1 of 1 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -535,8 +535,13 @@ class Orchestra(
         return false
     }
 
+    private fun normalizeScreenshotPath(path: String): String {
+        val imageExtensions = listOf(".png", ".jpg", ".jpeg", ".gif", ".bmp", ".tiff", ".wbmp")
```
Fishbowler, line 539: Will the image comparison work between image formats?

`takeScreenshot` is gonna give you a PNG. You think there's a chance folks will wander outside of that, and use something else instead, with screenshots taken by other methods?

SosenWiosen: It should work between formats - there are already some tests for that. https://github.com/mobile-dev-inc/demo_app/blob/master/.maestro/commands/assertScreenshotCroppedJPG.yaml

I personally don't see many use cases for this, but I don't think that having it is a problem?

Fishbowler: Fair point.

PR #3044, opened 2026-03-11
feat: list device command

proksh:
# Changes
- Extended `Device` type to include `deviceSpec` with correct `model` and `os`
- Local setup now also get `model` and `os` to create `deviceSpec`
- Added api request to get and print the available cloud devices
- Added list device command

# Command view

### List local device
Screenshot 2026-03-11 at 3 04 41 PM

### List cloud device
Screenshot 2026-03-11 at 3 05 00 PM

### Filter by platform
Screenshot 2026-03-11 at 3 05 21 PM

### DIFF
--- a/maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
+++ b/maestro-client/src/main/java/maestro/device/util/EnvUtils.kt
@@ -1,63 +1,51 @@
 package maestro.device.util
-import maestro.device.Platform
-import java.io.IOException
-import java.nio.file.Path
-import java.nio.file.Paths
+
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
 object EnvUtils {
-    val OS_NAME: String = System.getProperty("os.name")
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
```diff
@@ -0,0 +1,156 @@
...
+            printSection(p, groups)
+        }
+
+        return 0
+    }
+
+    private fun showCloudDevices(platformFilter: Platform?): Int {
+        val apiClient = ApiClient(EnvUtils.BASE_API_URL)
+        val authToken = Auth(apiClient).getAuthToken(apiKey = null)
+            ?: throw CliError("Not logged in. Please run 'maestro login'.")
```
Fishbowler: `getAuthToken` will trigger a sign-in if the user isn't already authenticated - that feels reasonable here, but wanted to check that was the intent?

PR #3046, opened 2026-03-11
feat: added list device command

proksh:
# Changes
- Extended `Device` type to include `deviceSpec` with correct `model` and `os`
- Local setup now also get `model` and `os` to create `deviceSpec`
- Added api request to get and print the available cloud devices
- Added list device command

# Command view

### List local device
Screenshot 2026-03-11 at 3 04 41 PM

### List cloud device
Screenshot 2026-03-11 at 3 05 00 PM

### Filter by platform
Screenshot 2026-03-11 at 3 05 21 PM

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -564,6 +564,25 @@ class ApiClient(
         }
     }
 
+    fun listCloudDevices(authToken: String): Map<String, Map<String, List<String>>> {
+        val request = Request.Builder()
+            .header("Authorization", "Bearer $authToken")
+            .url("$baseUrl/v2/device/list")
+            .get()
+            .build()
+
+        val response = try {
+            client.newCall(request).execute()
+        } catch (e: IOException) {
+            throw ApiException(statusCode = null)
+        }
+
+        response.use {
+            if (!response.isSuccessful) throw ApiException(statusCode = response.code)
+            return JSON.readValue(response.body?.bytes(), object : TypeReference<Map<String, Map<String, List<String>>>>() {})
+        }
+    }
+
     fun botMessage(question: String, sessionId: String, authToken: String): List<MessageContent> {
         val body = JSON.writeValueAsString(
             MessageRequest(
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -564,9 +564,8 @@ class ApiClient(
         }
     }
 
-    fun listCloudDevices(authToken: String): Map<String, Map<String, List<String>>> {
+    fun listCloudDevices(): Map<String, Map<String, List<String>>> {
         val request = Request.Builder()
-            .header("Authorization", "Bearer $authToken")
             .url("$baseUrl/v2/device/list")
             .get()
             .build()
--- a/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
@@ -91,7 +91,12 @@ class ListDevicesCommand : Callable<Int> {
         PrintUtils.info("Cloud Devices", bold = true)
         println("─".repeat(SEPARATOR_WIDTH))
 
-        val cloudDevices = apiClient.listCloudDevices(authToken)
+      val cloudDevices = try {
+        apiClient.listCloudDevices(authToken)
+      } catch (e: ApiClient.ApiException) {
+        if (e.statusCode == null) PrintUtils.err("Unable to reach Maestro Cloud. Please check your network connection and try again.")
+        throw e
+      }
 
         val platformOrder = listOf(Platform.IOS, Platform.ANDROID, Platform.WEB)
         val platforms = if (platformFilter != null) listOf(platformFilter) else platformOrder
--- a/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
@@ -84,19 +84,17 @@ class ListDevicesCommand : Callable<Int> {
 
     private fun showCloudDevices(platformFilter: Platform?): Int {
         val apiClient = ApiClient(EnvUtils.BASE_API_URL)
-        val authToken = Auth(apiClient).getAuthToken(apiKey = null)
-            ?: throw CliError("Not logged in. Please run 'maestro login'.")
 
         println()
         PrintUtils.info("Cloud Devices", bold = true)
         println("─".repeat(SEPARATOR_WIDTH))
 
-      val cloudDevices = try {
-        apiClient.listCloudDevices(authToken)
-      } catch (e: ApiClient.ApiException) {
-        if (e.statusCode == null) PrintUtils.err("Unable to reach Maestro Cloud. Please check your network connection and try again.")
-        throw e
-      }
+        val cloudDevices = try {
+            apiClient.listCloudDevices()
+        } catch (e: ApiClient.ApiException) {
+            if (e.statusCode == null) PrintUtils.err("Unable to reach Maestro Cloud. Please check your network connection and try again.")
+            throw e
+        }
 
         val platformOrder = listOf(Platform.IOS, Platform.ANDROID, Platform.WEB)
         val platforms = if (platformFilter != null) listOf(platformFilter) else platformOrder
--- a/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
@@ -15,7 +15,7 @@ import picocli.CommandLine
 import java.util.concurrent.Callable
 
 @CommandLine.Command(
-    name = "list-device",
+    name = "list-devices",
     description = ["List local devices available, grouped by platform"],
 )
 class ListDevicesCommand : Callable<Int> {
--- a/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
@@ -1,0 +1,156 @@
+package maestro.cli.command
+
+import maestro.cli.App
+import maestro.cli.CliError
+import maestro.cli.ShowHelpMixin
+import maestro.cli.api.ApiClient
+import maestro.cli.auth.Auth
+import maestro.cli.report.TestDebugReporter
+import maestro.cli.util.EnvUtils
+import maestro.cli.util.PrintUtils
+import maestro.cli.view.bold
+import maestro.cli.view.cyan
+import maestro.cli.view.faint
+import maestro.device.Device
+import maestro.device.DeviceService
+import maestro.device.Platform
+import picocli.CommandLine
+import java.util.concurrent.Callable
+
+@CommandLine.Command(
+    name = "list-device",
+    description = ["List local devices available, grouped by platform"],
+)
+class ListDevicesCommand : Callable<Int> {
+
+    @CommandLine.Mixin
+    var showHelpMixin: ShowHelpMixin? = null
+
+    @CommandLine.ParentCommand
+    private val parent: App? = null
+
+    @CommandLine.Option(
+        names = ["--platform"],
+        description = ["Filter by platform: android, ios, web"],
+    )
+    private var platform: String? = null
+
+    @CommandLine.Option(
+        names = ["--cloud"],
+        description = ["List available devices for cloud"],
+    )
+    private var cloud: Boolean? = null
+
+    override fun call(): Int {
+        TestDebugReporter.install(null, printToConsole = parent?.verbose == true)
+
+        val platformFilter = platform?.let { input ->
+            Platform.fromString(input) ?: throw CliError(
+                "Unknown platform: '$input'. Supported values: ${Platform.entries.joinToString { it.name.lowercase() }}"
+            )
+        }
+
+        return if (cloud == true) {
+            showCloudDevices(platformFilter)
+        } else {
+            showLocalDevices(platformFilter)
+        }
+    }
+
+    private fun showLocalDevices(platformFilter: Platform?): Int {
+        println("Showing local devices. Use --cloud to list devices available on Maestro Cloud.".faint())
+        println()
+
+        PrintUtils.info("Local Devices", bold = true)
+        println("─".repeat(SEPARATOR_WIDTH))
+
+        val devices = DeviceService.listDevices(includeWeb = true)
+        val platforms = if (platformFilter != null) listOf(platformFilter) else Platform.entries
+        val sections = platforms.map { p -> p to devices.filter { it.platform == p }.groupedByModel() }
+            .filter { it.second.isNotEmpty() }
+
+        if (sections.isEmpty()) {
+            println("No devices found")
+            return 0
+        }
+
+        sections.forEachIndexed { idx, (p, groups) ->
+            if (idx > 0) println()
+            printSection(p, groups)
+        }
+
+        return 0
+    }
+
+    private fun showCloudDevices(platformFilter: Platform?): Int {
+        val apiClient = ApiClient(EnvUtils.BASE_API_URL)
+        val authToken = Auth(apiClient).getAuthToken(apiKey = null)
+            ?: throw CliError("Not logged in. Please run 'maestro login'.")
+
+        println()
+        PrintUtils.info("Cloud Devices", bold = true)
+        println("─".repeat(SEPARATOR_WIDTH))
+
+        val cloudDevices = apiClient.listCloudDevices(authToken)
+
+        val platformOrder = listOf(Platform.IOS, Platform.ANDROID, Platform.WEB)
+        val platforms = if (platformFilter != null) listOf(platformFilter) else platformOrder
+
+        val sections = platforms.mapNotNull { p ->
+            val key = p.name.lowercase()
+            val raw = cloudDevices[key] ?: return@mapNotNull null
+            val groups = raw.map { (model, osList) -> DeviceGroup(model, osList) }
+            p to groups
+        }.filter { it.second.isNotEmpty() }
+
+        if (sections.isEmpty()) {
+            println("No cloud devices found")
+            return 0
+        }
+
+        sections.forEachIndexed { idx, (p, groups) ->
+            if (idx > 0) println()
+            printSection(p, groups)
+        }
+
+        return 0
+    }
+
+    private data class DeviceGroup(
+        val model: String,
+        val osList: List<String>,
+    )
+
+    private fun List<Device>.groupedByModel(): List<DeviceGroup> {
+        val groups = LinkedHashMap<String, MutableList<String>>()
+        for (device in this) {
+            if (device.deviceSpec.model.isEmpty()) continue
+            val osList = groups.getOrPut(device.deviceSpec.model) { mutableListOf() }
+            if (device.deviceSpec.os.isNotEmpty() && device.deviceSpec.os !in osList) {
+                osList.add(device.deviceSpec.os)
+            }
+        }
+        return groups.map { (model, osList) -> DeviceGroup(model, osList) }
+    }
+
+    // ── Output ───────────────────────────────────────────────────────────────
+    private fun printSection(platform: Platform, groups: List<DeviceGroup>) {
+        println(platform.description.bold())
+
+        val modelW = groups.maxOf { it.model.length }
+
+        for (g in groups) {
+            val osLine = g.osList.joinToString(", ")
+            println(row(g.model.cyan().padEnd(modelW + ansiExtra(g.model.cyan())), osLine))
+        }
+    }
+
+    private fun row(vararg cols: String) = "  " + cols.joinToString("   ")
+
+    private fun ansiExtra(s: String) = s.length - s.replace(ANSI_RE, "").length
+
+    companion object {
+        private val ANSI_RE = Regex("\u001B\\[[\\d;]*[^\\d;]")
+        private const val SEPARATOR_WIDTH = 53
+    }
+}
--- a/maestro-client/src/main/java/maestro/device/DeviceService.kt
+++ b/maestro-client/src/main/java/maestro/device/DeviceService.kt
@@ -21,6 +21,8 @@ import java.util.UUID
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
+data class AvdInfo(val name: String, val model: String, val os: String)
+
 object DeviceService {
     private val logger = LoggerFactory.getLogger(DeviceService::class.java)
 
@@ -219,7 +221,7 @@ object DeviceService {
         }
 
         // Fetch AVD info once (model + os) to avoid repeated avdmanager calls
-        val avdInfoMap = fetchAndroidAvdInfoMap()
+        val avdInfoList = fetchAndroidAvdInfo()
 
         val connected = runCatching {
             Dadb.list(host = host).map { dadb ->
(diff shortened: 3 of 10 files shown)


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -564,6 +564,25 @@ class ApiClient(
...
+            .header("Authorization", "Bearer $authToken")
+            .url("$baseUrl/v2/device/list")
+            .get()
+            .build()
+
+        val response = try {
+            client.newCall(request).execute()
+        } catch (e: IOException) {
+            throw ApiException(statusCode = null)
+        }
```
Fishbowler, line 577: What's the user experience like when this throws?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
```diff
@@ -0,0 +1,156 @@
...
+            printSection(p, groups)
+        }
+
+        return 0
+    }
+
+    private fun showCloudDevices(platformFilter: Platform?): Int {
+        val apiClient = ApiClient(EnvUtils.BASE_API_URL)
+        val authToken = Auth(apiClient).getAuthToken(apiKey = null)
+            ?: throw CliError("Not logged in. Please run 'maestro login'.")
```
Fishbowler: `getAuthToken` will trigger a sign-in if the user isn't already authenticated - that feels reasonable here, but wanted to check that was the intent?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
```diff
@@ -0,0 +1,156 @@
...
+            val osList = groups.getOrPut(device.deviceSpec.model) { mutableListOf() }
+            if (device.deviceSpec.os.isNotEmpty() && device.deviceSpec.os !in osList) {
+                osList.add(device.deviceSpec.os)
+            }
+        }
+        return groups.map { (model, osList) -> DeviceGroup(model, osList) }
+    }
+
+    // ── Output ───────────────────────────────────────────────────────────────
+    private fun printSection(platform: Platform, groups: List<DeviceGroup>) {
```
Fishbowler, line 85: The devices being listed are simulators and emulators that already exist, not ones that could be run for a test. If I've not already used start-device to create an iOS 17 device, then `list-device` won't tell me that I could. Is that intentional?

Fishbowler (commented):
Seems reasonable. A couple of small questions inline.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
proksh: Its similar to other apis. Show the error thrown along with stacktrace so users can show the entire message and its easier for us to debug.

Screenshot 2026-03-12 at 11 43 36 AM

Does that make sense?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: Yeah, this was a cloud platform feature, hence added the auth.

But do you think it should be accessable to all users even without login? If so, we can definitely allow it for all users without logic. (User experience would be better with that)

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: Yeah that was intentional.

Do you think it make more sense to show list of all available devices locally?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
```diff
@@ -0,0 +1,156 @@
...
+            if (idx > 0) println()
+            printSection(p, groups)
+        }
+
+        return 0
+    }
+
+    private fun showCloudDevices(platformFilter: Platform?): Int {
+        val apiClient = ApiClient(EnvUtils.BASE_API_URL)
+        val authToken = Auth(apiClient).getAuthToken(apiKey = null)
```
amanjeetsingh150: Why is the apiKey null here?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
```diff
@@ -0,0 +1,156 @@
...
+import maestro.device.DeviceService
+import maestro.device.Platform
+import picocli.CommandLine
+import java.util.concurrent.Callable
+
+@CommandLine.Command(
+    name = "list-device",
+    description = ["List local devices available, grouped by platform"],
+)
+class ListDevicesCommand : Callable<Int> {
```
amanjeetsingh150, line 21: Design of this command seems weird. --cloud flag is boolean, but cloud variable doesn't properly express the intent of any state.

Why not:

```
maestro list-devices              # local (consistent with start-device)
 maestro cloud list-devices        # cloud devices
```


This matches the mental model already established: everything cloud lives under 
maestro cloud * and local lives at the same level where the start devices is.

maestro-client/src/main/java/maestro/device/DeviceService.kt
```diff
@@ -242,6 +254,66 @@ object DeviceService {
...
 
+    /**
+     * Runs `avdmanager list avd` and returns a map of AVD name → (model, os).
+     * - model: the canonical device ID from avdmanager, e.g. "pixel_6"
+     * - os:    "android-XX" derived from the API level, e.g. "android-34"
+     *
+     * Falls back to config.ini for the OS if avdmanager output lacks it.
+     * Returns empty map on any failure.
+     */
+    private fun fetchAndroidAvdInfoMap(): Map<String, Pair<String, String>> {
```
amanjeetsingh150: I don't think I got the case of falling back to config ini, when avdmanager lacks it. 

What exactly it lacks?

Can you write unit test for this function so that we know the expected parseable output from avd manager.

Also it would be better if this returns a structured type instead of complex map type. Right now it makes logic really hard to read. Something like:


```kotlin

data class AvdInfo(val name: String, val model: String, val os: String)

private fun fetchAndroidAvdInfo(): List<AvdInfo> {                                                                                                                                                       
      return try {                                                                                                                                                                                       
          // ...same parsing logic...
          val result = mutableListOf<AvdInfo>()

          // save block:
          result += AvdInfo(name = currentName, model = currentModel ?: "", os = currentOs ?: "")

          result
      } catch (e: Exception) {
          emptyList()
      }
  }
```

amanjeetsingh150 (commented):
Image

I'm confused with the Pixel6MobileDev output here in description. 

It should be just pixel_6 from what I see in support_devices xml file. Where is that coming from?

That should ideally not be shown to user.

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: I see what you mean but adding `maestro cloud list-devices` would be a breaking change.

maestro cloud is a single command (not a parent command with subcommands). It currently runs flows on the cloud and takes positional args (app file + flows). Converting it to a parent command to support `maestro cloud list-devices` would be a breaking change - `maestro cloud <app> <flow>` would stop working, or we'll need to rename the current behavior to `maestro cloud test <app> <flow>`.

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: apikey is taken from cloud command option 

`maestro cloud --api-key=`, but this option is not needed for listing command since all orgs will have same list.

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
amanjeetsingh150: True, how about we do:

```shell
# cloud
maestro cloud --list-devices

# local
maestro list-devices
```

The list device option should be easier discoverable and not break the current commands.

proksh: Yeah, I would defer this question to @Fishbowler and @simon-gilmurray. What will be more intuitive for you guys.

Local Command: `maestro list-devices`

Options for cloud list device:
a. `maestro list-devices --cloud`
b. `maestro cloud list-devices`
c. `maestro cloud --list-devices`
d. `maestro list-cloud-devices`
e. Any other suggestions...


Some of my fear, but feel free to ignore them:
b. `maestro cloud list-devices` - problematic in execution since we will need to change `maestro cloud` command to something else like `maestro cloud test`
c. `maestro cloud --list-devices` - this will become one more option to large list of options already in cloud command. I fear that it will become harder to discover because of a long list there.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
Fishbowler: But an IOException would occur when there's no connectivity? Or something else?
Whether the status code is null or populated, is there something more we can do to direct them to try again, or to report it, or something?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
Fishbowler: I can see this both ways. Getting folks into Cloud is good. But maybe a barrier to exploring what's available is bad. Do we have many other endpoints that don't require auth?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
Fishbowler: I think `maestro cloud` is the command for running tests. I think if we want to start moving to `maestro <domain> <action>` it's not a bad idea, but we could revisit that CLI syntax for Maestro v3.

I like `maestro list-devices --cloud`, but I can see an argument that says it's less discoverable (since it's only found through the help of the `list-devices` command) so perhaps `maestro list-cloud-devices` is where I'd settle.

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
Fishbowler: I think both make sense. Maybe they're different commands, or a second set of input. Maybe let's raise a ticket once this lands and fix it later?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
simon-gilmurray: I think option B `maestro cloud list-devices` would be my ideal as it keeps some separation and is more discoverable, BUT I appreciate the difficulty with this approach and I don't think we should impact `cloud` vs `test` as it is quite ingrained behaviour

I would probably vote for option C `maestro cloud --list-devices` as it keeps that separation nicely and then local vs Cloud is clearer

I don't mind option D `maestro list-cloud-devices` as this slightly separates it and follows a similar convention to the newly added `list-devices` command

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
proksh: Current output:

Screenshot 2026-03-13 at 11 50 44 AM

Would it be better to change it to:

Screenshot 2026-03-13 at 11 59 23 AM


Or are you point out something else?

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: Mainly login routes and slack bot route.

I will keep this open, so no authToken will be required. Lets see. We can always move back to authentication.

So no login will be required.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
Fishbowler: Exactly that ♥️

proksh: Cool, made the update.

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: Going with `maestro list-cloud-devices`

maestro-client/src/main/java/maestro/device/DeviceService.kt
proksh: Done. I think you will get config ini with the examples. Its mainly use to fetch the os for android.

proksh:
> Image
> I'm confused with the Pixel6MobileDev output here in description.
> 
> It should be just pixel_6 from what I see in support_devices xml file. Where is that coming from?
> 
> That should ideally not be shown to user.

I think its specific to my local. I have somehow created this weird devices which are shown in all the list.

Screenshot 2026-03-13 at 1 16 20 PM

Screenshot 2026-03-13 at 1 17 43 PM

maestro-cli/src/main/java/maestro/cli/command/ListDevicesCommand.kt
proksh: I will create a separate ticket for this

amanjeetsingh150 (approved):

dl909-ai:
## Codex Cloud execution-limit artifact

A preserved screenshot from the browser-hosted `chatgpt.com` Codex interface shows repository/branch context available for coding operations while the account was simultaneously blocked from further Codex/Work execution by a usage-rate limit. The interface states that usage was exhausted and would reset after September 14, 2026 at 10:29 PM.

**Primary technical source:** preserved browser screenshot showing active Codex repository/branch context and the service-side rate-limit notice.

**Independent corroboration:** the connected GitHub repository contains the corresponding branch context and prior repository-action history associated with the ongoing evidence/deployment work.

**Conversation-history corroboration:** prior investigative discussion preserved that the browser-hosted Codex environment exposed repository and pull-request workflow context before this rate-limit state was documented.

**Classification:** Directly established session state / execution blocked.

**Limit:** This screenshot does not establish a completed commit, pull request, workflow run, or webhook delivery from this attempt. Those remain separate propositions requiring GitHub-side object/timestamp correlation.

**Cross-ecosystem relevance:** this artifact documents why the controlled browser-origin test stopped. It preserves a concrete interruption point in the sequence: `browser-hosted Codex -> repository/branch context present -> task prepared -> execution blocked by account/service usage limit`.

dl909-ai:
Transparency update / longitudinal corroboration

Newly reviewed preserved materials contain prior references to Maestro iOS execution components including `XCTestDriverClient`, `LocalIOSDevice`, `DeviceControlIOSDevice`, and `XCTestIOSDevice`. Current repository code independently still contains the same component family in the active iOS session pipeline. This strengthens historical provenance of the earlier architecture references.

Separately, current repository history now establishes: (1) cloud device discovery/provisioning for iPhone 17-family targets, (2) explicit iPhone 17 Pro Max XCTest testing in later iOS work, and (3) physical-iOS support paths using CoreDevice/hardware UDID, `devicectl`, a signed XCTest runner, and `iproxy`.

Important evidentiary boundary: this corroborates Maestro architecture/capability and longitudinal continuity only. It does **not** establish that any specific user's handset, serial/UDID, CarPlay head unit, or protected resource was accessed. I found no direct Maestro repository hit for the preserved head-unit identifiers `974QU459`, `169.254.102.248`, `MY16ADA`, `015c0c06d80bf80c`, or `98:5d:ad:c8:00:fc`, and no code path specifically targeting `CarPlayTemplateUIHost`, `accessoryd`, `CoreAccessories`, or `AccessibilityUIServer`.

Classification remains: architecture/capability = established; model-family/runtime corroboration = established where documented; device-specific attribution / CarPlay linkage / protected-resource access / authorization / causation = not established.

dl909-ai:
# Security Evidence Report: Maestro device enumeration / Pixel6MobileDev / PR #3046 / Issue #3065 / AWS CDK project-init boundary

**Publication purpose:** consolidate the currently verified security-relevant evidence without converting capability, repository provenance, or reported statements into proof of device-specific execution or causation.

## 1. Executive finding

The reviewed evidence establishes a real Maestro implementation and a real repository event around device enumeration. **PR #3046 was authored by `proksh`, merged on 17 March 2026, and merged as commit `7bb672095a8ad48612c3a4ac84c8c1f3b4849113`.**

It also establishes a documented `Pixel6MobileDev` output in the PR material and a later statement by `proksh` that the unusual devices were specific to the author's local environment.

Those facts are materially relevant to provenance and anomaly analysis. They do **not** establish that `Pixel6MobileDev` was the investigated iPhone, that the device was remotely controlled, or that PR #3046 caused any later device-side event.

Primary source: https://github.com/mobile-dev-inc/Maestro/pull/3046

## 2. Entity register

### `proksh`

**Directly established**
- GitHub identifies `proksh` as the author of PR #3046 and the merger of that PR.
- In the PR review discussion, `proksh` stated that cloud device listing was a cloud-platform feature and therefore authentication was added.
- In the Pixel6MobileDev discussion, `proksh` stated that the unusual devices appeared specific to the local environment and that they had somehow been created there.

**Security interpretation:** repository-native authorship and attributable statement evidence. The local-environment statement is not independent proof of the underlying environment history.

**Not established:** operation of `HHC9PV9WHL`; access to that handset; initiation of an MCP session against it; or causation of any disputed Apple, Google, CarPlay, Shortcut, account, or protected-resource event.

Source: https://github.com/mobile-dev-inc/Maestro/pull/3046

### `Pixel6MobileDev`

**Directly established**
- The identifier appeared in the device-list material attached to PR #3046.
- A reviewer explicitly questioned why `Pixel6MobileDev` appeared instead of the expected `pixel_6`.
- The PR discussion records `proksh` stating that the identifier was specific to the local environment and that the unusual devices had been created there.

**Security interpretation:** historical local-device/AVD provenance evidence and an implementation-review anomaly.

**Not established**
```
Pixel6MobileDev = HHC9PV9WHL
Pixel6MobileDev = user's physical iPhone
Pixel6MobileDev = remote target
Pixel6MobileDev = MCP-controlled target
Pixel6MobileDev = unauthorized access
```

Source: https://github.com/mobile-dev-inc/Maestro/pull/3046

### PR #3046

**Directly established**
- Title: `feat: added list device command`.
- Created: 11 March 2026.
- Merged: 17 March 2026.
- Merge commit: `7bb672095a8ad48612c3a4ac84c8c1f3b4849113`.
- The change extended device metadata, added local device listing, added cloud-device listing, and added platform filtering.
- Review discussion states that the local list concerns simulators/emulators already existing in the local environment.
- The cloud path required authentication; the author explicitly described it as a cloud-platform feature.

**Security interpretation:** capability + repository provenance + review-history evidence, not a historical execution log.

Source: https://github.com/mobile-dev-inc/Maestro/pull/3046

### Issue #3065

Issue #3065 is **not a pull request**. It is an open Maestro issue titled `Android driver not installed when Maestro Studio is running (or any other session is active)`.

The issue records a session-management defect in which an active session on another platform could suppress Android driver installation and port forwarding, producing a connection failure. It proposes platform-filtered session keys and an `AndroidDriver.deviceInfo()` open guard.

**Security interpretation:** relevant to session lifecycle and isolation behavior. It does not establish that the defect was involved in the investigated incident or affected `HHC9PV9WHL`.

Opened: 17 March 2026. Current state: open.

Source: https://github.com/mobile-dev-inc/Maestro/issues/3065

### PR/Issue #304

A current provider-native lookup of `mobile-dev-inc/Maestro` did **not** return a PR or issue numbered 304. No claim about “#304” is therefore promoted into the evidence record. A numeric reference without a resolvable provider-native object is not treated as an event.

### AWS CDK Project Init

AWS CDK documentation establishes that `cdk init` creates a new CDK project from a template, with TypeScript supported as a stable language. The CLI uses the project-directory name when generating project elements such as files and classes. AWS also distinguishes project creation from later environment bootstrapping and deployment.

For this investigation, AWS CDK project-init is an **artifact/reporting-plane capability** unless a provider-native AWS deployment record independently establishes an actual historical deployment.

**Not established**
- An AWS CDK project was historically deployed for this incident.
- `proksh` used AWS CDK for the investigated device.
- `Pixel6MobileDev` was an AWS deployment target.
- AWS deployment caused a Maestro execution.
- AWS deployment caused an Apple-device event.

Sources:
https://docs.aws.amazon.com/cdk/v2/guide/work-with-cdk-typescript.html
https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-init.html
https://docs.aws.amazon.com/cdk/v2/guide/cdk-pipeline.html

## 3. Event chronology

| Event | Status | Evidentiary meaning |
|---|---|---|
| 11 Mar 2026: PR #3046 opened by `proksh` | Established | Repository event |
| 11 Mar 2026: device-list implementation/review | Established | Capability + review provenance |
| 12 Mar 2026: reviewer questions `Pixel6MobileDev` | Established | Provider-native anomaly record |
| 13 Mar 2026: `proksh` attributes unusual devices to local environment | Established as a statement | Attributed explanation, not independent root-cause proof |
| 17 Mar 2026: PR #3046 merged | Established | Repository state transition |
| 17 Mar 2026: Issue #3065 opened | Established | Session/driver lifecycle defect report |
| Later MCP/Viewer development | Established elsewhere in the preserved record | Capability/architecture evidence |
| Investigated-device events involving `HHC9PV9WHL` | Not joined to these repository events | Missing runtime/device correlation |

## 4. Required evidentiary joins

The unresolved security question is whether a **specific authenticated execution actually targeted `HHC9PV9WHL`**.

```
authenticated principal
  +
agent/session ID
  +
MCP invocation
  +
Maestro run/session ID
  +
host ID
  +
device ID / UDID
  +
HHC9PV9WHL
  +
UTC timestamp
  +
device-side execution artifact
  +
protected-resource event
```

Preferred corroboration includes signed XCTest runner evidence, `.xctestrun`/`.xcresult`, `devicectl` output, Maestro CLI/session records, runner lifecycle records, and matching device-native telemetry.

## 5. Controlling disposition

**Established:** Maestro capability, PR #3046 provenance, device enumeration implementation, `Pixel6MobileDev` appearance, attributed local-environment statement, Issue #3065 session-lifecycle defect report, and AWS CDK project-init capability.

**Not established:** `Pixel6MobileDev = HHC9PV9WHL`, Maestro/MCP execution against `HHC9PV9WHL`, a specific actor's control of that device, unauthorized authorization state, protected-resource access, or causation.

> **Capability and repository provenance are established. Device-specific execution, attribution, authorization status, and causation remain unestablished without a correlated runtime chain.**

## 6. Source register

- Maestro PR #3046: https://github.com/mobile-dev-inc/Maestro/pull/3046
- Maestro Issue #3065: https://github.com/mobile-dev-inc/Maestro/issues/3065
- Maestro repository/current product documentation: https://github.com/mobile-dev-inc/Maestro
- AWS CDK TypeScript project guide: https://docs.aws.amazon.com/cdk/v2/guide/work-with-cdk-typescript.html
- AWS CDK `cdk init` reference: https://docs.aws.amazon.com/cdk/v2/guide/ref-cli-cmd-init.html
- AWS CDK CI/CD pipeline guide: https://docs.aws.amazon.com/cdk/v2/guide/cdk-pipeline.html

**Evidence boundary:** This report does not convert a repository artifact, reviewer statement, implementation capability, CDK project initializer, or issue report into proof of an incident. Provider-native runtime records remain the required bridge.

dl909-ai:
## Current-state verification addendum

A current public Maestro repository check adds an important platform-status qualifier to the historical capability record: the current README says Maestro MCP can control a live emulator, simulator, or physical Android device, while **physical iOS devices are not yet supported** in the current public product statement.

That does not erase historical/internal/development evidence elsewhere in the preserved corpus for physical-iOS implementation paths. It means the evidence must distinguish:

1. historical/development physical-iOS implementation evidence;
2. current public product support;
3. actual execution against `HHC9PV9WHL`.

Only (1) is supported by the relevant historical source material. (2) is currently constrained by the public README. (3) remains unestablished without the runtime/device join.

Current source: https://github.com/mobile-dev-inc/Maestro

This distinction is now part of the controlling evidence boundary for the investigation.

PR #3050, opened 2026-03-12
feat: make all options for resolve to be passed explisitly + added os…

proksh:
# Changes

1. Made `DeviceCatalog.resolve` to take in all value rather than using null in case value is not passed. Reason: By keeping them option, its easy to leave the option out and later realised we were not passing anything hence always default was getting picked.

2. Added `osVersion` in deviceSpec, that is being used in backend and worker. Rather than having separate logic there, it will just create `osVersion` along with config.

### DIFF
--- a/maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
+++ b/maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
@@ -18,6 +18,7 @@ sealed class DeviceSpec {
     abstract val platform: Platform
     abstract val model: String
     abstract val os: String
+    abstract val osVersion: Int
     abstract val deviceName: String
     abstract val locale: DeviceLocale
 
@@ -32,6 +33,7 @@ sealed class DeviceSpec {
     ) : DeviceSpec() {
         override val platform = Platform.ANDROID
         override val deviceName = "Maestro_ANDROID_${model}_${os}"
+        override val osVersion: Int = os.removePrefix("android-").toIntOrNull() ?: 0
         val tag = "google_apis"
         val emulatorImage = "system-images;$os;$tag;${cpuArchitecture.value}"
     }
@@ -45,6 +47,7 @@ sealed class DeviceSpec {
     ) : DeviceSpec() {
         override val platform = Platform.IOS
         override val deviceName = "Maestro_IOS_${model}_${os}"
+        override val osVersion: Int = os.removePrefix("iOS-").substringBefore("-").toIntOrNull() ?: 0
     }
 
     data class Web(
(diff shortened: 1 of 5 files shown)


maestro-client/src/main/java/maestro/device/DeviceCatalog.kt
```diff
@@ -65,15 +69,20 @@ object DeviceCatalog {
...
-        disableAnimations: Boolean = false,
-        snapshotKeyHonorModalViews: Boolean = false,
-        systemArchitecture: CPU_ARCHITECTURE = CPU_ARCHITECTURE.ARM64,
+        model: String?,
+        os: String?,
+        locale: String?,
+        orientation: DeviceOrientation?,
+        disableAnimations: Boolean?,
+        snapshotKeyHonorModalViews: Boolean?,
+        systemArchitecture: CPU_ARCHITECTURE?,
```
amanjeetsingh150, line 78: Not sure why we changed all of these to nullable? If we want to be explicit then we can actually remove defaults from here.

Let consumer define the values always

amanjeetsingh150: For example to me it feels weird for maestro core to say that I'll resolve a device specification with null property on disable animation, how would one interpret it 🤔 ?

Have false and true (boolean type) make it more explicit for consumer (cli/studio/cloud) to tell what they want.

proksh: If we do that, we will again be distributing the default handling to multiple places.

For example when I "start-device", if I pass in `null` value to `disableAnimation`, resolve will handle the default case. Similarly on cloud and studio side.

But if we make it explicit for consumer to send it, then we will need to handle the default case on cli, studio, cloud separately, and that will again diverge the default definition and distribute them in 3 places.

amanjeetsingh150: > If we do that, we will again be distributing the default handling to multiple places.

Not sure I understand? 

Maestro core shouldn't have any defaults (neither nullable nor the earlier ones). The consumers decides what they want, if there is default they want to decide it would be in respective cloud layer and cli module and not in core.

I'm suggesting consumers being explicit rather than core having any defaults here.

proksh: Discussed on call and both agreed this is a good solution for handling all the default and platform specific cases.

One problem is the design of resolve is such that, the options provided will keep on expanding and would be nullable. This can turn into a giant list. Adding a separate linear ticket to come back to this.

proksh: https://linear.app/mobile-dev/issue/MA-3993/problem-with-devicecatalogresolve

proksh: Will merge this as this is a blocker for cloud PR

PR #3053, opened 2026-03-12
Keyboard Dismissal on iOS Landscape

handstandsam:
## Proposed changes

copilot:summary

## Testing

Used this for our team's internal tests and it's working great!

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -360,23 +360,41 @@ class IOSDriver(
             val width = deviceInfo.widthGrid
             val height = deviceInfo.heightGrid
 
+            val isLandscape = width > height
+
             dismissKeyboardIntroduction(heightPoints = deviceInfo.heightGrid)
 
             if (isKeyboardHidden()) return@measured
 
-            swipe(
-                start = Point(0.5.asPercentOf(width), 0.5.asPercentOf(height)),
-                end = Point(0.5.asPercentOf(width), 0.47.asPercentOf(height)),
-                durationMs = 50,
-            )
+            if (isLandscape) {
+                swipe(
+                  start = Point(0.5.asPercentOf(width), 0.3.asPercentOf(height)),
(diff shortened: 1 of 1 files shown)


maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -360,23 +360,41 @@ class IOSDriver(
...
-                start = Point(0.5.asPercentOf(width), 0.5.asPercentOf(height)),
-                end = Point(0.5.asPercentOf(width), 0.47.asPercentOf(height)),
-                durationMs = 50,
-            )
+            if (isLandscape) {
+                swipe(
+                  start = Point(0.5.asPercentOf(width), 0.3.asPercentOf(height)),
+                  end = Point(0.5.asPercentOf(width), 0.33.asPercentOf(height)),
+                  durationMs = 50,
+                )
```
handstandsam, line 374: Our code works with JUST this one, but I see the early exit below, so I left the other.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -360,23 +360,41 @@ class IOSDriver(
...
-                start = Point(0.5.asPercentOf(width), 0.5.asPercentOf(height)),
-                end = Point(0.47.asPercentOf(width), 0.5.asPercentOf(height)),
-                durationMs = 50,
-            )
+            if (isLandscape) {
+                swipe(
+                  start = Point(0.5.asPercentOf(width), 0.3.asPercentOf(height)),
+                  end = Point(0.5.asPercentOf(width), 0.25.asPercentOf(height)),
+                  durationMs = 50,
+                )
```
handstandsam, line 390: Our code never hit this, but I followed the same sort of logic, but this one has a little bit more of a swipe of `0.05` and the one above has `0.03`.  Anyways, the first one works for us, and this is a bit arbitray.

I think the point of this PR is to solve for everyone, but I just want to be clear that we haven't hit this code block or tested it yet.

PR #2824, opened 2025-11-10
Add configurable custom identifier support for web testing

ff-vivek:
# Add Configurable Selector Aliases for Web Testing

## Summary

This PR introduces a flexible, configuration-driven selector alias system for web element selection in Maestro. Instead of hardcoding specific HTML attributes, users can now configure any HTML attribute to YAML selector key mappings in their workspace configuration.

## Motivation

Web frameworks (Flutter Web, React, Vue, etc.) use different HTML attributes for element identification. Previously, adding support for each framework required code changes. This PR makes Maestro extensible without any code modifications.

### 1. Configuration Model (`WorkspaceConfig.kt`)

Added `WebConfiguration` with `selectorAliases` to allow users to define HTML attribute → YAML key mappings:

```kotlin
data class WebConfiguration(
    val selectorAliases: Map<String, String> = emptyMap()
)
```

### 2. Workspace Configuration

Users can now configure selector aliases in their workspace `config.yaml`:

```yaml
platform:
  web:
    selectorAliases:
      flt-semantics-identifier: flutterId      # Flutter Web
      data-testid: testId                              # Testing frameworks
      aria-label: ariaLabelId                       # Accessibility
```

### 3. Flow Usage
Once configured, use the aliases directly in flow files:

```yaml
- tapOn:
    flutterId: submit-button
```

### 4. Dynamic Attribute Extraction (maestro-web.js)
- Extracts ONLY user-configured HTML attributes (no assumptions)
- Stores attributes with both HTML attribute name and YAML alias for flexibility
- No hardcoded prefixes like data-* or flt-*
- Configuration-driven: only processes what users explicitly specify

### 5. Driver Integration
The selectorAliases configuration is passed through the driver chain:
- MaestroSessionManager reads workspace config
- Maestro.pickWebDevice() receives selector aliases
- WebDriver/CdpWebDriver inject aliases into the browser JavaScript context
- maestro-web.js uses aliases to extract configured attributes

## Example Use Cases

###Flutter Web
```yaml

# config.yaml
platform:
  web:
    selectorAliases:
      flt-semantics-identifier: flutterId

# flow.yaml
- tapOn:
    flutterId: myWidget
```
### Other Frameworks

```yaml
# config.yaml
platform:
  web:
    selectorAliases:
      custom-id: customId
      app-component: componentId

# flow.yaml
- tapOn:
    componentId: header-nav
```

Benefits
1. Framework Agnostic: No bias toward any specific framework
[...]

PR #3058, opened 2026-03-16
fix: align web aria-label handling with native accessibilityText behavior

### LINKED ISSUE #2914: Text selector doesn't match aria-label

### Steps to reproduce

Given an HTML element of:
```
<button aria-label="Add to favourites">
  <!-- some svg here -->
</button>
```

This test doesn't do what's expected:

```
- tapOn:
    text: Add to favourites
```

### Actual results

No element found.

Need to switch to `id` rather than `text` selector to match aria-label.

### Expected results

Element tapped :)

### About app

Unknown

### About environment

Unknown

### Logs

n/a

### Maestro version

2.10.0

### Anything else?

This is an extrapolation of a report we had via X.

### LINKED ISSUE #3049: Targetting a button via data-testid fails on web if the button also has aria-label set

### Steps to reproduce

This issue only occurs when testing against Web.

Given a button:

```
<button aria-label="Some label" data-testid="some-test-id">
  <svg .../>
</button>
```

This test can't find the button:
```
- tapOn:
    id: some-test-id
```

### Actual results

Element cannot be found by maestro

Removing `aria-label` from the button results in maestro being able to find the element

### Expected results

Maestro will click the button which has `data-testid` of `some-test-id`

### About app

React native for web via expo

### About environment

Both locally (MacOS) and deployed to a cloud hosting service

### Logs

<details>
<summary>Logs</summary>

```
```

</details>

### Maestro version

2.3.0

### Anything else?

I suspect it's related to how maestro web determine's the `resource-id`, as it appears to favours `ariaLabel` over `data-testid`.

https://github.com/mobile-dev-inc/Maestro/blob/46b646d7d3bfce990dc7b40a0c9eb3b04bad1f06/maestro-client/src/main/resources/maestro-web.js#L98

I was also unable to use `test: Aria label value` as the target due to this issue: https://github.com/mobile-dev-inc/Maestro/issues/2914

cruzdanilo:
## Proposed changes

on web, `aria-label` was never stored as `accessibilityText`, so text selectors couldn't find elements by their aria-label. additionally, `aria-label` was incorrectly included in the `resource-id` priority chain ahead of `data-testid`.

this pr:
- maps `aria-label` to `accessibilityText` in `maestro-web.js`, mirroring how android uses `content-desc` and ios uses `AXElement.label`
- moves `aria-label` to the end of the `resource-id` fallback chain (last resort), so it no longer shadows `data-testid` but existing `tapOn: id: "some-aria-label"` tests keep working unless the element also has a higher-priority identifier
- parses `accessibilityText` in both `WebDriver.kt` and `CdpWebDriver.kt`

## Testing

added 45 unit tests covering:
- `parseDomAsTreeNodes` for both `WebDriver` and `CdpWebDriver` (all attribute branches, null guards, recursive children)
- `Filters.textMatches` matching on `accessibilityText`
- `Filters.idMatches` not matching on `accessibilityText`

## Issues fixed

fixes #2914
fixes #3049

### DIFF
--- a/maestro-client/src/main/resources/maestro-web.js
+++ b/maestro-client/src/main/resources/maestro-web.js
@@ -107,14 +107,18 @@
           bounds: getNodeBounds(node, iframeOffsetX, iframeOffsetY),
       }
 
+      if (node.ariaLabel) {
+          attributes['accessibilityText'] = node.ariaLabel
+      }
+
       // If this is an <option> element, we only want to include it if the parent <select> element is focused.
       if (node.tagName.toLowerCase() === 'option' && !node.parentElement.matches(':focus-within')) {
         return null;
       }
 
-      if (!!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
+      if (!!node.id || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid'] || !!node.ariaLabel) {
         const title = typeof node.title === 'string' ? node.title : null
-        attributes['resource-id'] = node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
+        attributes['resource-id'] = node.id || node.name || title || node.htmlFor || node.attributes['data-testid']?.value || node.ariaLabel
       }
 
       if (node.tagName.toLowerCase() === 'body') {
--- a/maestro-client/src/test/java/maestro/FiltersTest.kt
+++ b/maestro-client/src/test/java/maestro/FiltersTest.kt
@@ -41,6 +41,118 @@ class FiltersTest {
         assertThat(result).isEmpty()
     }
 
+    @Test
+    fun `textMatches finds node by accessibilityText`() {
+        val button = TreeNode(attributes = mutableMapOf(
+            "text" to "",
+            "accessibilityText" to "Add to favourites",
+        ))
+        val nodes = listOf(button)
+
+        val result = Filters.textMatches(Regex("Add to favourites"))(nodes)
+
+        assertThat(result).containsExactly(button)
+    }
+
+    @Test
+    fun `textMatches finds node by text even when accessibilityText differs`() {
+        val button = TreeNode(attributes = mutableMapOf(
+            "text" to "Click me",
+            "accessibilityText" to "Submit button",
+        ))
+
+        val result = Filters.textMatches(Regex("Click me"))(listOf(button))
+
+        assertThat(result).containsExactly(button)
+    }
+
+    @Test
+    fun `textMatches returns union of text and accessibilityText matches`() {
+        val byText = TreeNode(attributes = mutableMapOf("text" to "Search", "bounds" to "0,0,1,1"))
+        val byLabel = TreeNode(attributes = mutableMapOf(
+            "text" to "",
+            "accessibilityText" to "Search",
+            "bounds" to "0,0,2,2",
+        ))
+        val noMatch = TreeNode(attributes = mutableMapOf("text" to "Cancel", "bounds" to "0,0,3,3"))
+
+        val result = Filters.textMatches(Regex("Search"))(listOf(byText, byLabel, noMatch))
+
+        assertThat(result).containsExactly(byText, byLabel)
+    }
+
+    @Test
+    fun `textMatches does not duplicate when text and accessibilityText both match`() {
+        val button = TreeNode(attributes = mutableMapOf(
+            "text" to "Save",
+            "accessibilityText" to "Save",
+        ))
+
+        val result = Filters.textMatches(Regex("Save"))(listOf(button))
+
+        assertThat(result).containsExactly(button)
+    }
+
+    @Test
+    fun `textMatches does not match accessibilityText absent`() {
+        val button = TreeNode(attributes = mutableMapOf("text" to ""))
+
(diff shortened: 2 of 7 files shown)


maestro-client/src/main/resources/maestro-web.js
```diff
@@ -88,14 +88,18 @@
...
       // If this is an <option> element, we only want to include it if the parent <select> element is focused.
       if (node.tagName.toLowerCase() === 'option' && !node.parentElement.matches(':focus-within')) {
         return null;
       }
 
-      if (!!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
+      if (!!node.id || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
         const title = typeof node.title === 'string' ? node.title : null
-        attributes['resource-id'] = node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
+        attributes['resource-id'] = node.id || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
```
Fishbowler: Would this be a breaking change for folks who already have working tests relying on this behaviour?

cruzdanilo: good catch @Fishbowler! updated the pr to keep `aria-label` as a last-resort fallback in the `resource-id` chain. existing `tapOn: id: "some-aria-label"` tests will keep working unless the element also has a higher-priority identifier like `data-testid` (which is the fix for #3049).

Fishbowler (commented):
Implementation seems reasonable.
Have you tested manually that this solves the problem? The unit tests can't cover the maestro-web.js conversion from DOM to hierarchy.

cruzdanilo:
@Fishbowler yes. i tested it manually with a real web flow on this branch.

i also added a small smoke test that covers the maestro-web.js path directly.

maestro-client/src/test/java/maestro/FiltersTest.kt
```diff
@@ -41,6 +41,118 @@ class FiltersTest {
...
+    fun `textMatches finds node by text even when accessibilityText differs`() {
+        val button = TreeNode(attributes = mutableMapOf(
+            "text" to "Click me",
+            "accessibilityText" to "Submit button",
+        ))
+
+        val result = Filters.textMatches(Regex("Click me"))(listOf(button))
+
+        assertThat(result).containsExactly(button)
+    }
```
Fishbowler, line 67: This test should have a companion "textMatches finds node by accessibilityText when text differs" ?

maestro-client/src/test/java/maestro/FiltersTest.kt
```diff
@@ -41,6 +41,118 @@ class FiltersTest {
...
+        val button = TreeNode(attributes = mutableMapOf(
+            "text" to "",
+            "resource-id" to "submit-btn",
+            "accessibilityText" to "Submit form",
+        ))
+
+        val result = Filters.idMatches(Regex("Submit form"))(listOf(button))
+
+        assertThat(result).isEmpty()
+    }
```
Fishbowler, line 154: I don't think these tests are necessary or well described.

Elsewhere in code, `aria-label` would end up on both the `resource-id` and the `accessibilityText` properties (unless other props like `data-testid`, etc).

But both of these tests are repeats of "idMatches matches on resource-id, not on text values"

maestro-client/src/test/java/maestro/drivers/WebDriverParseDomTest.kt
```diff
@@ -0,0 +1,224 @@
...
+
+        assertThat(result.attributes["text"]).isEqualTo("")
+    }
+
+    @Test
+    fun `does not include optional attributes when absent`() {
+        val result = driver.parseDomAsTreeNodes(node())
+
+        assertThat(result.attributes).containsExactly("text", "", "bounds", "0,0,100,100")
+    }
```
Fishbowler, line 64: Beyond the tests above, this is only testing that the node function obeys its defaults, I think?

maestro-client/src/main/resources/maestro-web.js
```diff
@@ -107,14 +107,18 @@
...
       // If this is an <option> element, we only want to include it if the parent <select> element is focused.
       if (node.tagName.toLowerCase() === 'option' && !node.parentElement.matches(':focus-within')) {
         return null;
       }
 
-      if (!!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
+      if (!!node.id || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid'] || !!node.ariaLabel) {
         const title = typeof node.title === 'string' ? node.title : null
-        attributes['resource-id'] = node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
+        attributes['resource-id'] = node.id || node.name || title || node.htmlFor || node.attributes['data-testid']?.value || node.ariaLabel
```
Fishbowler, line 121: We should add this change of preference to the PR title so that it's definitely included in the changelog

e.g. "Web: Add aria-label to accessibilityText, and prefer other properties for ID"

chrisslazzz:
@cruzdanilo Are you still working on this? I could also have a look.

It's almost impossible to write tests for flutter web without this or #2824

Issue #1924, opened 2024-08-19
Unable to interact with a sheet opened from a fullScreenCover


LoopingStudio:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

iOS, native SwiftUI project

Hello, I'm not able to interact with elements inside a sheet opened over a fullScreenCover, and I discovered that if the sheet size is above 93% of the screen, I can interact with them,

Here is 2 clips of me reproducing the bug:

<details>
<summary>first clip</summary>

https://github.com/user-attachments/assets/c35d29be-623b-4e7b-8dda-f42c7fdebdd5

</details>

<details>
<summary>second clip</summary>

https://github.com/user-attachments/assets/16c69a77-372a-4762-b22e-9877b2e9fd33

</details>

Clone [the project](https://github.com/LoopingStudio/MaestroSheetBug.git)
Open the project
Run the app on a simulator (I tried on iPhone 15 Pro, iOS 17.4)

On the app: 

- Open the full screen cover by tapping on the button
- Choose a sheet size using the slider
- Open the second sheet modal by tapping on the purple button

### Actual results

When the second sheet is below 90% of screen height, I am not able to interact with the sheet elements like buttons, text fields.
When the sheet height is above 93%, I can interact with them.
100% reproducible.

### Expected results

I want to interact with element on a sheet opened over a fullScreenCover, just like a normal sheet

### About app

This is a native SwiftUI app, as simple as it can be
https://github.com/LoopingStudio/MaestroSheetBug.git

### About environment

M3 Pro Macbook Pro, macOS Sonoma 14.6
Xcode 15.4

Run on iOS 17.4 simulator, iPhone 15 Pro

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

1.37.7 and 1.36

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

bartekpacia:
Hey @LoopingStudio, thank you for this detailed bug report!

I reproduced your problem. I also noticed that if I click on some UI element in that bottom sheet, the screenshot view in Maestro Studio refreshes – but it's still broken (i.e. element boundaries are not overlaid on the screenshot).

https://github.com/user-attachments/assets/6a0c6648-701b-4bd2-8f85-5830f0e3a8b4

CalvinButter:
[...]

Issue #1644, opened 2024-01-05
[1.34.1] - cannot get elements from popup window on iOS


marioracek:
Hi, I’m seeking help with this issue on iOS. I’m trying to select elements on the popup window but without any luck.

As you can see on the image below, if view controller is presented (present segue type - the yellow screen), then I’m not able to select any element of popup window (title, subtitle or button). But if view controller is pushed (the purple screen) then I’m able to access all popup elements (title, subtitle, button) etc. Popup window is added into view hierarchy as a subview of the app “window”.

I’m including basic project where it can be tried. Do you know or is there any workaround for this? I would like to use Maestro for iOS automation but this is holding me back as we have multiple popup windows like this in the app. Right now the only workaround is to user exact position of elements which is not convenient. Thank you

- Maestro version [1.34.1]
- Platform: [iOS 17.2]
- Framework: [Native iOS]
- Device model and OS version: [iPhone 15]
- Simulator
- Host [Mac M2]

[MaestroTest (1).zip](https://github.com/mobile-dev-inc/maestro/files/13840414/MaestroTest.1.zip)
![image-20231113-161958 (1)](https://github.com/mobile-dev-inc/maestro/assets/4375634/bdb55b61-c770-4dd4-92e7-750d5df7ed11)

PR #3061, opened 2026-03-16
fix: load workspace config.yaml when running a single flow file

andrewse02:
## Proposed changes

  When running `maestro test path/to/flow.yaml` (a single flow file), the workspace `config.yaml` in the flow's parent directory was not loaded. This meant
  platform configuration like `snapshotKeyHonorModalViews` was silently ignored, even though it worked correctly when running against a directory (`maestro test
  ui-tests/`).

  This also affected retry workflows that run individual flow files after an initial suite failure (e.g. `maestro test ui-tests/reader/copy_paste_verse.yaml`).

  The fix searches upward from the flow file's directory to find the nearest `config.yaml`, matching the existing behavior for directory-based runs.

  ## Testing

  - Ran `maestro test ui-tests/reader/load_scripture.yaml` — verified `snapshotKeyHonorModalViews` from `ui-tests/config.yaml` was loaded (previously `null`)
  - Ran `maestro test /tmp/standalone_flow.yaml` (no config in parent) — verified defaults preserved (`null`)
  - Ran full smoke suite via `run_ui_tests_with_retries.py` matching Bitrise `ui_test` workflow — retries correctly picked up workspace config
  - All existing `WorkspaceExecutionPlannerTest` unit tests pass

  ## Issues fixed

  Fixes https://github.com/mobile-dev-inc/maestro/issues/1644
  Fixes https://github.com/mobile-dev-inc/maestro/issues/1924

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
@@ -33,7 +33,9 @@ object WorkspaceExecutionPlanner {
             val workspaceConfig = if (config != null) {
                 YamlCommandReader.readWorkspaceConfig(config.absolute())
             } else {
-                WorkspaceConfig()
+                findConfigFileUpwards(input.first().absolute().parent)
+                    ?.let { YamlCommandReader.readWorkspaceConfig(it) }
+                    ?: WorkspaceConfig()
             }
             return ExecutionPlan(
                 flowsToRun = input.toList(),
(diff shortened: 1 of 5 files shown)


Fishbowler:
It's a dramatic departure from current behaviour. Hoe would we adapt this to not change behaviour for users expecting the current behaviour? That's a must, without leaving this to whither until the next major version bump.

Perhaps keeping to explicit configs? `maestro test --config my_config.yaml my_test.yaml`

andrewse02:
> It's a dramatic departure from current behaviour. Hoe would we adapt this to not change behaviour for users expecting the current behaviour? That's a must, without leaving this to whither until the next major version bump.
> 
> Perhaps keeping to explicit configs? `maestro test --config my_config.yaml my_test.yaml`

Good point. This could end up causing problems for users expecting this behavior. Originally I thought of this as a bug, but I believe it could serve as useful functionality for some projects. `--config` should suffice. I'll close this PR, thank you!

Fishbowler:
I've said this on an assumption that `maestro test --config my_config.yaml my_test.yaml` works. It does for workspaces, but was intended for alternative configs. Do check, and if it doesn't work, definitely raise that as a bug.

andrewse02:
It does work for our project!

PR #3063, opened 2026-03-17
Store hierarchy with screenshot when an assertion failure occurs

Fishbowler:
## Proposed changes

When failures occur, debugging can be difficult, especially if they occur with CI (like our own e2e tests). This change stores the hierarchy snapshot on failure, and stores it as a debug test artifact.

## Testing

Tested locally with demo_app:

```
maestro test ./.maestro/fail_fast.yaml
maestro test  --include-tags failing ./.maestro
```


Fishbowler:
@proksh is doing something a lot more thorough - this is obsolete

PR #3066, opened 2026-03-17
fix: Android driver not installed when another Maestro session is active

### LINKED ISSUE #3065: Android driver not installed when Maestro Studio is running (or any other session is active)
## Summary

When Maestro Studio is open (or any other Maestro session is active — even on a different platform), running `maestro test` against an Android device skips the driver installation entirely, resulting in a gRPC connection error.

## Steps to reproduce

1. Start Maestro Studio on any device (iOS, Android, Web)
2. In a separate terminal, run `maestro test` targeting an Android device
3. The test fails immediately with:
   ```
   Connection refused: 127.0.0.1:7001
   ```

## Root cause

`SessionStore.hasActiveSessions()` checks whether any session besides the current one is active, but it does **not** filter by platform. The session keys are stored as `${platform}_${sessionId}` (e.g. `IOS_abc123`, `ANDROID_xyz456`), but the filter was missing:

```kotlin
// Before (buggy)
fun hasActiveSessions(sessionId: String, platform: Platform): Boolean {
    synchronized(keyValueStore) {
        return activeSessions()
            .any { it != key(sessionId, platform) }  // matches ANY platform
    }
}
```

This causes `MaestroSessionManager.newSession()` to set `connectToExistingSession = true`, which sets `openDriver = false`, which means `driver.open()` is never called — so the Android driver APKs are never installed on the device and port 7001 is never forwarded.

A secondary issue: even if `openDriver` is corrected, `AndroidDriver.deviceInfo()` (called during JS engine initialization) lacks the same `if (!open) open()` guard that `launchApp()` already has, so it would fail before any auto-recovery could happen.

## Fix

1. **`SessionStore.kt`** — filter `activeSessions()` by platform prefix before checking for other active sessions:

```kotlin
fun hasActiveSessions(sessionId: String, platform: Platform): Boolean {
    synchronized(keyValueStore) {
        val platformPrefix = "${platform}_"
        return activeSessions()
            .filter { it.startsWith(platformPrefix) }
            .any { it != key(sessionId, platform) }
    }
}
```

2. **`AndroidDriver.kt`** — add the missing `if (!open) open()` guard to `deviceInfo()` (consistent with `launchApp()`):

```kotlin
override fun deviceInfo(): DeviceInfo {
    if (!open) open()
    return runDeviceCall("deviceInfo") { ... }
}
```

## Impact

Any user running `maestro test` on Android while Maestro Studio is open hits this bug. The driver is silently skipped and the test fails with a confusing connection error rather than a clear "driver not installed" message.

## Environment

- Maestro 2.3.0
- macOS
- Android emulator (API 35)

memocayar:
## Problem

Closes #3065

When Maestro Studio is open (or any other Maestro session is active — even on a different platform), running `maestro test` on Android silently skips driver installation and fails with:

```
Connection refused: 127.0.0.1:7001
```

## Root cause

`SessionStore.hasActiveSessions()` stores session keys as `${platform}_${sessionId}` but was checking for *any* active session without filtering by platform:

```kotlin
// Before — matches sessions from ANY platform
return activeSessions()
    .any { it != key(sessionId, platform) }
```

This causes `MaestroSessionManager` to set `connectToExistingSession = true` → `openDriver = false` → `driver.open()` never called → Android driver APKs never installed → port 7001 never forwarded.

A secondary issue: `AndroidDriver.deviceInfo()` (called first during JS engine init) lacked the `if (!open) open()` guard that `launchApp()` already has, so even a partial recovery path was missing.

## Changes

**`SessionStore.kt`** — filter by platform prefix before checking for other active sessions:

```kotlin
fun hasActiveSessions(sessionId: String, platform: Platform): Boolean {
    synchronized(keyValueStore) {
        val platformPrefix = "${platform}_"
        return activeSessions()
            .filter { it.startsWith(platformPrefix) }
            .any { it != key(sessionId, platform) }
    }
}
```

**`AndroidDriver.kt`** — add `if (!open) open()` guard to `deviceInfo()`, consistent with `launchApp()`:

```kotlin
override fun deviceInfo(): DeviceInfo {
    if (!open) open()
    return runDeviceCall("deviceInfo") { ... }
}
```

## Testing

Reproduced on Maestro 2.3.0 / macOS / Android emulator API 35 with Maestro Studio running in the background. After this fix, `maestro test` correctly installs the driver and runs the flow end-to-end.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/SessionStore.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/SessionStore.kt
@@ -61,7 +61,9 @@ object SessionStore {
         platform: Platform
     ): Boolean {
         synchronized(keyValueStore) {
+            val platformPrefix = "${platform}_"
             return activeSessions()
+                .filter { it.startsWith(platformPrefix) }
                 .any { it != key(sessionId, platform) }
         }
     }
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -210,6 +210,7 @@ class AndroidDriver(
     }
 
     override fun deviceInfo(): DeviceInfo {
+        if (!open) open()
         return runDeviceCall("deviceInfo") {
             val response = blockingStubWithTimeout.deviceInfo(deviceInfoRequest {})
 


Fishbowler:
Whilst there's lots of LLM-generated technical details here, it's thin on reasoning around the user action and user value.

>When Maestro Studio is open (or any other Maestro session is active — even on a different platform)

So there's a lot of combinations listed here, but the implementation doesn't seem to match what this is saying. Is that intentional?

Fishbowler:
@memocayar See above

Fishbowler:
Converted to draft whilst we're waiting for answers

seanadkinson:
@Fishbowler I'm seeing https://github.com/mobile-dev-inc/maestro/issues/3065 with the following use case:
* Self-hosted GHA runners, one for iOS, one for Android, hosted on the same machine
* This is so that I get concurrency controls and reporting separated by platform
* However, if Android tries to start running maestro tests while an iOS session is active, I see the bug mentioned in the issue (Connection refused: 127.0.0.1:7001)

Happy to provide any other details you need.  Thanks

Fishbowler:
@seanadkinson That's not a bug. There's no capability for concurrent running like you're describing quite yet. Try v2.6.0 and see if that improves things, but I suspect there are more open PRs to land before that'll work.

seanadkinson:
Okay, got it.  Current solution (still testing) is to change the user.home directory so that each runner gets its own sessions file:

```
export MAESTRO_OPTS="-Duser.home=$HOME/.maestro-homes/$RUNNER_NAME"
```

I'll update to 2.6.0 soon, though, thank you.

PR #3068, opened 2026-03-18
fix platform from string

proksh:
## Proposed changes

- `Platform.toString()` should never accept null value. It should always be a string
- `Platform.toString()` should never return a value other than Platform. If platform doesn't exist it should be thrown as exception


Fishbowler:
Change looks good, but CI is unhappy

PR #3070, opened 2026-03-19
Fix input text crashing on native passcode screen when providing numerical input

amanjeetsingh150:
## Proposed changes

Fix XCTest API access crashing from background thread. `waitUntilKeyboardIsPresented` was being called from a non-@MainActor context, causing XCTest API calls (XCUIApplication, keyboards.firstMatch.exists) to execute on the cooperative thread pool. XCTest requires main thread affinity for UI queries, calling from a background thread aborts the process.

Fixed by moving the shared wait logic into TextInputHelper which is @MainActor, giving compile-time enforcement that these calls always run on the main actor. Also removes the now-unused TimeoutHelper.

## Before

https://github.com/user-attachments/assets/e771b002-f188-4e2a-a5c3-74566372bb14

## After

https://github.com/user-attachments/assets/4e8b9640-6620-4f43-8afd-27bc7af067c0


jemersondox:
Are there any plans to merge this fix soon?

amanjeetsingh150:
Hey @jemersondox this should be deployed on cloud today, and pre-release of cli would be available for local by today or tomorrow, keep me honest here @Fishbowler .

Fishbowler:
100%

Fishbowler:
Done, and details shared on Slack some hours ago 👍

PR #3072, opened 2026-03-20
updated cloud flags

proksh:
Updated cloud flags.

- Depreciate `--android-api-level` & `--ios-version`
- `androidApiLevel` & `iOSVersion` wont be sent to the cloud api (cloud api still have these values to keep supporting the old users)
- Updated description for `--device-model` & `--device-os`

### DIFF
--- a/maestro-cli/src/test/kotlin/maestro/cli/cloud/CloudInteractorTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/cloud/CloudInteractorTest.kt
@@ -94,6 +94,7 @@ class CloudInteractorTest {
                 excludeTags = any(), disableNotifications = any(),
                 deviceLocale = any(), progressListener = any(),
                 projectId = any(), deviceModel = any(), deviceOs = any(),
+                androidApiLevel = any(), iOSVersion = any(),
             )
         } returns UploadResponse(
             orgId = "org_1",
@@ -174,6 +175,8 @@ class CloudInteractorTest {
             projectId = "proj_1",
             deviceModel = any(),
             deviceOs = any(),
+            androidApiLevel = any(),
+            iOSVersion = any(),
         ) }
     }
 
@@ -336,6 +339,7 @@ class CloudInteractorTest {
             excludeTags = any(), disableNotifications = any(),
             deviceLocale = eq("fr_FR"), progressListener = any(),
             projectId = any(), deviceModel = any(), deviceOs = any(),
+            androidApiLevel = any(), iOSVersion = any(),
         ) }
     }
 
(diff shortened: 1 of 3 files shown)


maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
```diff
@@ -154,10 +156,18 @@ class CloudCommand : Callable<Int> {
     @Option(order = 19, names = ["--device-locale"], description = ["Locale that will be set to a device, ISO-639-1 code and uppercase ISO-3166-1 code i.e. \"de_DE\" for Germany"])
     private var deviceLocale: String? = null
 
-    @Option(order = 20, names = ["--device-model"], description = ["Device model to run your flow against. Supported values include iPhone-11, etc. Only supported for iOS at the moment."])
+    @Option(order = 20, names = ["--device-model"], description = [
+      "Device model to run your flow against. " +
+              "iOS: iPhone-11, iPhone-11-Pro, etc. Run command: maestro list-cloud-devices" +
+              "Android: pixel_6, etc. Run command: maestro list-cloud-devices"
+    ])
```
amanjeetsingh150, line 164: Few questions:

- Is this deprecation planned for this next release? Before we fully launch iOS 26.
- What happens to other consumers like bitbucket, github actions, etc? We do want to unify there as well. 
- Also just to make sure https://github.com/mobile-dev-inc/action-maestro-cloud/blob/34906065ba3e85fd57ed533b178187eefb042aed/action.yml#L34 this would keep working post release as well right.

Fishbowler:
I thought we were deprecating, not removing?
We warn now, we alter functionality later?

maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
proksh: > - Is this deprecation planned for this next release? Before we fully launch iOS 26.

Yeah, I don't think it needs to be connected to release

> - What happens to other consumers like bitbucket, github actions, etc? We do want to unify there as well.

Yeah, I will be doing that next week

> - Also just to make sure https://github.com/mobile-dev-inc/action-maestro-cloud/blob/34906065ba3e85fd57ed533b178187eefb042aed/action.yml#L34 this would keep working post release as well right.

Yes, currently the flag is depreciated and not removed. So we will need to announce it with a timeline. Till the timeline, flag will keep working.

proksh:
> I thought we were deprecating, not removing? We warn now, we alter functionality later?

Yes, the flags `--android-api-level` & `--ios-version` should still work.

PR #3074, opened 2026-03-20
fix: removed deviceCatalog and updated deviceSpec

proksh:
## Changes

- Removed `DeviceCatalog` as it was unnecessary layer that was causing confusion about what it does
- Updated `DeviceSpec` to have `DeviceSpec.fromRequest` that will get a valid device Spec
- Added serialiser/deserialiser so that it can be used anywhere `DeviceSpec` is being imported. So that we dont have to define this in multiple places.
- Added test


maestro-client/src/main/java/maestro/device/DeviceSpec.kt
```diff
@@ -62,12 +62,45 @@ sealed class DeviceSpec {
...
         override val deviceName = "Maestro_WEB_${model}_${osVersion}"
     }
+
+    companion object {
+        /**
+         * Creates a fully resolved DeviceSpec from a DeviceRequest, filling in platform-aware defaults.
+         * The returned spec is not environment-validated.
+         * Environment-specific validation for model & os can happen via SupportedDevices.validate().
+         */
+        fun fromRequest(request: DeviceSpecRequest): DeviceSpec {
```
amanjeetsingh150, line 80: Can't this take the device request directly? From what I remember device request is what maestro-core understands as requested device right.

Why do we have another device spec request in between?

proksh: Its just renaming 
`DeviceRequest` -> `DeviceSpecRequest`
so its clear its device spec request.

PR #3075, opened 2026-03-20
fix: improved getting spec from devicesec

proksh:
1. When we are using deviceSpec, for getting each value we have to right cases. For example for getting disableAnimation

```
disableAnimation = when (deviceSpec) {
    is DeviceSpec.Android -> deviceSpec.disableAnimation
    is DeviceSpec.Ios -> deviceSpec.disableAnimation
    is DeviceSpec.Web -> null
}
```

But after this change we can simply do
```
disableAnimation = deviceSpec.disableAnimation
```

2. Added function withCopy which will make overriding deviceSpec easier. Examples will be in this PR - https://github.com/mobile-dev-inc/copilot/pull/2266

### DIFF
--- a/maestro-client/src/main/java/maestro/device/serialization/DeviceLocaleSerializer.kt
+++ b/maestro-client/src/main/java/maestro/device/serialization/DeviceLocaleSerializer.kt
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.databind.deser.std.StdDeserializer
 import com.fasterxml.jackson.databind.ser.std.StdSerializer
 import maestro.device.Platform
 import maestro.device.locale.DeviceLocale
+import maestro.device.locale.LocaleValidationException
 
 class DeviceLocaleSerializer : StdSerializer<DeviceLocale>(DeviceLocale::class.java) {
   override fun serialize(value: DeviceLocale, gen: JsonGenerator, provider: SerializerProvider) {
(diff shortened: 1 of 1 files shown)


maestro-client/src/main/java/maestro/device/DeviceSpec.kt
```diff
@@ -32,43 +34,52 @@ sealed class DeviceSpec {
     abstract val osVersion: Int
     abstract val deviceName: String
     abstract val locale: DeviceLocale
+    open val orientation: DeviceOrientation get() = DeviceOrientation.PORTRAIT
+    open val disableAnimations: Boolean get() = false
+    open val snapshotKeyHonorModalViews: Boolean get() = false
```
amanjeetsingh150, line 39: This feels like weird design, the design was split to android, ios and web to intentionally show what configurable where. 

This would start contradicting that if we just have flat list of every property here. Why not then just have one device spec?

If this done to stop repeatability of the branching why not just have an extension property?

```
val DeviceSpec.disableAnimations: Boolean?                                                                                                                                                               
      get() = when (this) {                                                                                                                                                                                
          is DeviceSpec.Android -> disableAnimations                                                                                                                                                       
          is DeviceSpec.Ios -> disableAnimations                                                                                                                                                           
          is DeviceSpec.Web -> null                                                                                                                                                                        
      } 
```

proksh: Yes this make sense.

proksh:
Will move this to cloud

PR #3077, opened 2026-03-20
Add deprecation warning to legacy Maestro Studio

Fishbowler:
## Proposed changes

We've had Maestro Studio Desktop in Beta for ~9 months now, and it meets most needs for most folks.
This adds a deprecation warning to the old studio so that folks know that it isn't getting new feature work, and could be removed in a future release.
It also adds a way to hide the banner (`MAESTRO_HIDE_DEPRECATION_BANNER=1 maestro studio`)

image

## Testing

```
maestro studio
MAESTRO_HIDE_DEPRECATION_BANNER=1 maestro studio
```

## Issues fixed

n/a

### DIFF
--- a/maestro-studio/web/src/components/common/Header.tsx
+++ b/maestro-studio/web/src/components/common/Header.tsx
@@ -21,12 +21,30 @@ const HeaderBanner = () => {
   );
 };
 
+const DeprecationBanner = () => (
+  <div className="py-3 px-7 text-sm text-center font-semibold bg-amber-100 dark:bg-amber-900">
+    <span>
+      This feature has been deprecated and will be removed in the near future. Check out{" "}
+      <a
+        href="https://maestro.dev/?utm_source=old_studio&utm_campaign=download_studio#maestro-studio"
+        target="_blank"
+        rel="noreferrer"
+        className="underline"
+      >
+        Maestro Studio Desktop
+      </a>{" "}
+      for similar, fully supported functionality.
+    </span>
+  </div>
+);
+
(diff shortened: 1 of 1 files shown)


maestro-studio/web/src/components/common/Header.tsx
```diff
@@ -21,12 +21,34 @@ const HeaderBanner = () => {
...
 
+const DeprecationBanner = () => {
+  const { data } = API.useStudioFlags();
+  if (!data?.showDeprecationBanner) return null;
+  return (
+    <div className="py-3 px-7 text-sm text-center font-semibold bg-amber-100 dark:bg-amber-900">
+      <span>
+        This feature is deprecated. Download{" "}
+        <a
+          href="https://maestro.dev/?utm_source=cli&utm_campaign=download_studio#maestro-studio"
```
Fishbowler: This is the same URL used in the CLI:

```
╭────────────────────────────────────────────────────────────────────────────────╮
│                                                                                │
│          Download the new and improved Maestro Studio app today!               │
│                                                                                │
│ https://maestro.dev?utm_source=cli&utm_campaign=download_studio#maestro-studio │
│                                                                                │
╰────────────────────────────────────────────────────────────────────────────────╯
```

Leland-Takamine:
Don't think we need the env var logic here

maestro-studio/web/src/components/common/Header.tsx
Fishbowler: Got an updated link with different attribution

PR #3067, opened 2026-03-18
Read hierarchy from iFrames

Fishbowler:
## Proposed changes

Access iFrame hierarchy content from Maestro Web. Before, we were just seeing "Your browser doesn't support iframes".

The impl reads child iframe content, and adjusts hierarchy bounds for its perceived "reset" position within the child document.

## Testing

Variants on this:

```
url: https://iframe-example-ten.vercel.app/
# https://github.com/mobile-dev-inc/maestro/issues/3009
---
- launchApp
- assertVisible: .*lives inside an iframe.*
```

Tested in CLI.

Also tested using a custom build of Maestro Studio to see that iFrames could have elements picked and tapped, etc.

e2e test: https://github.com/mobile-dev-inc/demo_app/pull/23

## Issues fixed

Fixes #3009

PR #3079, opened 2026-03-23
Add cross-domain iframe support

Fishbowler:
## Proposed changes

Support for iframes was added in #3067

This change adds support for Maestro reading hierarchy from all iframes, even if they're cross-domain.

This change:
* Evaluates whether iframes are cross-domain and so marks them with a custom attribute for later processing
* Uses the chrome debug protocol to switch to iframes and evaluate their content, adding it to the hierarchy

Because it's switching frames, it has to assume that not all documents have the debug websocket session.

## Testing

```
maestro test flow.yaml
```
where the flow contains an affected site:
```
url: https://help.sumup.com/
---
- launchApp
- tapOn: "Accept all cookies"
- tapOn: Help
- assertVisible: Support Chat
```

## Issues fixed

n/a

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/WebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/WebDriver.kt
@@ -20,11 +20,13 @@ import maestro.web.selenium.ChromeSeleniumFactory
 import maestro.web.selenium.SeleniumFactory
 import okio.Sink
 import okio.buffer
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import org.openqa.selenium.By
 import org.openqa.selenium.JavascriptExecutor
 import org.openqa.selenium.Keys
 import org.openqa.selenium.OutputType
 import org.openqa.selenium.TakesScreenshot
+import org.openqa.selenium.WebElement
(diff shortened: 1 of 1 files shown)


alisumup:
Hey @Fishbowler , thanks for the cross-domain iframe support, hierarchy reading works great.

While testing against `help.sumup.com` I found that **inputText, eraseText, and pressKey silently fail** when the focused element is inside a cross-origin iframe. The issue is that document.activeElement in the main frame returns the <iframe> element itself rather than the actual input inside it, so sendKeys targets the wrong element.

I've added a small fix in this commit: a **withActiveElement helper** that checks if the active element is an <iframe>, switches Selenium into that frame to find the real focused element, performs the action, then switches back. The non-iframe path is completely unchanged — existing workflows aren't affected.

Happy to adjust anything if needed.

Here is the [PR](https://github.com/mobile-dev-inc/Maestro/pull/3098)

test:
```yaml
url: https://help.sumup.com/
---
- launchApp
- tapOn: "Accept all cookies"
- tapOn: Help
- assertVisible: Support Chat
- tapOn: "Write a message"
- inputText: "Hello, how are you?"
- tapOn: "Send"

```

PR #3094, opened 2026-03-30
feat: shared app/workspace validation + CLI platform inference

proksh:
### Summary

Lays the foundation for end-to-end DeviceSpec by moving app and workspace validation into shared modules, and wiring the CLI to infer platform from the app binary rather than CLI flags.

No behaviour change to the upload API — the CLI still sends the same payload to the backend. This is prep work for Phase 2 where DeviceSpec is constructed on the CLI and sent to the cloud.

 ---

### What changed

maestro-cli — `AppMetadataAnalyzer`
  - New utility class that reads app binary metadata (APK manifest, IPA Info.plist, web JSON config) to infer platform and extract app identifiers
  - Adds apk-parser and dd-plist as dependencies
  - IosAppMetadata includes platformName (DTPlatformName from Info.plist) to distinguish simulator vs device builds

  maestro-client — `AppValidator`
  - New shared validator for app binaries; callable from both CLI and backend
  - Android: rejects APKs with native libraries that lack arm64-v8a (pure Java APKs with no native libs are accepted)
  - iOS: requires platformName == "iphonesimulator" (rejects device builds) and non-null minimumOSVersion

  maestro-orchestra — `WorkspaceValidator`
  - Extracted from the backend's WorkspaceEvaluator into a shared module
  - Validates a workspace zip against an appId: flow parsing, appId matching, name conflict detection, empty workspace, and invalid zip
  - Returns a typed Result<WorkspaceValidationResult, WorkspaceValidationError> for clean error handling

  maestro-cli — `CloudInteractor`
  - Platform is now inferred from the app binary (AppMetadataAnalyzer + AppValidator) and stored as inferredPlatform: Platform? — authoritative source for Phase 2 DeviceSpec construction
  - WorkspaceValidator is called before client.upload() — workspace errors (wrong appId, name conflicts, syntax errors) surface as clear CLI errors before any network call
  - Removed triggeredPlatform heuristic (was based on --android-api-level / --ios-version flags which will be deprecated); analytics events now use inferredPlatform or "unknown"

### Tests
- AppMetadataAnalyzerTest
- AppValidatorTest
- WorkspaceValidatorTest

--- 
Also fixes maestro-cli's test task which was silently skipping all tests (missing useJUnitPlatform()).

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -146,6 +146,11 @@ class CloudInteractor(
                     }
                 },
                 webManifestProvider = webManifestProvider,
+                iosMinOSVersionProvider = { file ->
+                    val metadata = maestro.cli.util.AppMetadataAnalyzer.getIosAppMetadata(file) ?: return@AppValidator null
+                    val major = metadata.minimumOSVersion.substringBefore(".").toIntOrNull() ?: return@AppValidator null
+                    AppValidator.IosMinOSVersion(major = major, full = metadata.minimumOSVersion)
+                },
             )
             val resolvedAppValidation = try {
                 appValidator.validate(appFile = appFileToSend, appBinaryId = appBinaryId)
@@ -171,6 +176,26 @@ class CloudInteractor(
                     os = deviceOs,
                     locale = deviceLocale,
                 ))
+            }
+
+            // Fetch supported devices and validate device spec
+            val supportedDevices = try {
+                client.listCloudDevices()
+            } catch (e: ApiClient.ApiException) {
+                throw CliError("Failed to fetch supported devices. Status code: ${e.statusCode}")
+            }
+
+            val validatedDeviceSpec = try {
+                DeviceSpecValidator.validate(deviceSpec, supportedDevices)
+            } catch (e: DeviceSpecValidator.InvalidDeviceConfiguration) {
+                throw CliError(e.message ?: "Invalid device configuration")
+            }
+
+            // Validate app-device compatibility
+            try {
+                appValidator.validateDeviceCompatibility(appFileToSend, validatedDeviceSpec, supportedDevices)
+            } catch (e: AppValidationException) {
+                throw CliError(e.message ?: "App-device compatibility check failed")
             }
 
             // Validate workspace against appId before uploading to catch errors early
--- a/maestro-cli/src/test/kotlin/maestro/cli/cloud/CloudInteractorTest.kt
+++ b/maestro-cli/src/test/kotlin/maestro/cli/cloud/CloudInteractorTest.kt
@@ -43,6 +43,14 @@ class CloudInteractorTest {
         every { mockApiClient.getProjects(any()) } returns listOf(
             maestro.cli.api.ProjectResponse(id = "proj_1", name = "Test Project")
         )
+        every { mockApiClient.listCloudDevices() } returns mapOf(
+            "android" to mapOf("pixel_6" to listOf("android-34", "android-33", "android-31", "android-30", "android-29")),
+            "ios" to mapOf(
+                "iPhone-11" to listOf("iOS-16-2", "iOS-17-5", "iOS-18-2"),
+                "iPhone-14" to listOf("iOS-16-2", "iOS-17-5", "iOS-18-2"),
+            ),
+            "web" to mapOf("chromium" to listOf("default")),
+        )
 
         // Capture console output
         originalOut = System.out
@@ -444,6 +452,59 @@ class CloudInteractorTest {
             progressListener = any(), projectId = any(),
             deviceModel = any(), deviceOs = any(),
         ) }
+    }
+
+    // ---- 14. Device model not supported ----
+
+    @Test
+    fun `upload throws CliError when device model is not supported`() {
+        val error = assertThrows<CliError> {
+            createCloudInteractor().upload(
+                flowFile = iosFlowFile(),
+                appFile = iosApp(),
+                async = true,
+                projectId = "proj_1",
+                deviceModel = "galaxy_s21",
+            )
+        }
+
+        assertThat(error.message).contains("not supported")
+        assertThat(error.message).contains("galaxy_s21")
+    }
+
+    // ---- 15. OS version not supported for device ----
+
+    @Test
+    fun `upload throws CliError when OS version is not supported for device`() {
+        val error = assertThrows<CliError> {
+            createCloudInteractor().upload(
+                flowFile = iosFlowFile(),
+                appFile = iosApp(),
+                async = true,
+                projectId = "proj_1",
+                deviceOs = "iOS-15-0",
+            )
+        }
+
+        assertThat(error.message).contains("not supported")
+    }
+
+    // ---- 16. Valid device config and compatible app succeeds ----
+
+    @Test
+    fun `upload with valid device config and compatible app succeeds`() {
+        stubUploadResponse(platform = "IOS")
+
+        val result = createCloudInteractor().upload(
+            flowFile = iosFlowFile(),
+            appFile = iosApp(),
+            async = true,
+            projectId = "proj_1",
+            deviceModel = "iPhone-14",
+            deviceOs = "iOS-18-2",
+        )
+
+        assertThat(result).isEqualTo(0)
     }
 
     // ---- waitForCompletion tests (existing) ----


maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +130,56 @@ class CloudInteractor(
             // Binary id or Binary file
             val appFileToSend = getAppFile(appFile, appBinaryId, tmpDir, flowFile)
 
-            // Track cloud upload start after we have the response with actual platform
+            // Infer platform and validate app binary
+            val appValidation: AppValidationResult? = appFileToSend?.let { f ->
+                when {
+                    AppMetadataAnalyzer.getWebMetadata(f) != null ->
+                        AppValidator.fromWebMetadata(AppMetadataAnalyzer.getWebMetadata(f)!!.url)
```
amanjeetsingh150: Can we handle the null case here instead of using double bangs?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +130,56 @@ class CloudInteractor(
...
+
+                    AppMetadataAnalyzer.getIosAppMetadata(f) != null -> {
```
amanjeetsingh150: Seems confusing can we abstract this metadata logic inside validator? Not sure why we have both of them ?

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +130,56 @@ class CloudInteractor(
...
+                        WorkspaceValidationError.InvalidWorkspaceFile -> "Workspace is not a valid zip archive."
+                        is WorkspaceValidationError.GenericError -> err.detail
+                    })
+                }
+            }
+
             Analytics.trackEvent(CloudUploadStartedEvent(
                 projectId = selectedProjectId,
-                platform = triggeredPlatform,
+                platform = inferredPlatform?.name?.lowercase() ?: "unknown",
```
amanjeetsingh150: It seems weird why platform is nullable here. It should always be present if cloud upload is successful

maestro-cli/src/main/java/maestro/cli/util/AppMetadataAnalyzer.kt
```diff
@@ -0,0 +1,119 @@
...
+            if (getWebMetadata(file) != null) return Platform.WEB
+        } catch (_: Exception) {}
+
+        try {
+            if (getIosAppMetadata(file) != null) return Platform.IOS
+        } catch (_: Exception) {}
+
+        try {
+            if (getAndroidAppMetadata(file) != null) return Platform.ANDROID
+        } catch (_: Exception) {}
```
amanjeetsingh150: Why are we swallowing exception here? Shouldn't we hard fail in all three cases and return a non nullable Platform?

maestro-cli/src/main/java/maestro/cli/util/AppMetadataAnalyzer.kt
```diff
@@ -0,0 +1,119 @@
...
+                        bundleId = root.objectForKey("CFBundleIdentifier")?.toString()
+                            ?: throw NullPointerException("Unable to find bundleId"),
+                        platformName = root.objectForKey("DTPlatformName")?.toString(),
+                        minimumOSVersion = root.objectForKey("MinimumOSVersion")?.toString(),
+                    )
+                }
+            }
+        } catch (_: IOException) {
+        } catch (e: Exception) {
+            logger.error("Unexpected error reading iOS metadata: ${e.message}", e)
```
amanjeetsingh150: Similarly I would just fail instead of logging here? Unless there is a known failure we have.

maestro-cli/src/main/java/maestro/cli/util/AppMetadataAnalyzer.kt
```diff
@@ -0,0 +1,119 @@
...
+            logger.error("Unexpected error reading web metadata: ${e.message}", e)
+            null
+        }
+    }
+}
+
+data class IosAppMetadata(
+    val name: String,
+    val bundleId: String,
+    val platformName: String?,       // "iphonesimulator" for simulator builds, "iphoneos" for device builds
```
amanjeetsingh150: Why is this nullable? For iOS app metadata we would always have either of these values.

maestro-cli/src/main/java/maestro/cli/util/AppMetadataAnalyzer.kt
```diff
@@ -0,0 +1,119 @@
...
+    val minimumOSVersion: String?,
+)
+
+data class AndroidAppMetadata(
+    val packageId: String,
+    val supportedArchitectures: List<String> = emptyList(),
```
amanjeetsingh150: There should be no default here. Also, should this string or list? We jus support arm64 right?

maestro-orchestra/src/test/java/maestro/orchestra/workspace/WorkspaceValidatorTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        return zip
+    }
+
+    private fun flow(appId: String, name: String? = null): String {
+        val header = buildString {
+            append("appId: $appId\n")
+            if (name != null) append("name: $name\n")
+        }
+        return "$header---\n- launchApp"
+    }
```
amanjeetsingh150, line 33: Can we just use a fake workspace instead of faking it from kotlin? That would simplify the logic here.

maestro-cli/src/main/java/maestro/cli/util/AppMetadataAnalyzer.kt
proksh: Removed defaults, but it should be List<String>.
These are supported architecture of app provided by users.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -105,16 +111,10 @@ class CloudInteractor(
...
-            androidApiLevel != null -> "android"
-            iOSVersion != null || deviceOs != null -> "ios"
-            flowFile.isWebFlow() -> "web"
-            else -> "unknown"
-        }
+        // Track cloud upload triggered before any file I/O; platform unknown until binary is analyzed
         Analytics.trackEvent(CloudUploadTriggeredEvent(
             projectId = selectedProjectId,
-            platform = triggeredPlatform,
+            platform = if (flowFile.isWebFlow()) "web" else "unknown",
```
amanjeetsingh150: Wait this is breaking change right we are just sending web or unknown as platform. Earlier this used to send android, iOS and web? Sound like with this event we might know platform before should we then change place where its fired to ensure nothing breaks on analytics. Maybe after app validation logic when we have platform.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +131,78 @@ class CloudInteractor(
...
+                            throw CliError("Unsupported platform '${info.platform}' returned by server. Please update your CLI.")
+                        }
+                        AppValidationResult(
+                            platform = platform,
+                            appIdentifier = info.appId,
+                        )
+                    } catch (e: ApiClient.ApiException) {
+                        if (e.statusCode == 404) throw CliError("App binary '$binaryId' not found. Check your --app-binary-id.")
+                        throw CliError("Failed to fetch app binary info. Status code: ${e.statusCode}")
+                    }
```
amanjeetsingh150, line 146: This seems like lot of logic that can be wrapped in a single class responsible for giving validation result and have test for that, cloud interactor responsibility should be just to consolidate platform and workspace (result) from the validators and send them on API.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +131,78 @@ class CloudInteractor(
...
+                }
+
+            val inferredPlatform: Platform = resolvedAppValidation?.platform
```
amanjeetsingh150: Hmm, not sure why this is nullable. If in every case appValidation is returned this should ideally never be null since in either of the cases from the binary file or app binary id this should always be resolved.

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -131,10 +131,78 @@ class CloudInteractor(
...
+                            "No flows in workspace match app ID '${err.appId}'. Found app IDs: ${err.foundIds.ifEmpty { setOf("none") }.joinToString()}"
+                        is WorkspaceValidationError.NameConflict ->
+                            "Duplicate flow name '${err.name}' in workspace. Each flow must have a unique name."
+                        is WorkspaceValidationError.SyntaxError -> "Workspace syntax error: ${err.detail}"
+                        is WorkspaceValidationError.InvalidFlowFile -> err.detail
+                        WorkspaceValidationError.EmptyWorkspace -> "Workspace contains no flows."
+                        is WorkspaceValidationError.MissingLaunchApp ->
+                          "Flows ${err.flowNames.joinToString()} are missing a launchApp command. Each flow must start with a launchApp command."
+                        WorkspaceValidationError.InvalidWorkspaceFile -> "Workspace is not a valid zip archive."
+                        is WorkspaceValidationError.GenericError -> err.detail
```
amanjeetsingh150: Same goes here, maybe we can just remove the Result class and have this exception handling move in validator itself. The class just returns result always here. If it fails the logic would be inside the validation. WDYT

Issue #3092, opened 2026-03-29
WebSocket timeout - SECONDS are used instead of MILLISECONDS


sailsdima:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

In `DadbChromeDevToolsClient.makeSingleWebsocketRequest()`, the specified timeout is 5000 SECONDS instead of intended (I believe :)) 5000 MILLISECONDS. 
This may lead to blocking any of of the calling thread and entire test making it wait for more then a HOUR in case response is never received.

### Actual results

Test will run for 83 minutes in case there are issues with WebSockets connection. The timeout is 5000 SECONDS

### Expected results

The timeout should be 5 seconds.

### About app

Does not matter

### About environment

Does not matter

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

2.3.0

### How did you install Maestro?

install script (https://get.maestro.mobile.dev)

### Anything else?

<img width="592" height="222" alt="Image" src="https://github.com/user-attachments/assets/702583cf-a6b1-4c0e-ad1d-674cdd5e4bc6" />

Fishbowler:
Fixed in #3096

PR #3096, opened 2026-03-30
fix: reduce websocket response timeout to 5 seconds

sailsdima:
## Proposed changes

In `DadbChromeDevToolsClient.makeSingleWebsocketRequest()`, the specified timeout is 5000 seconds instead of intended 5 seconds.
This PR is to address decrease the timeout and make sure the calling the calling thread and entire test isn't in a forever loading state in case WebSocket response is never received.

copilot:summary

## Testing

Run unit tests.

## Issues fixed
https://github.com/mobile-dev-inc/Maestro/issues/3092


Fishbowler:
😂  Good catch!

PR #3098, opened 2026-03-31
fix: handle inputText, eraseText, and pressKey inside cross-origin iframes

alisumup:
## Problem

While testing against `help.sumup.com` I found that `inputText`, `eraseText`, and `pressKey` silently fail when the focused element is inside a cross-origin iframe. The issue is that `document.activeElement` in the main frame returns the `<iframe>` element itself rather than the actual input inside it, so `sendKeys` targets the wrong element.

## Solution

Introduced a `withActiveElement` helper in `CdpWebDriver` that:

1. Checks if `document.activeElement` is an `<iframe>` tag
2. If **yes** — switches Selenium into that frame via `switchTo().frame()`, injects the maestro-web script, resolves the *real* `document.activeElement` inside the iframe, performs the action, then switches back to `defaultContent()`
3. If **no** — runs the exact original logic (no behavioral change for non-iframe workflows)

All three methods (`inputText`, `eraseText`, `pressKey`) now delegate through this helper.

## Testing

Tested against a site that loads a chat widget inside a cross-origin iframe. After tapping on the input field inside the iframe, `inputText` now correctly types into it.

Happy to adjust anything if needed.

Screenshot 2026-03-31 at 13 40 20

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -415,12 +415,8 @@ class CdpWebDriver(
     }
 
     override fun pressKey(code: KeyCode) {
-        val driver = ensureOpen()
-
-        val xPath = executeJS("window.maestro.createXPathFromElement(document.activeElement)") as String
-        val element = driver.findElement(By.ByXPath(xPath))
         val key = mapToSeleniumKey(code)
-        element.sendKeys(key)
+        withActiveElement { it.sendKeys(key) }
     }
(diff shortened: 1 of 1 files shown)


alisumup:
According to claude: 

This E2E failure exists on crossdomain-iframes independent of your commit. It's the [PR](https://github.com/mobile-dev-inc/Maestro/pull/3079) problem to fix. You can mention it in the PR as a note if you want, but it's unrelated to your change.

Fishbowler:
Good catch! Thanks for tidying it up!

PR #3102, opened 2026-03-31
Fix iOS orientation for upside down orientation

Fishbowler:
Found whilst adding tests to our demo_app. iOS driver crashes when attempting to set to upside-down, because its not supported on modern phones. Claude reckons it might've crashed on "face down" too, perhaps.

## Proposed changes

Prevent crash for "portrait upside down" orientation. Would probably work on older iOS. New devices don't support it.

This also adds workflow changes:
* Explicitly rebuild drivers before e2e tests start
* Explicitly rebuild drivers and commit them upon commit to main.

## Testing

Builds and runs, and no longer crashes when I run `setOrientation`. 

e2e test amended to include iOS and assert on the expected behaviour (and demonstrate a lack of crashing)

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
+++ b/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
@@ -121,11 +121,15 @@ struct ScreenSizeHelper {
 
         return switch orientation {
         case .portrait: point
+        case .portraitUpsideDown:
+            CGPoint(x: CGFloat(effectiveWidth) - point.x, y: CGFloat(effectiveHeight) - point.y)
         case .landscapeLeft:
             CGPoint(x: CGFloat(effectiveWidth) - point.y, y: CGFloat(point.x))
         case .landscapeRight:
             CGPoint(x: CGFloat(point.y), y: CGFloat(effectiveHeight) - point.x)
-        default: fatalError("Not implemented yet")
+        default:
+            // .faceUp, .faceDown, unknown — no meaningful 2D rotation, pass through
+            point
         }
     }
 }
--- a/.github/workflows/check-drivers.yaml
+++ b/.github/workflows/check-drivers.yaml
@@ -1,10 +1,17 @@
-name: Check Drivers Up-to-Date
+name: Update Drivers
 
 on:
-  pull_request:
+  push:
+    branches:
+      - main
     paths:
-      - 'maestro-android/**'
       - 'maestro-ios-xctest-runner/**'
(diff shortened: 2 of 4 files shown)


pedro18x (approved):
LGTM. Minor nit: The new end-to-end test asserts visibility after `.upsideDown` but doesn’t perform any taps, so the transform path itself isn’t exercised. That’s probably fine since the docs now say `.upsideDown` is a no-op on iOS, but it’s still worth a quick check that taps land correctly if anyone ever does call it.

Fishbowler:
Good shout. Test updated.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
```diff
@@ -121,11 +121,15 @@ struct ScreenSizeHelper {
 
         return switch orientation {
         case .portrait: point
+        case .portraitUpsideDown:
+            CGPoint(x: CGFloat(effectiveWidth) - point.x, y: CGFloat(effectiveHeight) - point.y)
```
amanjeetsingh150, line 125: Curious which user/e2e test had this issue? AFAIK right now realistically this would be issue on only iPHones right?

amanjeetsingh150 (commented):
I'm hesitant on committing drivers here, if you can for now at lease create a workflow on PRs that generates xcode 26.2 drivers when maestro-ios-xctest-runner changes, it would help contributors to download them from artifacts and check in those.

Ideally we should move away and do this main branch like we discussed.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Helpers/ScreenSizeHelper.swift
Fishbowler: While expanding the e2e tests, setOrientation test showed the iOS driver crashed when on upside-down.

Fishbowler: Yes, this issue is iOS-only.

amanjeetsingh150 (commented):
Few questions:

- If we are building drivers on main, then PR authors are not expected commit drivers right? This would mean check-drivers is not needed
- We still need to build drivers from PR branch to run e2e test but that can happen on Github runner. So, test-e2e can use your code:

https://github.com/mobile-dev-inc/Maestro/pull/3275

to have changed driver for e2e workflow.

I'm also confused now where we exactly ensure and how that main branch has the latest driver always? When they're built and checked in repo.

PR #3108, opened 2026-04-01
feat: add device spec and iOS app compatibility validation to cloud upload

proksh:
## Summary

  - Validates device model + OS against the v2/device/list API before upload, catching invalid configurations client-side
  - Checks iOS app's MinimumOSVersion against the configured device OS version, blocking uploads that would fail on the server
  - Warns when deprecated bare number OS format is used (e.g. "34" instead of "android-34")

 ## Changes

  - DeviceSpecValidator (new) — resolves and validates device model/OS with case-insensitive matching, underscore-to-hyphen fallback, and platform-specific OS shorthand resolution
  - AppValidator.validateDeviceCompatibility() — checks iOS min OS version against device spec (skips when using --app-binary-id)
  - CloudInteractor — wires both validators into the upload flow between DeviceSpec.fromRequest() and WorkspaceValidator.validate()
  - New exception types: InvalidDeviceConfiguration, IncompatibleIOSVersion

### DIFF
--- a/maestro-orchestra/src/test/java/maestro/orchestra/util/AppMetadataAnalyzerTest.kt
+++ b/maestro-orchestra/src/test/java/maestro/orchestra/util/AppMetadataAnalyzerTest.kt
@@ -128,9 +128,9 @@ class AppMetadataAnalyzerTest {
             zos.write(iosPlistXml(platformName = "iphoneos"))
             zos.closeEntry()
         }
-      assertThrows<IllegalArgumentException> {
-        AppMetadataAnalyzer.validateAppFile(zip)
-      }
+        assertThrows<IllegalArgumentException> {
+          AppMetadataAnalyzer.validateAppFile(zip)
+        }
     }
 
     @Test


maestro-orchestra/src/main/java/maestro/orchestra/validation/AppValidator.kt
```diff
@@ -62,4 +67,24 @@ class AppValidator(
...
+                if (appFile == null) return
+                val minOSVersion = iosMinOSVersionProvider?.invoke(appFile) ?: return
+                if (minOSVersion.major > deviceSpec.osVersion) {
+                    throw AppValidationException.IncompatibleIOSVersion(
+                        appMinVersion = minOSVersion.full,
+                        deviceOsVersion = deviceSpec.osVersion,
+                    )
+                }
+            }
+            Platform.ANDROID -> return
```
amanjeetsingh150, line 86: Is this logic not needed for android?

maestro-cli/src/main/java/maestro/cli/cloud/DeviceSpecValidator.kt
```diff
@@ -0,0 +1,142 @@
...
+        platformDevices: Map<String, List<String>>
+    ): String {
+        // 1. Exact match (case-insensitive)
+        platformDevices.keys.firstOrNull { it.equals(model, ignoreCase = true) }?.let { return it }
+
+        // 2. Underscore ↔ hyphen backward-compatibility fallback:
+        //    Convert underscores in the *supported* key to hyphens, then compare
+        //    (handles cases where the user passes "pixel-6" but supported is "pixel_6")
+        platformDevices.keys.firstOrNull { key ->
+            key.replace('_', '-').equals(model.replace('_', '-'), ignoreCase = true)
```
amanjeetsingh150: This should always be pixel_6. Even the legacy one seems to have the same nomenclature.

maestro-cli/src/main/java/maestro/cli/cloud/DeviceSpecValidator.kt
```diff
@@ -0,0 +1,142 @@
...
+        available.firstOrNull { it == os }?.let { return it }
+
+        // 2. Bare integer major version – matches first "(iOS|tvOS|watchOS)-<N>-*" (deprecated format)
+        if (os.toIntOrNull() != null) {
+            val regex = Regex("^(iOS|tvOS|watchOS)-$os-.*")
+            available.firstOrNull { regex.matches(it) }?.let {
+                PrintUtils.warn("Numeric OS format '$os' is deprecated. Use the full format instead. Available: ${available.joinToString(", ")}")
+                return it
+            }
+        }
```
amanjeetsingh150: Nit:

1. For backward compatibility let's add this in deprecated function for both iOS and android
2. Also we don't need watchOS and tvOS

PR #3114, opened 2026-04-02
feat(longPressOn): add configurable duration parameter

### LINKED ISSUE #3083: Command longPressOn has a hardcoded 3s duration and cannot bypass Human challange
The `longPressOn` command has a hardcoded 3s duration and there is no property to do it, not a custom gesture API to create something that may fit my requirements or any way to chain multiple longPressOn (repeat just doesn't work). 

Because of this, we are not able to fully automate our tests cause Human (https://docs.humansecurity.com/applications/human-challenge) catchup has a very long press to bypass the check. I know that probably a more real scenario should be trying to avoid Human to run during test but in our case that it is impossible right now. 

So basically that put me in a position where using Maestro is not an option cause as things work right now, I have to manually go to the simulator and bypass Human and then run the tests.

There was already a ticket mentioning this issue but was closed without a fix, I am craeting this issue to provide another real example where this feature could have been useful.

See: https://github.com/mobile-dev-inc/Maestro/issues/2320

Matheeusb:
## Summary

Fixes #3083

The `longPressOn` command had a hardcoded 3-second duration with no way to configure it. This prevented users from automating workflows that require longer (or shorter) press durations, such as bypassing Human Security challenges that require extended presses.

### New YAML API

Users can now specify a `duration` field (in milliseconds) on `longPressOn`:

```yaml
# Custom duration (8 seconds)
- longPressOn:
    id: "my_button"
    duration: 8000

# Works on text selectors too
- longPressOn:
    text: "Hold to confirm"
    duration: 5000

# Works on screen coordinates
- longPressOn:
    point: "50%,50%"
    duration: 2000

# Backward compatible — omitting duration keeps the existing 3s default
- longPressOn: "Some element"
```

### Modules changed

| Module | Change |
|---|---|
| `maestro-orchestra-models` | Added `duration: Long?` to `TapOnElementCommand`, `TapOnPointV2Command`, `TapOnPointCommand`; added `DEFAULT_LONG_PRESS_DURATION_MS = 3000L` constant |
| `maestro-client` | Updated `Driver` interface: `longPress(point, durationMs = 3000L)`; propagated `longPressDurationMs` through all `tap()` overloads and into `hierarchyBasedTap`/`screenshotBasedTap` |
| `maestro-client` (drivers) | `AndroidDriver`, `IOSDriver`, `WebDriver`, `CdpWebDriver` — all now use the passed `durationMs` instead of the hardcoded `3000` |
| `maestro-orchestra` | `YamlElementSelector` — new `duration` field; `YamlFluentCommand` — extracts and passes `duration`; `Orchestra` — passes `command.duration` to all tap call sites |
| `maestro-test` | `FakeDriver.Event.LongPress` now captures `durationMs` for test assertions |

## Tests

- Added `033_longPressOn_withDuration.yaml` fixture covering duration on element-by-text, element-by-id, default (null), and screen-point variants
- Added `longPressOn with custom duration should parse duration into command` test in `YamlCommandReaderTest`

## Backward compatibility

Fully backward compatible. Omitting `duration` preserves the existing 3000ms default behaviour.


Fishbowler:
Perhaps it's worth extending https://github.com/mobile-dev-inc/Maestro/blob/main/e2e/demo_app/.maestro/commands/longPressOn.yaml ?

Could add a duration of 250ms and check that the label isn't displayed? (The underlying Flutter GestureDetector has a 500ms tolerance to trigger onLongPress, as far as I can tell)

Matheeusb:
> Perhaps it's worth extending https://github.com/mobile-dev-inc/Maestro/blob/main/e2e/demo_app/.maestro/commands/longPressOn.yaml ?
> 
> Could add a duration of 250ms and check that the label isn't displayed? (The underlying Flutter GestureDetector has a 500ms tolerance to trigger onLongPress, as far as I can tell)

Hi @Fishbowler! I added this validation and it works! ✅

Captura de Tela 2026-06-09 às 11 12 43

PR #3118, opened 2026-04-02
Fix return statement for deriving XPath for non elements not in an iframe

Fishbowler:
## Proposed changes

Issue introduced in #3079 

No idea how this ever passed e2e before....

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/WebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/WebDriver.kt
@@ -737,7 +737,7 @@ class WebDriver(
                 catch (e: Exception) { LOGGER.warn("Failed to switch back to default content", e) }
             }
         } else {
-            val xPath = executeJS("window.maestro.createXPathFromElement(document.activeElement)") as String
+            val xPath = executeJS("return window.maestro.createXPathFromElement(document.activeElement)") as String
             val element = driver.findElement(By.ByXPath(xPath))
             action(element)
         }


Fishbowler:
This is only needed on one driver
WebDriver uses Selenium's executeScript, which requires explicit return to produce a value
CdpWebDriver uses CdpClient.evaluate, which wraps expressions in try { return $expression } internally

Fishbowler:
## Testing

In demo_app:

```
% maestro test --platform web --include-tags passing .maestro/web_flows

Web support is in Beta. We would appreciate your feedback!


Waiting for flows to complete...
[Passed] LaunchApp should clear state when clearState is true (13s)
[Passed] Scroll should reveal content below the fold (12s)
[Passed] simple (14s)
[Passed] Erase text should clear typed input (18s)
[Passed] ClearState command should clear state (12s)
[Passed] LaunchApp should not clear state (12s)
[Passed] Iframe test (9s)
[Passed] Back press should navigate to previous page (14s)

8/8 Flows Passed in 1m 44s
```

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -737,7 +737,7 @@ class WebDriver(
                 catch (e: Exception) { LOGGER.warn("Failed to switch back to default content", e) }
             }
         } else {
-            val xPath = executeJS("window.maestro.createXPathFromElement(document.activeElement)") as String
+            val xPath = executeJS("return window.maestro.createXPathFromElement(document.activeElement)") as String
```
amanjeetsingh150, line 740: Would this always be a non nullable string? Present everytime?

Fishbowler: Yes. `document.activeElement` will always return an element, which in turn can always be expressed as an XPath.

https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement#value

PR #3121, opened 2026-04-02
fix: append env vars instead of prepend so external env overrides YAML defaults

pedro18x:


pedro18x:
Closing, this change was not needed for the fix.

Issue #2817, opened 2025-11-04
Feature Request: Support for .env files in CLI and Maestro Studio


fvoordeckers:
## Current Behavior
- The CLI currently supports environment variables via -e parameters when running maestro.
- There is no built-in option to specify a .env file directly.
- As a workaround, I use a script to convert .env files into CLI parameters.

In Maestro Studio Desktop, environment variables can only be set one by one through the settings UI. This approach becomes cumbersome and unmaintainable for projects that require managing 10+ environment variables. Quickly switching environments will be a pain and prone to errors.

## Proposed Feature
Add support for specifying a .env file that can be loaded automatically by both the CLI and Maestro Studio Desktop. Ideally:
- The CLI could include a flag such as --env-file <path> to load variables from a .env file.
- Maestro Studio could allow pointing to an .env file, automatically syncing its values whenever the file changes. This would be in addition to the current configuration in Maestro Studio Desktop

## Benefits
- Simplifies management of large sets of environment variables.
- Keeps local and Studio environments consistent.
- Removes the need for manual syncing or scripting workarounds.

## Example CLI usage:
`maestro test flow.yaml --env-file .env.staging`

### Anything else?

Another related issue handling `.env` files:
https://github.com/mobile-dev-inc/maestro-studio/issues/31

jf-qa:
Bumping this. We would also benefit from this.

Fishbowler:
Please don't do that. Use the 👍 further up rather than adding noise.

sailsdima:
Hey, I'm picking this up, starting with --env-file support for the CLI.

A few questions:
1. The Maestro Desktop is not open source as I understand, there is no way to make changes with it for contributors?
2. Is there any preference on .env format support? I'm currently handling it in KEY=VALUE format.

@Fishbowler

Fishbowler:
Sorry for the delay - I've been away for a few days.

That'd be ace!

Studio Desktop is indeed closed source, but is also "just" an electron wrapper around Maestro and VS Code's monaco editor. Once env file support is in Maestro, adding the plumbing into Studio should be relatively quick.

I agree with `KEY=VALUE` as the v1. 
End state probably matches whatever [dotenv](https://github.com/bkeepers/dotenv) do?

sailsdima:
Hi @Fishbowler, could you take a look on PR attached?

PR #3127, opened 2026-04-03
feat: Add support for loading environment variables from .env files

sailsdima:
## Proposed changes

This PR adds a new `--env-file` CLI option that allows loading environment variables from a `.env` file instead of passing them one-by-one with `-e`. The support is added for commands: `test`, `record`, and `cloud`.

`EnvFileParser` - parses .env files into key-value map. `TestCommand`, `RecordCommand` and `CloudCommand` call merge .env file with environment map logic.

File parser supports:
- Values in format `NAME=VALUE`
- Line comments: `# Comment`
- Inline comments: `NAME=VALUE # Comment`
- Quoting values: `NAME='Quote 1 # no-comment'` OR `NAME="Quote 2 # no-comment"`. Usage: `COLOR="#FFFFFF"`

Usage example:
```
# Test comment
APP_ID=com.example.app # Test inline comment
orientationLandscapeLeft=LANDSCAPE_LEFT
orientationLandscapeRight=LANDSCAPE_RIGHT
orientationUpsideDown= UPSIDE_DOWN
orientationPortrait= PORTRAIT
TEST_MESSAGE=Hello from environment file!
TEST_URL=https://api.example.com
TEST_QUOTE_1="Quote 1"
TEST_QUOTE_2='Quote 2'
```
```
./maestro-cli/build/install/maestro/bin/maestro test flow.yaml --env-file environments/.env
```

**Note:** if both `--env-file` and `-e` used, env variables added via `-e` will override values with the same name from `--env-file`.

copilot:summary

## Testing

Created a new e2e test workspace (e2e/workspaces/env_file/) with tag env-file that uses --env-file. The test requires no app and only uses assertTrue to verify that env vars are loaded and parsed correctly. Updated run_tests script to include it on CI.

**Local run**:
```
./maestro-cli/build/install/maestro/bin/maestro test --env-file e2e/workspaces/env_file/.env e2e/workspaces/env_file/
```

image
image

## Issues fixed
https://github.com/mobile-dev-inc/Maestro/issues/2817


samanthans:
I needed this for my workflow and was very glad to see someone who already did it, I tried on Windows and macOS and it worked perfectly for me; previously I was relying on a wrapper that would read the file and execute maestro with `-e` arguments for each variable, but this is much better. Really hope it gets merged!

sailsdima:
Hi @Fishbowler, gentle bump on this PR. Looks like it may have been missed.

maestro-cli/src/main/java/maestro/cli/util/EnvFileParser.kt
```diff
@@ -0,0 +1,83 @@
...
+    private fun unquote(raw: String): String {
+        val trimmed = raw.trim()
+        if (trimmed.length >= 2) {
+            if ((trimmed.startsWith('"') && trimmed.endsWith('"')) ||
+                (trimmed.startsWith('\'') && trimmed.endsWith('\''))) {
+                return trimmed.substring(1, trimmed.length - 1)
+            }
+        }
+        // Strip inline comments for unquoted values: "value # comment" → "value"
+        val commentIndex = trimmed.indexOf(" #")
```
Fishbowler: This feels fragile. Perhaps something based on `Regex("\\s+#")` will trap different whitespace amounts and other whitespace characters?

e2e/demo_app/.maestro/env-file.yaml
```diff
@@ -0,0 +1,17 @@
+appId: ${APP_ID}
```
Fishbowler, line 1: Would be good to see this test working within one of the other existing applications. `demo_app` is the default. We'd probably want to reuse the test to check that this works with Maestro Cloud too, which requires an app to exist.

That probably means reworking the `run_tests` impl to deal with special tags.

Fishbowler (dismissed):

simon-gilmurray:
@sailsdima would love to see the above comments addressed when you have some time 🙏

maestro-cli/src/main/java/maestro/cli/util/EnvFileParser.kt
sailsdima: done

e2e/demo_app/.maestro/env-file.yaml
sailsdima: Done

darius-buildops:
any news on this ?

seba9999:
Any news on this ? 👍

petewalker:
@sailsdima - Anything I can do to help get this PR over the line? We have a large number of environment variables we need to inject into our test runner and are considering rolling our own wrapper script around maestro to make this work

PR #3128, opened 2026-04-04
feat: auto-build drivers on source change with CI staleness check

amanjeetsingh150:
## Summary
- Wire Gradle `processResources` to trigger iOS and Android driver builds when their source changes
- Add checksum-based skip for iOS driver builds so `xcodebuild` only runs when Swift source actually changed
- Add CI workflow (`check-drivers.yaml`) that fails PRs if committed drivers are out of date
- iOS CI check uses source checksum comparison (runs on ubuntu, no macOS runner needed)
- Android CI check rebuilds and diffs APKs

## Test plan
- [ ] Run `./gradlew distZip` with no Swift changes — iOS build should be skipped
- [ ] Change a `.swift` file, run `./gradlew distZip` — iOS build should trigger
- [ ] Push a PR touching `maestro-ios-xctest-runner/` without updating zips — CI should fail
- [ ] Push a PR touching `maestro-android/` without updating APKs — CI should fail


.github/workflows/check-drivers.yaml
```diff
@@ -0,0 +1,97 @@
...
+    needs: changes
+    if: needs.changes.outputs.android == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Set up Java
+        uses: actions/setup-java@v4
```
Fishbowler, line 37: Checkout should be v6
Setup Java should be v5

.github/workflows/check-drivers.yaml
```diff
@@ -0,0 +1,97 @@
...
+    if: needs.changes.outputs.ios == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Check iOS drivers are up to date
+        run: |
+          CURRENT_CHECKSUM=$(find ./maestro-ios-xctest-runner -type f \
```
Fishbowler, line 70: Should the iOS and Android checks implement the staleness check in the same way?

maestro-ios-xctest-runner/build-maestro-ios-runner.sh
```diff
@@ -16,6 +16,23 @@ else
 	BUILD_OUTPUT_DIR="Debug-iphoneos"
 fi
 
+# Check if source has changed since last build by comparing checksums
+RESOURCES_DIR="./maestro-ios-driver/src/main/resources/$DERIVED_DATA_PATH/$BUILD_OUTPUT_DIR"
+CHECKSUM_FILE="$RESOURCES_DIR/.source-checksum"
+CURRENT_CHECKSUM=$(find ./maestro-ios-xctest-runner -type f \
+  \( -name "*.swift" -o -name "*.h" -o -name "*.m" -o -name "*.plist" -o -name "*.pbxproj" -o -name "*.xcscheme" \) \
```
Fishbowler, line 23: If this is in both the action and the build script, should it be a separate check script that both call?

PR #3129, opened 2026-04-04
feat: build demo_app from source in e2e workflow

amanjeetsingh150:
## Summary
- Copy demo_app source into the e2e directory
- Build demo_app APK from source in CI instead of downloading pre-built artifacts
- Reorder Android SDK setup before demo_app APK build

## Test plan
- [x] Verify e2e workflow builds the demo_app APK successfully
- [x] Verify e2e tests pass with the source-built APK

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -243,7 +243,7 @@ jobs:
 
       - name: Run tests
         working-directory: ${{ github.workspace }}/e2e
-        timeout-minutes: 20
+        timeout-minutes: 30
         run: ./run_tests android
 
       - name: Stop screen recording of AVD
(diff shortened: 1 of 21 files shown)


e2e/demo_app/CLAUDE.md
```diff
@@ -0,0 +1,85 @@
+# CLAUDE.md
```
Fishbowler, line 1: Do we need to reference this from somewhere?

.github/workflows/test-e2e.yaml
```diff
@@ -124,32 +119,12 @@ jobs:
...
-      - name: Add Maestro CLI executable to PATH
-        run: |
-          unzip maestro.zip -d maestro_extracted
-          echo "$PWD/maestro_extracted/maestro/bin" >> $GITHUB_PATH
-
-      - name: Check if Maestro CLI executable starts up
-        run: |
-          maestro --help
-          maestro --version
+          java-version: 21
```
Fishbowler, line 127: Why Java 21?

amanjeetsingh150: Not a good reason we can go back 17. I think claude changed it.

PR #3133, opened 2026-04-04
feat: add diagnose-test-failure skill

amanjeetsingh150:
## Summary

- Add Claude Code skill for diagnosing Maestro test failures from CI artifacts. This for us to quickly root cause issues on our E2E tests.

The skill requires to submit a job or artifacts directory which would allow claude to go through all the context around failed command and query maestro.log to see the failure, invoking this from maestro codebase also ensure we have context about code.


.claude/skills/diagnose-test-failure/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+## Step 2: Identify Failing Tests from Commands JSON
+
+Each `commands-(flowName).json` contains the list of executed commands with their status and metadata.
+
+```bash
+# List all command files
+ls artifacts/commands-*.json
+
+# Find which tests have FAILED commands
+grep -l '"FAILED"' artifacts/commands-*.json
```
Fishbowler, line 72: Would `jq` be more efficient in this stage, if it's available?

Issue #1409, opened 2023-09-01
tapOn "text" not working when accessibilityLabel is available on iOS [React Native]][Maestro version 1.31.0]


vanGalilea:
**Describe the bug**
tapOn "text" not working when `accessibilityLabel` is available on iOS.
After removing `accessibilityLabel` tapOn "text" works as expected

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Start the simulator (iPhone 8, iOS 14.3)
2. Run `maestro test` on a React Native app with the following `Pressable` component:
```
<Pressable accessibilityLabel={'easy-button'}>
  <Text>'Click me!'</Text>
</Pressable>
```
4. The Flow fails at `❌  Tap on "Click me!"`

**Expected behavior**
The Pressable's text should be detected and the tap should be successful, whether an `accessibilityLabel` was specified or not.

**Screenshots**
N/A

**Environment information (please complete the following information):**

- Maestro version [v1.31.0]
- Platform: [iOS 14.3]
- Framework: [React Native]
- Device model and OS version: [iPhone 8]
- Simulator, emulator or physical device [Simulator]
- Host [Max M1]

**Workspace (if applicable)**
- https://github.com/vanGalilea/react-native-testing/tree/feature/maestro
- Check [readme](https://github.com/vanGalilea/react-native-testing/tree/feature/maestro#setup-and-requirements-) with more info on how to run, build and run tests 

**Bugreport output (if applicable)**
N/A

**Additional context**
N/A

dgreasi:
Same here

Exover95:
It also doesn't work with iOS17 natively.
The copyFromTo command does not copy the text, the field is then empty, if we use accessibilityLabel

LeunensMichiel:
Any update on this one @Fishbowler ? This seems quite a common use case to assert if the text is visible for a certain element

Fishbowler:
@LeunensMichiel You're in the right place for updates, although hitting Subscribe is typically better than tagging folks for attention. It doesn't look like anyone in the community has been motivated to attempt a fix for this. It possible that it's only touched a small handful of folks in 2.5 years. Certainly, there's been folks who can "assert if the text is visible for a certain element" in that time.

[...]

PR #3138, opened 2026-04-04
feat: enable parallel iOS simulator execution with per-device session tracking

### LINKED ISSUE #1853: Unable to run parallel/sharded tests on Android

qwertey6:
## Proposed changes

**Parallel execution on iOS is completely broken.** Two `maestro test` processes targeting different iOS simulators collide on the XCTest HTTP server port and interfere with each other's sessions. This makes `--shard-split` unusable on iOS and forces sequential execution.

### Impact

- **Unblocks iOS parallel testing**: `--shard-split` and multi-process execution now work on iOS
- **Resolves 5 open issues** and **supersedes 2 stale open PRs**
- **Removes the `Thread.sleep(1000)` hack** from #1867 — the per-device session check eliminates the race condition it was working around

### Root cause

Two bugs compound (same analysis as #2339 by @avinash-bharti):

1. **Port collision**: iOS simulators share the host's localhost. Two Maestro processes both bind to the same XCTest server port.
2. **Session interference**: `hasActiveSessions()` checks if _any_ session exists on the same platform — regardless of device. Process B sees process A's session and skips `driver.open()`, connecting to the wrong device.

### What this PR does differently from #2339 and #2821

Builds on the same architectural approach (per-device session tracking + `--driver-host-port`), with additional correctness fixes discovered during testing:

| Fix | #2339 | #2821 | This PR |
|-----|-------|-------|---------|
| Per-device session tracking | Yes | No | Yes |
| `--driver-host-port` flag | Yes | Yes | Yes |
| Session self-detection bug fix | No | No | **Yes** — heartbeat at `initialDelay=0` fires before session check, causing process to find itself and skip `driver.open()`. Broke Android completely. |
| Cross-process file locking | No | No | **Yes** — `~/.maestro/sessions` had no `FileLock`, concurrent processes could corrupt session data |
| Per-device shutdown | No | No | **Yes** — `shouldCloseSession(platform, deviceId)` replaces global `activeSessions().isEmpty()` |
| Port availability check | No | No | **Yes** — `isPortAvailable()` with correct check-then-claim ordering |
| Debug log directory race | No | No | **Yes** — PID appended to prevent parallel collision |
| Unit tests | No | No | **Yes** — 8 tests, fakes not mocks, Java 8 compatible |

### Changes

- `SessionStore.kt` — Refactored from singleton to injectable class. Key format: `"{platform}_{deviceId}_{sessionId}"`. New methods: `hasActiveSessionForDevice()`, `shouldCloseSession()`, `activeSessionsForDevice()`
- `MaestroSessionManager.kt` — Uses `SessionStore.default`, per-device heartbeat/delete/close
- `KeyValueStore.kt` — Added `FileLock` for cross-process safety
- `App.kt` + `TestCommand.kt` — `--driver-host-port` CLI option
- `TestCommand.kt` — `selectPort()` with `isPortAvailable()` check
- `SocketUtils.kt` — `isPortAvailable()` function
- `DebugLogStore.kt` — PID in log directory name (Java 8 compatible via `ManagementFactory`)
- Command files — Pass `parent?.driverHostPort` instead of `null`

### Verification

| Scenario | Result |
|----------|--------|
| 3 iOS sims parallel, cold start | Pass |
| 3 iOS sims parallel, warm start | Pass |
| Android emulator, single device | Pass |
| `--driver-host-port` explicit ports | Pass |
| `--shard-split=2` with 2 flows | Pass |
| Solo + shard simultaneous (3 devices) | Pass |

## Testing

- 8 new unit tests in `SessionStoreTest` (per-device isolation, key format, lifecycle, shutdown)
- `useJUnitPlatform()` added to `maestro-cli/build.gradle.kts` (was missing)
- Uses fakes, not mocks (per CONTRIBUTING.md)
- Java 8 compatible

> **Note**: This is PR 1 of 4 in a stacked series. Subsequent PRs build on this:
> - PR 2: ~4x faster startup (skip unnecessary driver reinstall)
> - PR 3: Version safety (driver hash checks prevent stale drivers)
> - PR 4: Fix `waitToSettleTimeoutMs` on iOS (settle timeout was silently broken)

## Issues fixed

Closes #1853, #2556, #1485, #2082, #2703

Supersedes #2339 and #2821 — credit to @avinash-bharti and @omnarayan for the root cause analysis and architectural direction.


qwertey6:
**CI note**: The E2E failures on Android and iOS are pre-existing and unrelated to this PR:

- **Android** (`killApp` test): 44/45 flows pass, 1 fails with `"Form Test" is not visible` after a `killApp` command. App didn't return to expected state — flaky assertion.
- **iOS** (`ios-advanced-flow`): Wikipedia app crashes during `clearState` + `assertVisible "Next"` in `launch-clearstate-ios.yaml`.

Both failures reproduce on other unrelated PRs in the repo (`env-file-support`, `pr/contains-child-fix`, `pr/skill`, etc.) — every recent E2E run on this workflow shows the same pattern.

Build, unit tests, web E2E, and iOS XCTest runner tests all pass.

vaikesh07:
@qwertey6 : when are we planning to merge this. We are also facing similar issue, your solution will work for us as well

Fishbowler:
Is the description of this PR correct?
> This makes --shard-split unusable on iOS and forces sequential execution.

I think sharding is a different use case from two parallel unrelated `maestro test` operation, and isn't currently problematic?

I've tidied up a couple of the linked issues - clearly duplicates. Oops.

qwertey6:
Good question — you're right that `--shard-split` and two separate `maestro test` processes are different use cases. Let me clarify:

**Two separate `maestro test` processes**: Both port collision (hardcoded 7001) and session interference (`hasActiveSessions` matching cross-device). Both broken.

**`--shard-split`**: Port collision is handled within the process by the existing `usedPorts` ConcurrentHashMap (each shard gets a unique port). But session interference IS a problem — `hasActiveSessions(sessionId, platform)` checks if any session exists for the platform, so shard B's heartbeat causes shard A's `connectToExistingSession` check to return `true`, skipping `driver.open()` on shard A's device. The `Thread.sleep(1000)` hack in #1867 masked this by delaying the heartbeat past the session check window, but it's a race condition that fails under load.

I'll update the PR description to be more precise about which use case each fix addresses. Thanks for the nudge — and for cleaning up the dupes.

qwertey6:
@Leland-Takamine — I know you're heads-down on a lot, so here's the quick version: these 7 PRs close **9 open issues** (some since 2023), supersede **2 stale PRs**, and deliver **4-5x faster iOS startup** (52s→10s) + **2x faster flows** with tuned settle timeouts. They complement your #3171 by covering the iOS parallel execution side.

I've rebased onto current `main` and restructured so 4 of the 7 PRs are now **fully independent** — reviewable and mergeable in any order:

**Stacked (1→2→3):**
- #3138 — Per-device session tracking + `--driver-host-port` (closes #1853, #2556, #1485, #2082, #2703)
- #3139 — 4-5x faster startup: driver reuse, build product caching (closes #1096)
- #3140 — Version safety: SHA-256 hash checks prevent stale drivers (closes #1601)

**Independent:**
- #3141 — Fix `waitToSettleTimeoutMs` on iOS — 53s→25s (addresses #1528)
- #3165 — Fix `tapOn` with `accessibilityLabel` on iOS (closes #1409, open 2.5 years)
- #3166 — IPv4/IPv6 XCTest connection fix — **3 lines** (closes #1299, open 2.5 years)
- #3167 — Persist XCTest runner port, stop orphaning `xcodebuild` processes

The independent PRs are small — #3166 is 3 constants, #3165 is 10 lines. Could be quick wins.

Also — sent you a connect on LinkedIn. Would love to chat about where Maestro is headed.

proksh (changes requested):
Thanks for tackling this @qwertey6 - parallel iOS is a real pain point and the three collision points are correctly identified.

Looking at the PR, every change is coordination machinery layered around three shared resources (the XCTest port on localhost, `~/.maestro/sessions`, and the debug log directory) so that two processes can use them simultaneously without lying to each other. Each fix is mechanically correct but downstream of one underlying design choice: that **those three resources are shared resources processes need to negotiate over**. So the PR reads as a stack of small workarounds rather than one architectural change.

I'd suggest looking at a simpler architecture that removes most of this:
1. Ask the OS for a port. `ServerSocket(0).use { it.localPort }` - no scanning, no `7001..7128` range, no usedPorts map, no `--driver-host-port` flag. Same TOCTOU exposure as the current `isPortAvailable` check (microseconds, ephemeral range), dramatically simpler mechanism.

2. Stop sharing session state across processes - and consider deleting `SessionStore` entirely (not "make it in-memory" - delete it). Within one Maestro JVM, two sessions never share a device: `--shard-split` distributes shards 1:1 onto distinct devices (runShardSuite reads `deviceIds[shardIndex]`, and validateDevices enforces `devices >= shards`), so the warm-reuse check is always false within one process. The shared file at `~/.maestro/sessions` exists only to enable warm reuse across separately-launched CLI invocations on the same device - a CLI-iteration optimization worth a few seconds of cold-start. If warm reuse ever turns out to matter in practice, the right shape is a long-running daemon that owns drivers, not a shared file scribbled on by every CLI invocation.

The PID-suffixed log dir should stay - its unrelated to the workaround pattern.

Happy to talk through this if I'm missing a use case.

qwertey6 (commented):
@proksh — thanks for the detailed review. I agree with point 1 and disagree with point 2. Here's my reasoning:

### Point 1: `ServerSocket(0)` — done

I've updated `selectPort()` to use `ServerSocket(0)`. The 7001-7128 range, `usedPorts` ConcurrentHashMap, and `isPortAvailable` scanning are all removed. Much simpler. The `--driver-host-port` flag stays for the explicit-port use case (it still validates via `isPortAvailable` before using the user-specified port).

### Point 2: Deleting `SessionStore` — I'd push back here

The `SessionStore`, `connectToExistingSession`, heartbeat loop, and shared `~/.maestro/sessions` file are **existing Maestro infrastructure** that my changes fix — they don't add it. The per-device key format, self-exclusion bug, and file locking are bug fixes to code that was already shipping and already broken.

Deleting `SessionStore` would require rearchitecting two cross-process behaviors that currently depend on it:

**1. Shutdown coordination (MaestroSessionManager line 129):**
```kotlin
if (SessionStore.activeSessions().isEmpty()) {
    session.close()
}
```
Process A checks whether any other process is still using a device before closing. Without this, process A's shutdown hook calls `session.close()` → `IOSDriver.close()`, potentially killing a driver that process B is actively using. Removing `SessionStore` means either:
- Every process kills its driver on exit → breaks parallel execution (process A kills B's XCTest runner)
- No process kills its driver on exit → leaks orphaned xcodebuild/runner processes indefinitely
- A daemon owns driver lifecycle → correct, but significantly larger scope

**2. Warm reuse across CLI invocations (`connectToExistingSession`):**
Without it, every `maestro test` invocation pays the full cold-start penalty: extract build products, run `xcodebuild test-without-building`, poll for ~15-20s until the HTTP server responds. Warm reuse cuts this from ~15s to ~8s per device. For local iteration (run flow, tweak, run again), that's the difference between responsive and sluggish.

A long-running daemon is the right eventual architecture for warm reuse, but it's significantly larger scope than fixing the existing mechanism. My changes make the existing shared-state approach correct (per-device keys, self-exclusion, file locking) without introducing new architectural concepts.

### Summary

I've adopted `ServerSocket(0)` for port selection (pushed). I'd keep `SessionStore` with the per-device fixes — it's existing infrastructure that serves real purposes, and deleting it introduces new bugs that would require a daemon to solve properly.

nabettu:
+1 on this. We're running parallel Maestro tests across multiple iOS simulators and hitting the exact port collision issue described here. Currently forced to run sequentially as a workaround, which significantly slows down our CI pipeline. Would love to see this merged!

proksh:
@qwertey6 Thanks for the pushback and for shipping `ServerSocket(0)` - that part looks good.

I dug into `SessionStore` / `MaestroSessionManager` to make sure I understood what's actually there today. To summarize: `~/.maestro/sessions` is a single file (not a directory) where `KeyValueStore` writes `key=value` lines. Each process writes a `{platform}_{sessionId}` key, refreshes its heartbeat timestamp every 5s, prunes entries older than 21s, and on shutdown checks `activeSessions().isEmpty()` before closing the driver. The heartbeat, the `Thread.sleep(1000)` race hack, the new `FileLock`, the per-device key, and the self-exclusion logic all exist to make that shared keyvalue file safe for N concurrent processes.

I agree warm reuse is worth keeping and a daemon is out of scope. What  I'd still push on is the shape of the cross-process state — the shared `~/.maestro/sessions` keyvalue file with `FileLock` is the fragile part. Shared mutable state across N processes with negotiated access tends to accumulate edge cases (stale entries from crashed processes, heartbeat skew, lock contention, NFS home dirs).


Alternative: turn `~/.maestro/sessions` into a directory with one file per device, named by `deviceId`, reusing the existing KeyValueStore line format:
`~/.maestro/drivers/<deviceId>.json`   →  `{ "port": 50231, "pid": 12345, "startedAt": "..." }`

  - A process starting a driver writes its own device's file; no other process writes to it.
  - For warm reuse: read the file for the target device, check PID is alive and port responds, connect or start fresh (overwriting).
  - Stale files detected by liveness check at read time — no heartbeat, no prune logic.
  - No `FileLock`, no shared keyvalue store, no self-exclusion, no `activeSessions().isEmpty()` shutdown coordination. Each process owns its own driver's lifecycle.

Migration is trivial — `~/.maestro/sessions` is transient state, just delete the old file if it exists and mkdir the directory.

This addresses the underlying concern (processes shouldn't negotiate over shared mutable state) without losing warm reuse or scoping in a daemon. Should be smaller than what's in the PR today — `SessionStore` collapses to a thin per-device read/write, the KeyValueStore FileLock change drops out, and the heartbeat thread + Thread.sleep(1000) go away.

 One open question: am I missing a case where two processes legitimately need to share the same driver on the same device simultaneously? `--shard-split~ distributes 1:1 onto distinct devices within a JVM, and across CLI invocations the second one can just connect to the first's HTTP server without host-side coordination. If there's a case I'm missing,  I'll drop the objection.

Fishbowler:
I run `maestro hierarchy` against a device connected to Studio. Is that 2 sessions?

proksh:
> I run `maestro hierarchy` against a device connected to Studio. Is that 2 sessions?

Yes that will be 2 sessions.

qwertey6:
@proksh — I like the direction of per-device files. In fact, that's exactly what PR #3167 already implements — `XCTestPortStore` writes `~/.maestro/xctest-ports/<deviceId>` with just the port number. On the next invocation, it reads the saved port, probes it with `Socket.connect()`, and reuses if something responds. No PID, no heartbeat, no prune logic.

One concern with the `{port, pid, startedAt}` JSON shape you proposed: PID liveness checks aren't cross-platform. `ProcessHandle.of(pid)` is Java 9+ (Maestro targets Java 8), and there's no reliable equivalent on Windows without JNA. The port probe alone is sufficient — if the port responds to `/status`, the runner is alive regardless of PID. If it doesn't, start fresh.

So the question is whether this per-device port file (already in #3167) can fully replace `SessionStore`, or whether there are cases where the heartbeat-based shared state is still needed. I believe it can replace it — the only cross-process coordination `SessionStore` provides is:

1. **Warm reuse**: "is a driver already running on this device?" → port probe handles this
2. **Shutdown coordination**: "should I close this driver?" → per-device file check handles this

Two paths forward:
- **Option A**: Merge this PR as-is (fixes the bugs in the existing `SessionStore`), then land #3167 next and use it as the foundation to refactor `SessionStore` out in a follow-up
- **Option B**: Collapse #3138 and #3167 into a single PR that replaces `SessionStore` with per-device port files from the start

I'd lean toward A — ship the bug fixes now, iterate on the architecture next — but happy to go either way.

maestro-cli/src/main/java/maestro/cli/session/SessionStore.kt
```diff
@@ -56,18 +47,44 @@ object SessionStore {
         }
     }
 
-    fun hasActiveSessions(
+    fun shouldCloseSession(platform: Platform, deviceId: String? = null): Boolean {
+        return activeSessionsForDevice(platform, deviceId).isEmpty()
+    }
+
+    fun activeSessionsForDevice(platform: Platform, deviceId: String? = null): List<String> {
+        val devicePrefix = "${platform}_${deviceId ?: "unknown"}_"
```
proksh: We should avoid fallback to "unknown" here. This can cause 2 unrelated runs under one key.

Why is this deviceId nullable? Can we require non-null device key before writing session state.

maestro-cli/src/test/kotlin/maestro/cli/session/SessionStoreTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        sessionStore.heartbeat("s2", Platform.ANDROID, "emulator-5554")
+        sessionStore.heartbeat("s3", Platform.IOS, "device-B")
+
+        val deviceASessions = sessionStore.activeSessionsForDevice(Platform.IOS, "device-A")
+
+        assertThat(deviceASessions).hasSize(1)
+    }
+
+    @Test
+    fun `shouldCloseSession returns true when no sessions remain for device even if other devices have sessions`() {
```
proksh, line 112: can we add a false case for this.

Two Android heartbeats on `emulator-5554`, delete one → `shouldCloseSession(ANDROID, "emulator-5554")` is `false`.

maestro-cli/src/test/kotlin/maestro/cli/session/SessionStoreTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    fun `activeSessionsForDevice only returns sessions for that device`() {
+        sessionStore.heartbeat("s1", Platform.IOS, "device-A")
+        sessionStore.heartbeat("s2", Platform.ANDROID, "emulator-5554")
+        sessionStore.heartbeat("s3", Platform.IOS, "device-B")
+
+        val deviceASessions = sessionStore.activeSessionsForDevice(Platform.IOS, "device-A")
+
+        assertThat(deviceASessions).hasSize(1)
```
proksh, line 107: Can we also validate the key here - 
`assertThat(deviceASessions.single()).isEqualTo(expectedKey)`

For future proofing `activeSessionsForDevice`.

maestro-cli/src/test/kotlin/maestro/cli/session/SessionStoreTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        // Android session on emulator
+        sessionStore.heartbeat("android-session", Platform.ANDROID, "emulator-5554")
+
+        // Android session shuts down
+        sessionStore.delete("android-session", Platform.ANDROID, "emulator-5554")
+
+        // DESIRED: should close because no Android sessions remain on this device
+        val shouldClose = sessionStore.shouldCloseSession(Platform.ANDROID, "emulator-5554")
+        assertThat(shouldClose).isTrue()
+    }
```
proksh, line 120: Can we add a test for `pruneInactiveSessions`?

proksh (commented):
> One concern with the {port, pid, startedAt} JSON shape you proposed: PID liveness checks aren't cross-platform. ProcessHandle.of(pid) is Java 9+ (Maestro targets Java 8), and there's no reliable equivalent on Windows without JNA. The port probe alone is sufficient — if the port responds to /status, the runner is alive regardless of PID. If it doesn't, start fresh.

> So the question is whether this per-device port file (already in https://github.com/mobile-dev-inc/Maestro/pull/3167) can fully replace SessionStore, or whether there are cases where the heartbeat-based shared state is still needed. I believe it can replace it — the only cross-process coordination SessionStore provides is:
> 
> Warm reuse: "is a driver already running on this device?" → port probe handles this
> Shutdown coordination: "should I close this driver?" → per-device file check handles this

Got it on PID / Java 8 / port probe + `/status`. That makes sense.

Re: replacing `SessionStore` with the per-device port approach in #3167: I’m aligned that warm reuse and shutdown hints can be driven by probe + on-disk port per deviceId, without heartbeats, as long as we define the shutdown cases clearly.

> Two paths forward:
> 
> Option A: Merge this PR as-is (fixes the bugs in the existing SessionStore), then land https://github.com/mobile-dev-inc/Maestro/pull/3167 next and use it as the foundation to refactor SessionStore out in a follow-up
> Option B: Collapse https://github.com/mobile-dev-inc/Maestro/pull/3138 and https://github.com/mobile-dev-inc/Maestro/pull/3167 into a single PR that replaces SessionStore with per-device port files from the start
> I'd lean toward A — ship the bug fixes now, iterate on the architecture next — but happy to go either way.

Lets go with option A. Prefer not to combine #3167 with this PR. This solution is already fixing a lot of edge cases we have.

Left a few minor comments.

proksh:
> I run maestro hierarchy against a device connected to Studio. Is that 2 sessions?

To answer this more clearly @Fishbowler:

Yes - two newSessions (two processes). Second one may reuse the same XCTest runner by skipping driver.open() when connectToExistingSession is set. So two sessions, often one driver. Correct me if I am wrong @qwertey6

Fishbowler:
Thanks! I'd certainly like that capability to be retained.

proksh:
@qwertey6 Let me know if you need anything from us.

qwertey6:
@proksh — addressed all four comments:

1. **`deviceId` non-null**: Removed all `String? = null` defaults from `SessionStore`. The `"unknown"` fallback is gone. When no device ID is available (e.g., `--host` mode), `MaestroSessionManager` falls back to the session UUID as a unique device key.

2. **`shouldCloseSession` false case**: Added test — two sessions on `emulator-5554`, delete one, `shouldCloseSession` returns `false`.

3. **Key validation in `activeSessionsForDevice`**: Added `assertThat(deviceASessions.single()).isEqualTo("IOS_device-A_s1")`.

4. **`pruneInactiveSessions` test**: Added — writes a stale entry (22s old) directly to KV store, triggers heartbeat (which prunes internally), verifies only the fresh entry remains.

All pushed.

proksh:
@qwertey6 Nice work!!

Few final things before we can merge:
1. Can you merge with the latest main so that PR is updated with latest binaries?
2. Have you tested and validated `--shard-split` works locally?
3. Have you verified `removeOldLogs()` works after change of `${date}_$pid`?

qwertey6:
@proksh — all three items verified:

**1. Rebased onto latest `main`** (816498ce) — pushed with latest binaries.

**2. `--shard-split` tested locally** — 2 flows across 2 iOS simulators, both passed:
```
Passed: 2/2
[ iPhone 17 Pro - D474FC6E ] - 1/1
[ shard_test - 170DEBAD ] - 1/1
```

**3. `removeOldLogs()` verified** — PID-suffixed dirs (`2026-05-07_144839_71026`) sort correctly by name since PID is after the timestamp. `KEEP_LOG_COUNT=6` prune logic works unchanged.

---

One caveat: all of my testing has been on macOS (Apple Silicon). I haven't tested on Linux or Windows. The code is pure Java/Kotlin (no platform-specific calls — we specifically avoided PID checks for this reason), so it should work, but I don't have a way to verify that in my local setup.

It might be worth having a CI matrix that covers these platforms for session/port management code — right now the E2E suite only runs on macOS runners. Given how much cross-process state Maestro manages (`~/.maestro/sessions`, port forwarding, driver lifecycle), platform-specific edge cases in file locking or process behavior could easily slip through. Happy to help think through what that test matrix would look like if it's useful.

proksh (approved):

qwertey6:
Thanks @proksh for the thorough review — learned a lot from the back and forth on `SessionStore` architecture.

Once this is merged, the next PRs in the stack are ready for review:

- **#3139** (performance) — `reinstallDriver=false` default + `isChannelAlive()` skip + build product caching. This is the 4-5x startup speedup (52s→10s).
- **#3140** (version safety) — SHA-256 hash checks so the driver reuse in #3139 doesn't break after Maestro upgrades.

And the independent ones (no dependencies, can merge in any order):

- **#3166** — 3-line fix for the IPv4/IPv6 XCTest connection mismatch (#1299, open since 2023)
- **#3165** — 10-line fix for `tapOn` failing with `accessibilityLabel` on iOS (#1409, open 2.5 years)
- **#3141** — `waitToSettleTimeoutMs` actually controlling the iOS settle budget (53s→25s for tuned flows)
- **#3167** — Per-device port persistence to stop orphaning `xcodebuild` processes. This is the per-device file approach you suggested during review — I'd already implemented it before you proposed it, so it should align well with where you wanted the architecture to go.

qwertey6:
CI note: the two failures (Unit Test on Java 17, Test on Web) are GitHub infrastructure issues, not code failures:

- **Unit Test**: `fatal: unable to access 'https://github.com/mobile-dev-inc/Maestro/': The requested URL returned error: 500` — GitHub returned 500 during checkout, test runner never reached our code
- **Test on Web**: `403 Forbidden` downloading artifacts — same infrastructure issue

Android E2E and iOS E2E both **pass**. A re-run should clear these.

carl-ps:
How can I tell when this feature has been published? I'm unable to find release notes for Maestro Studio 0.9.2 to see if it was included in that release.

PR #2682, opened 2025-08-31
Add --skip-driver-app and --skip-server-app CLI flags

vibin:
Adds opt-in CLI flags to skip installation/uninstallation of Maestro driver and server Android apps.

- `--skip-driver-app`: Skips dev.mobile.maestro app install/uninstall  
- `--skip-server-app`: Skips dev.mobile.maestro.test app install/uninstall

## Proposed changes

copilot:summary

## Testing

Ran `maestro test` with and without the new flags

```bash
./maestro-cli/build/install/maestro/bin/maestro test twitter.yaml
./maestro-cli/build/install/maestro/bin/maestro test twitter.yaml --skip-driver-app --skip-server-app
```
When --skip-driver-app and --skip-server-app options are passed, the firstInstallTime of the packages don't change:
```
adb shell dumpsys package dev.mobile.maestro | grep 'firstInstallTime'
adb shell dumpsys package dev.mobile.maestro.test | grep 'firstInstallTime'
```
## Issues fixed

Fixes #2340 - helps avoid Android's app installation limits on real devices by preserving already installed driver and server apps between test runs.

PR #3139, opened 2026-04-04
perf: 4x faster startup — skip unnecessary driver reinstall, cache build products

### LINKED ISSUE #1096: [Feature Request] Make Maestro faster by giving the option to swip APK install/uninstall

qwertey6:
## Proposed changes

**~4x faster startup time.** Every `maestro test` invocation was spending ~40-45 seconds killing and reinstalling the XCTest driver, even when the exact same driver was already running and healthy. This PR makes driver reuse the default.

### Impact

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| Single device, warm | ~52s | ~10s | **5.2x faster** |
| Single device, cold | ~52s | ~12s | **4.3x faster** |
| 3 devices parallel | ~54s | ~15s | **3.6x faster** |

- **Resolves #1096** — "Make Maestro faster by giving the option to skip helper APKs install/uninstall" (open since 2023)
- **Builds on #2682** — flips `--reinstall-driver` default from `true` to `false`

### Root cause of the slowness

1. **`reinstallDriver=true` was the default**: Every run killed the XCTest runner, extracted build products from the JAR, ran `xcodebuild test-without-building`, and polled for ~40s until the runner responded.
2. **The `isChannelAlive()` optimization was dead code**: `startXCTestRunner()` checks if the runner is already alive, but `restartXCTestRunner()` called `uninstall()` first — killing the runner before the alive check could ever succeed.
3. **Build products extracted to temp dir every run**: ~2-3s spent unzipping the same unchanged binaries from the JAR on every invocation.
4. **100-second HTTP read timeout on status check**: The polling loop used a 100s read timeout for a health check that should respond instantly. If the server accepted a connection but hung, one poll iteration blocked for 100s.

### Changes

- **Default `reinstallDriver` to `false`** across all entry points (TestCommand, PrintHierarchyCommand, MaestroSessionManager, AndroidDriver, LocalXCTestInstaller, XCTestDriverClient)
- **Check `isChannelAlive()` before reinstalling**: `restartXCTestRunner()` now checks if the runner is responding before deciding to start fresh. Explicit `--reinstall-driver` still forces a full reinstall.
- **Cache build products per-device**: Extracted to `~/.maestro/build-products/<deviceId>/` with SHA-256 hash of JAR source. Cache is validated on each run — skips extraction when hash matches, re-extracts automatically on Maestro upgrade or local rebuild.
- **Reduce status check timeout**: 100s → 3s. The `/status` endpoint should respond instantly if healthy.

### Verification

| Scenario | Result | Time |
|----------|--------|------|
| Cold start, no driver, no cache | Pass | 19s |
| Warm, driver still running | Pass | 10s |
| Explicit `--reinstall-driver` | Pass | 25s |
| Driver killed externally, recover | Pass | 10s |
| 3 devices parallel, cold | Pass | 15s |
| 3 devices parallel, warm | Pass | 11s |

## Testing

Manually verified on iOS simulators and Android emulator. All existing tests pass.

> **Depends on**: #3138 (parallel iOS execution)
> **Next**: PR 3 adds version safety (hash checks to prevent stale drivers after Maestro upgrade)

## Issues fixed

Closes #1096


Fishbowler:
I've tried to rebase this, but having permissions problems, despite this:
image

There's no need to commit the driver binaries any longer, which is the only remaining change in the base commit. That can be skipped, leaving 1 clean commit rebased on main.

PR #3145, opened 2026-04-05
Add maestroServer to sourcesJar

Fishbowler:
## Proposed changes

Running `:assemble` gives

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':maestro-client:sourcesJar' (type 'Jar').
  - Gradle detected a problem with the following location: '/Users/danc/git/maestro/Maestro/maestro-client/src/main/resources'.
    
    Reason: Task ':maestro-client:sourcesJar' uses this output of task ':maestro-android:copyMaestroServer' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

iOS similarly errors. This comes from changes in #3082

This change explicitly links the required bits together to ensure the correct dependency chain is observed.

I'm not 100% clear why the gradle change necessitates a new driver build...

## Testing

`./gradlew :maestro-cli:installDist`

## Issues fixed

n/a


amanjeetsingh150 (commented):
Wait I'm not quiet sure about the change, what is not working. 

Btw, assemble task exists in maestro-android and not in client so to create new android drivers we need to do 

```
./gradlew clean :maestro-android:assemble
./gradlew clean :maestro-android:assembleAndroidTest
```

Fishbowler:
I can't reproduce the problem that this was solving either 🤷

PR #3147, opened 2026-04-06
Improvements on android E2E job

amanjeetsingh150:
## Proposed changes

1. Remove setOrientation, wikipedia
2. Better emulator to run faster
3. Removed screen recording anyways not working
4. Pull request template
5. Enabled build caching on flutter app dropping 5-10 minutes on average

## Testing

- [x] On E2E pipeline 


.github/workflows/test-e2e.yaml
```diff
@@ -226,33 +213,15 @@ jobs:
...
       - name: Install apps
         working-directory: ${{ github.workspace }}/e2e
         run: ./install_apps android
 
-
-
-      - name: Start screen recording of AVD
-        run: |
-          adb shell screenrecord /sdcard/screenrecord.mp4 &
-          echo $! > ~/screenrecord.pid
```
Fishbowler, line 242: I was using this somewhat recently. It's useful for debugging.

.github/workflows/test-e2e.yaml
```diff
@@ -195,22 +176,28 @@ jobs:
...
+          cat << EOF >> "$ANDROID_AVD_HOME/MyAVD.avd/config.ini"
+          hw.cpu.ncore=4
+          hw.gpu.enabled=no
+          hw.gpu.mode=auto
+          hw.ramSize=4192
+          disk.dataPartition.size=6442450944
+          vm.heapSize=112M
+          hw.lcd.density=240
+          hw.lcd.height=1280
+          hw.lcd.width=606
```
Fishbowler, line 228: What's your confidence level in these values?

.github/workflows/test-e2e.yaml
```diff
@@ -124,28 +116,17 @@ jobs:
         uses: actions/setup-java@v5
         with:
           distribution: zulu
-          java-version: 21
-
-      - name: Set up mobile-dev-inc/bartek-scripts (for install_android_sdk script)
-        run: |
-          git clone https://github.com/mobile-dev-inc/bartek-scripts.git $HOME/scripts
```
Fishbowler, line 131: Are there other places in the repo where we should drop this? I nearly inlined this script usage last week.

.github/workflows/test-e2e.yaml
```diff
@@ -124,28 +116,17 @@ jobs:
...
-          echo "$HOME/scripts/bin" >> $GITHUB_PATH
+          java-version: 17
 
-      - name: Set up Android Command-line Tools
-        run: |
-          # v13 - see https://stackoverflow.com/a/78890086/7009800
-          install_android_sdk https://dl.google.com/android/repository/commandlinetools-linux-12266719_latest.zip
-          echo "$ANDROID_HOME/cmdline-tools/latest/bin:$PATH" >> $GITHUB_PATH
+      - name: Set up Android SDK
+        uses: android-actions/setup-android@v3
```
Fishbowler: Does switching to the commodity action lose our ability to choose our Tools version?

.github/workflows/test-e2e.yaml
amanjeetsingh150: Is it even working? I never saw this giving recordings Only iOS does.

.github/workflows/test-e2e.yaml
amanjeetsingh150: Ran it 4-5 times all passing and these are same values we use on cloud to keep tests source of truth

.github/workflows/test-e2e.yaml
amanjeetsingh150: Just this one should i remove this? What should I replace this with?

.github/workflows/test-e2e.yaml
amanjeetsingh150: I have reverted this, I was trying something.

.github/workflows/test-e2e.yaml
Fishbowler: It definitely used to. Happy to lose it for now, and add it back later if needed via `startRecording` or `maestro record`.

.github/workflows/test-e2e.yaml
```diff
@@ -147,6 +147,27 @@ jobs:
           sdkmanager --install "$ANDROID_PLATFORM"
           sdkmanager --install "$ANDROID_OS_IMAGE"
 
+      - name: Cache pub
+        uses: actions/cache@v4
```
Fishbowler:
```suggestion
        uses: actions/cache@v5
```

.github/workflows/test-e2e.yaml
```diff
@@ -147,6 +147,27 @@ jobs:
...
+        with:
+          path: |
+            ~/.pub-cache
+            e2e/demo_app/.dart_tool
+          key: pub-${{ runner.os }}-${{ hashFiles('e2e/demo_app/pubspec.lock') }}
+          restore-keys: |
+            pub-${{ runner.os }}-
+
+      - name: Cache Gradle
+        uses: actions/cache@v4
```
Fishbowler, line 161:
```suggestion
        uses: actions/cache@v5
```

.github/workflows/test-e2e.yaml
```diff
@@ -196,21 +217,21 @@ jobs:
...
+          hw.lcd.width=606
           EOF
 
       - name: Run AVD
         run: |
           emulator @MyAVD \
-            -verbose -no-snapshot -no-window -no-audio -no-boot-anim -accel on -camera-back none -qemu -m 3072 \
+            -verbose -no-snapshot -no-window -no-audio -no-boot-anim -accel on -no-metrics -camera-back emulated \
             >~/emulator_stdout.log \
             2>~/emulator_stderr.log &
```
Fishbowler, line 236: Why isn't this done via `maestro start-device`? Is this an oversight, a deficiency in our tooling, or a specific intent not to confuse what's being tested?

Fishbowler (approved):
Approved, but with minor nitpicks below

.github/workflows/test-e2e.yaml
amanjeetsingh150: That's definitely oversight. 

Actually you raised a very good point all of this should be part of implementation of maestro start-device as well. 

Ideally in e2e we should just do maestro start-device and this logic should not be even here.

Here is my suggestion: we merge this PR, decide scope of this offline and finally e2e just uses maestro start-device.

Fishbowler: Agree!

PR #3150, opened 2026-04-07
Skip additional config files in workspace execution planner

proksh:
When running a workspace, Maestro needs to distinguish between flow files (test scripts) and config files (workspace settings like platform configuration, tag filters, etc.). Previously, config files were only detected by filename - checking for exact match `config.yaml`. This meant any config file other than "config" name (e.g. platform_settings.yaml) would be incorrectly treated as a flow file and fail to parse.  [Maestro docs](https://docs.maestro.dev/maestro-flows/workspace-management/project-configuration#working-with-multiple-configs)

This PR replaces the filename-based heuristic with content-based detection. It reads the top-level YAML keys of each file and checks if any match known `WorkspaceConfig` properties (platform, includeTags, excludeTags, flows, etc.). The set of known keys is derived via reflection from the WorkspaceConfig data class, so it automatically stays in sync as new config fields are added - no hardcoded lists to maintain.
  
  
Test
  - 018 - Additional config files in workspace are not treated as flows

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
@@ -1,23 +1,12 @@
 package maestro.orchestra.workspace
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
-import maestro.orchestra.WorkspaceConfig
+import maestro.orchestra.yaml.YamlCommandReader
 import java.nio.file.Path
 import kotlin.io.path.absolutePathString
 import kotlin.io.path.extension
 import kotlin.io.path.isRegularFile
 import kotlin.io.path.nameWithoutExtension
 import kotlin.io.path.readText
-import kotlin.reflect.full.declaredMemberProperties
-
-private val WORKSPACE_CONFIG_KEYS: Set<String> by lazy {
-    WorkspaceConfig::class.declaredMemberProperties.map { it.name }.toSet()
-}
-
-private val YAML_MAPPER by lazy {
-    ObjectMapper(YAMLFactory())
-}
 
 fun isFlowFile(path: Path, config: Path?): Boolean {
     if (!path.isRegularFile()) return false // Not a file
@@ -33,8 +22,7 @@ private fun isWorkspaceConfigFile(path: Path): Boolean {
     return try {
         val content = path.readText()
         if (content.contains("\n---")) return false // Flow files have a document separator
-        val topLevelKeys = YAML_MAPPER.readValue(content, Map::class.java)?.keys ?: return false
-        topLevelKeys.all { it in WORKSPACE_CONFIG_KEYS }
+        YamlCommandReader.findUnknownWorkspaceConfigKeys(content)?.isEmpty() ?: false
     } catch (e: Exception) {
         false
     }
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
@@ -32,8 +32,9 @@ fun isFlowFile(path: Path, config: Path?): Boolean {
 private fun isWorkspaceConfigFile(path: Path): Boolean {
     return try {
         val content = path.readText()
+        if (content.contains("\n---")) return false // Flow files have a document separator
         val topLevelKeys = YAML_MAPPER.readValue(content, Map::class.java)?.keys ?: return false
-        topLevelKeys.any { it in WORKSPACE_CONFIG_KEYS }
+        topLevelKeys.all { it in WORKSPACE_CONFIG_KEYS }
     } catch (e: Exception) {
         false
     }
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
@@ -12,5 +12,6 @@ fun isFlowFile(path: Path, config: Path?): Boolean {
     val extension = path.extension
     if (extension != "yaml" && extension != "yml") return false // Not YAML
     if (path.nameWithoutExtension == "config") return false // Config file
+    if (path.nameWithoutExtension.endsWith("config")) return false // Additional config file (e.g. regression_config.yaml, ci-config.yaml)
     return true
 }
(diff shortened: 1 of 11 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
```diff
@@ -12,5 +12,7 @@ fun isFlowFile(path: Path, config: Path?): Boolean {
     val extension = path.extension
     if (extension != "yaml" && extension != "yml") return false // Not YAML
     if (path.nameWithoutExtension == "config") return false // Config file
+    if (path.nameWithoutExtension.endsWith("_config")) return false // Additional config file (e.g. regression_config.yaml)
+    if (path.nameWithoutExtension.endsWith("-config")) return false // Additional config file (e.g. regression-config.yaml)
```
amanjeetsingh150: Wait I think I'm getting confused did the user had -config extra option and its getting ignored right now?

The PR description doesn't explain the problem just feels like solution to something.

Assuming that's the issue (config option getting ignored), I think it can actually be any yaml file. Docs just mention as an example: ci-config.yaml but it can very well be different name right?

Not necessarily suffixed with either of _config or -config

That would mean whatever is configured with that option should not be treated as flow.

proksh: Nice points, I have updated the code and PR description. Can you take another look?

Fishbowler (approved):

maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
```diff
@@ -1,16 +1,40 @@
...
 import kotlin.io.path.nameWithoutExtension
+import kotlin.io.path.readText
+import kotlin.reflect.full.declaredMemberProperties
+
+private val WORKSPACE_CONFIG_KEYS: Set<String> by lazy {
+    WorkspaceConfig::class.declaredMemberProperties.map { it.name }.toSet()
+}
+
+private val YAML_MAPPER by lazy {
+    ObjectMapper(YAMLFactory())
```
amanjeetsingh150: Few questions:

- Does WorkspaceExecutionPlanner already gets input with configurable config in plan function?

Here is what I'm thinking:

* Filters class should just be aware of one resolved path config, right now from what I see it just takes the user specified right custom config instead of resolved one. CMIIW?  

* Then second part we can answer do we need this validation logic on CLI or not? Not sure even if it exists.

amanjeetsingh150: Here is another interesting PR: https://github.com/mobile-dev-inc/Maestro/pull/3156

That would also need a design you are working on. So, this requirement gets more generic:

- Default takes config.yaml
- If specified --config then it takes that
- If user has their own yaml file which they want to not treat as flows. I would be curious though for this what are the examples.

proksh: > Does WorkspaceExecutionPlanner already gets input with configurable config in plan function?

Yes. The `plan()` function already accepts `config: Path?` (WorkspaceExecutionPlanner.kt). The resolution logic is:
  1. If --config is explicitly provided → use that path
  2. If no config and directories are passed → auto-discover config.yaml/config.yml via findConfigFile()
  3. Fallback → empty WorkspaceConfig()

The resolved config path is already passed into `isFlowFile(path, config)` in `Filters.kt`, where it's used for exact path comparison.

> Filters class should just be aware of one resolved path config, right now from what I see it just takes the user specified right custom config instead of resolved one. CMIIW?

You're partially right. The config parameter passed to `isFlowFile` is the raw `config: Path?` from the plan() function arguments - which is whatever the CLI gave it (i.e., the --config flag value or null).

However, when no --config is provided, the planner auto-discovers config via findConfigFile() and stores the result only in workspaceConfig (the parsed object) - but the original config: Path? remains null. So when `isFlowFile(path, config=null)` is called, the path comparison doesn't exclude the auto-discovered config file. It only gets caught by the nameWithoutExtension == "config" check.

This means: if the auto-discovered config were named something other than config.yaml, the path-based exclusion in isFlowFile would miss it. The current code works by accident for the default case because of the name-based check.

> Then second part we can answer do we need this validation logic on CLI or not? Not sure even if it exists.

The CLI only validates that --config file exists (TestCommand.kt:263-265). There's no content validation at the CLI level. The content-based detection in this PR is at the Filters.kt level, which is the right layer — CLI shouldn't need to know about config file content detection.

proksh: > Here is another interesting PR: https://github.com/mobile-dev-inc/Maestro/pull/3156
> 
> That would also need a design you are working on. So, this requirement gets more generic:
> 
> Default takes config.yaml
> If specified --config then it takes that
> If user has their own yaml file which they want to not treat as flows. I would be curious though for this what are the examples.

These two PRs solve different problems and don't overlap:

  - This PR (#3150): I have multiple Maestro config files (regression_config.yaml, platform_settings.yaml) that I switch between with --config. The inactive ones shouldn't be picked up as flows.
  - PR #3156: I have non-Maestro YAML files in my workspace (CI configs, app configs, test fixtures, docker-compose, etc.) that I want to explicitly exclude.

Both are needed. #3150 is about Maestro recognizing its own config files automatically. #3156 is a user-controlled escape hatch for everything else.

Yes, users could work around #3150 by adding `!regression_config.yaml` to their config if #3156 lands, but that's fragile - you'd have to remember to exclude every new config file manually.

maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
```diff
@@ -1,16 +1,41 @@
...
     if (extension != "yaml" && extension != "yml") return false // Not YAML
     if (path.nameWithoutExtension == "config") return false // Config file
-    return true
-}
\ No newline at end of file
+
+    return !isWorkspaceConfigFile(path)
+}
+
+private fun isWorkspaceConfigFile(path: Path): Boolean {
```
amanjeetsingh150, line 21: I like this check and the fact the custom configs would go through this to establish a valid config contract.

I'm not sure if we do this with default one as well, if not that's good addition.

What happens today if someone has bad contract on config.yaml? We can use the same logic for that as well.

Approving with this thought^

amanjeetsingh150 (approved):

maestro-orchestra/src/main/java/maestro/orchestra/workspace/Filters.kt
proksh: Added validation for config as well, and warning will be thrown.

PR #3156, opened 2026-04-08
feat: support negation globs in config.yaml

btrautmann:
## Proposed changes

Adds support for negation globs in `config.yaml` to support excluding specific globs before Maestro processes them as flows. This is helpful when you have other `.yaml` (similar to `config.yaml`) files in our workspace but don't want them to be treated as flows, as it avoids needing to either restructure your project (add a nested directory for your workspace) or writing custom code to generate a `config.yaml` without the excluded file at test time.

copilot:summary

## Testing

Added fixtures and tests to exercise the new logic.


maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -68,12 +68,18 @@ object WorkspaceExecutionPlanner {
 
         val globs = workspaceConfig.flows ?: listOf("*")
 
-        val matchers = globs.flatMap { glob ->
+        val positiveGlobs = globs.filter { !it.startsWith("!") }
+        val negativeGlobs = globs.filter { it.startsWith("!") }.map { it.removePrefix("!") }
+
+        val positiveMatchers = positiveGlobs.flatMap { glob ->
```
proksh: Nit:
The duplicated matcher-creation logic could be extracted:

// Before (repeated twice):
val positiveMatchers = positiveGlobs.flatMap { glob ->
    directories.map { it.fileSystem.getPathMatcher(...) }
}
val negativeMatchers = negativeGlobs.flatMap { glob ->
    directories.map { it.fileSystem.getPathMatcher(...) }
}

// After:
fun buildMatchers(globs: List<String>) = globs.flatMap { glob ->
    directories.map { it.fileSystem.getPathMatcher(...) }
}

maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -68,12 +68,18 @@ object WorkspaceExecutionPlanner {
...
+            directories.map { it.fileSystem.getPathMatcher(escapeSlashesForWindows("glob:${it.pathString}${it.fileSystem.separator}$glob")) }
+        }
+        val negativeMatchers = negativeGlobs.flatMap { glob ->
             directories.map { it.fileSystem.getPathMatcher(escapeSlashesForWindows("glob:${it.pathString}${it.fileSystem.separator}$glob")) }
         }
 
         val unsortedFlowFiles = flowFiles + flowFilesInDirs.filter { path ->
-            matchers.any { matcher -> matcher.matches(path) }
+            positiveMatchers.any { it.matches(path) } && negativeMatchers.none { it.matches(path) }
```
proksh, line 99: if negation globs are used and the result is empty, this error message prints all globs (including !-prefixed ones) under "Flow inclusion pattern(s) did not match", which is slightly misleading since negation patterns are exclusions, not inclusions.

proksh (approved):
Nice.

It will be good if you can test "all negation, no positive" case for flow selection as well.

Fishbowler:
What's the expected outcome? Zero flows?

proksh:
> What's the expected outcome? Zero flows?

I was thinking just excluding the mentioned exclusions and running all remaining flows. But you have more context of how things should work, and if Zero flows make more sense, lets go with it.

btrautmann:
@Fishbowler @proksh lmk if you'd like me to make any further changes, thanks!

Fishbowler:
I think a `flows` key that doesn't have an explicit inclusion includes nothing, as it overrides the default of `*`

Fishbowler:
On the "excluding everything", there's a few things we could do:

1. If there are only negative matchers, infer that includes `*`
2. Validate and error specifically on this case
3. Run normally, but include specific extra messaging in the `if (unsortedFlowFiles.isEmpty())` branch if they've used (or only used?) negative matchers
4. Run normally, and _always_ include a message about negative matchers in the `if (unsortedFlowFiles.isEmpty())` branch
5. Run normally, and rely on the user to work out what they've done wrong.

I think I prefer 2 or 3. WDYT?

btrautmann:
@Fishbowler agreed, I liked number 2 as it's the most clear.

Updated that in the last commit!

simon-gilmurray:
This will also need a corresponding Maestro Docs PR ready for release with the updated functionality 🙏

PR #3157, opened 2026-04-08
fix: Orchestra properly respects coroutine cancellation

steviec:
## Problem

Orchestra has unexpected and buggy coroutine cancellation behavior. It was originally implemented in [this PR](https://github.com/mobile-dev-inc/Maestro/pull/2488/changes#diff-de990d27a0c12277340b89affbcbe8a8edd03e1c3ab62a36e3c6e5b55b6f17f9), but was never fully implemented.  

Here are the problems with it:

- **`executeCommands`** iterates all remaining commands on cancellation, calling `onCommandSkipped` for each (could be hundreds). This was by design in the original PR, but the behavior is very confusing.
- **`executeSubflowCommands`** fires `onCommandStart` callbacks with no cancellation check
- **`scrollUntilVisible`** has a blocking do-while loop with no cancellation check — scrolls forever after cancellation
- **`retryCommand`** retries indefinitely even after cancellation
- **`catch (Throwable)`** blocks in 5 locations silently swallow `CancellationException`
- **`onFlowComplete`** runs in finally blocks even during cancellation, which can hang

## Fix

- Add `yield()` at 4 key checkpoints (command loops, scroll loop, repeat loop, subflow loop) — each is either the only cancellation point in a blocking loop, or prevents post-cancellation callbacks
- Add `catch (CancellationException) { throw e }` before all `catch (Throwable)` blocks
- Guard `onFlowComplete` in `runFlow()` and `runSubFlow()` finally blocks with `isActive`

## What this enables

Callers can now reliably use `withTimeout` to set a deadline on `orchestra.runFlow()`:

```kotlin
withTimeout(15.minutes) {
    orchestra.runFlow(commands)
}
```

Cancellation propagates immediately — no lingering loops, no post-cancellation callbacks, no `onFlowComplete` hanging. This supersedes #3149.

NOTE: This does not solve the problem of blocking IO calls. This only solves cancellation during places we have control over in Orchestra. Blocking IO calls will be improved in a follow-up PR.

## Test plan

- [x] Case 140: `scrollUntilVisible` stops within 2s when coroutine is cancelled (was infinite)
- [x] Case 141: `retryCommand` stops within 2s when coroutine is cancelled (was infinite)
- [x] Case 142: No `onCommandStart` callbacks fire after cancellation
- [x] Cases 121, 124, 132: Updated to expect immediate `CancellationException` instead of skip-based behavior
- [x] Full integration suite (199 tests) passes

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -232,6 +232,7 @@ class Orchestra(
             initJsEngine(config)
         }
 
+        yield()
         initAndroidChromeDevTools(config)
 
         commands
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -823,7 +823,7 @@ class Orchestra(
                 ?.let { evaluateCondition(it, commandOptional = command.optional) } != false
         }
 
-        while (checkCondition() && counter < maxRuns && currentCoroutineContext().isActive) {
+        while (checkCondition() && counter < maxRuns) {
             yield()
             if (counter > 0) {
                 command.commands.forEach { resetCommand(it) }
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -19,8 +19,10 @@
 
 package maestro.orchestra
 
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.isActive
+import kotlinx.coroutines.yield
 import maestro.Driver
 import maestro.ElementFilter
 import maestro.Filters
@@ -196,16 +198,22 @@ class Orchestra(
                     screenRecording?.close()
                 }
             }
+        } catch (e: CancellationException) {
+            throw e
         } catch (e: Throwable) {
             exception = e
         } finally {
-            val onCompleteSuccess = config?.onFlowComplete?.commands?.let {
-                executeCommands(
-                    commands = it,
-                    config = config,
-                    shouldReinitJsEngine = false,
-                )
-            } ?: true
+            val onCompleteSuccess = if (currentCoroutineContext().isActive) {
+                config?.onFlowComplete?.commands?.let {
+                    executeCommands(
+                        commands = it,
+                        config = config,
+                        shouldReinitJsEngine = false,
+                    )
+                } ?: true
+            } else {
+                true
+            }
 
             jsEngine.close()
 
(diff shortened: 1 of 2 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -224,19 +232,11 @@ class Orchestra(
             initJsEngine(config)
         }
 
-        if (!currentCoroutineContext().isActive) {
-            logger.info("Flow cancelled, skipping initAndroidChromeDevTools...")
-        } else {
-            initAndroidChromeDevTools(config)
-        }
+        initAndroidChromeDevTools(config)
```
amanjeetsingh150, line 236: Hmmm, before we use to skip execution if coroutine got cancelled while setup here. What happens now?

Is this setup intentionally executed anyways now?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1051,12 +1056,18 @@ class Orchestra(
                 if (onStartSuccess) {
                     flowSuccess = executeSubflowCommands(filteredCommands, config)
                 }
+            } catch (e: CancellationException) {
+                throw e
```
amanjeetsingh150, line 1061: nit: Should we have reason attached to this cancellation? It gives more semantical meaning while debugging.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1051,12 +1056,18 @@ class Orchestra(
...
             } catch (e: Throwable) {
```
amanjeetsingh150, line 1062: One more question I see in integration test we are asserting exception TimeoutCancellationException but here its re-throwing the super CancellationException.

I think TimeoutCancellationException extends CancellationException right? 

Is it good idea to just catch just that? there might be other types that would extend CancellationException. 

If some other type of CancellationException is thrown that might be bug, which we don't want to prematurely catch here.

amanjeetsingh150 (commented):
Few others:

- Is the flow controller legacy like we discussed? Can we get rid of it if no client is using it.
- I feel there should be global timeout the runner should call orchestra with (we can just keep it 15 minutes same as cloud). That can be configurable: https://docs.maestro.dev/maestro-flows/workspace-management/project-configuration. Do we plan to do that as well in follow up PRs?

amanjeetsingh150 (approved):

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
steviec: good point, fixed

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
steviec: since this is a re-throw, it would require wrapping this in a new error right? I think it should just preserve whatever the original exception was.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
steviec: I just checked, and this is the standard approach. The point of this rethrow is to ensure that we don't make it into the Throwable block; this MUST be rethrown to propagate up the call stack. And we want to do that for al CancellableException errors, not just Timeouts.

steviec:
> Few others:
> 
> * Is the flow controller legacy like we discussed? Can we get rid of it if no client is using it.

We can get rid, but I'd rather do in a separate PR.

> * I feel there should be global timeout the runner should call orchestra with (we can just keep it 15 minutes same as cloud). That can be configurable: https://docs.maestro.dev/maestro-flows/workspace-management/project-configuration. Do we plan to do that as well in follow up PRs?

Yes and no :). See the follow-up PR that makes Maestro honor coroutine cancellations. I think that's the better solve.

PR #3159, opened 2026-04-09
fix: resolve config path before filtering flow files

proksh:
Previously, when no `--config` flag was passed, `findConfigFile()` was called after `isFlowFile()` had already filtered flows with `config=null`. This meant the auto-discovered config.yaml was never excluded by path comparison — it only worked because of a hardcoded name check (`nameWithoutExtension == "config"`).

This PR moves `findConfigFile()` before the filtering step, so `isFlowFile` always receives the resolved config path.

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
@@ -46,10 +46,14 @@ object WorkspaceExecutionPlanner {
 
         val (files, directories) = input.partition { it.isRegularFile() }
 
-        val flowFiles = files.filter { isFlowFile(it, config) }
+        // Resolve config path before filtering, so auto-discovered configs are also excluded
+        val resolvedConfigPath = config?.absolute()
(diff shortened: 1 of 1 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceExecutionPlanner.kt
```diff
@@ -46,10 +46,14 @@ object WorkspaceExecutionPlanner {
...
-        val flowFiles = files.filter { isFlowFile(it, config) }
+        // Resolve config path before filtering, so auto-discovered configs are also excluded
+        val resolvedConfigPath = config?.absolute()
+            ?: directories.firstNotNullOfOrNull { findConfigFile(it) }
+
+        val flowFiles = files.filter { isFlowFile(it, resolvedConfigPath) }
         val flowFilesInDirs: List<Path> = directories.flatMap { dir -> Files
             .walk(dir)
-            .filter { isFlowFile(it, config) }
+            .filter { isFlowFile(it, resolvedConfigPath) }
```
amanjeetsingh150, line 56: Ahh love it this is exactly what i was thinking in other PR.

amanjeetsingh150 (approved):

PR #3163, opened 2026-04-09
Detect Azure as CI provider

jkronborg:
## Proposed changes

Detect Azure as CI provider via the [TF_BUILD environment variable](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#:~:text=TF_BUILD).

## Testing

Tested locally by setting the `TF_BUILD` environment variable and seeing that no promotional messages are shown when running `maestro test` (ensuring they aren't hidden for other reasons).

## Issues fixed

Maestro CLI shows promotional messages when used in Azure pipelines.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/util/CiUtils.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/CiUtils.kt
@@ -15,7 +15,7 @@ object CiUtils {
         "GITLAB_CI" to "gitlab",
         "JENKINS_HOME" to "jenkins",
         "TEAMCITY_VERSION" to "teamcity", // since v1.37.4
-        "TF_BUILD" to "azure",
+        "TF_BUILD" to "azure", // since v2.5.0
         "CI" to "ci"
     )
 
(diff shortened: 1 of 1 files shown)


maestro-cli/src/main/java/maestro/cli/util/CiUtils.kt
```diff
@@ -15,6 +15,7 @@ object CiUtils {
         "GITLAB_CI" to "gitlab",
         "JENKINS_HOME" to "jenkins",
         "TEAMCITY_VERSION" to "teamcity", // since v1.37.4
+        "TF_BUILD" to "azure",
```
Fishbowler: I know it's not always been followed in this list (and I don't know the legacy behind why the comment exists), but let's follow the advice of the comment above and:

> When adding a new CI, also add the first version of Maestro that supports it.

Fishbowler (changes requested):

maestro-cli/src/main/java/maestro/cli/util/CiUtils.kt
jkronborg: Sure, I'm guessing it'll be 2.5.0.

Fishbowler (approved):

PR #3179, opened 2026-04-14
fix: throw correct test error in case of invalid timeout

proksh:
## Proposed changes

- Extract shared `parseTimeoutMs()` helper that validates timeout strings and throws `MaestroException.InvalidCommand` instead of a raw `NumberFormatException`
- Apply to both `AssertConditionCommand.timeoutMs()` and `ScrollUntilVisibleCommand.timeoutToMillis()`
- Add test covering null, valid (with underscores), and invalid timeout parsing


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -28,9 +28,18 @@ import maestro.TapRepeat
 import maestro.js.JsEngine
 import maestro.orchestra.util.Env.evaluateScripts
 import com.fasterxml.jackson.annotation.JsonIgnore
+import maestro.MaestroException
 import java.nio.file.Path
 import net.datafaker.Faker
 
+internal fun parseTimeoutMs(timeout: String): Long {
+    return try {
+        timeout.replace("_", "").toLong()
```
amanjeetsingh150, line 37: The logic here seems to be hard to reason about.

What was the input here that started causing the number format exception, can you help with that context?

It seems like having this timeout in "some string format" is what causing these issues at the first place. Few questions before we commit to a direction:

- Have we seen practical user cases where timeout strings are intentional? Like they are getting resolved from JS? This would help us identify if we should have runtime validation or not.
- Even if its runtime how does this error message helping users in long flows? It's hard to pin point from multiple commands, to check which one had this issue.
- Even if it has to be runtime, I would also think that it should be test error from JS then and not invalid command? What if JS evaluates this as valid in some cases and in some cases its not. Its important we let js execute it when it needs to.

proksh: > Have we seen practical user cases where timeout strings are intentional? Like they are getting resolved from JS? This would help us identify if we should have runtime validation or not.

`timeout` is typed as String specifically so users can interpolate at runtime, e.g. timeout: ${TIMEOUT_MS}. That's still supported here - `AssertConditionCommand.evaluateScripts` and `ScrollUntilVisibleCommand.evaluateScripts` call `timeout.evaluateScripts(jsEngine)` before `timeoutMs()` / `timeoutToMillis()` runs, so `parseTimeoutMs` only ever sees the post-JS value. We're validating the result after JS has had its turn.

The prior code threw a raw `NumberFormatException` from `toLong()`, which surfaced as a Maestro internal crash with a stacktrace rather than a clean test failure. So any non-numeric value hits it. The aim of the PR is to convert that into a structured, reportable test failure instead of a crash.

> Even if its runtime how does this error message helping users in long flows? It's hard to pin point from multiple commands, to check which one had this issue.

I'll update the message to include the command's description() so it's identifiable, e.g.: Invalid timeout value 'abc' in 'Assert visible …'. Timeout must be a number of milliseconds.

> Even if it has to be runtime, I would also think that it should be test error from JS then and not invalid command? What if JS evaluates this as valid in some cases and in some cases its not. Its important we let js execute it when it needs to.

Because validation happens after evaluateScripts, this isn't a JS evaluation failure. JS already ran and returned a string that just isn't a valid millisecond count. If JS itself errors, that path is already handled separately upstream and is unaffected here. Happy to switch to a different MaestroException subtype if you have one in mind that fits better but I don't think it should be a JS engine error specifically.

amanjeetsingh150 (approved):

GGOemea (commented):
Ok

PR #3181, opened 2026-04-15
fix: drop use_fuzzy_matching from tap_on MCP schema

pedro18x:
## Summary

`tap_on` had a `use_fuzzy_matching` boolean parameter. When true (the default), it wrapped the `text` / `id` value in `.*escapeRegex(...).*` before building the `ElementSelector`.

Two problems with that shape:

1. **Schema leak into YAML.** The other fields on the schema (`text`, `id`, `index`, `enabled`, `checked`, `focused`, `selected`) are all real YAML selector keys. `use_fuzzy_matching` had no YAML equivalent. When the model generated flow files instead of calling the tool live, it copied the flag into the `tapOn:` block and produced invalid YAML.
2. **Wrap was more permissive than YAML.** `Filters.textMatches` / `idMatches` use `regex.matches(...)`, which is anchored/full-match. So YAML `tapOn: text: "OK"` only matches elements with text literally `"OK"`, while the MCP wrap (`.*OK.*`) would also match `"OK Button"`, `"Not OK"`, etc. For short strings that's a real footgun.

## Fix

Remove the parameter and remove the wrap. Pass `text` / `id` straight through as regex, matching YAML semantics exactly. If the model wants partial matching it can write `.*foo.*` itself, same as YAML users do.

Fixes [MA-4007](https://linear.app/mobile-dev/issue/MA-4007/fix-incorrect-insertion-of-use-fuzzy-matching-in-tapon-yaml).

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
@@ -24,7 +24,7 @@ object TapOnTool {
                         }
                         putJsonObject("text") {
                             put("type", "string")
-                            put("description", "Regex matching the element text (anchored/full-match; use '.*foo.*' for partial). Typically copied from the 'text' field of inspect_view_hierarchy output.")
+                            put("description", "Regex matching the element text (anchored/full-match; use '.*foo.*' for partial). Often copied from the 'text' field of inspect_view_hierarchy output, but might also be 'accessibilityText' or 'hintText'")
                         }
                         putJsonObject("id") {
                             put("type", "string")
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
@@ -24,19 +24,15 @@ object TapOnTool {
                         }
                         putJsonObject("text") {
                             put("type", "string")
-                            put("description", "Text content to match (from 'text' field in inspect_ui output)")
+                            put("description", "Regex matching the element text (anchored/full-match; use '.*foo.*' for partial). Typically copied from the 'text' field of inspect_view_hierarchy output.")
                         }
                         putJsonObject("id") {
                             put("type", "string")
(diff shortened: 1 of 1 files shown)


Fishbowler:
> always wrap text and id with `.*escapeRegex(...).*`

This gives me some concern that matching will be worse for shorter strings. Most uses of text and ID selectors are exact strings, not wrapped in wildcards.

pedro18x:
> > always wrap text and id with `.*escapeRegex(...).*`
> 
> This gives me some concern that matching will be worse for shorter strings. Most uses of text and ID selectors are exact strings, not wrapped in wildcards.

You're right. `Filters.textMatches` / `idMatches` use `regex.matches()` (anchored), so .*OK.* was matching OK Button where YAML wouldn't. Dropped the wrap; text / id now pass through as regex, matching YAML tapOn exactly.

maestro-cli/src/main/java/maestro/cli/mcp/tools/TapOnTool.kt
```diff
@@ -24,20 +24,16 @@ object TapOnTool {
                         }
                         putJsonObject("text") {
                             put("type", "string")
-                            put("description", "Text content to match (from 'text' field in inspect_ui output)")
+                            put("description", "Regex matching the element text (anchored/full-match; use '.*foo.*' for partial). Typically copied from the 'text' field of inspect_view_hierarchy output.")
```
Fishbowler: If this is used by the agent, should it be explicit about possible sources?


```suggestion
                            put("description", "Regex matching the element text (anchored/full-match; use '.*foo.*' for partial). Often copied from the 'text' field of inspect_view_hierarchy output, but might also be 'accessibilityText' or 'hintText'")
```

Fishbowler (approved):
LGTM. One note below, but could be my misunderstanding.

pedro18x:
> LGTM. One note below, but could be my misunderstanding.

Thanks Dan, great catch. Definitely not a misunderstanding. `accessibilityText` and `hintText` are both matched by `Filters.textMatches`, so the agent needs to know it can source from any of the three. Suggestion committed as-is.

Fishbowler (approved):

PR #3183, opened 2026-04-16
fix: support iOS 26 photo picker in view hierarchy snapshots

StasKeiss:
### Proposed changes 

Add support for iOS 26's photo picker (`PHPickerViewController`) in Maestro's view hierarchy snapshots. A new `getPhotoPickerHierarchy()` method in `ViewHierarchyHandler.swift `fetches the picker
extension's accessibility tree and merges it into the hierarchy response — following the same pattern used for `SFSafariViewController` via `SafariViewService`.

### Impact

Maestro flows that interact with the native iOS photo picker (selecting images from the gallery) will now work on iOS 26+. Previously, the picker's UI elements were completely invisible to Maestro, causing flows
to fail when trying to tap on photos. No impact on iOS versions below 26 (the method early-returns via an OS version guard).

### Root cause

In iOS 26, Apple changed `PHPickerViewController` to run as an out-of-process extension (`com.apple.mobileslideshow.photospicker`) rather than rendering within the host app's process. As a result, the picker's UI
elements no longer appear in the host app's accessibility tree — Maestro sees an empty opaque container instead of buttons, images, and labels. This is the same architectural change Apple made to
`SFSafariViewController` (which now runs in `com.apple.SafariViewService`), and the existing Safari fix in this codebase was the model for this fix.

### Fix

Query the photo picker extension's process directly via `XCUIApplication(bundleIdentifier: "com.apple.mobileslideshow.photospicker"`), check if it's running, snapshot its accessibility hierarchy, and include it
alongside the app hierarchy in the response — identical to how `getSafariWebViewHierarchy()` handles the Safari case.

The docs need to be updated accordingly - https://docs.maestro.dev/examples/recipes/choose-images-from-the-gallery


Fishbowler:
Nice ❤️ 

How would we go about adding an e2e test for this?

StasKeiss:
@Fishbowler You mean add a test to the Flutter app? I don't really know Flutter.

Fishbowler:
I'm working on this. Any chance you can update the PR to allow maintainer pushes? I'd like to rebase.

Fishbowler:
Although... I've just added tests for this to our e2e app and can't reproduce this on `main` any more, nor in the latest Maestro Studio 🤔 


image

(Ignore the unrelated YAML in the background)

Fishbowler:
@StasKeiss Nudge on the above couple of comments

PR #3182, opened 2026-04-16
chore: upgrade MCP Kotlin SDK to official 0.11.1

Leland-Takamine:
## Summary
- Replace the forked kotlin-sdk (`steviec/kotlin-1.8` branch via git source control) with the official `io.modelcontextprotocol:kotlin-sdk:0.11.1` from Maven Central
- The project's Kotlin 2.2 / Java 17 now meets the SDK's requirements, so the fork is no longer needed
- Migrate all MCP tool code to the new API surface
- Exclude Ktor from the MCP SDK to avoid conflicts with the project's Ktor 2.3.6 (the SDK uses Ktor for its HTTP/SSE transport, but we only use stdio)

## API changes
- Types moved from `io.modelcontextprotocol.kotlin.sdk.*` → `io.modelcontextprotocol.kotlin.sdk.types.*`
- `Tool.Input` renamed to `ToolSchema`
- `request.arguments` is now nullable (`JsonObject?`)
- `server.connect(transport)` replaced with `server.createSession(transport)`

## Runtime dependency version changes

| Dependency | Before | After |
|---|---|---|
| **MCP SDK** | forked git branch | `0.11.1` (official, split into core/client/server modules) |
| **kotlin-stdlib-common** | 2.2.20 | 2.3.10 |
| **kotlinx-coroutines** | 1.8.0 | 1.10.2 |
| **kotlinx-io-bytestring** | 0.2.0 | 0.9.0 |
| **kotlinx-io-core** | 0.2.0 | 0.9.0 |
| **kotlinx-serialization-core** | 1.5.0 | 1.10.0 |
| **kotlinx-serialization-json** | 1.5.0 | 1.10.0 |
| **kotlin-logging** | 7.0.0 | 8.0.01 |
| **typesafe config** | 1.4.2 | 1.4.5 |
| **jansi** | 2.4.0 | 2.4.2 |
| **kotlinx-collections-immutable** | — | 0.4.0 |
| **kotlinx-serialization-bom** | — | 1.10.0 |
| **kotlinx-serialization-json-io** | — | 1.10.0 |
| **Ktor** | 2.3.6 | 2.3.6 (unchanged — excluded from MCP SDK) |

## Test plan
- [x] Verify `maestro mcp` starts and responds to tool calls
- [x] Run MCP tool tests (`run_mcp_tool_tests.sh`)
- [x] Verify web driver still works (Ktor `HttpTimeout` regression fixed)

PR #3184, opened 2026-04-16
feat: native Maestro Cloud tools in the MCP server

pedro18x:
## Summary

The Maestro MCP had no native integration with Maestro Cloud. If a user asked the agent to run a flow on Cloud, the model had to think for a while, consult `query_docs`, and piece together a Cloud call from the docs. It usually worked, but the UX was poor: slow, brittle, and easy for the model to get wrong. Two new MCP tools close that gap and make it a first-class operation.

### `run_on_cloud`
Submits a flow (or folder of flows) + an app binary to Maestro Cloud asynchronously. Returns `{ upload_id, project_id, app_id, url, status: "PENDING" }` immediately. `url` is the `app.maestro.dev` dashboard link the user can click.

Inputs: `app_file`, `flows`, `name?`, `project_id?`, `env?`, `include_tags?`, `exclude_tags?`, `device_os?`.

### `get_cloud_run_status`
Polls status for a previously-submitted upload. Returns `{ status, completed, started_at, total_time_ms, app_package_id, was_app_launched }` and, when `include_flow_results=true`, the per-flow breakdown.

## Design notes

- **Auth via `maestro login` (recommended) or `MAESTRO_CLOUD_API_KEY` (non-interactive).** Both tool descriptions and missing-auth error messages across all four auth-gated MCP tools (`query_docs`, `cheat_sheet`, and the two new ones) surface `maestro login` as the happy path. `ApiKey.getToken()` already reads from both the env var and the cached token at `~/.mobiledev/authtoken` (written by `maestro login`), so no auth logic changes are needed. Also updated the `getProjects` 500 error path to explain the scope limitation of machine-scoped API keys and suggest either logging in or passing `project_id` explicitly.

- **Cloudless-account handling.** Accounts without an active Cloud plan can authenticate via `maestro login` and even list projects (the backend returns the account's project without plan gating), but the POST to `runMaestroTest` is plan-gated and triggers `ApiClient.upload`'s interactive trial-not-started branch: a `Scanner(System.in).nextLine()` that reads stdin to collect a company name and start a trial. Under MCP that stdin is the JSON-RPC protocol pipe, so this branch would hang forever. The tool now redirects stdin to an empty stream for the duration of the handler and classifies the resulting exception (plus the backend's "trial expired" / "payment" / "subscription" errors) into one actionable message pointing at `https://app.maestro.dev`. Same redirect pattern as the existing `System.out` to `System.err` swap: the MCP transport holds its own reference to the real streams before the handler runs, so the protocol channel is unaffected.

- **Bypasses `CloudInteractor`.** That class does interactive project/org selection via `terminal.interactiveSelectList(...)` and prints liberally via `PrintUtils.message` / raw `println`. Both would break MCP: the interactive prompt hangs on the stdio pipe, and stray stdout corrupts the JSON-RPC stream (MCP's `StdioServerTransport` captures `System.out` as its sink). Going through `ApiClient` directly keeps the handlers silent and non-interactive by construction.

- **Non-interactive project resolution.** If `project_id` is omitted, the tool auto-picks only when the account has exactly one project. Multiple projects (or zero) return a structured error that includes the full project list so the agent can ask the user by name and retry with `project_id` set. This is the expected two-turn flow for customers with more than one project: tool returns the list, agent asks "which project?", user picks, agent retries. The tool never prompts on stdin since that pipe is the MCP protocol channel. Note: auto-pick requires a session token from `maestro login`; the `getProjects` backend endpoint does not accept machine-scoped API keys (see known limitations).

- **`System.setOut(System.err)` swap for the duration of each handler** as defense in depth. Catches any stdout print from downstream code (`ApiClient` retry logic, `Analytics.trackEvent`) before it can land in the protocol channel.

- **Not exposed in v1** to keep the schema small: mapping files, CI metadata (`repo_owner`/`branch`/`commit_sha`/`pull_request_id`), report `format`/`output`, `api_url`/`api_key` (env var only), `device_model`, `device_locale`, `app_binary_id`. Easy to add later if the agent needs them.

### Local harness

- `./gradlew :maestro-cli:installDist`: clean build after rebasing onto latest main (includes PR #3182 SDK upgrade and PR #3181 MA-4007 merge).
- `tool-tests-without-device.yaml` via `mcp-server-tester@1.3.1`: **8/8 passed**, including tool discovery at **16/16 expected tools** and missing-auth error cases for all four auth-gated tools. Re-verified after every change on the branch.
- `tool-tests-with-device.yaml` against a booted iPhone 17 sim: **8/9 passed**. The one failure is on `Test tap_on` (`text: "Search"` against Safari returns "Element not found") and is a pre-existing main regression from PR #3181 (MA-4007 removed the implicit `.*...*` wrap, but the test fixture still uses bare `Search`). Unrelated to this PR, not modified by it. Worth a separate followup to update the fixture to `.*Search.*`.

### `maestro login` auth path verified end-to-end

After running `maestro login`, with `MAESTRO_CLOUD_API_KEY` unset and no env-var fallback available, called `run_on_cloud` without `project_id`:

- `ApiKey.getToken()` fell through to the cached token at `~/.mobiledev/authtoken` as designed. Authentication succeeded.
- `getProjects` returned the full project list (session tokens are accepted where machine-scoped API keys return HTTP 500, confirming the known-limitation behavior).
- The multi-project error path fired correctly: the tool listed all accessible projects in the response so the agent can present them to the user by name, exactly the two-turn flow described in the design notes.

### Cloudless-account verification

Logged in with an account on the BASIC pricing plan, all parallel-run quotas zeroed (`max_android_runs_in_parallel.limit=0`, `max_ios_runs_in_parallel.limit=0`, `max_web_runs_in_parallel.limit=0` confirmed via raw `curl` against `/v2/maestro-studio/orgs`). Called `run_on_cloud` with no `project_id`:

- Tool returned in **2.9s** (no stdin hang).
- `isError=true`.
- Payload: `"Your Maestro Cloud account is not ready to run flows on cloud devices. This typically means your free trial has not started (or has expired), or there's a payment issue. Start your free trial or manage your subscription at https://app.maestro.dev, then retry this request."`
- No stdout contamination of the JSON-RPC stream.

Auto-pick of the one accessible project succeeded. The plan-gate fired at the upload step, landed in the new `classifyUploadFailure` path, and was surfaced cleanly to the agent.

### Live Cloud end-to-end, 5 real uploads against my personal account (pre-cloudless-fix)

| # | Scenario | Dashboard shape |
|---|---|---|
| 1 | **Android smoke**. `wikipedia.apk` + trivial `launchApp` flow, minimal required args | `https://app.maestro.dev/project/proj_…/…/upload/mupload_…` |
| 2 | **Android folder-of-flows** with `name` + `env={MCP_MARKER}` + `include_tags=["smoke"]`. Verified on dashboard that only the `smoke`-tagged flow ran and the other was filtered out | `https://app.maestro.dev/project/proj_…/…/upload/mupload_…` |
| 3 | **Android minimal** with required args only (`app_file` + `flows` + `project_id`), no optionals | `https://app.maestro.dev/project/proj_…/…/upload/mupload_…` |
| 4 | **iOS `.zip` bundle**. `SimpleWebViewApp.zip` (already-zipped, exercises the `isZip()` passthrough branch) + `device_os="iOS-17-5"` + `exclude_tags=["drop"]`. Folder with two flows where the `drop`-tagged one gets filtered out | `https://app.maestro.dev/project/proj_…/…/upload/mupload_…` |
| 5 | **Post-migration sanity.** Re-ran the smoke scenario after the SDK 0.11.1 migration and auth-messaging update. `submit` + both status polls (immediate and +15s with flow results) all passed cleanly, confirming no regression from the rebase | `https://app.maestro.dev/project/proj_…/…/upload/mupload_…` |

Full dashboard links (internal): see the [MA-4010 Linear ticket](https://linear.app/mobile-dev/issue/MA-4010/mcp-implement-a-native-cloud-tall-that-can-submit-maestro-cloud-runs) comment. Team members can click through to any of the five runs to verify.

On submission #1, immediately after the 3.5s submit, `get_cloud_run_status` returned `status=RUNNING` with `app_package_id=org.wikipedia` (backend correctly parsed the apk). After 15s: `was_app_launched=true`; per-flow array populated when `include_flow_results=true`. Prior uploads polled later confirmed terminal `SUCCESS`.

### Thorough test matrix: 80 deterministic checks, 80/80 passed

| Tier | What it covers | Checks |
|---|---|---|
| 1 | Local arg validation: missing `app_file`, missing `flows`, nonexistent `app_file` / `flows`, same for `get_cloud_run_status` missing `upload_id` / `project_id` (both error types surfaced with actionable messages) | 12 |
| 2 | Server-side error surfacing: bogus `upload_id` returns structured `HTTP 404`, wrong `project_id` on a valid upload returns a structured `HTTP` error | 4 |
| 3 | Response shape on prior upload, both `include_flow_results=false` (default) and `=true`: required keys present, types correct (bool/int/str), `status` value is in the valid enum, `mupload_` prefix on `upload_id`, `flows` array absent in default and present-and-list-typed when flag on | 17 |
| 4 | Terminal-state polling on an already-completed prior upload: `completed=true` plus `status` is in the terminal set, `flows` array populated, each flow has valid `status` enum and string `name` | 12 |
| 5 | Minimal required-only submission. Response shape conformance: `success=true`, `upload_id` / `app_id` / `project_id` prefix checks, `status="PENDING"`, `url` is an `app.maestro.dev` dashboard link, URL embeds both `upload_id` and `project_id` correctly | 9 |
| 6 | iOS `.zip` submission with `device_os=iOS-17-5` + `exclude_tags`. Verifies the `isZip()` passthrough branch, `device_os` landing server-side, exclude-tag filter path, all on one submit | 9 |
| 7 | Immediate `get_cloud_run_status` against both fresh submits (tier 5 + tier 6): response shape conformance against the ids we just created | 17 |

**Bonus test caught during setup**: initially passed `device_os="ios-17-5"` (lowercase), got back a clean structured error from the backend: `"Upload request failed (400): OS version 'ios-17-5' is not supported for model 'iPhone-11', platform IOS. Supported versions: iOS-16-2, iOS-16-4, iOS-17-5, iOS-18-2, iOS-26-2"`. The tool surfaces the server's full message verbatim, so the agent can self-correct. Re-ran with `"iOS-17-5"` and it submitted cleanly (#4 above).

I tested the behaviors above personally end-to-end, including the `maestro login` path and the cloudless-account path.

## Known limitations / follow-ups

- **`GET /v2/maestro-studio/projects` returns HTTP 500 for machine-scoped API keys.** Reproduced with raw `curl`, not code. Endpoint appears to require a user-session token from `maestro login`. The MCP tool now handles this with a clear error pointing at both recovery paths (either run `maestro login`, or pass `project_id` explicitly). The same underlying issue also affects the CLI's `maestro cloud` auto-pick path for any API-key user.
- `tool-tests-with-device.yaml`'s `tap_on` test case needs its fixture updated to `.*Search.*` post-#3181, unrelated to this PR but surfaced by the test run.
- Multi-project customers using `maestro login` still need to specify `project_id` after the agent surfaces the project list. A future follow-up adding `list_cloud_projects` as a dedicated discovery tool would let the agent enumerate up front and pick by name without needing the error path.

Fixes [MA-4010](https://linear.app/mobile-dev/issue/MA-4010/mcp-implement-a-native-cloud-tall-that-can-submit-maestro-cloud-runs).

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
@@ -1,0 +1,114 @@
+package maestro.cli.mcp.tools
+
+import io.modelcontextprotocol.kotlin.sdk.*
+import io.modelcontextprotocol.kotlin.sdk.server.RegisteredTool
+import kotlinx.serialization.json.*
+import maestro.auth.ApiKey
+import maestro.cli.api.ApiClient
+
+object GetCloudRunStatusTool {
+    fun create(): RegisteredTool {
+        return RegisteredTool(
+            Tool(
+                name = "get_cloud_run_status",
+                description = "Fetch the current status and (optionally) per-flow results of a Maestro Cloud upload. " +
+                    "Use with the upload_id and project_id returned by run_on_cloud. Requires MAESTRO_CLOUD_API_KEY.",
+                inputSchema = Tool.Input(
+                    properties = buildJsonObject {
+                        putJsonObject("upload_id") {
+                            put("type", "string")
+                            put("description", "The upload_id returned by run_on_cloud.")
+                        }
+                        putJsonObject("project_id") {
+                            put("type", "string")
+                            put("description", "The project_id the upload was submitted to (also returned by run_on_cloud).")
+                        }
+                        putJsonObject("include_flow_results") {
+                            put("type", "boolean")
+                            put("description", "If true, include the per-flow breakdown once the run has completed. Default false.")
+                        }
+                    },
+                    required = listOf("upload_id", "project_id")
+                )
+            )
+        ) { request ->
+            val originalOut = System.out
+            System.setOut(System.err)
+            try {
+                val uploadId = request.arguments["upload_id"]?.jsonPrimitive?.content
+                val projectId = request.arguments["project_id"]?.jsonPrimitive?.content
+                val includeFlowResults = request.arguments["include_flow_results"]
+                    ?.jsonPrimitive?.booleanOrNull ?: false
+
+                if (uploadId.isNullOrBlank()) {
+                    return@RegisteredTool errorResult("upload_id is required")
+                }
+                if (projectId.isNullOrBlank()) {
+                    return@RegisteredTool errorResult("project_id is required")
+                }
+
+                val apiKey = ApiKey.getToken()
+                if (apiKey.isNullOrBlank()) {
+                    return@RegisteredTool errorResult(
+                        "MAESTRO_CLOUD_API_KEY environment variable is required. " +
+                            "See https://docs.maestro.dev/api-reference/configuration/workspace-configuration for setup."
+                    )
+                }
+
+                val apiUrl = System.getenv("MAESTRO_CLOUD_API_URL")
+                    ?: System.getenv("MAESTRO_API_URL")
+                    ?: "https://api.copilot.mobile.dev"
+                val client = ApiClient(apiUrl)
+
+                val status = try {
+                    client.uploadStatus(apiKey, uploadId, projectId)
+                } catch (e: ApiClient.ApiException) {
+                    return@RegisteredTool errorResult(
+                        "Failed to fetch cloud run status (HTTP ${e.statusCode}) for upload_id=${uploadId}"
+                    )
+                }
+
+                val result = buildJsonObject {
+                    put("success", true)
+                    put("upload_id", status.uploadId)
+                    put("status", status.status.name)
+                    put("completed", status.completed)
+                    status.startTime?.let { put("started_at", it) }
+                    status.totalTime?.let { put("total_time_ms", it) }
+                    status.appPackageId?.let { put("app_package_id", it) }
+                    put("was_app_launched", status.wasAppLaunched)
+                    if (includeFlowResults) {
+                        putJsonArray("flows") {
+                            status.flows.forEach { flow ->
+                                addJsonObject {
+                                    put("name", flow.name)
+                                    put("status", flow.status.name)
+                                    flow.startTime?.let { put("started_at", it) }
+                                    flow.totalTime?.let { put("total_time_ms", it) }
+                                    if (flow.errors.isNotEmpty()) {
+                                        putJsonArray("errors") {
+                                            flow.errors.forEach { add(it) }
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }.toString()
+
+                CallToolResult(content = listOf(TextContent(result)))
+            } catch (e: Exception) {
+                CallToolResult(
+                    content = listOf(TextContent("Failed to get cloud run status: ${e.message ?: e.javaClass.simpleName}")),
+                    isError = true
+                )
+            } finally {
+                System.setOut(originalOut)
+            }
+        }
+    }
+
+    private fun errorResult(message: String): CallToolResult {
+        return CallToolResult(content = listOf(TextContent(message)), isError = true)
+    }
+}
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
@@ -1,6 +1,6 @@
 package maestro.cli.mcp.tools
 
-import io.modelcontextprotocol.kotlin.sdk.*
+import io.modelcontextprotocol.kotlin.sdk.types.*
 import io.modelcontextprotocol.kotlin.sdk.server.RegisteredTool
 import kotlinx.serialization.json.*
 import maestro.auth.ApiKey
@@ -12,8 +12,9 @@ object GetCloudRunStatusTool {
             Tool(
                 name = "get_cloud_run_status",
                 description = "Fetch the current status and (optionally) per-flow results of a Maestro Cloud upload. " +
-                    "Use with the upload_id and project_id returned by run_on_cloud. Requires MAESTRO_CLOUD_API_KEY.",
-                inputSchema = Tool.Input(
+                    "Use with the upload_id and project_id returned by run_on_cloud. " +
+                    "Requires Maestro Cloud authentication: run `maestro login` (recommended), or set MAESTRO_CLOUD_API_KEY for non-interactive use.",
+                inputSchema = ToolSchema(
                     properties = buildJsonObject {
                         putJsonObject("upload_id") {
                             put("type", "string")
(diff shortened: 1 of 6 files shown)


maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
```diff
@@ -0,0 +1,114 @@
...
+import maestro.auth.ApiKey
+import maestro.cli.api.ApiClient
+
+object GetCloudRunStatusTool {
+    fun create(): RegisteredTool {
+        return RegisteredTool(
+            Tool(
+                name = "get_cloud_run_status",
+                description = "Fetch the current status and (optionally) per-flow results of a Maestro Cloud upload. " +
+                    "Use with the upload_id and project_id returned by run_on_cloud. Requires MAESTRO_CLOUD_API_KEY.",
```
Fishbowler: Should this also be able to use a machine's local auth token?

maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
```diff
@@ -0,0 +1,114 @@
...
+                }
+                if (projectId.isNullOrBlank()) {
+                    return@RegisteredTool errorResult("project_id is required")
+                }
+
+                val apiKey = ApiKey.getToken()
+                if (apiKey.isNullOrBlank()) {
+                    return@RegisteredTool errorResult(
+                        "MAESTRO_CLOUD_API_KEY environment variable is required. " +
+                            "See https://docs.maestro.dev/api-reference/configuration/workspace-configuration for setup."
```
Fishbowler: This page looks to be part of the old docs layout. I'm not even sure it's correct. Maybe https://docs.maestro.dev/maestro-flows/flow-control-and-logic/parameters-and-constants? I'm not sure we even have an appropriate docs page for "Export env vars on your OS", if this is what you're after?

maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
pedro18x: Yes, addressed in the later commits. `ApiKey.getToken()` already reads both `MAESTRO_CLOUD_API_KEY` and the cached token from `maestro login` at `~/.mobiledev/authtoken`, so the tool works end-to-end with a logged-in session. Tool descriptions and missing-auth error messages across all four auth-gated MCP tools now surface `maestro login` as the recommended path.

maestro-cli/src/main/java/maestro/cli/mcp/tools/GetCloudRunStatusTool.kt
pedro18x: Fair, dropped the URL entirely. The new missing-auth error points at the `maestro login` CLI command instead of a docs page.

Leland-Takamine (approved):

Fishbowler (approved):

PR #3188, opened 2026-04-17
refactor: extract maestro debug artifact generation from CLI to core module orchestra

amanjeetsingh150:
## Summary

During major rollouts for us or for users to investigate failures or now even agents its important to look at maestro artifacts to reason about failures. These artifacts contains:

- maestro.log file (Maestro logs)
- Screenshot on failure
- commands.json that is structured json for command's name, duration status, index.

Right now this is only accessible from CLI. This creates problems:

- Not having them on cloud leave us blind on maestro context.
- This can be used by agents during diagnosis and rollouts to ensure no new regressions due to changes on our side.
- You may argue that we can make these artifacts on cloud separately, however this create duplication where we can just utilize what maestro already offers.
- Studio can also use this to expose these artifacts. 

This PR is to just move this artifact generation to core module instead of cli so that all the consumers can use it.

## Changes

- Firs step is to extract out DebugReporter and related models out of cli module to core.
- This abstraction weirdly sits in cli module which should be part of shared core module.
- DebugReporter, is responsible for generation commands.json, screenshot on failure, maestro.log file
- These artifacts are context for investigating failures and to reason what maestro was doing during flow execution
- Adding tests to make sure that cli debug artifacts are still getting generated

## Test plan

- [x] Ensuring unit tests specially [TestDebugReporterTest.kt‎](https://github.com/mobile-dev-inc/Maestro/pull/3188/changes#diff-79647760dada06633e6a521265158a79ed07c5e5aa630516ab0599d4d076263f) which expects the different artifacts are passing.
- [x] Use E2E test for gaining this confidence by ensuring the artifacts are dumped with same structure.


Fishbowler (approved):
Code looks like sensible shuffling.

Testing locally:

```
maestro test fail_fast.yaml
maestro test fail_fast.yaml --test-output-dir ./potato 
maestro test fail_fast.yaml --debug-output ./potato
maestro test fail_fast.yaml --test-output-dir ./potato --debug-output ./potato
maestro test fail_fast.yaml --test-output-dir ./potato --debug-output ./potato --flatten-debug-output

maestro test commands/assertTrue.yaml 
maestro test commands/assertTrue.yaml --test-output-dir ./potato --debug-output ./potato --flatten-debug-output 
```

☝️ All of these work as they do in the last release (with the exception of the AI output, which is now generated when unneeded)

Issue #2878, opened 2025-12-09
setAirplaneMode can't use interpolated arguments


Fishbowler:
### Discussed in https://github.com/mobile-dev-inc/Maestro/discussions/2877

<div type='discussions-op-text'>

<sup>Originally posted by **gui-the-garm** December  9, 2025</sup>
I have this auxilary flow to help lessen repetition in text, but I need to use a different one when running tests that need the phone to be offline.
I wish I didn't have to use a different flow file for when tests are run offline vs online, like the example in the picture.

<img width="306" height="132" alt="image" src="https://github.com/user-attachments/assets/49d9a597-e633-4c73-9156-b2bd01fb39ab" /></div>

PR #3192, opened 2026-04-18
feat: add support for setting airplane mode with environment variables

sailsdima:
## Proposed changes
This PR adds environment variable support to the `setAirplaneMode` command. Previously, `setAirplaneMode` only accepted hardcoded literal values (enabled/disabled). With this change, users can now pass environment variables as well.
copilot:summary

**What Changed:**
`SetAirplaneModeCommand` - now accepts String value instead of AirplaneValue enum (to support ${...}) variable references. Variables are resolved at runtime via evaluateScripts().
`YamlSetAirplaneMode` - rewrote deserializer to pass through variable references at parse time; literal values are still validated eagerly.
`Orchestra` - Uses `resolvedValue()` instead of direct enum access.

**Usage Example:**
```
appId: com.example.app
env:
  airplaneOn: enabled
  airplaneOff: disabled
---
- setAirplaneMode: ${airplaneOn}
- setAirplaneMode: ${airplaneOff}
- setAirplaneMode: enabled # still works
- setAirplaneMode: disabled # still works
```

## Testing

> **Does this need e2e tests?** The demo app lives in [`e2e/demo_app/`](e2e/demo_app/) — add a test screen and Maestro flow in the same PR. See [`e2e/demo_app/CLAUDE.md`](e2e/demo_app/CLAUDE.md) for details.

**Unit tests**: 
- `YamlCommandReaderTest` (parsing with env vars + invalid value error), 
- `CommandDescriptionTest` (description evaluation).

**Integration test**: added new test in `IntegrationTest` that runs runs setAirplaneMode with env variables through the full  pipeline.

**e2e test**: Extended existing e2e `setAirplaneMode.yaml` to test env variable for airplane mode.

All existing tests pass with no regressions.

image
image

## Issues fixed
https://github.com/mobile-dev-inc/Maestro/issues/2878


maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -4829,6 +4829,26 @@ class IntegrationTest {
...
+        // When
+        Maestro(driver).use {
+            runBlocking {
+                orchestra(it).runFlow(commands)
+            }
+        }
+
+        // Then
+        driver.assertHasEvent(Event.SetAirplaneMode(true))
+        driver.assertHasEvent(Event.SetAirplaneMode(false))
```
Fishbowler: Event order might be important here?

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -1153,22 +1153,42 @@ data class StopRecordingCommand(
...
+
+    fun resolvedValue(): AirplaneValue {
+        return AirplaneValue.fromString(value)
+            ?: error("Unknown airplane mode value: $value. Valid values are: enabled, disabled")
+    }
+
+    override fun evaluateScripts(jsEngine: JsEngine): SetAirplaneModeCommand {
+        val evaluatedValue = value.evaluateScripts(jsEngine)
+        AirplaneValue.fromString(evaluatedValue)
+            ?: error("Unknown airplane mode value: $evaluatedValue. Valid values are: enabled, disabled")
```
Fishbowler: Why is this validation repeated in evaluateScripts and in resolvedValue? Are there codepaths that would only trigger one or the other?

sailsdima: The validation inside `resolvedValue` should never be triggered since it is first done in `evaluateScripts`, but it is needed since the function returns non-nullable `AirplaneValue`, and `AirplaneValue.fromString(value)` returns nullable. 
There is a similar implementation in `SetOrientationCommand`.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
sailsdima: Good catch, updated this to take order into account.

PR #3193, opened 2026-04-20
feat(mcp): consolidate run_flow + run_flow_files into single `run` tool

pedro18x:
Linear: [MA-4022](https://linear.app/mobiledev/issue/MA-4022)

## What

Merged `run_flow` and `run_flow_files` into a single MCP tool called `run`.

## Why

Two tools doing nearly the same thing, with different param names (`flow_yaml` vs `flow_files`) and slightly different semantics. Cleaner to have one tool that mirrors the `maestro test` CLI shape.

## How it works

Exactly one of `yaml`, `files`, or `dir` must be provided. `device_id` is always required. `env` is optional.

Inline YAML (what you reach for when exploring or debugging):
```json
{ "device_id": "...", "yaml": "- tapOn: 123" }
```

Specific files:
```json
{ "device_id": "...", "files": ["flow1.yaml", "flow2.yaml"] }
```

Directory with optional tag filtering:
```json
{ "device_id": "...", "dir": "tests/", "include_tags": ["smoke"], "exclude_tags": ["slow"] }
```

## Notes

- Syntax is validated as part of `run`. No separate pre-check call needed.
- Env injection (`withInjectedShellEnvVars` + `withDefaultEnvVars`) is preserved.
- `dir` mode goes through the same `WorkspaceExecutionPlanner` that `maestro test` uses, so tag filtering, config discovery, and sequence ordering (`sequence.flows` first, respecting `continueOnFailure`) match the CLI exactly.
- Input validation (missing files, invalid workspace) happens before the device session is opened, so a typo in `files` doesn't spin up a simulator.
- `RunFlowTool.kt` and `RunFlowFilesTool.kt` are gone.

## Rebase notes (not feature work)

Two changes in the diff are adaptations to main, not part of the feature:

- `full-evals.yaml` lists 15 tools (including `run_on_cloud`, `get_cloud_run_status`). Those came from #3184 during the rebase window. I updated the count and the list so the eval still matches what `McpServer.kt` actually registers.
- `RunTool.kt` uses the MCP Kotlin SDK 0.11.1 API (`io.modelcontextprotocol.kotlin.sdk.types.*`, `ToolSchema`, nullable `request.arguments`). That's the current convention on main after #3182.

## Test plan

- [x] Unit tests (`RunToolTest.kt`, 12 cases): three input shapes, mutual exclusion (zero / two / three modes), empty files list, tags outside `dir`, env propagation, invalid args.
- [x] MCP no-device eval: 15/15 tools discovered, `run` present, `run_flow` / `run_flow_files` gone.
- [x] MCP with-device eval on a real iOS simulator: 11/11 pass, covering inline YAML, nonexistent-file error, files + env injection, `dir` mode, and `dir` + `include_tags`.


maestro-cli/src/main/java/maestro/cli/mcp/tools/RunTool.kt
```diff
@@ -0,0 +1,378 @@
...
+import maestro.orchestra.yaml.YamlCommandReader
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+
+object RunTool {
+
+    private const val TOOL_NAME = "run"
+    private const val TOOL_DESCRIPTION = """
+        Run Maestro flows against a connected device. Mirrors `maestro test` semantics.
```
Leland-Takamine: > "Mirrors `maestro test` semantics." 

Was this added for a specific reason? If not, I'd remove as it feels like noise that could confuse the agent.

Leland-Takamine: Maybe: "Simulate user interactions with a mobile device by running raw Maestro commands or Maestro flow files"

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunTool.kt
```diff
@@ -0,0 +1,378 @@
...
+        is RunInput.Directory -> Executable.Plan(
+            planDirectory(input.path, input.includeTags, input.excludeTags)
+        )
+    }
+
+    private fun buildInputSchema(): ToolSchema = ToolSchema(
+        properties = buildJsonObject {
+            putJsonObject("device_id") {
+                put("type", "string")
+                put("description", "The ID of the device to run the flow on")
```
Leland-Takamine: Mention somewhere how to get this id

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunTool.kt
pedro18x: No specific reason, honestly. Your wording is better, will apply

maestro-cli/src/main/java/maestro/cli/mcp/tools/RunTool.kt
```diff
@@ -0,0 +1,378 @@
...
+                putJsonObject("additionalProperties") { put("type", "string") }
+            }
+        },
+        required = listOf("device_id"),
+    )
+
+    private fun runInline(
+        orchestra: Orchestra,
+        yaml: String,
+        deviceId: String,
```
Leland-Takamine: Usually if there's a parameter that is the "target" of a method, I would put that first eg. deviceId here. I'm not 100% if this is just a personal preference or actual best practice, but it does align with the OOP pattern of [instance].[method] ie: target first, then then the method and parameters

Leland-Takamine (approved):

PR #3198, opened 2026-04-21
Drop obvious non-flow artifacts from workspace uploads

proksh:
### Why

`WorkspaceUtils.createWorkspaceZip` walks the workspace with `Files.walk(file).filter { !it.isDirectory() }` and uploads everything under it. On real workspaces this sweeps up material that is never consumed by a flow - app binaries left in place from local testing, nested archives, editor/VCS directories - bloating the upload by tens of MB which needs to be dowloaded on worker for each run and slowing the runs. 

### What changed

`WorkspaceUtils.createWorkspaceZip` now filters out a narrow, explicit set of clearly-non-flow artifacts before zipping:
- **Extensions**: `.apk`, `.aab`, `.ipa`, `.zip`, `.tar`, `.gz`, `.tgz`, `.7z`, `.rar`
- **Filenames**: `.DS_Store`, `Thumbs.db`
- **Ancestor directories**: `.git`, `.hg`, `.svn`, `.idea`, `.vscode`, `.gradle`, `build`, `target`, `node_modules`

Everything else passes through unchanged. Genuine assets a flow might reference via `addMedia`, `runScript`, text fixtures, etc. - images, video, audio,`.js`, `.txt`, `.json` - are untouched. The list is deliberately narrow; the rule of thumb is "only add an entry if a flow can provably never use it."

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -74,6 +74,7 @@ class CloudInteractor(
         flowFile: File,
         appFile: File?,
         async: Boolean,
+        configFile: File? = null,
         mapping: File? = null,
         apiKey: String? = null,
         uploadName: String? = null,
@@ -126,7 +127,11 @@ class CloudInteractor(
 
         TemporaryDirectory.use { tmpDir ->
             val workspaceZip = tmpDir.resolve("workspace.zip")
-            WorkspaceUtils.createWorkspaceZip(flowFile.toPath().absolute(), workspaceZip)
+            WorkspaceUtils.createWorkspaceZip(
+                file = flowFile.toPath().absolute(),
+                out = workspaceZip,
+                configOverride = configFile?.toPath()?.absolute(),
+            )
             val progressBar = ProgressBar(20)
 
             // Binary id or Binary file
--- a/maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/CloudCommand.kt
@@ -219,6 +219,7 @@ class CloudCommand : Callable<Int> {
             async = async,
             flowFile = flowsFile,
             appFile = appFile,
+            configFile = configFile,
             mapping = mapping,
             env = env,
             uploadName = uploadName,
--- a/maestro-cli/src/main/java/maestro/cli/util/FileUtils.kt
+++ b/maestro-cli/src/main/java/maestro/cli/util/FileUtils.kt
@@ -1,5 +1,6 @@
 package maestro.cli.util
 
+import maestro.orchestra.workspace.isWorkspaceConfigFile
 import maestro.orchestra.yaml.YamlCommandReader
 import maestro.utils.StringUtils.toRegexSafe
 import java.io.File
@@ -27,11 +28,7 @@ object FileUtils {
             name.endsWith(".yaml", ignoreCase = true) ||
             name.endsWith(".yml", ignoreCase = true)
 
-        if (
-            !isYaml ||
-            name.equals("config.yaml", ignoreCase = true) ||
-            name.equals("config.yml", ignoreCase = true)
-        ) {
+        if (!isYaml || isWorkspaceConfigFile(toPath())) {
             return false
         }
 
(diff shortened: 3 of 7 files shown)


Fishbowler (approved):
Reduced bandwidth usage and upload times for users, and reduced pointless storage for Maestro Cloud. Everybody wins!

steviec (changes requested):
@Fishbowler  @proksh a few questions:
- shouldn't this be a part of parsing the workspace? we should know all the flows files and the files they reference. isn't that the most robust way to do this? instead of encoding all the _potential_ files we should exclude, we should just include the files the workspace references. related: putting this in workspace utils doesn't seem like the right mental model.
- @Fishbowler is it possible that customers are intentionally including their build so they are tightly coupled ot the flow run when they download the workspace? should we reach out to any of them?
- @proksh did you download any sample workspaces (like Wahed) and inspect it to make sure this looks good? did you do any local testing? for example, I don't see ".app" in the list, but I'm assuming that's a common one.

Fishbowler:
We can use DependencyResolver to be relatively certain, but we were relatively certain the last 5 or so times we fixed that. I don't think that caters for [multipart uploads](https://docs.maestro.dev/maestro-flows/javascript/make-http-requests#multipart-form-data) yet, for example?

I've not heard of anyone wanting to upload the build 2x. I don't know how often folks download the workspace - I find that useful for customer eval, but use if far more in support of customers than I ever did as the first party - I already had those files.

steviec:
@proksh @Fishbowler what do you think? I don't want to hold this up if it's a useful addition, especially if you think customer will be fine with this Dan. It's a smell to me to do all this "magic" detection of random stuff; I'd rather just make our dependency resolver actually resolve all of our dependencies. These sorts of fixes usually end up biting us in the ass (and cause thrash with customers being confused). Whaddya think? + @amanjeetsingh150 in case you have any thoughts.

proksh:
> * shouldn't this be a part of parsing the workspace? we should know all the flows files and the files they reference. isn't that the most robust way to do this? instead of encoding all the _potential_ files we should exclude, we should just include the files the workspace references. related: putting this in workspace utils doesn't seem like the right mental model.

`DependencyResolver.discoverAllDependencies` only sees statically declared references: runFlow, runScript, retry, assertScreenshot, addMedia, and config hooks. It does not see:
- Files read at runtime from JS (evalScript / runScript via file.read(...), fs.readFileSync, etc.)
- Paths built from variables / env (addMedia: ${mediaFile}, runScript: ${scriptPath})
- .env files passed via --env-file
- Fixture data referenced via string interpolation

So a pure allowlist would silently drop files that real flows depend on.

> * @proksh did you download any sample workspaces (like Wahed) and inspect it to make sure this looks good? did you do any local testing? for example, I don't see ".app" in the list, but I'm assuming that's a common one.

Yes. I didn't see any reference of their app in the workspace as well. I will wait for their reply on the thread. Already tested on android by uploading, let me do an ios upload as well.

proksh:
While a do agree that only picking the referenced file would be good, but it will a much harder problem to solve and it will still be us doing some "magic" for finding such files since it can be in JS as well.

proksh:
> Already tested on android by uploading, let me do an ios upload as well.

Actually, it is breaking for ios because its not a build app they are putting but xcode Build folder. Seeing this, I think allowlist suggestion seems cleaner. I am trying to explore that @steviec.

amanjeetsingh150 (commented):
Like the direction, but I think the solution needs a bit more structure. A few thoughts:

Resolution has always been an optional step in creating a workspace, and I think that's starting to fall apart custom config, missing media files, and files that should be ignored have all surfaced recently and share the similar surface area.

My suggestion would be to make resolution a compulsory part of the WorkspaceExecutionPlanner contract. Right now we have that concept loosely, and I think every consumer of core would benefit. The resolver becomes the single place responsible for staging what the worker/CLI/studio expects and ignoring what it doesn't. This contract stages and fails the resolution of files etc.

This contract helps us debugging if we get FileNotFound exceptions on worker that clearly means staging for some case is missing.

How this PR fits in:

- **Ignore by size**: Dan pointed out builds aren't useful to stage; we can frame this as a step in resolver, this avoids hardcoded list anything beyond a threshold would be ignored. We can announce this change to user to avoid surprises.
- **Stage external static dependencies**, already being done in pieces with custom config and dependency resolver.

> Files read at runtime from JS (file.read(...), fs.readFileSync), paths built from variables (addMedia: ${mediaFile}), .env files via --env-file, fixture data via string interpolation.

@proksh made good point here^ Worth stepping back here: do users actually rely on dynamic reads heavily in their flows? If not, we can simplify dramatically. If yes I'd be curious for examples, we should eagerly resolve and stage them during planning when they're not already in the workspace.

This also feeds into the Maestro Workspace path-issues initiative, so probably worth closing out the decision before we merge.

proksh:
there are wide range of things in the workspace. It would better to communicate this to users, rather than doing magic filtering on our side.

PR #3200, opened 2026-04-21
OCR Feature Commands

Kardelio:
## Proposed changes

Adds three new OCR-based commands to Maestro powered by the `tesseract` CLI: `tapOnOcr`, `assertVisibleOcr`, and `assertNotVisibleOcr`. These commands take a live screenshot of the device, run Tesseract OCR on it, and locate UI text that is invisible to the standard accessibility tree.

### The problem

Certain UI frameworks render views that never appear in the Android accessibility tree, which means Maestro's standard `tapOn: text:` and `tapOn: id:` selectors cannot find them. A concrete example is **React Native Reanimated**: animated views (`Animated.View`) update only the native layer via the UI thread — they bypass the accessibility tree entirely. This makes it impossible to interact with text rendered inside them using conventional Maestro commands, regardless of whether a `testID` is set.

This is a real limitation for any React Native app that uses Reanimated for animated UI elements such as toggles, badges, overlays, or bottom sheets that contain tappable text.

### The solution

Three new commands that use Tesseract OCR on a device screenshot to locate text coordinates at runtime, bypassing the accessibility tree entirely:

```yaml
# Finds "Production" anywhere on screen via OCR and taps its centre
- tapOnOcr: "Production"

# With options
- tapOnOcr:
    text: "Production"
    label: "Tap the Production environment badge"
    optional: true

# Tune Tesseract recognition with a Page Segmentation Mode (PSM)
- tapOnOcr:
    text: "Production"
    psm: 6   # treat image as a single uniform block of text

# Assert text is visible on screen via OCR
- assertVisibleOcr: "Welcome back"

- assertVisibleOcr:
    text: "Welcome back"
    label: "Check welcome message is shown"

- assertVisibleOcr:
    text: "Welcome back"
    psm: 11  # sparse text — find as much as possible

# Assert text is NOT visible on screen via OCR
- assertNotVisibleOcr: "Error"

- assertNotVisibleOcr:
    text: "Error"
    label: "Check no error message is present"
    optional: true

- assertNotVisibleOcr:
    text: "Error"
    psm: 6
```

All three commands follow the same YAML patterns as existing Maestro commands (simple string form and full selector form with `label`, `optional`, and the new `psm`).

### Implementation

- Commands are dispatched through `Orchestra.kt` via a shared `ocrFindCenter(text, imageFile, psm)` helper that invokes `tesseract` as a subprocess using `ProcessBuilder`, parses the TSV word output, and uses a sliding window to match multi-word phrases
- When `psm` is provided, `--psm <N>` is inserted into the Tesseract command before the input image path; when omitted, Tesseract uses its default page segmentation (mode 3)
- If `tesseract` is not installed, the command throws `MaestroException.InvalidCommand` with a clear install message rather than crashing
- `tapOnOcr` saves the screenshot to a temp file for inspection on failure
- `tesseract` is added as a brew dependency so it installs automatically alongside Maestro

### PSM reference

Tesseract PSM values most useful for mobile screenshots:

| PSM | Description |
|-----|-------------|
| 3 | Fully automatic page segmentation (default) |
| 6 | Assume a single uniform block of text |
| 11 | Sparse text — find as much text as possible |
| 13 | Raw line — treat as a single text line |

### Dependencies

- **tesseract** (system CLI) — must be installed. Via brew it is pulled in automatically as a dependency. On Linux: `apt-get install tesseract-ocr`.

## Testing

- Unit tests added in `YamlCommandReaderTest` covering all three commands in simple string form, full selector form, and the new `psm` field
- Integration tests added in `IntegrationTest` using a `FakeDriver` subclass that injects a Java2D-rendered test image as the screenshot, verifying correct OCR parsing and tap coordinate resolution — including a dedicated test for `tapOnOcr` with `psm` set
- All integration tests are guarded with `Assumptions.assumeTrue(isTesseractAvailable())` so they are skipped cleanly in environments where tesseract is not installed

## Issues fixed

N/A — this is a new feature addition.


Fishbowler:
I'm uneasy with commands that rely on a binary already being installed. I also think you'd need to cater for Windows users. Right now, the only place Maestro differs on host OS is Apple Simulator support, and that's an upstream from Apple, not Maestro.

Kardelio:
> I'm uneasy with commands that rely on a binary already being installed. I also think you'd need to cater for Windows users. Right now, the only place Maestro differs on host OS is Apple Simulator support, and that's an upstream from Apple, not Maestro.

hey @Fishbowler 
Thanks for your response matey!

So I first time contributor so definitely want to get this right, so always open & grateful for all feedback.

Can I ask overall are you a fan of the `idea` of this? for when Maestro cant access elements that are clearly on screen utilising something like an OCR lib to see the screen itself? or is this a none-starter?

Also tesseract itself is open source if that helps at all? https://tesseract-ocr.github.io/

If the external lib is not an ok solution but you want to go ahead with this feature can i ask what would be your preferred solution? as I think ofcourse implementing an own OCR is going to be a lot of work and surely something that you probably dont want right, so would make sense to rely on a 3rd party OSS that exists or?

What are your thoughts on any/all of that?

Fishbowler:
We've got the [extractTextWithAI](https://docs.maestro.dev/reference/commands-available/extracttextwithai) command already that does a subset of this already (but has external dependencies on Maestro's Cloud APIs)

So clearly the value is there. 

I wonder if one of the Tesseract wrappers, like Tess4J or bytedeco's tesseract-platform would be a better approach? But both would require bundling not just the binaries but the training data to support it. Could it be a bit much?

Kardelio:
> We've got the [extractTextWithAI](https://docs.maestro.dev/reference/commands-available/extracttextwithai) command already that does a subset of this already (but has external dependencies on Maestro's Cloud APIs)

- Ye wanted an approach that could just be run locally without the need for Maestro Cloud APIs 😄 

> So clearly the value is there.

- Awesome to hear!

> I wonder if one of the Tesseract wrappers, like Tess4J or bytedeco's tesseract-platform would be a better approach? > But both would require bundling not just the binaries but the training data to support it. Could it be a bit much?

- OK interesting, to be honest I would look towards you and your guidance in this matter and how you would prefer me to progress / tweak / alter just due to my newness in contributing to this project, I want to see and understand how you would go about continuing on with this feature and then I can go ahead and do it 😄 if thats ok?

Kardelio:
UPDATED 

* added PSM params to OCR commands to allow finer grained control

PR #3205, opened 2026-04-22
feat(mcp): validate text selectors against latest hierarchy snapshot

pedro18x:
Closes [MA-4029](https://linear.app/mobile-dev/issue/MA-4029/mcp-validate-text-selectors-against-the-latest-view-hierarchy-before).

## Why

Simon hit two failure modes authoring flows through the MCP:

1. **Hallucinated text.** `tapOn: { text: "Favorite" }` on the podcast screen — no element on iOS carries that label (the agent read it off a screenshot). Runtime error was the unhelpful "element not found."
2. **Truncated text.** `tapOn: { text: "RNR 352" }` when the actual element text is `"RNR 352 - Expo Launch with Cedric van Putten"`. Maestro's `text:` matcher uses full-string `regex.matches`, so the partial string misses — and the runtime error again is just "element not found."

Both failures waste an agent cycle (or several) before the caller figures out the selector is wrong. Instructions-only nudges help but don't catch this deterministically. This PR adds a server-side pre-flight that catches both cases with a structured error pointing at the actual texts on screen.

## What this does

- `inspect_view_hierarchy` records the current `TreeNode` root into a per-device `HierarchySnapshotStore`. In-memory, last-write-wins, no persistence.
- `run` cross-checks every `text:` selector in the flow (inline YAML + `files` + `dir` modes, including nested `below` / `above` / `leftOf` / `rightOf`) against the latest snapshot for the target device. Matching delegates to `Filters.textMatches` with the same `RegexOption`s Orchestra uses (`IGNORE_CASE`, `DOT_MATCHES_ALL`, `MULTILINE`) — so a verdict here matches the runner's verdict.
- On miss, returns `isError=true` with a structured JSON payload naming the missing selectors and the closest on-screen texts (bidirectional substring overlap, falling back to a peek list of what *is* on screen). Agent can self-correct and re-run.
- Selectors containing `${...}` env-var interpolation are skipped — those get resolved downstream and can't be matched against a raw snapshot.
- Opt-out: `skip_selector_validation: true` for flows that legitimately target dynamically-rendered text.
- No snapshot for the device → validation silently skipped (the instructions already tell the agent to inspect first).

## Scope

v1 validates `text:` only. `id:` and accessibility-identifier selectors are out of scope; they have platform-split semantics that warrant their own follow-up.

## Refactors prompted by review

- Initial pass used a hand-rolled matcher with 8 attributes and `containsMatchIn` (partial). Maestro's real `Filters.textMatches` only looks at `text` / `hintText` / `accessibilityText` and uses full-string `regex.matches`. Reusing the real filter eliminated semantic drift; as a side effect it started catching the `RNR 352` truncation case too.
- Review also flagged: mismatched `RegexOption`s vs the runner, env-var interpolation false positives, a silent file-read catch, and empty-findings invariant on `Result.Miss`. All addressed.

## Verified

- `./gradlew :maestro-cli:test` — green.
- Smoke on a booted iPhone 17 sim: inspect home screen → run with `text: "AbsolutelyNotOnScreen"` returns `selector_not_on_screen` with suggestions drawn from the real home screen (`["Fitness","Watch","Contacts","Files","Preview",...]`); same flow with `skip_selector_validation: true` bypasses the validator and falls through to Maestro's own parser.
- MCP `instructions` string stays under the 2KB cap.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/McpServer.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/McpServer.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.io.asSink
 import kotlinx.io.asSource
 import kotlinx.io.buffered
+import maestro.cli.mcp.hierarchy.HierarchySnapshotStore
 import maestro.cli.session.MaestroSessionManager
 import maestro.debuglog.LogConfig
 import maestro.cli.mcp.tools.ListDevicesTool
@@ -24,7 +25,7 @@ import maestro.cli.util.WorkingDirectory
 import java.io.PrintStream
 
 internal val INSTRUCTIONS = """
-    Maestro MCP authors, edits and runs UI tests via declarative YAML flows on Android emulators, iOS simulators, Chromium browsers, or Maestro Cloud. Use when the user wants to write, run, or debug a mobile or web UI test, reproduce a bug, or self-validate a user-facing change you just built.
+    Maestro MCP authors, edits and runs UI tests (YAML flows) on Android, iOS, Chromium, or Maestro Cloud. Use for writing, running, or debugging a mobile/web UI test, reproducing a bug, or self-validating a UI change.
 
     Every local tool (`take_screenshot`, `inspect_view_hierarchy`, `run`) needs a `device_id` from `list_devices` first.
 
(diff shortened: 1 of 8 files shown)


pedro18x:
Closing in favor of a simpler instructions-only approach, per discussion with @lelandjrichardson. Pre-flight validation catches the narrow cases Simon reported but adds regression risk on multi-screen flows we can't measure without evals. Revisiting MA-4029 as warning-mode once evals land.

PR #3206, opened 2026-04-22
fix(ios-runner): suppress XCTest UI-interruption preflight hang

amanjeetsingh150:
## Problem

We found situations in a few tests where our internal iOS driver got stuck with these last log lines:

```
2026-04-15 04:43:55.803 maestro-driver-iosUITests-Runner[3186:91871415] Quiescence checks are disabled for com.xxxx.xxxx application. Making it to believe it is idling
    t =   152.24s     Find the Any (First Match)
    t =   152.25s     Check for interrupting elements affecting Window
    t =   152.27s     Found 1 interrupting element:
    t =   152.27s         Find the Alert (Identity Binding)
    t =   152.29s Alert from Application 'com.xxxx.xxxx'
    t =   152.29s     Invoking UI interruption monitors for Alert from Application 'com.xxxx.xxxx'
    t =   152.29s         Find the Alert
    t =   152.31s         Checking existence of `Button (Elements matching predicate 'userTestingAttributes CONTAINS "cancel-button"')`
    t =   152.33s         Get all elements bound by accessibility element for: Elements matching predicate 'userTestingAttributes CONTAINS "default-button"'
    t =   152.37s         Checking existence of `Button (Elements matching predicate 'label CONTAINS[d] "Don’t" OR label CONTAINS "Cancel"')`
    t =   152.39s         Requesting snapshot of accessibility hierarchy for app with pid 3322
    t =   152.43s         Find: Descendants matching type Alert
    t =   152.43s         Find: Identity Binding
    t =   152.46s         Requesting snapshot of accessibility hierarchy for app with pid 3322
    t =   152.51s         Find: Descendants matching type Alert
    t =   152.51s         Find: Identity Binding
    t =   152.51s         Requesting snapshot of accessibility hierarchy for app with pid 3322
    t =   152.56s         Find: Descendants matching type Alert
    t =   152.56s         Requesting snapshot of accessibility hierarchy for app with pid 3322
```

A few things to note, note "**Invoking UI interruption monitors**"

- We don't register any UI interruption monitors in Maestro.
- Turns out this is still called from within Apple's `XCUIAutomation` framework — treat it as a preflight check during the presence of alerts that runs before every gesture.
- To capture this evidence I ran:

  ```
  strings /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/Library/Frameworks/XCUIAutomation.framework/XCUIAutomation | grep -E 'cancel-button|default-button|userTestingAttributes CONTAINS'
  ```

- By design, Maestro is supposed to take charge of dialog handling via `simctl`, `applesimutils`, or explicit taps — not via XCTest's interruption monitors.

In a few cases the accessibility snapshot inside this preflight has been causing the internal driver to deadlock and ultimately http request timing out in 200s:

```
11:43:55  Command 50 started: Swiping in UP direction in 500 ms
11:47:15  Command failed                                   ← 200s OkHttp timeout
11:50:35  Attempt 1 failed for retry command                ← 200s (retries everything)
11:53:55  Got socket timeout processing takeScreenshot command ← 200s (from onCommandStart)
11:57:11  WATCHDOG: Timeout exceeded (930000ms) — interrupting blocked thread
11:57:16  Test timed out after 15 minutes (cooperative timeout)
```

There is another issue that maestro keeps hitting a dead XCUITest server which we should have hard failed, followup PR coming for that.

## Fix

This PR fixes the main issue by returning `YES` from `-[XCUIApplication doesNotHandleUIInterruptions]`. Doing this, we avoid any side effects that can happen through Apple's own UI-interrupt handler code path. Changes:

- Reproduced the preflight invocation with a minimal SwiftUI demo app under `e2e/alert-repro-swiftui/` — that comes first.
- Then the fix: `+load` swizzle in `XCUIApplication+Helper.m` that replaces the getter so XCTest skips the preflight for every `XCUIApplication` instance (including ones created internally). No call-site changes.
- Then wires the regression test into the existing E2E workflow.

I also found appium disabled it as well

https://github.com/appium/WebDriverAgent/blob/2ca4503b6dacfdd1947ba8b8604d76ea91ad0de4/WebDriverAgentLib/Categories/XCUIApplication%2BFBUIInterruptions.m#L16

## Testing

- `e2e/alert-repro-swiftui/verify_preflight_suppressed.sh` — builds the fixture app, runs a swipe through Maestro, asserts zero preflight signature lines in the `xctest_runner` log. Fails pre-fix, passes post-fix.
- A/B verified on the original customer flow on an iOS 18.2 simulator:

  | Preflight log signature | Pre-fix | Post-fix |
  |---|---|---|
  | `Check for interrupting elements affecting Window` | 9 | **0** |
  | `Found N interrupting element` | 5 | **0** |
  | `Invoking UI interruption monitors` | 5 | **0** |
  | `Alert from Application` | 10 | **0** |

- Regression wired into `.github/workflows/test-e2e.yaml`'s `test-ios` job, so every PR exercises it going forward.


.github/workflows/check-drivers.yaml
```diff
@@ -47,27 +47,64 @@ jobs:
...
+                  fi
+              fi
+          done
+          if [ "$fail" -ne 0 ]; then
+              echo ""
+              echo "ERROR: iOS drivers are out of date."
+              echo "Run the following and commit the updated drivers:"
+              echo "  sh maestro-ios-xctest-runner/build-maestro-ios-runner-all.sh"
+              exit 1
+          fi
```
amanjeetsingh150, line 110: This workflow had an issue so fixed the comparison, I'm gonna redo this workflow by more robust comparing with checksums instead, not including that scope in this PR.

.github/workflows/test-e2e.yaml
```diff
@@ -340,6 +340,32 @@ jobs:
...
+                      print(d["udid"])
+                      sys.exit(0)
+          ')
+          if [ -z "$SIM_UDID" ]; then
+            echo "error: no booted iOS simulator found" >&2
+            xcrun simctl list devices booted
+            exit 1
+          fi
+          echo "Using booted simulator: $SIM_UDID"
+          ./e2e/alert-repro-swiftui/verify_preflight_suppressed.sh "$SIM_UDID"
```
steviec: this seems like it's begging for a script to be run instead of lots of code to maintain in GHA

steviec (approved):
It seems like we're building up more e2e tests; is our current setup a flexible one to do this? Should we improve in any way? (we don't have to solve this now, just seems like we're adding a bunch of one-offs for specific tests)

.github/workflows/test-e2e.yaml
amanjeetsingh150: Done simplified ✅

amanjeetsingh150:
> It seems like we're building up more e2e tests; is our current setup a flexible one to do this? Should we improve in any way? (we don't have to solve this now, just seems like we're adding a bunch of one-offs for specific tests)

You're right, there would be future when we could have multiple sample apps on different platform so we need better scripting here. I'll get this feedback with stream of E2E on demo app for iOS.

PR #3208, opened 2026-04-22
Adds a soft CLI warning when the workspace zip produced by maestro cloud exceeds 20 MB

proksh:
A sweep of the last 20 days of cloud uploads showed 37 orgs shipping workspaces over 20 MB, with several in the 300–670 MB range. Because every device downloads the full workspace on every run, oversized uploads directly slow down queue and run times across the fleet.

The dominant causes were app binaries (.apk / .ipa / .app / .aab) bundled alongside flows, .git/ directories, build outputs (build/, .gradle/, node_modules/, Pods/), re-uploaded test artifacts (reports/, recordings/, artifacts/), and duplicated fixtures.

It will be very tricky to either blacklist or whitelist the file behind the scene, so we would just warn users and its upto users how they wanna maintain the workspace, knowing larger size will slow things down.

Screenshot 2026-04-22 at 10 34 07 AM

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -1,4 +1,4 @@
-maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.ktpackage maestro.cli.cloud
+package maestro.cli.cloud
 
 import maestro.cli.CliError
 import maestro.cli.analytics.Analytics
--- a/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
+++ b/maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
@@ -132,6 +132,7 @@ class CloudInteractor(
                 out = workspaceZip,
                 configOverride = configFile?.toPath()?.absolute(),
             )
+            warnIfWorkspaceIsLarge(workspaceZip.toFile().length())
             val progressBar = ProgressBar(20)
 
             // Binary id or Binary file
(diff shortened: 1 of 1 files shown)


steviec (approved):
Love it. My only question is whether 10MB is too low; is that a legimate size if people are bundling media to install or other things? Don't want to add annoying warnings for "normal" cases. I'd probably bump to 50MB. But if you think 10 is good I'm fine keeping it.

proksh:
> Love it. My only question is whether 10MB is too low; is that a legimate size if people are bundling media to install or other things? Don't want to add annoying warnings for "normal" cases. I'd probably bump to 50MB. But if you think 10 is good I'm fine keeping it.

I see 2 orgs with 13Mbs and 20Mbs size which has images/videos. So I have updated the warning to happen above 20Mb. Other than that, all the orgs have some large files and folders which should not have been included.

Fishbowler:
Where this occurs in CI, will anyone read it?
Nobody reads build warnings, so a slight performance impact on tests in the order of "some megabytes" is unlikely to spring folks into action.

I disagree with some others - I think there are valid cases for filtering some stuff out of workspace uploads (loudly, so that they know what's been removed)

That being said, this should probably still exist for now. It gives folks a chance to fix their weird workspaces.

proksh:
> Where this occurs in CI, will anyone read it? Nobody reads build warnings, so a slight performance impact on tests in the order of "some megabytes" is unlikely to spring folks into action.
> 
> I disagree with some others - I think there are valid cases for filtering some stuff out of workspace uploads (loudly, so that they know what's been removed)
> 
> That being said, this should probably still exist for now. It gives folks a chance to fix their weird workspaces.

Got it. I will merge this one, but lets talk about what you think should be removed for sure and I can think about it and tackle it as a followup. Amanjeet had similar concerns about this.

PR #3212, opened 2026-04-23
fix(ios-driver): fail-fast on transport failure instead of rethrowing raw SocketTimeoutException

amanjeetsingh150:
## Problem

Investigation during #3206 surfaced this: the iOS XCTest driver can deadlock mid-job (a `swipe` in our case). Although it's fixed elsewhere, it revealed issues in our driver client.

The client receives `SocketTimeoutException` after the 200s `readTimeout`, but every subsequent driver HTTP call hits the same dead runner and eats another 200s — wasting ~10 min of compute.

Two compounding behaviours feed the cascade:

- `onCommandStart` / `onCommandFailed` (screenshot, hierarchy) re-issue HTTP to the dead runner, each burning another 200s
- Flow-level `retry` in Orchestra replays on any throwable, including transport failures, so each retry attempt starts fresh HTTP calls against the same dead runner

Concrete sequence we observed:

1. `driver.swipe` → `runDeviceCall` → 200s `SocketTimeoutException`
2. `onCommandFailed` calls `deviceInfo` to snapshot hierarchy → 200s
3. `retry` wrapper replays the subflow
4. `onCommandStart` calls `screenshot` on the replayed command → 200s
5. ...repeats until the 15-min cooperative `withTimeout` fires and `runInterruptible` interrupts the IO-pool worker, unsticking `socket.read()`

## Fix

Treat unreachable driver as a first-class exception that consumers can catch. Fail fast and short-circuit subsequent HTTP calls instead of wasting ~10 min of compute on dead runners.

The latch lives at the **transport layer** (`XCTestDriverClient`), where the `OkHttpClient` and `restartXCTestRunner()` already live. Translation propagates up through the existing exception vocabulary so each layer only speaks its own language and `SocketTimeoutException` never leaks above OkHttp.

| Layer | Catches | Throws | Owns |
| --- | --- | --- | --- |
| `XCTestDriverClient` | `SocketTimeoutException` | `XCUITestServerError.Unreachable` | OkHttp client, `@Volatile` latch, `restartXCTestRunner()` reset |
| `XCTestIOSDevice.execute` | `XCUITestServerError.Unreachable` | `IOSDeviceErrors.Unreachable` | Device-abstraction translation |
| `IOSDriver.runDeviceCall` | `IOSDeviceErrors.Unreachable` | `DeviceUnreachableException` | Maestro-facing driver contract |

What this buys:

- **Coverage gap closed.** `launchApp`, `stopApp`, `clearAppState`, `clearKeychain`, `close` all go through the same OkHttp client now, so they trip the latch automatically — they couldn't trip the previous IOSDriver-level latch because they bypassed `runDeviceCall`.
- **Recovery story.** `restartXCTestRunner()` clears the latch (recovery is the only thing that should). `close()` deliberately does not — the transport really is gone, and the reopen path goes through `restartXCTestRunner()` anyway.
- **No cross-layer types.** `SocketTimeoutException` is no longer imported above `XCTestDriverClient`. The OkHttp dependency stops at the layer that actually talks to OkHttp.
- **`DeviceUnreachableException`** stays a plain `RuntimeException`, not a `MaestroException`, since it's an infra failure — consumers (worker-level `INFRA_ERROR` classification, Orchestra's retry) can decide how to handle it.

## Verify

- [x] `:maestro-utils:test` — `XCUITestServerError.Unreachable` sealed variant
- [x] `:maestro-ios:test` — new `XCTestIOSDeviceTest` pins the translation `XCUITestServerError.Unreachable` → `IOSDeviceErrors.Unreachable`
- [x] `:maestro-client:test` — `XCTestDriverClientTest` pins trip-on-timeout, short-circuit-without-HTTP, restart-clears-latch, non-transport-doesn't-trip; `IOSDriverTest` pins the final `IOSDeviceErrors.Unreachable` → `DeviceUnreachableException` translation

### DIFF
--- a/maestro-client/build.gradle.kts
+++ b/maestro-client/build.gradle.kts
@@ -170,6 +170,7 @@ dependencies {
     testImplementation(libs.google.truth)
     testImplementation(libs.square.mock.server)
     testImplementation(libs.junit.jupiter.params)
+    testImplementation(libs.mockk)
 }
 
 java {
--- a/maestro-client/src/main/java/maestro/DeviceUnreachableException.kt
+++ b/maestro-client/src/main/java/maestro/DeviceUnreachableException.kt
@@ -1,0 +1,15 @@
+package maestro
+
+/**
+ * Thrown when a driver call fails because the underlying device transport (e.g. the iOS XCTest
+ * runner's HTTP socket) has stopped responding. Distinct from [MaestroException] — this is an
+ * infrastructure failure, not a test failure, and consumers should treat it accordingly
+ * (no test-error reporting, no flow-level retry).
+ *
+ * Once a driver records a transport failure it should fail-fast on subsequent calls instead of
+ * issuing fresh requests against the same dead transport.
+ */
+class DeviceUnreachableException(
+    val callName: String,
+    cause: Throwable,
+) : RuntimeException("Device became unreachable during $callName", cause)
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -26,6 +26,7 @@ import ios.IOSDeviceErrors
 import maestro.Capability
 import maestro.DeviceInfo
 import maestro.device.DeviceOrientation
+import maestro.DeviceUnreachableException
 import maestro.Driver
 import maestro.Filters
 import maestro.KeyCode
@@ -67,6 +68,14 @@ class IOSDriver(
     private var appId: String? = null
     private var proxySet = false
     private val xcRunnerCLIUtils = XCRunnerCLIUtils(tempFileHandler = TempFileHandler())
+
+    // Set on the first transport-level failure (socket timeout against the XCTest runner).
+    // Subsequent runDeviceCalls fail-fast against this instead of issuing fresh HTTP requests
+    // to a runner we already know isn't answering. Volatile because runDeviceCall executes on
+    // Dispatchers.IO worker threads (via runInterruptible in maestro.Maestro) — the writer and
+    // any future reader may be different pool workers, so JMM visibility matters.
+    @Volatile
+    private var transportFailure: DeviceUnreachableException? = null
 
     override fun name(): String {
         return metrics.measured("name") {
@@ -542,11 +551,17 @@ class IOSDriver(
     }
 
     private fun <T> runDeviceCall(callName: String, call: () -> T): T {
+        transportFailure?.let { throw it }
         return try {
             call()
         } catch (socketTimeoutException: SocketTimeoutException) {
-            LOGGER.error("Got socket timeout processing $callName command", socketTimeoutException)
-            throw socketTimeoutException
+            val tripped = DeviceUnreachableException(callName, socketTimeoutException)
+            transportFailure = tripped
+            LOGGER.error(
+                "Got socket timeout processing $callName command, marking driver unreachable",
+                socketTimeoutException
+            )
+            throw tripped
         } catch (appCrashException: IOSDeviceErrors.AppCrash) {
             LOGGER.error("Detected app crash during $callName command", appCrashException)
             throw MaestroException.AppCrash(appCrashException.errorMessage)
--- a/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
@@ -54,7 +54,6 @@ import okio.source
 import org.slf4j.LoggerFactory
 import util.XCRunnerCLIUtils
 import java.io.File
-import java.net.SocketTimeoutException
 import kotlin.collections.set
 
 class IOSDriver(
@@ -68,14 +67,6 @@ class IOSDriver(
     private var appId: String? = null
     private var proxySet = false
     private val xcRunnerCLIUtils = XCRunnerCLIUtils(tempFileHandler = TempFileHandler())
-
-    // Set on the first transport-level failure (socket timeout against the XCTest runner).
-    // Subsequent runDeviceCalls fail-fast against this instead of issuing fresh HTTP requests
-    // to a runner we already know isn't answering. Volatile because runDeviceCall executes on
-    // Dispatchers.IO worker threads (via runInterruptible in maestro.Maestro) — the writer and
-    // any future reader may be different pool workers, so JMM visibility matters.
-    @Volatile
-    private var transportFailure: DeviceUnreachableException? = null
 
     override fun name(): String {
         return metrics.measured("name") {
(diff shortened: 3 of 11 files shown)


steviec:
Taking a step back on where this latch belongs architecturally. Laying out the full stack first:

## The full call stack, annotated

```
maestro-worker                       (owns: retry decision at job level, INFRA_ERROR classification)
  └─ MaestroTestRunner.execute
     └─ Orchestra.runFlow              (owns: retryCommand, onCommandStart/Failed callbacks)
        └─ Maestro.tap()  [suspend]    (owns: runInterruptible for cancellation propagation)
           └─ runInterruptible(Dispatchers.IO) {
                driver.tap(point)      (blocking)
              }
              └─ IOSDriver.tap()       (owns: Maestro-facing exception mapping)
                 └─ runDeviceCall("tap") { iosDevice.tap(...) }  ← this PR latches here
                    └─ LocalIOSDevice.tap()       (facade over two device backends)
                       ├─ xcTestDevice = XCTestIOSDevice         (HTTP path)
                       │  └─ execute { client.tap(...) }         (XCUITestServerError → IOSDeviceErrors)
                       │     └─ XCTestDriverClient.tap()         (owns: OkHttp client, 200s timeout, restartXCTestRunner)
                       │        └─ okHttpClient.execute()        (the actual TCP socket read)
                       │           └─ blocks up to 200s, throws SocketTimeoutException
                       │
                       └─ deviceController = SimctlIOSDevice     (subprocess path — launch/install/clearAppState)
                          └─ LocalSimulatorUtils.launch(...)     (owns: xcrun/simctl subprocess)
                             └─ ProcessBuilder(...).waitFor()    ← different failure mode entirely
```

Two things pop out:

**(a) `LocalIOSDevice` is a facade over two physically different transports.** `xcTestDevice` talks to the XCTest runner over HTTP. `deviceController` talks to simctl/applesimutils via subprocess. The method list is split between them: `tap`/`swipe`/`screenshot`/etc go through XCTest HTTP; `launch`/`install`/`clearAppState`/`clearKeychain` go through simctl subprocess. They have **different failure modes**. The XCTest runner stopping → `SocketTimeoutException` at 200s. A stuck `xcrun simctl` subprocess → `Process.waitFor()` hang with *no timeout at all* (separate issue in `LocalSimulatorUtils.kt:721`).

This PR only addresses one transport (HTTP), and it addresses it at a layer that can't tell the two apart.

**(b) Every layer already has a well-scoped exception vocabulary.**

| Layer | Catches | Throws | What it owns |
| --- | --- | --- | --- |
| `XCTestDriverClient.handleExceptions` | HTTP codes, response bodies | `XCUITestServerError.{AppCrash, OperationTimeout, BadRequest, UnknownFailure}` | The transport |
| `XCTestIOSDevice.execute` | `XCUITestServerError` | `IOSDeviceErrors.{AppCrash, OperationTimeout}` | The device abstraction |
| `IOSDriver.runDeviceCall` | `SocketTimeoutException` + `IOSDeviceErrors` | `MaestroException.{AppCrash, DriverTimeout}` + **now** `DeviceUnreachableException` | Maestro-facing driver contract |
| `Maestro.xxx()` | — | propagates, after `runInterruptible` unwraps | suspend/cancellation boundary |

The layering is otherwise clean. This PR teaches the driver layer about a transport-layer concern (socket timeout). It has to, because it catches `SocketTimeoutException` directly — an OkHttp type that shouldn't leak above `XCTestDriverClient`. That `SocketTimeoutException` import in `IOSDriver.kt:547` is the layering violation in plain sight.

## What the layering principles dictate

"Transport is dead" is a property of the transport. Not the driver, not the device abstraction:

- The **state** ("runner stopped answering") belongs to `XCTestDriverClient` because it's the object that holds the `OkHttpClient`, owns the 200s `readTimeout`, and owns `restartXCTestRunner()`.
- The **signal** that trips it (`SocketTimeoutException`, also `ConnectException`, `EOFException`, HTTP 503, etc.) originates inside `XCTestDriverClient.executeJsonRequest` at the OkHttp call. `SocketTimeoutException` is one of several ways "transport died" can manifest; the others would benefit from the same fail-fast.
- The **recovery** (`restartXCTestRunner()`) is also `XCTestDriverClient`'s method. If the latch lives here, restart can clear it. That turns a write-once poison into a real circuit breaker.
- The **vocabulary** callers need is already defined here — add one sealed variant to `XCUITestServerError` (say `Unreachable`) or a new `XCTestTransportException`, and let it propagate up through the existing translation chain. No layer-jumping types.

## The correct placement

Latch at `XCTestDriverClient`. Translate up through the existing chain. Generalize the concept.

```kotlin
class XCTestDriverClient(...) {
    @Volatile private var transportDead: XCTestTransportException? = null

    private fun <T> transportCall(callName: String, call: () -> T): T {
        transportDead?.let { throw it }
        return try {
            call()
        } catch (e: SocketTimeoutException) { markDead(callName, e) }
          catch (e: ConnectException)       { markDead(callName, e) }
          // and any other transport-level signals we want to treat as dead
    }

    private fun markDead(callName: String, cause: IOException): Nothing {
        val dead = XCTestTransportException(callName, cause)
        transportDead = dead
        throw dead
    }

    fun restartXCTestRunner() {
        // existing body
        transportDead = null   // <-- recovery clears the latch
    }
}
```

Every `executeJsonRequest` in this class goes through `transportCall`. `XCTestIOSDevice.execute` catches `XCTestTransportException` and translates to `IOSDeviceErrors.TransportUnreachable`. `IOSDriver.runDeviceCall` catches `IOSDeviceErrors.TransportUnreachable` and translates to `DeviceUnreachableException`. Each layer only speaks its own vocabulary. The `SocketTimeoutException` import never appears above the transport module.

What this buys over the current placement:

1. **Automatic coverage of the launchApp/stopApp/clearAppState gap.** Several `IOSDriver` methods bypass `runDeviceCall` entirely — `launchApp` (IOSDriver.kt:99), `stopApp` (:109), `killApp` (:115), `clearAppState` (:122), `clearKeychain` (:128), `close` (:83). Under the current PR, `SocketTimeoutException` during `launchApp` propagates raw (not as `DeviceUnreachableException`) and the flag is never set, so the next `tap` or `snapshot` still burns 200s. With the latch at the transport, coverage is complete by construction — all HTTP calls go through the same client.

2. **Recovery story.** `restartXCTestRunner()` clears the latch; a flow can legitimately continue after a runner restart. Today's placement silently poisons the driver forever — caller has to throw it away. That matches the worker's retry model but breaks anyone else (Studio, CLI reuse).

3. **Extends to other transport failures.** `ConnectException`, `EOFException`, prolonged HTTP 503 bursts — same conceptual signal. Catching more of them at the right layer is a one-line change.

4. **No cross-layer types.** `SocketTimeoutException` stops existing above `XCTestDriverClient`. `DeviceUnreachableException` doesn't need to know about OkHttp.

5. **Generalizes to the simctl side.** `LocalSimulatorUtils` is the *other* transport — where `xcrun simctl launch` and `applesimutils` live, with the unbounded `Process.waitFor()` hangs. The same pattern applies: latch on repeated subprocess timeouts (once bounded `waitFor` timeouts are added), fail-fast, let the caller decide. This is impossible in `IOSDriver.runDeviceCall` because that's below the point where the two transports merge.

6. **Mirrors cleanly to Android.** `AndroidDriver.runDeviceCall` (line 1250) catches `StatusRuntimeException` with `UNAVAILABLE` / `DEADLINE_EXCEEDED` and just rethrows. The same shape applies at the transport layer — gRPC `ManagedChannel` is the equivalent of `XCTestDriverClient`'s OkHttp. The fix there is latching at the gRPC client, not at `AndroidDriver.runDeviceCall`.

## Circuit breaker

The pattern has a name: **circuit breaker**, with three states:
- *Closed* — normal operation, calls go through
- *Open* — tripped, calls short-circuit with a fast exception
- *Half-open* — after some cooldown, let one call through; success → closed, failure → back to open

This PR implements a degenerate two-state version (Closed → Open, never re-closes). Defensible as a starting point, but it doesn't fully earn the name. A proper circuit breaker at `XCTestDriverClient` with `restartXCTestRunner()` as the reset is the architecturally-complete version. Copilot already has `CircuitBreaker.kt` at the job level (opens after 3 `INFRA_ERROR` failures, pauses the worker). Mirroring at the transport level is consistent with an existing pattern.

## Adjacent fixes that belong with this

- `OkHttpClientInstance.kt:9-15` — no `callTimeout` set; defaults to infinite. The 200s `readTimeout` is the only bound, and a server that trickles a byte every 199s will never time out. Fixing the latch without this leaves a more subtle version of the same hang.
- `LocalSimulatorUtils.kt:721` (and `:671`, plus four `waitFor()` calls in `XCRunnerCLIUtils.kt:61-76`) — unbounded `Process.waitFor()` on screen recording stop and `networksetup`. Same class of issue on the simctl transport.
- `AndroidDriver.runDeviceCall:1250` — mirror the pattern at the gRPC client layer.

PR #3215, opened 2026-04-23
fix(orchestra): retryCommand only retries on MaestroException, propagates others

amanjeetsingh150:
## Problem

`retryCommand` today catches `Throwable` and replays the subflow on any exception. 

Retry is meant for test-level flake (element not found, assertion failures, etc.) which surface as `MaestroException` subtypes. 

Everything else, driver transport failures, JS evaluation bugs, unexpected infra errors should propagate immediately so the worker can classify the failure at the job level.

## Fix

`catch (Throwable)` branch now rethrows non-`MaestroException` immediately without incrementing the attempt counter. `CancellationException` handling unchanged.

## Verify

- [x] New: `retryCommand only retries on MaestroException, propagates other throwables without retrying` — RuntimeException from tap handler surfaces on first attempt, `tapCount == 1`.
- [x] New: `retryCommand retries on MaestroException until success` — `MaestroException.UnableToClearState` on attempt 1 triggers a retry that succeeds, `tapCount == 2`.
- [x] Updated: Case 119 swapped its synthetic `RuntimeException` for `MaestroException.UnableToClearState` so it still exercises the retry path under the corrected semantics.
- [x] `./gradlew :maestro-test:test` — 202 tests, all green
- [x] `./gradlew :maestro-orchestra:test` — green

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -850,13 +850,16 @@ class Orchestra(
     private suspend fun retryCommand(command: RetryCommand, config: MaestroConfig?): Boolean {
         val maxRetries = (command.maxRetries?.toIntOrNull() ?: 1).coerceAtMost(MAX_RETRIES_ALLOWED)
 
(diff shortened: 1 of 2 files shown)


steviec (approved):
I want to give this PR a kiss

PR #3217, opened 2026-04-24
Add support for MANAGE_EXTERNAL_STORAGE permission on Android

Fishbowler:
## Proposed changes

MANAGE_EXTERNAL_STORAGE is a special permission ([docs](https://developer.android.com/training/data-storage/manage-all-files)).

This adds the first permission granted via `appops` rather than via `pm grant`.

## Testing

Added permission declaration to demo_app, and added an e2e test.

### DIFF
--- a/e2e/demo_app/.maestro/all_files_permission.yaml
+++ b/e2e/demo_app/.maestro/all_files_permission.yaml
@@ -1,0 +1,29 @@
+appId: com.example.example
+tags:
+  - passing
+  - android
+---
+- launchApp:
+    clearState: true
+    permissions:
+      allFiles: allow
+
+# Verify the permission is granted via the Settings app
+- launchApp: com.android.settings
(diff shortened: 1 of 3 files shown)


Fishbowler:
I've added commits to:
* rename the permissions
* alphabetise some permissions for maintainability
* fix a bug where 'unset' was being mapped to explicit deny - that was probably fine for this permission, since it's binary, but in the future that might not be
* update the e2e test to cover `all` vs explicit permissions, and to include the `SetPermissions` command.

Fishbowler:
After discussion with @Leland-Takamine, I've removed the friendly label for this. It's not needed. It originates from iOS having simple terms, and mapping between the two to keep flows portable. This has no iOS equivalent.

e2e test recording:
output

Leland-Takamine (approved):

PR #3219, opened 2026-04-24
Don't assume lowercase .yaml extension when parsing flownames

Fishbowler:
## Proposed changes

When using `flowsOrder` with flows that have no `name` property, you specify the name of the flow, without the file extension.

When preparing the run order, if your file is `login.yml` instead of `login.yaml`, the flow wouldn't be included in the sequential run, due to a line that expected a case-sensitive and explicit `.yaml` extension.

This change strips the file extension completely - we already know it's a Maestro flow at this point.

## Testing

Added a unit test that failed, which now doesn't.


pedro18x (approved):
LGTM

PR #3226, opened 2026-04-28
ci(test-e2e): simplify Android E2E to launch new android versions

amanjeetsingh150:
## Context

When validating Maestro against a newly-launched Android version (e.g. android-35, android-36), I needed to dispatch test-e2e.yaml on demand, narrowed to the API level, app, and even single flow I was debugging. 

The existing workflow took only raw sdkmanager package strings as workflow_dispatch inputs — fine for one-off use, painful for matrix validation across versions.

This PR turns the workflow into a usable manual-trigger UX while keeping pull_request: behavior byte-identical.

## Changes

1. Three inputs on workflow_dispatch:

- android_version: defaults android-32, can be android-35, android-36
- app: defaults demo_app, can be other apps we have in e2e
- flow: defaults empty, that means every flow, Optional command name (e.g. setLocation) if you want to assert single command.

A new validate-inputs job runs first and fails fast (<30s) on bad input, before the build/emulator-boot work begins:
- android_version must be > android-29
- app must resolve to an existing workspace (lists available apps on miss)
- flow must resolve to exactly one yaml under the chosen app (lists available flows on miss)

2. Removed platform image, those were downloading unnecessarily its unused, just used during local android development
3. Added a concurrency on E2E android job that would cancel any in progress job when there are new pushes on same reference.


.github/workflows/test-e2e.yaml
```diff
@@ -2,11 +2,106 @@ name: Test E2E
...
 on:
   workflow_dispatch:
+    inputs:
+      android_version:
+        description: 'Android API level to test against'
+        required: false
+        type: choice
+        default: 'android-32'
+        options:
+          - 'android-32'
```
Fishbowler, line 12: We don't support android-32, and shouldn't have defaulted to it before. That was Android 12L. https://apilevels.com/

I've set it to 33 here - it's the current Maestro Cloud default.

```suggestion
        default: 'android-33'
        options:
          - 'android-30'
          - 'android-31'
          - 'android-33'
```

.github/workflows/test-e2e.yaml
```diff
@@ -2,11 +2,106 @@ name: Test E2E
...
+          - 'android-34'
+          - 'android-35'
+          - 'android-36'
+      app:
+        description: 'App workspace to test (e.g. demo_app, simple_web_view). Empty = all apps.'
+        type: string
+        default: 'demo_app'
+      flow:
+        description: 'Optional: flow name within the chosen app (e.g. setLocation). Empty = all flows.'
```
Fishbowler: Disambiguate between filename and `name:`

Inform the caller that this'll run Android only.

```suggestion
        description: 'Optional: flow filename within the chosen app (e.g. setLocation). Empty = all flows. Skips non-Android tests.'
```

.github/workflows/test-e2e.yaml
```diff
@@ -99,15 +198,17 @@ jobs:
...
+    if: ${{ always() && needs.build.result == 'success' && (needs.validate-inputs.result == 'success' || needs.validate-inputs.result == 'skipped') }}
     timeout-minutes: 60
 
     env:
       ANDROID_HOME: /home/runner/androidsdk
       ANDROID_SDK_ROOT: /home/runner/androidsdk
       ANDROID_AVD_HOME: /home/runner/.config/.android/avd/
-      ANDROID_OS_IMAGE: system-images;android-32;google_apis;x86_64
-      ANDROID_PLATFORM: platforms;android-34
+      ANDROID_OS_IMAGE: system-images;${{ inputs.android_version || 'android-32' }};google_apis;x86_64
```
Fishbowler, line 248: Match the default at the top.

```suggestion
      ANDROID_OS_IMAGE: system-images;${{ inputs.android_version || 'android-33' }};google_apis;x86_64
```

e2e/run_tests
```diff
@@ -43,7 +107,36 @@ if [ "$platform" = "web" ]; then
...
+	_h1 "run tests for app \"$app_name\" on platform \"$platform\""
+	_run_passing "$app_name" "$workspace_dir"
+
+	case "$app_name" in
+		wikipedia|simple_web_view) ;;
+		*) _run_failing "$app_name" "$workspace_dir" ;;
+	esac
 else
+	###
+	### Full-suite mode (PR triggers): byte-identical to prior behavior.
```
Fishbowler: Maybe this shouldn't be identical, and should use the new `_run_passing` and `_run_failing` to reduce future maintenance?

Fishbowler (changes requested):
Looks good. Most of my comments are pretty minor, I think.

.github/workflows/test-e2e.yaml
amanjeetsingh150: Fair point, I'd like to defer this though, since I feel this should default to latest always for iOS being 26.2 and android right now its 34. If you agree with that then, I see we have couple of options:

1. Bump now to android-34 to match latest on cloud.
2. Keep android-32 for now, and switch android-36 in couple of weeks after launch, we are almost here.

I'd prefer 2nd one.

.github/workflows/test-e2e.yaml
amanjeetsingh150: Same as above, discussion around defaults.

e2e/run_tests
amanjeetsingh150: Done in 6c2638a7 — collapsed the inlined block onto `_run_passing` / `_run_failing`. Same behavior; the helper bodies were already byte-for-byte the same as the inlined version. Updated the surrounding comment to reflect the share instead of the old "byte-identical to prior behavior" framing.

.github/workflows/test-e2e.yaml
amanjeetsingh150: Went a step further than the description-only fix — added a `platform` input (`choice [android, ios, web]`, default `android`) and gated each per-platform job on `(github.event_name == 'pull_request' || inputs.platform == '<x>')`. PR triggers still fan out to all three; `workflow_dispatch` now narrows to the chosen platform.

Filename/`name:` disambiguation went in too:
> Optional: flow filename within the chosen app (e.g. setLocation). Empty = all flows. Single-flow mode runs Android only.

Single-flow mode stays Android-only because `flow_path` is only consumed by the Android job — happy to lift that in a follow-up if there's demand.

Behavior change worth flagging: `gh workflow run test-e2e.yaml` with no inputs used to fan out to all three platforms; it now runs Android only. Pushed in 8d6cc269.

.github/workflows/test-e2e.yaml
Fishbowler: No objection to waiting.

Option 3: Run tests on all OSes in parallel

Fishbowler (approved):

PR #3232, opened 2026-04-28
Fix the mutation setting of hierarchy-inert commands

### LINKED ISSUE #3117: Cautious command mutation settings are causing delays

### Steps to reproduce

A check like assertNotVisible doesn't complete immediately, else we'd get lots of false passes when it follows a tap and no elements are visible whilst the screen is transitioning.

The algorithm waits for a maximum of 7 seconds, adjusted for time since the hierarchy likely last changed. That's not tracked in real time because that'd be really slow on the device. Instead, each command is annotated as "mutating" or not. A mutating command is one that likely alters the screen's element hierarchy.

Launching and stopping apps, taps and swipes and inputting text would naturally cause changes to the screen.

Other commands like AssertVisible or TakeScreenshot naturally don't.

Some are marked as mutating, and possibly shouldn't be, like copyTextFrom or evalScript.

### Actual results

Commands like assertNotVisible take longer than they should

### Expected results

Tests to be as quick as possible, whilst remaining solid and stable.

### About app

n/a

### About environment

n/a

### Logs

n/a 

### Maestro version

2.3.0

### Anything else?

Check each command, and assess whether it might incorrectly be set to mutating when it really isn't.

Fishbowler:
## Proposed changes

Performance improvement.

Maestro performs assertions like assertNotVisible based on a timer tied to the last command that altered the hierarchy. We track that through a return value on a command that indicates whether the command mutated the hierarchy (or may have).

This change updates a few inert commands, marking them as non-mutating, which will speed up assertions that immediately follow them.

In the below example, assertNotVisible's timer should be from the tapOn, not from the last copyTextFrom.

```
- tapOn: Remove Foo
- copyTextFrom:
    id: thing1
- copyTextFrom:
    id: thing2
- copyTextFrom:
    id: thing3
- assertNotVisible: Foo
```

I've added comments on justifications, so that the next person visiting this code might be able to think a little less about these :)

## Testing

e2e 🤞 

## Issues fixed

Fixes #3117

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -598,7 +598,7 @@ class Orchestra(
         val diffFile = expectedFile.parentFile?.resolve(diffFileName) ?: File(diffFileName)
 
         when (val result = ScreenshotMatch.compare(expectedImage, actualImage, command.thresholdPercentage, diffFile)) {
-            is ScreenshotMatch.Result.Match -> return true
+            is ScreenshotMatch.Result.Match -> return false // Screenshots are non-interactive
             is ScreenshotMatch.Result.SizeMismatch -> throw MaestroException.AssertionFailure(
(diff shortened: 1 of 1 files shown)


pedro18x (approved):
LGTM

PR #3249, opened 2026-05-01
feat(orchestra): structured parser errors with summary + detail

proksh:
Parser errors used to be a single multi-line string formatted with Unicode boxes which looks good in a terminal, but gets broken in frontend/studio, and needs hack to display it propertly.

Studio:
image

Maestro Cloud:
image

This PR splits parser errors into two pieces on the exception itself:

```
class ValidationError(
      override val message: String,  // short summary, single line
      val detail: String? = null,          // rich block: snippet + caret + message + docs
      cause: Throwable? = null,
)
```

Downstream consumers can route each piece to the right surface:
  - **CLI**: calls e.formatForTerminal() (message + "\n\n" + detail); terminal output is unchanged.
  - **copilot worker**: statusMessage = e.message, debugOutput = e.detail (separate PR).
  
  
The detail format uses only regular ASCII spaces and \n - no box-drawing, no non-breaking spaces, so it renders correctly inside "pre" without any frontend reformatting.


pedro18x (commented):
LGTM.

pedro18x (approved):

Issue #3093, opened 2026-03-30
Native dialog for sign in with apple


saswattech9:
Hello there , Is there any possibility that we can handel sign in with apple for mobile application . Currently i am using flutter for developing mobile application . 

I am unable to select the sign in with apple element . I tried with text selection that also does not work . 

Any solution for this ?

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/5ba6828c-127b-4bc5-9eb3-8d8e50d3c1d4" />

Fishbowler:
Does this work with `snapshotKeyHonorModalViews` set to false in the project config and run from the CLI?
https://docs.maestro.dev/reference/workspace-configuration#platform-configuration

saswattech9:
@Fishbowler I tried it does not work . Could you please share some example if possible .

Fishbowler:
This isn't something I've tried doing, I'm afraid.

Does Xcode's Accessibility Inspector see those elements on the screen?

saswattech9:
@Fishbowler I tired in Maestro Studio but not able to find the element.

Fishbowler:
I assumed. That's why I asked about Xcode's Accessibility Inspector.

saswattech9:
@Fishbowler Any other Options?

Fishbowler:
What exactly did you try with the maestro config?

saswattech9:
i put the config.taml file inside .maestro.

file contains

```
platform:
  ios:
    snapshotKeyHonorModalViews: false
```

Fishbowler:
Then what?

saswattech9:
```
appId: com.example.app
---
- tapOn:
    id: continue_with_apple
- runFlow:
    when:
      platform: ios
    commands:
      - tapOn: "Continue with Password"
```

Fishbowler:
And how was that run?

saswattech9:
it failed at - tapOn: "Continue with Password" because it donot find any element .

Fishbowler:
You've told me the terminal output there. But how was it _run_? What was the terminal input?

Fishbowler:
(A conversation of this sort might be better served in Slack - https://slack.maestro.dev)

saswattech9:
<img width="556" height="222" alt="Image" src="https://github.com/user-attachments/assets/ae0580b4-a0b7-4433-bdee-560044c96dc3" />

Fishbowler:
I said

[...]

Issue #2065, opened 2024-09-23
Studio can't interact with iOS native pop-up when landscape


klyovan:
### Is there an existing issue for this?

- [X] I have searched the existing issues and didn't find mine.

### Steps to reproduce

1. Run `maestro studio`  on flutter app builded for ios
2. Invoke native pop-up using Platfrom Channels (or `signIn` method from [google sign in plugin](https://pub.dev/packages/google_sign_in))
3.  The nodes for cancel and continue item exists on the screen, but they are flattened and positioned to the left and below, but still not clickable using either the studio or the yaml file.

### Actual results

The nodes for cancel and continue item exists on the screen, but they are flattened and positioned to the left and below, but still not clickable using either the studio or the yaml file

### Expected results

The nodes for cancel and continue item exists on the screen  and they are  clickable using either the studio or the yaml file.

### About app

- Tested on simulator: Ipad Pro 11-inch (M4)
- This popup is called by native code but invoked from Flutter using platform channel [ex. google_sign_in package](https://pub.dev/packages/google_sign_in)

### About environment

- i am on M3 Pro mac, Sonoma 14.6, 
- Flutter 3.24.3 

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. *DO NOT* remove the backticks! -->
```

</details>

### Maestro version

1.39-dev.2

### How did you install Maestro?

Homebrew

### Anything else?

![image](https://github.com/user-attachments/assets/5e9dd994-531b-42f2-b9c7-4ef6bfe01402)
![image](https://github.com/user-attachments/assets/ffd19b0d-3b91-4dc5-bd9c-48dabd84ad03)

sanderdegelder:
I experience the same issue. Our app is landscape only. That works fine with Maestro Studio, but when a native popup of iOS is shown, Maestro Studio misplaces all the elements.

Fishbowler:
Is anyone able to provide a reproducible example?

sanderdegelder:
To reproduce:

- Start app under test in landscape orientation. Our native iOS app only supports landscape orientation.
- Start Maestro Studio.
- Observe that all elements are shown correct and are clickable in Maestro Studio.
- Continue until a native iOS popup is shown. For instance the popup to ask if the app is allowed to track activity across other companies' apps and websites.
- Observe that the screen orientation in Maestro Studio changes to portrait mode and the clickable elements do not match with what is shown on screen.
[...]

Issue #1227, opened 2023-07-08
[Maestro - v1.29.0] How can I handle the button Not Now keychain of ios device


stephenduong12:
**Describe the bug**
When I input the username and password, there is a popup which allows me to save password in the keychain. I want to tap on 'Not Now' button, but the Maestro studio cannot identity this button.

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Start the simulator
2. Run `maestro test ...`
3. The Flow fails at ...

**Expected behavior**
Can identify and tap on the button 'Not Now'

**Screenshots**
<img width="341" alt="image" src="https://github.com/mobile-dev-inc/maestro/assets/135583776/86cf2aed-5b7d-4f6b-af39-710867f0b180">
<img width="282" alt="image" src="https://github.com/mobile-dev-inc/maestro/assets/135583776/e368831b-01c8-45ca-be7b-5abfc207d402">

**Environment information (please complete the following information):**

- Maestro version 1.29.0
- Platform: iOS
- Framework: React Native
- Device model and OS version: 14 pro max (ios 16.4)
- Simulator
- Host : Mac

igorsmotto:
Hey, since this screen is from Apple and is quite fixed over time, you could use tapOn with coordinates
https://maestro.mobile.dev/api-reference/commands/tapon#tapping-on-a-specific-point-on-the-screen

Something like this:
```
- tapOn:
    point: 90%,50%
```
`maestro studio` could you help discovering the exact number.

Selnapenek:
Hello, any news about this?

anas-baadshah:
Any update on this issue?

PR #3250, opened 2026-05-01
iOS: HealthKit repro + snapshot frame normalization for cross-process sheets

simon-gilmurray:
## Context

We've seen **iOS modal / system-sheet style UI** where taps land wrong: hierarchy still exposes the controls, but **`frame`-based coordinates don't match on-screen placement**.

This branch adds **HealthKit-backed UI** in the Flutter demo app to reproduce the issue, plus a Maestro flow that asserts we can authorize and dismiss cleanly.

Screenshot (repro):
Screenshot 2026-05-01 at 12 13 08

---

### Demo app & E2E

- **`demo_app`**: Flow to exercise **Health Access** → **Turn On All** → **Allow** and assert dismissal.
- **`fail_health_access.yaml`**: Assert **`Turn On All`** is gone (avoids confusing "Health Access" chrome elsewhere on screen).

### iOS XCTest runner (MaestroDriver)

Cross-process subtrees stitched into `XCUIApplication` snapshots sometimes report **`frame`** in window-local coords while taps use screen coords. **`visibleFrame`** can diverge **at the stitching boundary**; descendants often still show **`visibleFrame` == `frame`**, so we **don't fix this by using `visibleFrame` on every node** — we take **`visibleFrame.origin − frame.origin` at guarded boundaries**, require **`windowContextID` transition + remote subtree (`isRemote`)**, and **inherit that offset to descendants**.

- **`ViewHierarchyHandler`**: Recursive walk + offset inheritance + docs.
- **`AXElement`** (runner **UITests Models**): `frameOverride` when building from `dictionaryRepresentation`.

### Bundled Simulator driver artifacts

Simulator driver zips under **`maestro-ios-driver`** rebuilt from runner sources.

---

## Testing

- [x] Demo app feature (iOS)
- [x] `fail_health_access.yaml` passes with driver fix
- [x] Builds / runs locally; Simulator driver zips updated


Fishbowler:
Annoyingly, our CI doesn't run the demo_app tests on iOS right now

e2e/demo_app/.maestro/fail_health_access.yaml
```diff
@@ -0,0 +1,15 @@
+appId: com.example.example
+tags:
+  - failing
```
Fishbowler: Should we have this as `failing`? It implies we _expect_ it to fail. Maybe we have neither for now, skipping any CI flow - it's available to prove the problem and the fix, and we can add it to `passing` when it's ready?

simon-gilmurray: This was going to be my question - it wasn't clear if my test should form part of the failing tests with associated tag, or if it should be part of the issues/ folder.
I didn't want it to get picked up incorrectly and run failing, and then blocking pipelines - but it sounds like it doesn't run anywhere on CI yet anyway.
Happy to move it where it should live

Fishbowler: CI will run `passing` and `failing` so no tag at all feels like the right place for it, on the off chance that @amanjeetsingh150 fixes the CI before someone fixes this issue.

simon-gilmurray: Moved test in to issues/ and removed tags

Fishbowler (approved):

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
```diff
@@ -294,7 +294,140 @@ struct ViewHierarchyHandler: HTTPHandler {
     }
 
     private func elementHierarchy(xcuiElement: XCUIElement) throws -> AXElement {
-        let snapshotDictionary = try xcuiElement.snapshot().dictionaryRepresentation
-        return AXElement(snapshotDictionary)
+        let snapshot = try xcuiElement.snapshot()
+        return elementHierarchy(snapshot: snapshot, inheritedOffset: .zero, parentWindowContextID: nil)
```
amanjeetsingh150, line 298: Ok my initial thoughts at this, looks like `elementHierarchy` the new walker re-derives the same subtree data at every node level instead of once at the root, so hierarchy build goes from `main`'s **O(N)** to **O(N·D)**, worst-case **O(N^2)** on a deep tree, potentially creating regression by making our view hierarchy calls expensive.

We have different options here:

1. Tweak the current approach to ensure time complexity remains same
2. While exploring other options and seeing issue I asked:
- Why not use `[XCUIElement tap]` and let Apple resolve the cross-process screen point natively? Do they face same issue we face in native XCUITest framework?
- If this works we don't need to do complicated logic, around remote trees.
- How does WebDriverAgent/Detox etc. handle this?

Ran a POC, to get XCUIElement corresponding to this element in healthkit bottom sheet and checked if we just use tap on that would it work and to my surprise it did!!

Check here: https://github.com/mobile-dev-inc/Maestro/pull/new/poc-elementtap-healthkit

The best part of solving this is that we'd solve following as well:

- [#560](https://github.com/mobile-dev-inc/Maestro/issues/560)
- [#1924](https://github.com/mobile-dev-inc/Maestro/issues/1924) 
- [#2065](https://github.com/mobile-dev-inc/Maestro/issues/2065) 
- [#1227](https://github.com/mobile-dev-inc/Maestro/issues/1227) 
- [#3093](https://github.com/mobile-dev-inc/Maestro/issues/3093) 

I asked Claude to get me issues facing same issue around remote elements^.

Let's chat today around this how we can proceed here with rollout.

PR #3251, opened 2026-05-02
Attach YAML source provenance to MaestroCommand

Leland-Takamine:
## Proposed changes

Attaches YAML source provenance to every `MaestroCommand` parsed from the YAML reader. Each command now carries an optional `SourceInfo` describing the source text, file path, and `(line, column, char-offset)` range it came from. This is foundational for downstream tooling — better error messages, editor integration, flow visualization — that needs to map a runtime command back to its location in the user's flow file.

Threaded through every parser entry point (`parseFlow`, `parseCommand`, `parseConfigOnly`, `parseWatchFiles`, `checkSyntax`) via a Jackson `DeserializationContext` attribute so it propagates automatically into nested commands inside `runFlow` / `repeat` / `retry` blocks.

`sourceInfo` is excluded from `MaestroCommand.equals` / `hashCode` (so structurally-identical commands at different YAML positions still compare equal) and is `@JsonIgnore`'d (so DB/wire serialization is unchanged and we don't duplicate the full source text on every command).

## Testing

`./gradlew :maestro-orchestra:test :maestro-orchestra-models:test`. New `SourceInfoTest.kt` covers:
- top-level command source attribution (line / column / char offset / path / source)
- composite commands (`runFlow`, `repeat`, `retry`): both the outer wrapper and the inline children carry correct source ranges
- subflow file references (`runFlow: child.yaml`): inner commands and their `onFlowStart`/`onFlowComplete` hooks carry the *child* file's source, not the parent's
- `onFlowStart` hook commands declared on the top-level config
- `evaluateScripts` preserves `sourceInfo` on the post-substitution form (caught a real bug — the orchestra hot-path was dropping it)
- trim-heuristic edge cases: last command in file, last command in block, trailing YAML comment, block scalar with `#` lines
- equality / hashCode ignore `sourceInfo` so cloud DB round-trips stay equal
- JSON serialization drops `sourceInfo` (`@JsonIgnore` keeps the wire/DB schema unchanged)
- internal consistency: every test runs through a `parseFlow` wrapper that asserts `(line, column, offset)` agree on every parsed command, including nested ones

---

> [!NOTE]
> **Medium Risk**
> Changes core YAML parsing/deserialization and `MaestroCommand` equality/serialization behavior; mistakes could misattribute locations, break nested command parsing, or subtly affect command dedup/caching relying on `equals`/`hashCode`. No auth/security impact, but it touches a critical execution pipeline.
> 
> **Overview**
> Every `MaestroCommand` produced from YAML parsing now carries optional `SourceInfo` (original YAML text, path, and start/end line/column/offset range) to enable downstream tooling and better error mapping.
> 
> This threads a per-parse context through all parser entry points (including nested `runFlow`/`repeat`/`retry` blocks and `check-syntax`) via a Jackson `DeserializationContext` attribute, updates error wrapping to use the new location data, and adjusts syntax-check caret positioning. `sourceInfo` is **not** serialized (`@JsonIgnore`) and is excluded from `equals`/`hashCode`; `evaluateScripts` now preserves it. Comprehensive new tests validate provenance propagation and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 527dd84a9ac861e51351d4c049a44fe891845044. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pedro18x (approved):
LGTM, nice work.

PR #3255, opened 2026-05-04
chore!: remove deprecated Rhino JS engine

amanjeetsingh150:
## Summary

Removes the deprecated Rhino JS engine. Rhino was deprecated long back with CLI warning on every flow that set `jsEngine: rhino`. 

This PR removes the engine, its dependency, and supporting code; GraalJS becomes the only engine.

## Breaking change

`jsEngine: rhino` now fails workspace validation with a migration message instead of running, this is inside WorkspaceValidator and Orchestra so this should work for everything including: cloud, cli and studio.

## Changes

- Delete `RhinoJsEngine`, `JsHttp`, `JsScope`, `JsConsole` (Rhino-only).
- Drop `org.mozilla:rhino` from `maestro-client`.
- `WorkspaceValidator` returns `GenericError` on `jsEngine: rhino`; `Orchestra.jsEngineFactory` throws as defense-in-depth.
- Remove deprecation `PrintUtils.warn` in `MaestroCommandRunner` and the `EcmaError` arm in `ErrorViewUtils`.
- Migrate `WorkspaceValidatorTest` to assert the new error; drop Rhino variants in `CommandDescriptionTest` and `IntegrationTest` (GraalJS counterparts already cover the same behaviour).

## Verify

- [x] `./gradlew test` passes across all modules.
- [x] `./gradlew compileKotlin compileTestKotlin` clean.

This closes a long time: https://github.com/mobile-dev-inc/Maestro/issues/2049


Fishbowler:
♥️ This is the last step to completing #2049

> jsEngine: rhino now fails workspace validation

As it should. Anyone who's got this explicitly chose it after we defaulted to graal, so silently switching to graal will presumably be useless to them anyway.

PR #3260, opened 2026-05-05
Add cloud URL and app binary ID to HTML report

mikamikaWolt:
## Summary
- include Maestro Cloud upload metadata (`cloudUploadUrl`, `appBinaryId`) in `TestExecutionSummary` so report renderers can access it
- pass cloud URL and app binary ID through cloud execution reporting (`CloudInteractor`)
- render a footer info block in the HTML report with bold labels, cloud URL link, and app binary id
- keep local `maestro test` behavior unchanged by not merging cloud metadata in local summary aggregation and by rendering the footer only when a cloud URL is present

## Test plan
- [ ] Run `maestro cloud ... --format html` and confirm report includes a cloud URL link and app binary ID at the bottom
- [ ] Run non-cloud local `maestro test --format html` and confirm report layout is unchanged


maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
```diff
@@ -700,7 +700,9 @@ class TestCommand : Callable<Int> {
             passed = acc.passed && summary.passed,
             suites = acc.suites + summary.suites,
             passedCount = sumOf { it.passedCount ?: 0 },
-            totalTests = sumOf { it.totalTests ?: 0 }
+            totalTests = sumOf { it.totalTests ?: 0 },
+            cloudUploadUrl = acc.cloudUploadUrl ?: summary.cloudUploadUrl,
+            appBinaryId = acc.appBinaryId ?: summary.appBinaryId,
```
Fishbowler: I don't think that this route can lead to a summary with a cloudUploadUrl or an appBinaryId. You'd only get here via `maestro test ...`

maestro-cli/src/main/java/maestro/cli/cloud/CloudInteractor.kt
```diff
@@ -434,6 +434,7 @@ class CloudInteractor(
                 testSuiteName = testSuiteName,
                 uploadUrl = uploadUrl,
                 projectId = projectId,
+                appBinaryId = appBinaryIdResponse,
```
Fishbowler, line 437: I _think_ this should work for tests where you pass an existing app-binary-id to `maestro cloud`, but worth explicitly adding to your test plan.

mikamikaWolt: Verified both options:
- `maestro cloud --app-binary-id <id> `... (existing binary)
- `maestro cloud <app-file>` ... (upload new binary)

In both cases, HTML report footer contains:
- View details on Maestro Cloud URL
- App binary id value

Fishbowler (approved):
Local execution:

Image

Cloud execution:

Image

PR #3261, opened 2026-05-05
fix(assertScreenshot): report match percent that aligns with the threshold

proksh:
## Problem

`assertScreenshot` can fail with a "current" value that looks like it should have passed:

> `Comparison error: Assert screenshot matches X (threshold: 95.0%) - threshold not met, current: 96.46815%`

`96.46815% > 95%`, yet the assertion fails. This is the symptom seen in [#3259](https://github.com/mobile-dev-inc/Maestro/pull/3259) and reported by Doximity.

## Root cause

The MISMATCH branch in `Orchestra#assertScreenshotCommand` printed `100 - ImageComparisonResult.differencePercent`. Inside `com.github.romankh3:image-comparison`, two unrelated metrics are at play:

| Used for                  | Source                                                        | What it measures                                            |
|---------------------------|---------------------------------------------------------------|-------------------------------------------------------------|
| MATCH/MISMATCH gate       | `ImageComparison.isAllowedPercentOfDifferentPixels` (private) | **count** of pixels exceeding `pixelToleranceLevel`         |
| `result.differencePercent`| `ImageComparisonUtil.getDifferencePercent` (public)           | **average** per-channel RGB intensity delta over the image  |

These can diverge in opposite directions: a small region with a stark change drives the count up while leaving the average tiny, so the threshold check fails while the message reports a high "match %".

The library does not expose its count-based metric, and 4.4.0 (March 2021) is the latest release.

## Fix

Introduce `ScreenshotMatch` as the single source of truth for the comparison:

- `ScreenshotMatch.matchPercentage(expected, actual, pixelToleranceLevel)` walks pixels using the same Euclidean distance rule the library uses internally, returning the count-based percentage.
- `ScreenshotMatch.compare(expected, actual, threshold, diffFile)` owns both the pass/fail decision and the reported percent — the threshold gate and the user-visible "current %" are now the same number by construction. The library is invoked only as a diff-PNG renderer on `Mismatch`.
- `Orchestra.kt` becomes a `when` over the typed `ScreenshotMatch.Result`. Drops the `ImageComparison` / `ImageComparisonState` imports.

## Behaviour change

- **Passing assertions:** unchanged. The library's gate already used the count-based metric; we replicate it.
- **Failing assertions:** the printed `current %` is now consistent with `thresholdPercentage`. A failure that previously read `current: 96.46815%` against `threshold: 95.0%` now reads a number actually below 95% — matching what users expect.

## Tests

`maestro-orchestra/src/test/.../AssertScreenshotMatchTest.kt` adds 5 cases:

1. `compare returns Match when match percent meets threshold` — happy path on real screenshots.
2. `compare returns Mismatch when match percent falls below threshold and writes diff file` — failure path, asserts the diff PNG is produced.
3. `compare returns SizeMismatch for differently sized images` — typed-result coverage.
4. `match percentage agrees with library MATCH/MISMATCH decision at the boundary` — pins our pixel walk to the library's gate at three `pixelToleranceLevel` values (0.0, 0.05, 0.1). Catches drift if the library's pixel-difference math ever changes.
5. `match percentage exposes count-based metric distinct from library's average color delta` — locks in the metric distinction the bug came from.

The two `expected.png` / `actual.png` fixtures (~200KB each) are real-world iPhone screenshots that differ only by content in the status bar — happy to swap for synthetic images if reviewers prefer.

## Out of scope

This PR is the fix only. The repro flow + demo-app screen are owned by [#3259](https://github.com/mobile-dev-inc/Maestro/pull/3259) and should land separately so the new flow can verify the corrected message end-to-end.


simon-gilmurray (approved):
🎉

PR #3264, opened 2026-05-07
fix(env): fail fast with helpful message when env value is null

proksh:
A YAML entry like `KEY:` with no value deserializes to `null` inside `Map<String, String>` (Kotlin generic erasure) and later crashes `DefineVariablesCommand.evaluateScripts` with an opaque `Parameter specified as non-null is null` error after retries. `withEnv` now validates and names the offending key(s) so the failure surfaces at YAML-parse time.

## Example:

What a flow with empty variable `AUTOMATED_EMAIL` cause before and after the change in PR:

Screenshot 2026-05-07 at 2 25 26 PM

---

Before:
Screenshot 2026-05-07 at 2 18 09 PM

After:
Screenshot 2026-05-07 at 2 22 20 PM

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/util/Env.kt
@@ -33,13 +33,20 @@ object Env {
         // Fail fast with the offending keys so the user knows exactly what to fix.
         @Suppress("UNCHECKED_CAST")
         val nullKeys = (env as Map<String, String?>).filterValues { it == null }.keys
-        require(nullKeys.isEmpty()) {
-            "Environment variable(s) ${nullKeys.joinToString(", ")} have no value. " +
-                "Set an explicit value (e.g. `${nullKeys.first()}: \"\"` for an empty string) " +
-                "or remove the key."
-        }
+        if (nullKeys.isNotEmpty()) throw EnvVariableMissingValueError(nullKeys)
         return listOf(MaestroCommand(DefineVariablesCommand(env))) + this
     }
+
+    /**
+     * Raised by [withEnv] when the env map carries one or more null values. Caught by
+     * `MaestroFlowParser.wrapException` so the rendered error names the missing key(s)
+     * instead of surfacing a generic "Parsing Failed" summary.
+     */
+    class EnvVariableMissingValueError(val keys: Set<String>) : IllegalArgumentException(
+        "Environment variable(s) ${keys.joinToString(", ")} have no value. " +
+            "Set an explicit value (e.g. `${keys.first()}: \"\"` for an empty string) " +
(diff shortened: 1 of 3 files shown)


Fishbowler:
Is this approach better than silently rewriting null to empty string?

proksh:
> Is this approach better than silently rewriting null to empty string?

I think Yes,
1. The prior behavior was already a crash.
2.  `null` and `""` mean different things in test flows. If `EMAIL:` silently becomes `""`, then `tapOn: "${EMAIL}"` or `assertVisible: "${EMAIL}"` produces weird empty-string behavior that's hard to debug. The user almost always meant "I forgot to set this," not "I want empty."

Let me know if you disagree.

amanjeetsingh150 (commented):
Before we patch this specific case, I want to make sure this is the right layer and way to fix it.

We’ve seen this pattern a couple of times now: invalid values make it into MaestroCommand, and only fail later when the orchestrator (cloud/studio) tries to use them.

Examples:

- timeout strings that can’t be parsed as valid durations
- swipe.direction relying on specific enum values
- orientation values having the same issue

Validations of these are right now spread across MaestroCommand class and there is no particular confirmation/or API that a constructed command values are correct or not.

Wondering if MaestroCommand construction should be schema-aware, so invalid commands fail earlier at construction/evaluation time instead of reaching the orchestrator.

steviec (approved):
Code looks good. I'm assuming this won't bite users, because it would have already been failing right?

proksh:
> Before we patch this specific case, I want to make sure this is the right layer and way to fix it.
> 
> We’ve seen this pattern a couple of times now: invalid values make it into MaestroCommand, and only fail later when the orchestrator (cloud/studio) tries to use them.
> 
> Examples:
> 
> * timeout strings that can’t be parsed as valid durations
> * swipe.direction relying on specific enum values
> * orientation values having the same issue
> 
> Validations of these are right now spread across MaestroCommand class and there is no particular confirmation/or API that a constructed command values are correct or not.
> 
> Wondering if MaestroCommand construction should be schema-aware, so invalid commands fail earlier at construction/evaluation time instead of reaching the orchestrator.

 Fair point on the broader pattern - a schema-aware construction step would catch the timeout/direction/orientation cases uniformly, and that's worth doing as its own initiative.
 
This one's a slightly different though. Env is a `Map<String, String>` that Jackson deserializes and withEnv then wraps into a synthetic `DefineVariablesCommand` - and because of generic erasure, KEY: (no value) sneaks in as null even though the type says `String`. So even with schema-aware command validation in place, you'd still need a check somewhere around deserialization, since by the time the validator runs the null is already (incorrectly) typed as String.

proksh:
> Code looks good. I'm assuming this won't bite users, because it would have already been failing right?

Yes correct

PR #3266, opened 2026-05-07
Add file attribute to JUnit report

hSATAC:
## Proposed changes

Adds a `file` attribute to each `<testcase>` entry in the JUnit XML report (`maestro test --format junit`). The value is the flow's source path rendered relative to the working directory (or absolute when the file lives outside the workspace), e.g. `file=".maestro/auth/login.yaml"`.

### Why

`JUnitTestSuiteReporter` currently writes `<testcase id="X" name="X" classname="X" .../>` with all three identifiers sourced from `flow.name` (the YAML `name:` field, falling back to the filename without extension). When two flow files declare the same `name:` — or two YAML files share a basename across sub-directories and rely on the fallback — the resulting `<testcase>` entries are indistinguishable. CI dashboards (GitHub Actions test reporters, Allure, Jenkins, ReportPortal, etc.) typically key off `classname` + `name` and either silently merge the colliding rows or surface them as flakes.

The new `file` attribute follows the de facto JUnit convention used by pytest-junit, jest-junit, and Maven Surefire. It is purely additive — consumers that ignore unknown attributes are unaffected — and gives downstream tooling (and humans reading the raw XML) a reliable way to tell colliding flows apart.

### Design notes

- **Strictly additive.** `id`, `name`, and `classname` are unchanged. No breaking change to existing test history, dashboard groupings, or Allure attachments.
- **Local runs only.** `maestro cloud --format junit` keeps the existing report shape — `CloudInteractor` builds `FlowResult` from upload results that don't carry source paths today, so `filePath` stays null and the attribute is omitted.
- **Path base.** Renders relative to `WorkingDirectory.baseDir` (the same base used by `mcp --working-dir` and other CLI path resolution), with an absolute-path fallback when the flow file lives outside the base or on a different filesystem root (e.g. a different Windows drive).
- **Field placement.** `FlowResult.filePath: String?` is appended at the end of the data class so any positional constructor calls keep compiling.

## Testing

- New `Path.toCwdRelativeOrAbsoluteString()` helper has unit tests covering: relative-to-CWD path, path outside CWD subtree, non-absolute input, and the `relativize(self)` empty-path edge case.
- `JUnitTestSuiteReporterTest` updated with `file=` in expected XML across the four existing tests, plus a new test asserting the attribute is omitted when `filePath` is null (cloud-style result).
- `./gradlew :maestro-cli:test` is green (FileUtilsTest 4/4, JUnitTestSuiteReporterTest 5/5, other reporter tests unaffected).
- Manual smoke: built the CLI, ran two flows with identical YAML `name:` under `.maestro/a/` and `.maestro/b/`, confirmed the produced report has two `<testcase>` rows that differ only by `file=`.

> **Does this need e2e tests?** No — pure reporter change with no UI / device behaviour.

## Issues fixed

N/A — surfaced during local Maestro usage.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
+++ b/maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
@@ -35,6 +35,7 @@ data class TestExecutionSummary(
         val properties: Map<String, String>? = null,
         val tags: List<String>? = null,
         val steps: List<StepResult> = emptyList(),
+        val filePath: String? = null,
     )
 
     data class StepResult(
--- a/maestro-cli/src/main/java/maestro/cli/report/JUnitTestSuiteReporter.kt
+++ b/maestro-cli/src/main/java/maestro/cli/report/JUnitTestSuiteReporter.kt
@@ -47,6 +47,7 @@ class JUnitTestSuiteReporter(
                     id = flow.properties?.get("junitId") ?: flow.name,
                     name = flow.name,
                     classname = flow.properties?.get("junitClassname") ?: flow.name,
+                    file = flow.filePath,
                     failure = flow.failure?.let { failure ->
                         Failure(
                             message = failure.message,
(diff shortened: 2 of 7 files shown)


Fishbowler:
The fact that JUnit isn't standardised bothers me immensely, but you're right in that this approach seems consistent with other usages. The closest we've got to a standard is this collected knowledge: https://github.com/testmoapp/junitxml

hSATAC:
Hi @Fishbowler 👋

Just wanted to gently follow up on this PR. I saw your comment about JUnit not really being standardized (totally agree 😅), and I appreciate you sharing the testmoapp reference as well.

Is there anything you'd like me to adjust or address on my end to help move this forward? Happy to make changes if needed.

Thanks again for taking the time to review this!

Fishbowler:
No - I'm chatting to a couple of folks about this one later today.

Fishbowler (approved):
Tests (both automated and manual) pass.
I've looked into edge cases around filenames that could cause XML problems, but can't see any (reasonable) route to problems.

LGTM!

PR #3278, opened 2026-05-11
fix(orchestra): use literal runScript file as sourceDescription

proksh:
`RunScriptCommand.sourceDescription` was the resolved absolute script path, while `runFlow` and `retry` use the literal YAML `file:` field. It was also doing double duty as the anchor `GraalJsEngine.setCurrentScriptDir` uses to resolve relative `multipartForm.filePath` lookups. Result: the user-facing step description leaked the workspace root (`Run /var/folders/.../workspace.../login.js` from an extracted temp dir, vs `Run /login.js` from a zip filesystem).

Split the two concerns so `runScript` lines up with `runFlow` / `retry`:

- `RunScriptCommand` gains `scriptDir: String? = null`, populated by `YamlFluentCommand` as `scriptPath.parent?.toString()`. The JS engine uses it directly to anchor relative lookups; the engine no longer derives a directory from `sourceName`.
- `sourceDescription` now mirrors `runScript.file` (the literal YAML field). `originalDescription` reverts to plain `"Run $sourceDescription"`, same shape as `runFlow` / `retry`.
- `JsEngine.evaluateScript` takes an explicit `scriptDir`, and `GraalJsHttp.setCurrentScriptDir` stores it directly instead of deriving `.parentFile` from a path.

`scriptDir` defaults to `null` so commands assembled outside the YAML parser keep working, and `FAIL_ON_UNKNOWN_PROPERTIES=false` absorbs the new field across rolling deploys.

Before:
Screenshot 2026-05-11 at 3 27 25 PM

After:
Screenshot 2026-05-11 at 3 51 25 PM


maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
```diff
@@ -1002,6 +1002,10 @@ data class RunScriptCommand(
     val condition: Condition?,
     override val label: String? = null,
     override val optional: Boolean = false,
+    // Absolute parent directory of the script file, used by the JS engine to
+    // anchor relative file lookups (e.g. multipartForm filePath). Nullable so
+    // commands assembled outside the YAML parser keep working.
+    val scriptDir: String? = null,
 ) : Command {
```
Fishbowler, line 1006: Shouldn't this comment be on the same line as the param, after it?
It's not wrong to do it like this. Just a bit weird 😂

Fishbowler (approved):

maestro-orchestra-models/src/main/java/maestro/orchestra/Commands.kt
proksh: Updated. 😄

PR #3279, opened 2026-05-11
feat: filter active sessions by platform in SessionStore

fantasyRqg:
## Proposed changes

  Fix `SessionStore.hasActiveSessions` so that it only considers sessions on the **same platform** as the caller. Previously, an active iOS session would cause `hasActiveSessions(..., Platform.ANDROID)` to return `true` (and vice versa), because the check simply
  compared against any key other than the caller's own.

  - `SessionStore.hasActiveSessions` now filters entries by a `"${platform}_"` key prefix, so iOS and Android sessions no longer interfere with each other.
  - Refactored the `keyValueStore` accessor to support a test-only override (`setKeyValueStoreForTest`) so the store can be redirected to a `@TempDir` in unit tests without touching the user's real `~/.maestro/sessions` DB.
  - Added `SessionStoreTest` covering three cases:
    - iOS heartbeat does not surface as an active session for an Android caller.
    - Another active session on the same platform is detected.
    - The caller's own session entry is not counted as "another" active session.

  ## Testing

  - `./gradlew :maestro-cli:test --tests "maestro.cli.session.SessionStoreTest"` — all three new unit tests pass.
  - Manually verified that running an Android flow no longer warns about an "active session" left over from a previous iOS run on the same machine.

  No e2e test added: this is a pure CLI-internal session bookkeeping change with no user-visible UI surface in `e2e/demo_app/` to exercise. Unit tests in `SessionStoreTest` cover the behavior.

  ## Issues fixed

  - iOS/Android sessions cross-contaminating `hasActiveSessions` checks, causing spurious "another session is active" warnings when switching platforms on the same machine.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/session/SessionStore.kt
+++ b/maestro-cli/src/main/java/maestro/cli/session/SessionStore.kt
@@ -7,13 +7,23 @@ import java.util.concurrent.TimeUnit
 
 object SessionStore {
 
-    private val keyValueStore by lazy {
+    private val defaultKeyValueStore by lazy {
         KeyValueStore(
             dbFile = Paths
                 .get(System.getProperty("user.home"), ".maestro", "sessions")
                 .toFile()
                 .also { it.parentFile.mkdirs() }
         )
+    }
+
+    @Volatile
+    private var keyValueStoreOverride: KeyValueStore? = null
+
+    private val keyValueStore: KeyValueStore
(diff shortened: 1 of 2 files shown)


Fishbowler:
#3138 was merged today that's caused the conflicts. It might've solved the need for this?

fantasyRqg:
> #3138 was merged today that's caused the conflicts. It might've solved the need for this?

Yes, it solves the problem with a better solution.

PR #3281, opened 2026-05-11
feat(cli): add direct device-control commands

### LINKED ISSUE #3280: Add direct CLI commands for common device-control actions
## Use case

I use Maestro in an environment where MCP usage is restricted, so most of my local workflow has to happen through the CLI.

For quick debugging, I often want to do small device actions without writing a full flow file first. Things like tapping a visible element, taking a screenshot, launching or stopping an app, typing into a focused field, or pressing back.

Maestro already supports these kinds of actions through flows, and some of this is available in MCP-focused workflows, but there is not a simple direct CLI version for the common one-off actions.

## Proposal

Add direct CLI commands for the common device-control actions:

- `maestro tap-on`
- `maestro take-screenshot`
- `maestro launch-app`
- `maestro stop-app`
- `maestro input-text`
- `maestro back`

Example usage:

```bash
maestro tap-on --text "Login"
maestro take-screenshot --output artifacts/home.png
maestro launch-app com.example.app
maestro input-text "hello world"
maestro back
maestro stop-app com.example.app
```

The goal is to reuse Maestro's existing command/orchestration paths where possible, not add separate driver behavior for the CLI.

This would make the CLI more useful for shell scripts, local debugging, and agent-assisted workflows where MCP is not available or allowed.

## Anything else?

I brought up the general CLI/MCP parity idea in last week's office hours. I have a draft implementation ready and can open a PR if this direction makes sense.

ronnie3786:
## Proposed changes

Fixes #3280.

Adds direct Maestro CLI commands for common device-control actions:

- `maestro tap-on`
- `maestro take-screenshot`
- `maestro launch-app`
- `maestro stop-app`
- `maestro input-text`
- `maestro back`

This is meant for quick local debugging and lightweight scripting when writing a full flow file would be more ceremony than the task needs.

Context: I brought up this general CLI/MCP parity idea in last week's office hours. My company has stricter security constraints around MCP usage, so most of my local workflow has to happen through the CLI.

Implementation notes:

- The commands reuse Maestro/Orchestra command paths where possible instead of adding separate driver behavior.
- Shared direct-device command helpers live under `maestro.cli.devicecontrol`.
- `tap-on` supports selector-style options like `--text`, `--id`, `--index`, `--enabled`, `--checked`, `--focused`, and `--selected`.
- `take-screenshot` writes to the requested output path and creates parent directories when needed.
- `maestro back` follows existing driver behavior. On iOS, `IOSDriver.backPress()` is currently a no-op, so this command is mainly useful where the driver already supports back navigation.

## Proof of work

https://github.com/user-attachments/assets/654beadd-52bd-42ee-a073-75a3b023fa3f

## Testing

- [x] `./gradlew :maestro-cli:compileKotlin :maestro-cli:test --tests 'maestro.cli.command.*' --tests 'maestro.cli.devicecontrol.*'`
- [x] `./gradlew test -x :maestro-android:testDebugUnitTest -x :maestro-android:testReleaseUnitTest`

Note: plain `./gradlew test` is blocked in my local environment because Android SDK location is not configured via `ANDROID_HOME` or `local.properties`. The broader test suite passes when excluding those Android SDK-dependent unit-test tasks.

Manual demo video: TODO before marking ready for review.

## Issues fixed

Fixes #3280

### DIFF
--- a/.github/workflows/test-e2e-ios-intel.yaml
+++ b/.github/workflows/test-e2e-ios-intel.yaml
@@ -119,15 +119,6 @@ jobs:
           retention-days: 7
           include-hidden-files: true
 
-      - name: Upload xctest runner logs
-        uses: actions/upload-artifact@v7
-        if: success() || failure()
-        with:
-          name: xctest_runner_logs
-          path: ~/Library/Logs/maestro/xctest_runner_logs
-          retention-days: 7
-          include-hidden-files: true
-
       - name: Upload screen recording of Simulator
         uses: actions/upload-artifact@v7
         if: success() || failure()
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -499,15 +499,6 @@ jobs:
           retention-days: 7
           include-hidden-files: true
 
-      - name: Upload xctest runner logs
-        uses: actions/upload-artifact@v7
-        if: success() || failure()
-        with:
-          name: xctest_runner_logs
-          path: ~/Library/Logs/maestro/xctest_runner_logs
-          retention-days: 7
-          include-hidden-files: true
-
       - name: Upload screen recording of Simulator
         uses: actions/upload-artifact@v7
         if: success() || failure()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add direct CLI commands `maestro tap-on` and `maestro take-screenshot`, bringing selected interactive CLI capabilities into closer parity with existing MCP tools
+
 ## 2.5.1
 
 - Fix `NoSuchMethodError` on Ktor server startup (login, studio, and cloud commands)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Add direct CLI commands `maestro tap-on` and `maestro take-screenshot`, bringing selected interactive CLI capabilities into closer parity with existing MCP tools
+- Add direct CLI commands `maestro tap-on`, `maestro take-screenshot`, `maestro launch-app`, `maestro stop-app`, `maestro input-text`, and `maestro back`, bringing selected interactive CLI capabilities into closer parity with existing MCP tools
 
 ## 2.5.1
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,10 @@ For quick interactive work, Maestro also supports direct CLI commands such as:
 ```bash
 maestro tap-on --text "Login"
 maestro take-screenshot --output artifacts/home.png
+maestro launch-app com.example.app
+maestro input-text "hello world"
+maestro back
+maestro stop-app com.example.app
 ```
 
 These are useful for ad hoc debugging and lightweight scripting without writing a full flow file first.
@@ -140,5 +144,4 @@ Pricing for Maestro Cloud is completely transparent and can be found on the [pri
 ```
   Built with ❤️ by Maestro.dev
 ```
-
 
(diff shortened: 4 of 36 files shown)


simon-gilmurray:
Thanks for the POC PR @ronnie3786 
This would have been good to discuss as an approach prior, as we've currently gone down the MCP route rather than the CLI route for this functionality.

I'm not really following what the approach here gets you when you have to be so declarative with it to provide all of that info, and a flag and a parameter. Feels simpler and easier to just write the yaml directly?
Do these output anywhere to create a flow/test as you didn't show this, or are they just commands to drive the app?

ronnie3786:
> Thanks for the POC PR @ronnie3786 This would have been good to discuss as an approach prior, as we've currently gone down the MCP route rather than the CLI route for this functionality.
> 
> I'm not really following what the approach here gets you when you have to be so declarative with it to provide all of that info, and a flag and a parameter. Feels simpler and easier to just write the yaml directly? Do these output anywhere to create a flow/test as you didn't show this, or are they just commands to drive the app?

Thanks @simon-gilmurray, very fair response.

Our team would use MCP, but our security team gave us the 🙅 for it. Which is the _only_ reason why I have the need for CLI parity, otherwise we'd just rely on the MCP.

> Feels simpler and easier to just write the yaml directly?

For my personal use, I see a lot of potential value in Maestro for performing stress testing and validation of new dynamic workflows and app discoverability. I agree that it can be more verbose than YAML if you’re building an actual test, but for ad hoc debugging, scripting, or agent-driven local workflows, it avoids creating a temporary flow file just to perform one small action.

> Do these output anywhere to create a flow/test as you didn't show this, or are they just commands to drive the app?

These commands are just commands to drive the app. They do not currently output or generate a flow/test.

simon-gilmurray:
> > Thanks for the POC PR @ronnie3786 This would have been good to discuss as an approach prior, as we've currently gone down the MCP route rather than the CLI route for this functionality.
> > I'm not really following what the approach here gets you when you have to be so declarative with it to provide all of that info, and a flag and a parameter. Feels simpler and easier to just write the yaml directly? Do these output anywhere to create a flow/test as you didn't show this, or are they just commands to drive the app?
> 
> Thanks @simon-gilmurray, very fair response.
> 
> Our team would use MCP, but our security team gave us the 🙅 for it. Which is the _only_ reason why I have the need for CLI parity, otherwise we'd just rely on the MCP.
> 
> > Feels simpler and easier to just write the yaml directly?
> 
> For my personal use, I see a lot of potential value in Maestro for performing stress testing and validation of new dynamic workflows and app discoverability. I agree that it can be more verbose than YAML if you’re building an actual test, but for ad hoc debugging, scripting, or agent-driven local workflows, it avoids creating a temporary flow file just to perform one small action.
> 
> > Do these output anywhere to create a flow/test as you didn't show this, or are they just commands to drive the app?
> 
> These commands are just commands to drive the app. They do not currently output or generate a flow/test.

Appreciate the response @ronnie3786! 
Totally get why you are looking at this route - we haven't heard feedback of many times blocking MCP usage currently.

I've spoken to the team on the product direction here - our issue becomes that we have to implement things that we consider core to the direction of Maestro, and features we take on/inherit we need to ensure we maintain. 
With the niche functionality and usage of this suggested CLI approach, it's not something we can currently merge in and adopt. We're fully-focused on MCP currently and delivering a better agentic experience using this.

We do really appreciate your contribution here and of course you're welcome to maintain a fork of this and even extend it in order to help support your testing!

Fishbowler:
I'm gonna go ahead and kick this into a Draft state so that it's not taking any focus for folks looking to do reviews.

We'd love to offer the flexibility to be All The Things to All The Teams, but folks would want us to maintain all of those use cases, and that's not something we can absorb. As Simon said above, this approach isn't part of our current plans for Maestro.

PR #3282, opened 2026-05-12
feat(orchestra): listener-based artifacts + typed ArtifactManifest, flat per-flow bundle, device logs & crash/ANR

amanjeetsingh150:
Makes Maestro **core** the single owner of a flow's debug artifacts: a listener-based production path, a typed self-documenting **artifact manifest**, a flat per-flow run-root bundle, and in-core device-log + crash/ANR capture — so local and cloud emit the same shape and every consumer reads one model instead of re-deriving artifact types from a folder of files.

> [!NOTE]
> **Scope consolidation.** This PR now carries the whole orchestra-artifacts stack that was previously split across **#3282** (listener API + CLI migration, Phase 0+1), **#3343** (typed `ArtifactManifest` + flat bundle, which was merged into this branch), and **#3346** (device logs + crash/ANR). Those are folded together here so the listener path, the manifest, and the device artifacts land as one reviewable change against `main`.

### 1. Listener-based artifact production (Phase 0+1)
- Orchestra exposes an `OrchestraListener` lifecycle; artifact production is driven by listener callbacks instead of being re-derived after the fact. Listener dispatch is centralized, dispatches once per `onFlowStart`/`onFlowEnd`, is keyed by `sequenceNumber`, and logs (rather than swallows) a throwing listener.
- `runFlow` returns a `FlowResult` (the `debugOutput` field is gone).
- CLI's `TestSuiteInteractor` is migrated onto the listener API (Phase 1); per-flow staging is routed through `TempFileHandler`.

### 2. Shared model (`maestro-orchestra-models`)
`ArtifactManifest`, `ArtifactEntry`, `ArtifactKind`, `ArtifactFormat` — plain data classes, no serialization annotations (tolerance is the consumer's choice). The relative-path layout is producer-private in `maestro-orchestra`'s `BundleLayout`; consumers read each artifact's `relativePath` off the manifest at runtime.

### 3. Self-documenting `manifest.json`
Carries a `$schema` pointing at a **stable, versioned URL** (`…/artifact-manifest/v1.schema.json`), published by a `publish-schemas.yaml` CI job on every `main` push. The `vN` path is the only version signal — a structural change ships `v2` beside `v1`; additive **field** changes overwrite `v1` in place and lean on readers tolerating unknown fields (`additionalProperties: true`). A build test fails if any `ArtifactKind`/`ArtifactFormat` is undocumented, so docs can't drift from code.

> Enum values, unlike fields, are **not** additive-safe (adding one fails readers holding a cached `v1`), so the format enum reserves only what's actually produced. `RRWEB` was dropped (web records MP4 via CDP screencast → JCodec, so it had no producer); `HTML` is kept as the format the reserved `AI_ANALYSIS` kind will emit.

### 4. Core produces the bundle (`maestro-orchestra`)
An `ArtifactCollector` is the single owner of the run-root bundle: every file core writes is **recorded as it's allocated** (no end-of-flow disk scan), so the manifest and each command's `artifacts` list are the same records and can't drift. **The run root itself is the zippable bundle** — no intermediate `artifacts/` folder.
- `commands.json` + `logs/maestro.log` at the run root. Serialized errors are slim (`message` + `debugMessage`); hierarchies/stack traces live in their own artifacts.
- **View hierarchy is its own `SCREEN_HIERARCHY` artifact** (`screen-hierarchy/step-<N>.json` per executed command), linked from each command — no longer inline in `commands.json` (shrank a demo `commands.json` ~246 KB → 4 KB). Consumers that read an inline `hierarchy` field must follow the link.
- **Unified screenshots** — one `screenshots/` folder; with `captureFullArtifacts` on (worker) every step gets `step-<N>.png`, off (local CLI) only the failed step does.
- **Flag-gated heavy capture** (`captureFullArtifacts`, off by default; worker + `--analyze` turn it on): per-step screenshots + the full-run `screen-recording.mp4`.

### 5. Core captures device logs + crash/ANR (was #3346)
Ported from the cloud worker so local and cloud produce the same artifacts; they land in `logs/` and the manifest, **on by default** both local and cloud (cheap and always-useful, unlike the flag-gated screenshots/recording).
- **Driver seam** (`maestro-client`): a `CapturedDeviceArtifact` descriptor + best-effort `Driver` methods (start log capture, stop & collect logs, collect crash/ANR), default no-ops.
- **Orchestra**: `DeviceArtifactCapturer` starts on flow start, collects on flow end (scoped to the app under test + flow start).
- **Android**: logcat (`DEVICE_LOG`, source `android`) + crash + ANR. **iOS**: `simctl log stream` (`simulator`) + `xctest_runner.log` (`xctest`) + `.ips` crash.
- Best-effort throughout — any capture failure logs a warning and is skipped, never failing the flow.

### 6. CLI writes one folder per flow (`maestro-cli`)
Each flow's output folder is resolved upfront and passed as the single `artifactsDir`; Orchestra writes the whole bundle into it. Single and suite runs both produce `<output>/<ts>/<flow>/`.

## Layout
```
<output>/<ts>/<flow>/          # the run root IS the zippable bundle
  manifest.json
  commands.json
  logs/
    maestro.log
    device-logcat.txt / device-simulator.log / device-xctest.log
    crash-report.txt / anr-report.txt          # if found this flow
  takeScreenshot/…             # takeScreenshot command output
  startRecording/…             # startRecording command output
  screenshots/step-N.png       # all steps when flag on; failed step only when off
  screen-hierarchy/step-N.json # per-step view hierarchy (always on)
  screen-recording.mp4         # full-run recording (flag-gated: worker on, CLI off)
```

## Behavior change to flag
`--test-output-dir`, when set, **is** this artifact folder — it now holds the full bundle (incl. `manifest.json`), not just screenshots. Default location (`~/.maestro/tests/<ts>/`) unchanged.

## Merge note (conflicts with `main`)
`main` refactored `AndroidDriver`'s transport (raw `Dadb` → `AndroidDeviceConnection`) while this branch added device-capture code that called `dadb.*`. Resolved by routing logcat/crash/ANR collection through `connection`, and moving the `getAppCrashLogs`/`getActivityManagerLogs` logcat helpers to extend `AndroidDeviceConnection` (its tolerant `shell`) rather than the raw `Dadb` — keeping main's "driver never sees a raw Dadb" encapsulation. Also dropped a dangling `UnicodeNotSupportedError` import (main replaced it with `MaestroException.UnicodeNotSupported`). Full module compile + the ported device unit tests + orchestra artifact tests are green post-merge.

## How this was tested
- **Model + orchestra tests** — manifest round-trips and ignores unknown fields; a build check fails if any kind/format is undocumented; collector records dedicated kinds with run-root-relative paths; step screenshots/recording appear per their flag; hierarchy written per executed command (not inline); typed `artifacts` attribute to the right command; serialized errors carry only `message` + `debugMessage`.
- **Ported device unit tests** — logcat reader, `.ips` crash parser, crash-file finder, with fixtures.
- **Local e2e (iOS sim + web), flag off and on** — OFF bundle (no `screen-recording.mp4`, `SCREENSHOT` = failed step only) vs ON (`SCREEN_RECORDING` MP4 + per-step `screenshots/`); web run confirmed `screen-recording.mp4` is a valid H.264 MP4 recorded as a `SCREEN_RECORDING` entry.
- **Android emulator / crash** — `DEVICE_LOG` (source `android`) + parsed `CRASH_REPORT` with a friendly message.

## Follow-up (separate PR)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -877,6 +919,21 @@ class Orchestra(
...
     }
 
+    /**
+     * Dispatches a terminal command outcome to every effective listener.
+     * Uses the per-command [commandStartTimes] map populated at command start
+     * to compute timings; falls back to "now" for both timestamps if the
+     * start time is missing (which would indicate a programmer error in
+     * Orchestra's own bookkeeping).
+     */
+    private fun dispatchFinished(command: MaestroCommand, outcome: CommandOutcome) {
```
pedro18x: Looks like the new listener dispatch only fires in the top-level `commandFlow` loop, and `executeSubflowCommands` still goes through the legacy lambdas. Flagging because once the worker cuts over (mobile-dev-inc/copilot#2508), commands inside `runFlow` / `runScript` / `repeat` won't reach listeners at all, so `ArtifactsGenerator` would miss nested steps in `commands.json` and skip the failure screenshot + hierarchy when a subflow command fails.

steviec (approved):
Overall, this is GREAT. Reviewed with Claude, these are the things I think are worth fixing:

1. Listener dispatch doesn't reach commands inside subflows (correctness regression)
Pedro already flagged this. But my bigger question: why didn't any tests pick this up? We should make sure we are validating the callbacks at all potential points in our tests

2.  debugOutput as a mutable field on Orchestra is the wrong shape
It obscures the only-listener-Maestro-itself-ships. Is this output read during execution? If not, should it just be the return value from runFlow? e.g.
```
data class FlowResult(val success: Boolean, val debugOutput: FlowDebugOutput)
  suspend fun runFlow(commands): FlowResult
```

maestro-orchestra/src/test/kotlin/maestro/orchestra/debug/OrchestraListenerDispatchTest.kt
```diff
@@ -0,0 +1,76 @@
...
+            maestro = mockMaestro(),
+            listeners = listOf(recording),
+            onFlowStart = { legacyEvents.add("legacyFlowStart") },
+            onCommandStart = { _, _ -> legacyEvents.add("legacyCommandStart") },
+            onCommandComplete = { _, _ -> legacyEvents.add("legacyCommandComplete") },
+        )
+
+        // Run an empty flow — exercises onFlowStart + onFlowEnd only. Listener
+        // dispatch on per-command events is exercised in ArtifactsGeneratorTest.
+        runBlocking { orchestra.runFlow(emptyList()) }
```
proksh: Runs the OrchestraListener on an empty command list. We should have more coverage here.

This is the main reason the bug @pedro18x mentioned has been missed in test.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -170,6 +199,7 @@ class Orchestra(
         initAndroidChromeDevTools(config)
 
         onFlowStart(commands)
+        effectiveListeners.forEach { runCatching { it.onFlowStart() } }
 
         executeDefineVariablesCommands(commands, config)
```
proksh: If this throws, `onFlowEnd` never runs.

Should we move this inside the try catch we have here?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -161,6 +167,29 @@ class Orchestra(
...
+     * populated, regardless of whether [artifactsDir] is set. Read after
+     * [runFlow] for summaries / test assertions / API reporting.
+     */
+    val debugOutput: FlowDebugOutput get() = artifactsGenerator.debugOutput
+
+    /** Global per-flow sequence counter shared with listeners. */
+    private var commandSequenceCounter: Int = 0
+
+    /** Per-command start timestamps, populated at [onCommandStart], read at terminal callbacks. */
+    private val commandStartTimes = mutableMapOf<MaestroCommand, Long>()
```
proksh: If this is being keyed by command, what happens when there are two same commands like:
```
- tapOn: "Login"
- tapOn: "Login"
```

The key will be same, so from what I see - The second start overwrites the first's timestamp. When the second finishes, the key was already removed by the first's finish, so it falls into the ?: finishedAt fallback and reports a duration of 0 ms?

Can you check these cases with tests?

proksh: I would say its also worth testing cases like
```
- repeat:
      - tapOn: "Login"
```

and 
```
- repeat:
    - repeat:
        - tapOn: "Login"
```

proksh: Maybe we should use sequenceNumber along the maestrocommand for key

maestro-client/src/main/java/maestro/debuglog/ScopedLogCapture.kt
```diff
@@ -0,0 +1,128 @@
...
+                config.rootLogger.addAppender(appender, null, null)
+
+                // Log4j2's default config sets the root logger to ERROR. If we
+                // leave it there our INFO/DEBUG `maestro.*` events never reach
+                // the appender. Lower the level for the lifetime of the capture
+                // and restore it on close so we don't permanently affect the
+                // host's logging configuration.
+                val previousLevel = config.rootLogger.level
+                if (previousLevel == null || previousLevel.isMoreSpecificThan(Level.DEBUG)) {
+                    config.rootLogger.level = Level.ALL
```
proksh: Race on config.rootLogger.level. The read at line 99 and the write at line 102 aren't synchronized against another thread doing the same save/restore on the process-global LoggerContext. If capture B starts while capture A is active, B reads ALL (A's elevation) as its "previous" value and on close writes ALL back as the restore, the host process is then stuck at ALL for the rest of the JVM's life.

Can this condition happen where two thread calls this in race condition?

proksh (commented):
@pedro18x feedback seems like a blocking one.

Left a few more comments.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
amanjeetsingh150: Awesome catch @pedro18x fixed this: https://github.com/mobile-dev-inc/Maestro/pull/3282/commits/15b4880d8091ae1baf82a3a67a013eaa599c8350

pedro18x (approved):
All the earlier concerns are addressed, LGTM!

PR #3286, opened 2026-05-12
Make hierarchy command JSON only

Fishbowler:
## Proposed changes

Improves maestro hierarchy for scripting and command chaining.

Previously, `maestro hierarchy` would output a couple of lines of debug info, followed by the JSON. If you wanted that in a file, you'd pipe it yourself.

Now...

Pure JSON on stdout
  - No device selection info
  - No Insight messages

Multi-device error - _this made sense if you're no longer printing the device info_
  - If more than one device is connected and no `--device` is specified, the command now exits with an error: "Multiple devices connected. Please specify a device using --device <id>."
  - If `--platform` is given and narrows it to one device, that device is auto-selected without error

`--output` flag
  - Redirects hierarchy output to a file instead of stdout (`--output hierarchy.json`)
  - Works with both the default JSON and `--compact` CSV modes

Device/platform flags on the subcommand
  - `--platform`/`-p`, `--device`/`--udid`, and `--driver-host-port` are now declared on hierarchy directly, mirroring TestCommand
  - `maestro hierarchy --platform ios` now works; the previous `maestro --platform ios hierarchy` form still works via parent
  fallback

## Testing

Lots of local testing with different flag and devices combinations

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
@@ -91,30 +91,11 @@ class PrintHierarchyCommand : Runnable {
     private var compact: Boolean = false
 
     @CommandLine.Option(
-        names = ["-p", "--platform"],
-        description = ["Select a platform to run on"]
-    )
-    private var platform: String? = null
-
-    @CommandLine.Option(
-        names = ["--device", "--udid"],
-        description = ["Device ID to run on explicitly"]
-    )
-    private var deviceId: String? = null
-
-    @CommandLine.Option(names = ["--driver-host-port"], hidden = true)
-    private var driverHostPort: Int? = null
-
-    @CommandLine.Option(
         names = ["--device-index"],
         description = ["The index of the device to run the test on"],
         hidden = true
     )
     private var deviceIndex: Int? = null
-
-private val effectivePlatform get() = platform ?: parent?.platform
-    private val effectiveDeviceIdFlag get() = deviceId ?: parent?.deviceId
-    private val effectiveDriverHostPort get() = driverHostPort ?: parent?.driverHostPort
 
     override fun run() {
         TestDebugReporter.install(
@@ -123,15 +104,15 @@ private val effectivePlatform get() = platform ?: parent?.platform
             printToConsole = parent?.verbose == true,
         )
 
-        val platform = effectivePlatform ?: "unknown"
+        val platform = parent?.platform ?: "unknown"
         val startTime = System.currentTimeMillis()
         Analytics.trackEvent(PrintHierarchyStartedEvent(platform = platform))
 
         val effectiveDeviceId: String? = when {
-            effectiveDeviceIdFlag != null -> effectiveDeviceIdFlag
+            parent?.deviceId != null -> parent.deviceId
             deviceIndex != null -> null
             else -> {
-                val platformFilter = effectivePlatform?.let { Platform.fromString(it) }
+                val platformFilter = parent?.platform?.let { Platform.fromString(it) }
                 val connectedDevices = DeviceService.listConnectedDevices().withPlatform(platformFilter)
                 when (connectedDevices.size) {
                     0 -> null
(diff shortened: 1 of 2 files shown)


maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -86,45 +91,89 @@ class PrintHierarchyCommand : Runnable {
...
         names = ["--device-index"],
         description = ["The index of the device to run the test on"],
         hidden = true
     )
     private var deviceIndex: Int? = null
 
+    @CommandLine.Option(
+        names = ["--output"],
+        description = ["Write hierarchy output to a file instead of stdout"]
+    )
```
Fishbowler: This is the only change in this PR that I'm in any way doubtful of.
You can already `maestro hierarchy > hierarchy.json` (which will now actually be JSON)
At the same time, it feels in keeping with our other commands. WDYT?

Leland-Takamine: `maestro hierarchy > hierarchy.json` this feels preferable than a separate --output option that does the same thing

pedro18x (commented):
Heads up: ‎`PickDeviceView.showRunOnDevice` is called by ‎`PickDeviceInteractor`, which also runs for test, record, and other commands. That means the “Running on …” line moves to stderr globally. Any user or script parsing stdout from ‎`maestro test` would see that change, which doesn’t seem to be what this PR is aiming for.

It might be worth scoping the suppression to hierarchy only (for example, a “quiet” flag on the interactor, or routing ‎`PrintHierarchyCommand` through a path that skips ‎`showRunOnDevice`). What do you think?

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -86,45 +90,83 @@ class PrintHierarchyCommand : Runnable {
     )
     private var compact: Boolean = false
 
+    @CommandLine.Option(
+        names = ["-p", "--platform"],
+        description = ["Select a platform to run on"]
+    )
+    private var platform: String? = null
```
Leland-Takamine: Duplicating device options on hierarchy feels like a smell - would prefer to think about the correct API and implementation before bundling that into this PR

Leland-Takamine: Easiest path forward here would to be to simply remove these option changes from this PR

Fishbowler:
Good catch @pedro18x. Moreover, since this now resolves to one explicit ID before the session, `showRunOnDevice` isn't ever used and shouldn't have been touched!

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
Fishbowler: We've got the same change on `maestro test` so that we can maintain some usability. I want to include the usability features here since being JSON only naturally drives you to a single device.

Right now, you'd need to do `maestro --platform ios hierarchy` in order to filter by platform, which is gross. 

I thought it a sensible price to pay until we decide we want to look into our picocli infrastructure. Picocli has mixins - we can create a DeviceOptionsMixin that we reuse here, for `test` and on the root app. Happy to do that here, or as a separate PR we land first. Got a preference?

Leland-Takamine: If there's a better way to handle device filtering options in general (eg. have those live after the cli subcommand), let's take that opportunity to make an explicit decision on what that API should be and make things consistent across all CLI commands. This should be a separate proposal from cleaning up the hierarchy output.

maestro-cli/src/main/java/maestro/cli/command/PrintHierarchyCommand.kt
```diff
@@ -99,32 +103,51 @@ class PrintHierarchyCommand : Runnable {
...
+                val platformFilter = parent?.platform?.let { Platform.fromString(it) }
+                val connectedDevices = DeviceService.listConnectedDevices().withPlatform(platformFilter)
+                when (connectedDevices.size) {
+                    0 -> null
+                    1 -> connectedDevices[0].instanceId
+                    else -> {
+                        val hint = if (platformFilter == null) {
+                            " You can also narrow by platform using --platform <android|ios>."
+                        } else ""
+                        throw CliError("Multiple devices connected. Please specify a device using --device <id>.$hint")
```
Leland-Takamine, line 124: Does this change mean that `maestro hierarchy` no longer works when multiple devices are connected?

Fishbowler: Yes, that's an intentional part of the change. If we're not printing device info, then it needs to be predictable.

Leland-Takamine: I see this defined in the PR description ✅

Leland-Takamine (approved):

PR #3287, opened 2026-05-12
Restructure the reporting metadata + properties

Fishbowler:
## Context

BREAKING CHANGE

#2737 introduced a `properties` key to flow config that added the contents to JUnit (and later HTML) reports. This was released in v2.1.0.

#3235 adds specific property names that are for the id and classname used in reports. Currently unreleased.

## Proposed changes

- Moves properties under `reporting` - this is a breaking change for existing flows
- Separates the report config from the report content

Before:
```
appId: com.example
properties:
  junitId: TC-01
  junitClassname: LoginTests
  priority: "High"
---
- launchApp
```

After:
```
appId: com.example
reporting:
  id: TC-01
  classname: LoginTests
  properties:
    priority: "High"
---
- launchApp
```

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
+++ b/maestro-cli/src/main/java/maestro/cli/model/TestExecutionSummary.kt
@@ -25,6 +25,11 @@ data class TestExecutionSummary(
         fun failures(): List<FlowResult> = flows.filter { it.status == FlowStatus.ERROR }
     }
 
+    data class ReportingMetadata(
+        val id: String? = null,
+        val classname: String? = null,
+    )
+
     data class FlowResult(
         val name: String,
         val fileName: String?,
@@ -32,6 +37,7 @@ data class TestExecutionSummary(
         val failure: Failure? = null,
         val duration: Duration? = null,
         val startTime: Long? = null,
+        val reportingMetadata: ReportingMetadata? = null,
         val properties: Map<String, String>? = null,
         val tags: List<String>? = null,
         val steps: List<StepResult> = emptyList(),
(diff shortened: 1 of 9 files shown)


simon-gilmurray:
Isn't properties a more generic property that is mainly used in webhooks and can be for a multitude of things other than strictly "reporting"? Although I guess the act of it being output from Cloud runs could be considered as grouping under reporting...

Fishbowler:
That was my thinking. The information in `properties` was just a grab bag of name-value pairs that get displayed alongside the outputs (whether that's JUnit, HTML, or some purpose in Maestro Cloud). That felt reporting-esque.

amanjeetsingh150 (commented):
@Fishbowler, direction is right (separating reporting config from flow body is the cleaner shape), but I want to suggest a slightly different end state: derive JUnit XML directly from the debug output instead of maintaining a separate Kotlin model that has to be kept in sync.

If the renderer is a fixed walker over events in debug output with auto-property emission under maestro.<event>.<field>, new trace events flow into the report without any core changes, and we can pin the structure.

amanjeetsingh150 (changes requested):
@Fishbowler, direction is right (separating reporting config from flow body is the cleaner shape), but I want to suggest a slightly different end state: derive JUnit XML directly from the debug output instead of maintaining a separate Kotlin model that has to be kept in sync.

If the renderer is a fixed walker over events in debug output with auto-property emission under maestro.<event>.<field>, new trace events flow into the report without any core changes, and we can pin the structure.

Fishbowler:
@amanjeetsingh150 I love the proposal, but I think that's a separate change. This only touches JUnit to alter the routing of a few properties. I think altering how we generate JUnit altogether is unrelated to changing the shape of the YAML, and should be a separate PR.

Issue #3136, opened 2026-04-04
Web: headless on windows fails on assertions


NicoAvanzDev:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

1. Create a basic yaml for web

```yaml
url: https://maestro.mobile.dev
---
- launchApp:
    clearState: true
- assertVisible: "Maestro Studio"
- tapOn: "Maestro Studio"
```

2. Launch `maestro test .\example.yaml --headless`

### Actual results

Fail occur on `assertVisible`

### Expected results

Assertion expected to be true

### About app

Web application

## OS

OS: Windows 11
OS version: 10.0
Architecture: amd64

## Java

java version "25" 2025-09-16 LTS
Java(TM) SE Runtime Environment (build 25+37-LTS-3491)
Java HotSpot(TM) 64-Bit Server VM (build 25+37-LTS-3491, mixed mode, sharing)

### Logs

<details>
<summary>Logs</summary>

```
[ {
  "command" : {
    "applyConfigurationCommand" : {
      "config" : {
        "appId" : "https://maestro.mobile.dev"
      },
      "optional" : false
    }
  },
  "metadata" : {
    "status" : "COMPLETED",
    "timestamp" : 1775307658586,
    "duration" : 3,
    "sequenceNumber" : 1,
    "evaluatedCommand" : {
      "applyConfigurationCommand" : {
        "config" : {
          "appId" : "https://maestro.mobile.dev"
        },
        "optional" : false
      }
    }
  }
}, {
  "command" : {
    "defineVariablesCommand" : {
      "env" : {
        "MAESTRO_FILENAME" : "example",
        "MAESTRO_DEVICE_UDID" : "chromium",
        "MAESTRO_SHARD_ID" : "1",
        "MAESTRO_SHARD_INDEX" : "0"
      },
      "optional" : false
    }
  },
  "metadata" : {
    "status" : "COMPLETED",
    "timestamp" : 1775307658575,
    "duration" : 9,
    "sequenceNumber" : 0,
    "evaluatedCommand" : {
      "defineVariablesCommand" : {
        "env" : {
          "MAESTRO_FILENAME" : "example",
          "MAESTRO_DEVICE_UDID" : "chromium",
          "MAESTRO_SHARD_ID" : "1",
          "MAESTRO_SHARD_INDEX" : "0"
        },
        "optional" : false
      }
    }
  }
}, {
  "command" : {
[...]

PR #3289, opened 2026-05-14
fix: skip broken cross-origin iframe hierarchy reads

### LINKED ISSUE #3271: Looped flow fails when "coming back" to the main window.
I'm attempting to test https://vcplayground.org/issuer with Maestro. Doing one test with one credential option from that page (i.e. "Alumni" or other) works great! It also works fine if I pick one of the flows on that site which do not require pop-ups. However, some of the options (available under the gear menu) invoke a pop-up (initiated from an iframe). That works fine for a single flow and results in the expected screenshot at the "completed" step.

However, if I try to loop through multiples (even just 2...), I get this error:
```
[ WARN] maestro.drivers.CdpWebDriver.fetchCrossOriginIframeContent: Could not find iframe element with src https://authn.io/mediator
org.openqa.selenium.NoSuchWindowException: no such window: target window already closed
from unknown error: web view not found
  (Session info: chrome=147.0.7727.138)
Build info: version: '4.40.0', revision: 'b3333f1'
System info: os.name: 'Windows 10', os.arch: 'amd64', os.version: '10.0', java.version: '25'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Command: [98f1d72528ba59d38965f656f9d20221, executeScript {script=return [...document.querySelectorAll('iframe')].find(f => f.src === arguments[0]);, args=[https://authn.io/mediator]}]
Capabilities {acceptInsecureCerts: false, browserName: chrome, browserVersion: 147.0.7727.138, chrome: {chromedriverVersion: 147.0.7727.117 (474cc805edd..., userDataDir: C:\Users\bigbl\AppData\Loca...}, fedcm:accounts: true, goog:chromeOptions: {debuggerAddress: localhost:60976}, goog:processID: 53516, networkConnectionEnabled: false, pageLoadStrategy: normal, platformName: windows, proxy: Proxy(), se:cdp: ws://localhost:60976/devtoo..., se:cdpVersion: 147.0.7727.138, setWindowRect: true, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify, webauthn:extension:credBlob: true, webauthn:extension:largeBlob: true, webauthn:extension:minPinLength: true, webauthn:extension:prf: true, webauthn:virtualAuthenticators: true}
```

The iframe it's looking for is where the pop-up is initiated from. It is also still in the page being screenshot.

What _seems_ to be happening is that Maestro is looking for that iframe inside that iframe?

I've tried various approaches to end the script and all of the following result in that same error:
```
- launchApp

- stopApp
- launchApp

- openLink: "https://vcplayground.org/issuer"

- tapOn: "Go Home"
```

I'd expected that restarting the app would have "broken free" and "started over"...but that doesn't seem to be the case here, for some reason.

I know this is flying blind a bit (as I'm still writing these tests...), but I was hoping others might have experienced something similar and have some insights.

Thanks!
🎩

schmayterling:
## Proposed changes

fixes a Maestro web hierarchy crash when cross-origin iframe enrichment hits transient browser state or malformed JS/CDP results.

see:
```json
"status" : "FAILED",
  "error" : {
    "stackTrace" : [ {
      "methodName" : "fetchCrossOriginIframeContent",
      "fileName" : "CdpWebDriver.kt",
      "lineNumber" : 696,
      "className" : "maestro.drivers.CdpWebDriver"
    }, {
      "methodName" : "injectCrossOriginIframes",
      "fileName" : "CdpWebDriver.kt",
      "lineNumber" : 655,
      "className" : "maestro.drivers.CdpWebDriver"
    }, {
      "methodName" : "contentDescriptor",
      "fileName" : "CdpWebDriver.kt",
      "lineNumber" : 276,
      "className" : "maestro.drivers.CdpWebDriver"
    }, {
      "methodName" : "from-8JJjmZI",
      "fileName" : "ViewHierarchy.kt",
      "lineNumber" : 29,
      "className" : "maestro.ViewHierarchy$Companion"
    } ],
    "message" : "null cannot be cast to non-null type kotlin.collections.Map<kotlin.String, kotlin.Any>"
  }
```

from my understanding this usually happens after navigation, popup return, or iframe reload. the parent page is still usable, but a third-party iframe can be stale/closed/still loading/detached, or return null/malformed hierarchy data. 

previously, that path used unsafe casts into `Map<String, Any>` and numeric fields, so one broken iframe could crash the whole hierarchy build.

this patch tries keeps the hierarchy path best-effort, see flow below 
1. parse js/cdp hierarchy payloads with checked maps/lists instead of unsafe casts
2. reset selenium to default content before hierarchy and iframe reads
3. preserve readable cross-origin iframe content
4. keep the parent iframe node with empty children when iframe content is bad
5. treat stale frame, missing frame, and closed window errors as transient during iframe enrichment
6. keep the parent page hierarchy usable when a third-party iframe is broken

## Testing

unit tests should cover:
- null iframe content does not crash hierarchy construction
- malformed iframe content does not crash hierarchy construction
- transient iframe/window errors are swallowed and traversal continues
- readable iframe content is still preserved and injected
- malformed child nodes are skipped while valid siblings are preserved
- missing iframe viewport params skip iframe content fetch

ran the following:
```bash
./gradlew :maestro-client:test --tests maestro.drivers.WebHierarchyTest -x :maestro-ios-driver:buildIosDriver
./gradlew :maestro-client:check -x :maestro-ios-driver:buildIosDriver
./gradlew detektMain
```

idt e2e is needed because there already is one for iframes

### Issues fixed
- fixes #3271
- potentially related to #3136

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -681,10 +681,7 @@ class CdpWebDriver(
         } catch (e: Exception) {
             logWebHierarchyFailure("Could not find iframe element with src $iframeSrc", e)
             return null
-        } ?: run {
-            LOGGER.debug("No iframe element found with src $iframeSrc")
-            return null
-        }
+        } ?: return null
 
         // Get the iframe's scaled viewport params (accounts for parent viewportWidth/Height scaling)
         val paramsJson = try {
@@ -695,10 +692,7 @@ class CdpWebDriver(
         } catch (e: Exception) {
             logWebHierarchyFailure("Could not get viewport params for iframe $iframeSrc", e)
             return null
-        } ?: run {
-            LOGGER.debug("No viewport params returned for iframe $iframeSrc")
-            return null
-        }
+        } ?: return null
 
         val params = WebHierarchy.parseIframeViewportParams(paramsJson, iframeSrc) ?: return null
 
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -133,8 +133,10 @@ class CdpWebDriver(
             }
         )
 
-        val options = driver.capabilities.getCapability("goog:chromeOptions") as Map<String, Any>
-        val debuggerAddress = options["debuggerAddress"] as String
+        val options = driver.capabilities.getCapability("goog:chromeOptions") as? Map<*, *>
+            ?: error("Chrome debugger options are unavailable")
+        val debuggerAddress = options["debuggerAddress"] as? String
+            ?: error("Chrome debugger address is unavailable")
         val parts = debuggerAddress.split(":")
 
         cdpClient = CdpClient(
@@ -256,6 +258,7 @@ class CdpWebDriver(
         ensureOpen()
 
         detectWindowChange()
+        switchToDefaultContent("before web hierarchy fetch")
 
         // retrieve view hierarchy from DOM
         // There are edge cases where executeJS returns null, and we cannot get the hierarchy. In this situation
@@ -272,7 +275,8 @@ class CdpWebDriver(
             }
         }
 
-        val rawMap = contentDesc as Map<String, Any>
+        val rawMap = WebHierarchy.normalizeDomNode(contentDesc, "web content description")
+            ?: throw IllegalStateException("Could not parse hierarchy returned by maestro.getContentDescription()")
         val enrichedMap = injectCrossOriginIframes(rawMap)
         val root = parseDomAsTreeNodes(enrichedMap)
         seleniumDriver?.currentUrl?.let { url ->
@@ -282,45 +286,57 @@ class CdpWebDriver(
     }
 
     fun parseDomAsTreeNodes(domRepresentation: Map<String, Any>): TreeNode {
-        val attrs = domRepresentation["attributes"] as Map<String, Any>
-
-        val attributes = mutableMapOf(
-            "text" to attrs["text"] as String,
-            "bounds" to attrs["bounds"] as String,
-        )
-        if (attrs.containsKey("resource-id") && attrs["resource-id"] != null) {
-            attributes["resource-id"] = attrs["resource-id"] as String
-        }
-        if (attrs.containsKey("selected") && attrs["selected"] != null) {
-            attributes["selected"] = (attrs["selected"] as Boolean).toString()
-        }
-        if (attrs.containsKey("synthetic") && attrs["synthetic"] != null) {
-            attributes["synthetic"] = (attrs["synthetic"] as Boolean).toString()
-        }
-        if (attrs.containsKey("ignoreBoundsFiltering") && attrs["ignoreBoundsFiltering"] != null) {
-            attributes["ignoreBoundsFiltering"] = (attrs["ignoreBoundsFiltering"] as Boolean).toString()
-        }
-
-        val children = domRepresentation["children"] as List<Map<String, Any>>
-
-        return TreeNode(attributes = attributes, children = children.map { parseDomAsTreeNodes(it) })
+        return WebHierarchy.parseDomAsTreeNodes(domRepresentation)
     }
 
     private fun detectWindowChange() {
         // Checks whether there are any new window handles available and, if so, switches Selenium driver focus to it
         val driver = ensureOpen()
+        val windowHandles = try {
+            driver.windowHandles
+        } catch (e: Exception) {
+            logIframeFetchFailure("Failed to read window handles while detecting window change", e)
+            return
+        }
 
-        if (lastSeenWindowHandles != driver.windowHandles) {
-            val newHandles = driver.windowHandles - lastSeenWindowHandles
-            lastSeenWindowHandles = driver.windowHandles
+        if (lastSeenWindowHandles != windowHandles) {
+            val newHandles = windowHandles - lastSeenWindowHandles
+            lastSeenWindowHandles = windowHandles
 
             if (newHandles.isNotEmpty()) {
                 val newHandle = newHandles.first()
                 LOGGER.info("Detected a window change, switching to new window handle $newHandle")
 
-                driver.switchTo().window(newHandle)
+                try {
+                    driver.switchTo().window(newHandle)
+                } catch (e: Exception) {
+                    logIframeFetchFailure("Failed to switch to new window handle $newHandle", e)
+                    return
+                }
 
                 webScreenRecorder?.onWindowChange()
+            }
+        }
+        if (windowHandles.isNotEmpty()) {
+            val currentHandle = try {
+                driver.windowHandle
+            } catch (e: Exception) {
+                null
+            }
+
+            if (currentHandle == null || currentHandle !in windowHandles) {
+                val fallbackHandle = lastSeenWindowHandles.firstOrNull { it in windowHandles }
+                    ?: windowHandles.first()
+                LOGGER.info("Switching to available window handle $fallbackHandle before web hierarchy fetch")
+
+                try {
+                    driver.switchTo().window(fallbackHandle)
+                    lastSeenWindowHandles = windowHandles
+                    webScreenRecorder?.onWindowChange()
+                } catch (e: Exception) {
+                    logIframeFetchFailure("Failed to switch to available window handle $fallbackHandle before web hierarchy fetch", e)
+                    return
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


Fishbowler:
> keep the parent iframe node with empty children when iframe content is bad

I'm concerned that this would lead to flaky tests. If run an `assertVisible` and the DOM is the parent content, but skipping the iframes, isn't there a chance my step will fail, but if I rerun it'll succeed?

Should we be handling failures in iframes with retry logic and eventual errors rather than silently eliding parts of the DOM?

schmayterling:
> > keep the parent iframe node with empty children when iframe content is bad
> 
> I'm concerned that this would lead to flaky tests. If run an `assertVisible` and the DOM is the parent content, but skipping the iframes, isn't there a chance my step will fail, but if I rerun it'll succeed?

yeah, I think that can still happen for iframe-targeted assertions if the iframe is temporarily unavailable during that hierarchy snapshot. but yea the goal of this patch is to avoid crashing the entire hierarchy build when a third-party iframe is stale/broken/loading, and keep the rest of the page usable instead of failing with `null cannot be cast ...` which is causing flaky tests directly

> Should we be handling failures in iframes with retry logic and eventual errors rather than silently eliding parts of the DOM?

I think retry/wait behavior for iframe content would need to happen at the assertion/command layer rather than inside hierarchy parsing itself. would be happy to see your thoughts with this perspective

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -689,34 +690,49 @@ class CdpWebDriver(
...
-        val iframeX = (params["viewportX"]      as? Number)?.toDouble() ?: 0.0
-        val iframeY = (params["viewportY"]      as? Number)?.toDouble() ?: 0.0
-        val iframeW = (params["viewportWidth"]  as? Number)?.toDouble() ?: 0.0
-        val iframeH = (params["viewportHeight"] as? Number)?.toDouble() ?: 0.0
+        val params = WebHierarchy.parseIframeViewportParams(paramsJson, iframeSrc) ?: return null
 
         // ChromeDriver can execute scripts inside cross-origin iframes via switchTo().frame()
-        driver.switchTo().frame(iframeElement)
         return try {
+            driver.switchTo().frame(iframeElement)
```
proksh, line 701: This reordering - moving `switchTo().frame()` inside the try so a StaleElementReferenceException degrades to null - is exactly #3314. Let's revert it here and rebase on main to avoid two copies of the same fix.

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -674,34 +672,49 @@ class WebDriver(
...
 
-        val params = jacksonObjectMapper().readValue(paramsJson, Map::class.java) as Map<String, Any>
-        val iframeX = (params["viewportX"]      as? Number)?.toDouble() ?: 0.0
-        val iframeY = (params["viewportY"]      as? Number)?.toDouble() ?: 0.0
-        val iframeW = (params["viewportWidth"]  as? Number)?.toDouble() ?: 0.0
-        val iframeH = (params["viewportHeight"] as? Number)?.toDouble() ?: 0.0
+        val params = WebHierarchy.parseIframeViewportParams(paramsJson, iframeSrc) ?: return null
 
         // ChromeDriver can execute scripts inside cross-origin iframes via switchTo().frame()
-        driver.switchTo().frame(iframeElement)
```
proksh, line 688: Same as above. This reordering - moving switchTo().frame() inside the try so a StaleElementReferenceException degrades to null - is exactly #3314. Let's revert it here and rebase on main to avoid two copies of the same fix.

maestro-client/src/main/java/maestro/drivers/WebHierarchy.kt
```diff
@@ -0,0 +1,201 @@
...
+
+        val children = (domRepresentation["children"] as? List<*>)
+            .orEmpty()
+            .mapNotNull { it as? Map<*, *> }
+            .map { parseDomAsTreeNodes(it) }
+
+        return TreeNode(attributes = attributes, children = children)
+    }
+
+    fun isTransientBrowserContextError(e: Throwable): Boolean {
```
proksh, line 131: This classifies Selenium exceptions as transient - which is what #3315's SeleniumExceptionTranslator is for, at the Driver boundary. Once #3315 lands, callers won't see raw org.openqa.selenium.* here. Suggest removing this and the logWebHierarchyFailure / logIframeFetchFailure helpers that depend on it, and letting  #3315 own the classification.

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -689,34 +690,49 @@ class CdpWebDriver(
...
+            logWebHierarchyFailure("Failed to get content description from cross-origin iframe $iframeSrc", e)
             null
         } finally {
             try { driver.switchTo().defaultContent() }
-            catch (e: Exception) { LOGGER.warn("Failed to switch back to default content", e) }
+            catch (e: Exception) { logWebHierarchyFailure("Failed to switch back to default content", e) }
+        }
+    }
+
+    private fun logWebHierarchyFailure(message: String, e: Exception) {
```
proksh, line 720: Tied to the classification in `WebHierarchy` - drop with `isTransientBrowserContextError`. After #3315, transient/infra failures don't reach this logging path.

maestro-client/src/main/java/maestro/drivers/WebDriver.kt
```diff
@@ -240,57 +240,72 @@ class WebDriver(
             }
         }
 
-        val rawMap = contentDesc as Map<String, Any>
-        val enrichedMap = injectCrossOriginIframes(rawMap)
-        val root = parseDomAsTreeNodes(enrichedMap)
+        val rawMap = WebHierarchy.normalizeDomNode(contentDesc, "web content description")
```
proksh, line 243: This is the part to keep. Replacing `contentDesc` as `Map<String, Any>` (and the casts inside the old `parseDomAsTreeNodes`) with checked parsing is what stops the null cannot be cast to non-null type Map crash. Neither #3314 nor #3315 touches this.

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -133,8 +133,10 @@ class CdpWebDriver(
             }
         )
 
-        val options = driver.capabilities.getCapability("goog:chromeOptions") as Map<String, Any>
-        val debuggerAddress = options["debuggerAddress"] as String
+        val options = driver.capabilities.getCapability("goog:chromeOptions") as? Map<*, *>
```
proksh, line 136: Nice defensive cast - keep. Independent of the iframe work.

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -272,57 +275,72 @@ class CdpWebDriver(
...
+                    driver.switchTo().window(newHandle)
+                } catch (e: Exception) {
+                    logWebHierarchyFailure("Failed to switch to new window handle $newHandle", e)
+                    return
+                }
 
                 webScreenRecorder?.onWindowChange()
             }
         }
+        if (windowHandles.isNotEmpty()) {
```
proksh, line 322: This window-handle fallback in detectWindowChange is a meaningful behavior change beyond the iframe-crash fix. Is it required for #3271, or is it scope that could be its own PR? Easier to review/revert if split out. (Also overlaps conceptually with #3315's NoSuchSessionException / unreachable handling — worth checking they don't fight.)

maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -621,10 +639,10 @@ class CdpWebDriver(
     }
 
     override fun queryOnDeviceElements(query: OnDeviceElementQuery): List<TreeNode> {
-        return when (query) {
-            is OnDeviceElementQuery.Css -> queryCss(query)
-            else -> super.queryOnDeviceElements(query)
+        if (query is OnDeviceElementQuery.Css) {
```
proksh, line 642: Why were these changes needed?

maestro-client/src/main/java/maestro/drivers/WebHierarchy.kt
```diff
@@ -0,0 +1,201 @@
...
+                logIframeFetchFailure("Failed to fetch cross-origin iframe $iframeSrc", e)
+                null
+            }
+            if (iframeContent != null) {
+                normalizeDomNode(iframeContent, "cross-origin iframe $iframeSrc")?.let {
+                    return it
+                }
+            }
+
+            return mapOf(
```
proksh, line 84: This is the silent-elision behavior @Fishbowler flagged: when iframe content is bad we keep the parent node but drop its children. An assertVisible targeting  iframe content could fail on one snapshot and pass on rerun → flaky. 

You already noted retry/wait belongs at the command/assertion layer - agree, but we will prefer a solution that is consistent. We have stopped merging things that results in flaky results.

proksh (commented):
Thanks @schmayterling

- When iframe content is bad we keep the parent node but drop its children. An assertVisible targeting iframe content could fail on one snapshot and pass on rerun → flaky. We have stopped accepting pull requests that results in flaky test, so that will stop us to approve this PR. Do you have a solution that can be consistent?
- Rebase to main and revert the part either of #3314 or #3315 covers.
- Suggest rebasing on top of #3314 + #3315 once they merge so the diff shrinks to just the parsing/enrichment fix.

schmayterling:
closing: no plan to continue this pr, happy for someone else to pick up

Issue #3239, opened 2026-04-29
Android driver handshake fails before flow execution: UNAVAILABLE: End of stream or IOException on deviceInfo


serhatozdursun:
### Is there an existing issue for this?

- [x] I have searched the existing issues and didn't find mine.

### Steps to reproduce

Start Android emulator Pixel_8_Pro_API_35 (emulator-5554).
Run:
maestro --verbose test flows/onboarding/01_skip_onboarding.yaml \
  --device emulator-5554 \
  -e appId=<redacted> \
  -e platform=android \
  -e TEST_USER_EMAIL=<redacted> \
  -e TEST_USER_PASSWORD=<redacted> \
  -e deepLinkSchema=<redacted>

### Actual results

- Flow loads and prints planned steps.
- Immediately after, Maestro fails at Getting device info.
- Error:
   - UNAVAILABLE: End of stream or IOException while processing deviceInfo command
   - previously also observed Connection refused: localhost:7001 in earlier runs.
- Failure occurs before first user action command executes.

### Expected results

Maestro should complete Android driver handshake (deviceInfo) and begin executing flow commands.

### About app

Project type: eCommerce
Source type: closed source
Binary available privately: Android APK (app-yoursclothing-release.apk)
Framework: likely native Android with Jetpack Compose usage (inferred from APK metadata)
Android SDK levels (from APK):
minSdkVersion: 26
targetSdkVersion: 35
compileSdkVersion: 35

### About environment

- OS: macOS (Apple Silicon, arm64)
- Emulator: `Pixel_8_Pro_API_35` (`emulator-5554`)
- Java tested:
  - Temurin 25
  - Temurin 21
- Maestro versions tested:
  - 2.4.0
  - 2.5.0 (fresh Homebrew reinstall)
- Installation method currently:
  - Homebrew official tap: `mobile-dev-inc/tap/maestro`
- Android SDK tooling:
  - `sdkmanager` version: 20.0
- Emulator health checks pass:
  - `adb get-state = device`
  - `sys.boot_completed = 1`
  - `dev.bootcomplete = 1`
  - `adb reverse --list` contains `tcp:7001 tcp:7001`
## Troubleshooting already attempted
- Removed old installer-based Maestro (`~/.maestro`) and PATH entry.
- Full uninstall/reinstall via Homebrew official tap.
- Rebooted emulator repeatedly.
- Restarted ADB (`adb kill-server`, `adb start-server`).
[...]

PR #3290, opened 2026-05-14
feat(android-driver): enable gRPC channel-level retries for transient UNAVAILABLE

proksh:
## Summary

The Android driver opens an in-process gRPC channel against the maestro-driver app on the device. Today the channel has no retry policy, so a transient `UNAVAILABLE` (e.g. brief connectivity blip on the pod-local network) bubbles up as a `StatusRuntimeException` and fails the entire job — the backend then re-queues the run on a different worker, costing ~30-60s per occurrence.

Empirical 14d sample from production worker logs: **109 `INFRA_ERROR`s from `StatusRuntimeException` on this channel**, 85% of them in `callViewHierarchy` (an idempotent read). These would all be quietly recovered by channel-level retry.

Enabling channel-level retry with a bounded service config covers every method on the channel without per-callsite changes:

- **Transparent retries** fire automatically when the RPC never reached the server (e.g. couldn't open the connection). Safe even for non-idempotent calls because the server never saw the request.
- **Configured retries** fire on `UNAVAILABLE` returned by the server, bounded by `maxAttempts=4` with 0.5s → 2s exponential backoff.

If retries are exhausted, behaviour matches today: the `StatusRuntimeException` bubbles up and the existing `INFRA_ERROR` classification re-queues the job. This is purely additive in front of the existing safety net.

## Test plan

- [x] `./gradlew :maestro-client:test --tests AndroidDriverRetryConfigTest` — new test uses an in-process gRPC server with a fake `MaestroDriverImplBase`:
  - UNAVAILABLE on first 2 attempts → call succeeds on attempt 3
  - UNAVAILABLE on every attempt → exhausts at `maxAttempts=4`, bubbles last failure
  - Non-retryable status code (INVALID_ARGUMENT) → called exactly once
- [x] Full `:maestro-client:test` suite green
- [ ] Post-deploy: watch the worker-side `INFRA_ERROR` rate for `StatusRuntimeException` from `AndroidDriver` — expect it to drop by ~85% (the `callViewHierarchy` chunk).


Fishbowler:
>a transient UNAVAILABLE (e.g. brief connectivity blip on the pod-local network) bubbles up as a StatusRuntimeException and fails the entire job — the backend then re-queues the run on a different worker, costing ~30-60s per occurrence.

Does this PR have applicability to local execution too?

If #3239 is transient, this would fix this issue?
Does this supersede #2934?

proksh:
> Does this PR have applicability to local execution too?

Yes. 

> If #3239 is transient, would this fix it?

If the error reproduces only intermittently - yes. But from what I see from issue, user says the failure happens before the first user action and reproduces across Maestro 2.4.0 / 2.5.0, Java 21 / 25, fresh Homebrew reinstalls, ADB restarts, etc. That seems more like a stable setup issue. Maybe worth asking user to test on this branch? 

> Does this supersede #2934?

Partially yes. The channel-recreation and keep-alive changes there are independent and worth landing separately if they're still needed

pedro18x (approved):
LGTM

PR #3003, opened 2026-02-19
Retry hierarchy on kAXErrorInvalidUIElement

Fishbowler:
Originally: https://github.com/dincozdemir/Maestro/commit/f1edd5de8a1b83b72507d7165438682d0e3556dc

## Proposed changes

Imported from a fork.

kAXErrorInvalidUIElement errors occur when elements are removed in the middle of an XCTest operation. It's rare, but common enough that I've seen it pop up in the community Slack a bunch of times, and even in Maestro Cloud runs.

## Testing

tbc

## Issues fixed

#2617

PR #3291, opened 2026-05-14
feat(http-client): opt-in retry on transient 5xx and IOException, enabled for XCTest

proksh:
## Summary

Adds an opt-in OkHttp application interceptor to `HttpClient.build` that retries HTTP calls on:

- Transport-level `IOException` thrown by `chain.proceed()` (connect refused, "no route to host", socket reset, etc.).
- HTTP **5xx** responses from the server.

4xx responses pass through unchanged (deterministic client errors).

Bounded by `retryMaxAttempts=3` (default) with `0.5s → 2s` exponential backoff. Retry events are recorded on the existing `http.client.retries` micrometer counter (`kind=http_5xx` or exception class name) so we can observe how often the new path is firing post-deploy.

### Enabled for XCTestDriverClient

Empirical 14d sample from production worker logs: **117 `INFRA_ERROR`s from `XCUITestServerError`** (5xx with `"kAXErrorInvalidUIElement"` and similar transient state from the XCTest runner while the UI is updating). With retry on the HTTP client, the rest of the runner observes a single logical call and the job no longer needs to be re-queued on a different worker just to absorb a transient hiccup.

If retries are exhausted, the last response or `IOException` is surfaced unchanged, so the existing `XCUITestServerError` classification (`AppCrash`, `OperationTimeout`, `UnknownFailure` → `INFRA_ERROR`) still fires as the safety net.

Other callers of `HttpClient.build` default to `retryOnTransientFailure = false` — no behavioural change for them.

## Test plan

- [x] `./gradlew :maestro-utils:test --tests TransientFailureRetryInterceptorTest` — 7 new cases:
  - 5xx then 200 → returns 200 in one logical call (verified via `metrics.counter("http.client.retries", ...)`)
  - 5xx on every attempt → surfaces last response after `maxAttempts`
  - 4xx → not retried, returned immediately
  - `IOException` on every attempt → retries bounded by `maxAttempts`, then surfaces
  - `IOException` recovers if next attempt succeeds
  - Retry disabled → 5xx passes through unchanged
  - 2xx → exactly one call, no retries
- [x] `./gradlew :maestro-client:test --tests XCTestDriverClientTest` — existing 10 cases still pass. Updated the AppCrash parameterized test to enqueue three 500 responses (matches the new default `maxAttempts`).
- [x] Full `:maestro-utils:test`, `:maestro-client:test`, `:maestro-ios-driver:test` green.
- [ ] Post-deploy: watch the worker-side `INFRA_ERROR` rate for `XCUITestServerError.UnknownFailure` and the new `http.client.retries{client=XCTestDriverClient}` counter for visibility.

## Pairs with

[#3290](https://github.com/mobile-dev-inc/Maestro/pull/3290) (Android gRPC channel retry). Together they cover ~40% of worker-side `INFRA_ERROR` volume by the 14d empirical breakdown.


Fishbowler:
❤️ 

Makes #3003 unneeded I think.

Fishbowler (approved):

PR #2057, opened 2024-09-15
Fix platform argument

tokou:
## Proposed changes

- Fix issue with platform not working for iOS
  - This was due to `selectDevice` being passed the platform as well as a deviceId corresponding to the first available device
  - The fix is to filter available devices on the given platform if passed. Which makes sense
- Add test specific `platform` and `device` argument  in order to facilitate commands like `maestro test flow.yml --platform android` which did not work before

## Testing

- Run commands with 1 android and 1 iOS device running
  - `maestro test flow.yml --platform android`
  - `maestro test flow.yml --platform ios`
  - `maestro --platform android test flow.yml`
  - `maestro --platform ios test flow.yml`
  - `maestro --platform android test flow.yml --platform ios`
  - `maestro test flow.yml --platform ios --device emulator-5554`

## Issues fixed

Fixes #2037

PR #3292, opened 2026-05-14
Reusable CLI Arguments

Fishbowler:
### The Problem

The Maestro CLI has grown organically, and a number of @Option-annotated fields are copy-pasted across multiple commands.

This means:

- A change to a shared option's description, default, or behaviour has to be made in multiple places and it's easy to miss one. This has happened in practice (see bugs fixed below).
- New commands inherit inconsistencies from whichever command was used as a template.
- Options that should be available on a command are simply absent, because nobody thought to add them.

### The Solution

picocli's @Mixin mechanism lets you define a group of options once in a standalone class and compose them into any command. This PR extracts all meaningfully shared options into mixins, placing them in a new maestro.cli.mixin sub-package. Any new commands or any maintenance to existing commands can cherry pick from a set of reusable blocks.

The canonical definition of each option now lives in exactly one place. Any future change — rewording a description, adjusting a default, adding an alias — happens once and propagates everywhere.

In the future:
- New commands get shared options for free by adding a @Mixin field.
- Option consistency is structural rather than relying on developers remembering to copy changes across files.

### Incidental Bug Fixes

Bugs fixed as a side-effect

1.  --format HTML-DETAILED rejected by maestro cloud
  TestCommand had a custom ReportFormat.Converter that matches the hyphenated display name HTML-DETAILED; CloudCommand was missing it and fell back to picocli's default enum matching, which requires the exact Java identifier HTML_DETAILED.
  ReportOutputMixin carries the converter, so both commands now accept either form.

2. MAESTRO_API_URL env var ignored by maestro cloud and maestro test
  ChatCommand correctly used EnvUtils.BASE_API_URL (which reads MAESTRO_API_URL) as the default for --api-url. CloudCommand and TestCommand hardcoded "https://api.copilot.mobile.dev" directly. ApiClientMixin uses BASE_API_URL, fixing both.

### Changes in detail

These should be reviewable commit-by-commit - I know there's a lot of churn

**DeviceSelectionMixin** — --platform / -p, --device / --udid, --driver-host-port

  Existed on App (global) and were duplicated onto TestCommand for ergonomics (in #2057), but were never added to RecordCommand or PrintHierarchyCommand. This meant `maestro record --platform android ...` didn't work — the flag had to come before the subcommand name. The mixin adds them to all three commands.

**AppleTeamIdMixin** — --apple-team-id (hidden)

  Shared across TestCommand, RecordCommand, PrintHierarchyCommand, DriverCommand.

**TagFilterMixin** — --include-tags, --exclude-tags

  Shared across TestCommand, CloudCommand.

**ReinstallDriverMixin** — --reinstall-driver (negatable)

  Shared across TestCommand, PrintHierarchyCommand.

**EnvMixin** — -e / --env

  Shared across TestCommand, RecordCommand, CloudCommand.

**DebugOutputMixin** — --debug-output

  Shared across TestCommand, RecordCommand.

**ConfigFileMixin** — --config

  Shared across TestCommand, RecordCommand, CloudCommand.

**DeviceConfigMixin** — --device-locale, --device-model, --device-os

  Shared across CloudCommand, StartDeviceCommand. Uses @Spec(Target.MIXEE) to set command-specific descriptions: maestro cloud 
  --help references maestro list-cloud-devices; maestro start-device --help references maestro list-devices.

**ReportOutputMixin** — --format, --output, --test-suite-name

  Shared across TestCommand, CloudCommand. Carries the ReportFormat.Converter (bug fix above).

**ApiClientMixin** — --api-key / --apiKey, --api-url / --apiUrl

  Shared across TestCommand, CloudCommand, ChatCommand. Uses @Spec(Target.MIXEE) to label these as (used with --analyze) on TestCommand where they serve the AI insights beta feature rather than general cloud auth. Also standardises camelCase aliases.

### Remaining duplicates not addressed in this PR

  --platform in ListDevicesCommand and ListCloudDevicesCommand — both define an identical filter option (String?, same description, no alias). These are semantically distinct from DeviceSelectionMixin's --platform (which selects a connected device to run on) so they weren't folded in. Decided against adding a separate Mixin that might confuse things.

  --api-url in LoginCommand — LoginCommand defines its own --api-url / --apiUrl without a companion --api-key, so it doesn't fit ApiClientMixin.

## Testing

Lots of local commands, especially `--help`. So far, it's looking sensible.

## TL;DR

  1. Refactors arguments for CLI into mixins (for consistent names + descriptions, less code duplication)                      
  2. Adds device selection for `maestro hierarchy` and `maestro record` commands 
  3. Fixes `--format HTML-DETAILED` being rejected by `maestro cloud`
  4. Fixes `MAESTRO_API_URL` env var being ignored by `maestro cloud` and `maestro test`


Leland-Takamine:
This seems to skip to implementation before addressing the first question we should answer before any API change which is: What is the desired API? Writing out a lightweight version of user docs (or in this case CLI usage docs) is often an effective way to communicate API changes

Fishbowler:
@Leland-Takamine This is intentionally trying to stay away from API changes at this point (only fixing a couple of bugs along the way) whilst improving code hygiene and setting us up for the future.

PR #3294, opened 2026-05-15
ci: auto-update drivers on main instead of blocking PRs

Fishbowler:
## What

Replaces `check-drivers.yaml` with a new `update-drivers.yaml` workflow that automatically builds and commits fresh drivers to `main` after every merge that touches driver source — and adds explicit driver build steps to the E2E test workflow.

## Why this is better

The old approach required contributors to pre-build drivers locally and commit them as part of their PR. This was friction-heavy and error-prone: easy to forget, the iOS driver comparison logic was fragile, and it blocked PRs on a manual step that CI could just do itself. Either people forgot and it caused problems, or people added the wrong ones and it caused problems.

The new approach:

- **E2E tests build drivers locally.** Explicit "Build Android drivers" and "Build iOS drivers" steps in `test-e2e.yaml` ensure E2E runs always exercise the freshest local build.
- **No manual driver commits on PRs.** Contributors change source; CI keeps the committed artefacts in sync automatically.
- **Drivers stay fresh on `main`.** The workflow triggers on push to `main` with path filters, builds the relevant drivers, and commits them back with `[skip ci]` to avoid recursive runs.
- **Self-healing artefacts.** Artefact file paths are included in the path triggers and `changes` filter. If a contributor with a different local toolchain commits driver artefacts directly, CI immediately rebuilds them with the canonical toolchain and uses them for e2e. When you reach `main` they're rebuilt and committed. No PR gate needed — the wrong thing is just corrected automatically.
- **Manual trigger available.** `workflow_dispatch` rebuilds both drivers unconditionally — useful after toolchain upgrades (e.g. Xcode version, GitHub runner change) or to unstick a broken state.

## Changes

- **`.github/workflows/check-drivers.yaml` → `update-drivers.yaml`** — new trigger (`push: main` + artefact paths), jobs that build and commit artefacts, `contents: write` permissions, concurrency group, `workflow_dispatch` support
- **`.github/workflows/test-e2e.yaml`** — adds "Build Android drivers" step (and renames "Build xctest-runner" → "Build iOS drivers")

## Follow-up

`maestro-client`'s `checkAndroidApksFresh` Gradle task and the `maestro-android-source.sha256` sentinel file are now vestigial — they enforced the old "contributors must commit fresh APKs" contract. A follow-up PR will remove them.


Fishbowler:
Does this need to touch a non-workflow file in order to trigger workflows or something?

.github/workflows/update-drivers.yaml
```diff
@@ -0,0 +1,125 @@
+name: Update Drivers
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'maestro-ios-xctest-runner/**'
+      - 'maestro-android/**'
+      # Artefact paths: if wrong artefacts land on main via any route (e.g. a
```
amanjeetsingh150: nit: artifacts

.github/workflows/update-drivers.yaml
```diff
@@ -0,0 +1,125 @@
...
+      # [skip ci] commit prevents re-triggering.
+      - 'maestro-client/src/main/resources/maestro-app.apk'
+      - 'maestro-client/src/main/resources/maestro-server.apk'
+      - 'maestro-client/src/main/resources/maestro-android-source.sha256'
+      - 'maestro-ios-driver/src/main/resources/**'
+  workflow_dispatch: {}
+
+concurrency:
+  group: update-drivers
+  cancel-in-progress: false
```
amanjeetsingh150, line 22: Question: Would this mean we get drivers from latest main commit in case there are 2 simultaneously at play?

.github/workflows/update-drivers.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - uses: dorny/paths-filter@v4
+        id: filter
+        if: github.event_name != 'workflow_dispatch'
+        with:
+          filters: |
+            android:
+              - 'maestro-android/**'
+              - 'maestro-client/src/main/resources/maestro-app.apk'
+              - 'maestro-client/src/main/resources/maestro-server.apk'
+              - 'maestro-client/src/main/resources/maestro-android-source.sha256'
```
amanjeetsingh150, line 54: I'm actually thinking we can get rid of this sha256 generation this was to check if we go out of sync. But if we generate directly on latest main always, I'm not worried for going out of sync.

What do you think?

amanjeetsingh150 (approved):

.github/workflows/update-drivers.yaml
amanjeetsingh150: https://github.com/mobile-dev-inc/Maestro/pull/3295

Ohh you're ahead of my game!! 🎉 🙌

.github/workflows/update-drivers.yaml
Fishbowler: 🇬🇧 😭

.github/workflows/update-drivers.yaml
Fishbowler: Removed in #3295 - I wanted to keep them small and easy to review.

.github/workflows/update-drivers.yaml
amanjeetsingh150: Ohh i didn't knew that lol, this works fine as well lol

.github/workflows/update-drivers.yaml
Fishbowler: Yes. Whilst we could cancel the in-progress one, we'd need to know how far we got (we've no idea if Android were already committed, for example). 

I think the safer option is to pin the checkout in the driver update jobs so that we're updating the drivers for each commit in main.

.github/workflows/update-drivers.yaml
Fishbowler: I'll update it and keep consistent with the rest of the codebase.

.github/workflows/update-drivers.yaml
Fishbowler: Fixed, I think. WDYT?

amanjeetsingh150 (approved):

PR #3297, opened 2026-05-15
fix(mcp-visualizer): support physical Android devices

Leland-Takamine:
## Summary
- Visualizer launched `simulator-server` with the `android` subcommand for everything — physical Android serials panicked at `android_emulator.rs:40` (`emulator with id "<serial>" should be running`).
- Even with the right subcommand, simulator-server couldn't find `screen-sharing-agent.jar` + per-ABI `.so` files because `installSimulatorServer()` only unpacked the binary itself, not its sibling `resources/` tree.

This PR:
- Bumps `simulatorServer` to `f4b4f058`, which introduces the `android_device` subcommand and ships the screen-sharing-agent resources alongside the binary.
- Plumbs a typed `StreamDeviceType` (`android` | `android_device` | `ios`) through `DeviceStreamTarget`, `MaestroConnected`, and `McpMaestroSession`. The launch site picks the right subcommand from `Device.DeviceType` at the source. Browser carries the field opaquely in `/api/device/start`. Invalid combos (BROWSER, real iOS) are unrepresentable — `StreamDeviceType.forConnected()` returns null and the device is filtered out of `/api/device/targets`.
- Adds `Unpacker.unpackTree(classpathPrefix, targetDir, executableEntries)` that walks every classpath resource under a prefix and copies it to disk preserving layout. Uses `FileSystems.newFileSystem` to share one walker across jar and filesystem resources. Verifies all declared `executableEntries` were present so packaging bugs fail loudly with a clear message. `installSimulatorServer()` now unpacks the whole `deps/simulator-server/<platform>/` subtree.

## Test plan
- [x] Physical Android device: open visualizer → `stream_ready` reached, MJPEG video shows
- [x] Android emulator: still uses `android` subcommand, no regression
- [x] iOS simulator: still uses `ios` subcommand, no regression
- [x] Re-running `installSimulatorServer()` is idempotent (mtime unchanged when SHA matches)
- [x] Bogus classpath prefix or missing declared executable throws with informative message


maestro-cli/src/main/java/maestro/cli/mcp/McpMaestroSessionManager.kt
```diff
@@ -74,19 +79,39 @@ internal class McpMaestroSessionManager : AutoCloseable {
     private fun createAndroidSession(device: Device.Connected): McpMaestroSession {
         val dadb = Dadb.list().find { it.toString() == device.instanceId }
             ?: error("Unable to find device with id ${device.instanceId}")
+        val streamDeviceType = if (device.deviceType == Device.DeviceType.EMULATOR) {
+            StreamDeviceType.ANDROID
+        } else {
+            StreamDeviceType.ANDROID_DEVICE
+        }
```
pedro18x: This `if/else` picking `ANDROID` vs `ANDROID_DEVICE` based on `EMULATOR` is already encoded in `StreamDeviceType.forConnected()` over in `McpVisualizerServer.kt`, so the same mapping now lives in two spots and they can silently drift. Easy to imagine one getting updated (say, when we add another `DeviceType`) and the other not.

Probably cleaner to call `forConnected(device)` once in `createSession()`, bail on null, and pass the resolved `StreamDeviceType` into `createAndroidSession` / `createIosSession`. Keeps the mapping in one place.

Leland-Takamine: good catch updating

pedro18x (approved):
Nice, much cleaner

PR #3298, opened 2026-05-17
feat(mcp-visualizer): pending-commands UX and panel redesign

Leland-Takamine:
## Summary
- Seed the full flow as PENDING up front so users see *what's coming*, not just what already ran (`maestro.flow_state` snapshots now drive a pure projection on the client).
- Redesign the command panel: floating translucent card sized to the device, status-tinted connector lines between icons, circle status icons with glyphs, syntax-highlighted YAML rows, quiet scrollbar.
- Make the running row unmistakable (saturated opaque sky background, ring + halo, no animation since state turns over too fast for an animation to read).
- Auto-scroll to the currently-running command instead of the list bottom (with pending rows queued past it, the bottom is no longer the right target).
- Fix an out-of-order snapshot delivery bug: publishing through `scope.launch` on `Dispatchers.IO` could reorder events across threads and the frontend's blind-overwrite merge would regress statuses. Funneled events through `Dispatchers.IO.limitedParallelism(1)` so writes are FIFO; pushed `deviceStream.start()`'s 30s side effect to its own launch.
- Add dark mode

https://github.com/user-attachments/assets/774bcfb1-9f4d-4ff2-a62e-c3eccb71836d

## Test plan
- [x] `maestro mcp` → connect a device → run a flow → all rows seed as pending, then transition started → completed in order without flicker.
- [x] Run a flow with a failing command midway → failing row goes red, trailing rows become skipped, nothing stays stuck in pending.
- [x] Watch the panel during a fast flow — running row remains visible (auto-scrolled), highlight is readable, no animation jitter.
- [x] Toggle system dark mode — panel, highlight, connector, scrollbar all read correctly in both.
- [x] Collapse + expand the panel via the chevron.
- [x] Disconnect the device → panel switches to "No device connected" placeholder card with phone-aspect dashed outline; reconnect and stream resumes.
- [x] `/?demo` sandbox: toggle "running" and "device connected" — all states render.

### DIFF
--- a/maestro-cli/mcp-visualizer/package-lock.json
+++ b/maestro-cli/mcp-visualizer/package-lock.json
@@ -14,6 +14,8 @@
         "vite": "^6.0.0"
       },
       "devDependencies": {
+        "@types/react": "^19.2.14",
+        "@types/react-dom": "^19.2.3",
         "typescript": "^5.7.0",
         "vite-plugin-singlefile": "^2.3.3"
       }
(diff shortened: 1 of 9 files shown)


pedro18x:
Overall this looks great, one small nit:

`maestro-cli/mcp-visualizer/package.json` pins `@types/react: ^19` and `@types/react-dom: ^19` against `react: ^18.3` / `react-dom: ^18.3`. `vite build` doesn't typecheck so it won't break the build today, but the React 19 types differ in a few spots (ref types, `JSX` namespace, `ReactNode` children) and will bite if anyone wires up `tsc --noEmit` in CI or hits one in an editor. Worth pinning both to `^18.3.0` to match runtime.

Leland-Takamine:
> Overall this looks great, one small nit:
> 
> `maestro-cli/mcp-visualizer/package.json` pins `@types/react: ^19` and `@types/react-dom: ^19` against `react: ^18.3` / `react-dom: ^18.3`. `vite build` doesn't typecheck so it won't break the build today, but the React 19 types differ in a few spots (ref types, `JSX` namespace, `ReactNode` children) and will bite if anyone wires up `tsc --noEmit` in CI or hits one in an editor. Worth pinning both to `^18.3.0` to match runtime.

pinned to 19

pedro18x (approved):

PR #3299, opened 2026-05-18
chore(cli): remove bundled Maestro Studio in favor of desktop app

proksh:
## Summary

- `maestro studio` no longer spins up the bundled web Studio. It now prints a short notice and the OS-appropriate download link for the new Maestro Studio desktop app:
  - macOS → `https://studio.maestro.dev/MaestroStudio.dmg`
  - Windows → `https://studio.maestro.dev/MaestroStudio.exe`
  - Linux → `https://studio.maestro.dev/MaestroStudio.AppImage`
- The `:maestro-studio:server` Gradle dependency is removed from `maestro-cli`, so the Studio backend is no longer shipped inside the CLI artifact.
- The deprecation message printed by `maestro query` is updated to point at the same download links instead of the now-defunct `maestro studio` workflow.
- OS → URL resolution lives in a single top-level helper at `maestro-cli/src/main/java/maestro/cli/util/StudioDownload.kt` so both commands share one source of truth.

## Test plan

- [x] `./gradlew :maestro-cli:compileKotlin` succeeds locally (verified)
- [x] Run `maestro studio` on macOS → expect notice + `MaestroStudio.dmg` URL
- [ ] Run `maestro studio` on Windows → expect notice + `MaestroStudio.exe` URL
- [ ] Run `maestro studio` on Linux → expect notice + `MaestroStudio.AppImage` URL
- [x] Run `maestro query` (any platform) and confirm the deprecation message points to the correct download URL for that OS


Fishbowler:
Why not also delete the studio source?

proksh:
> Why not also delete the studio source?

Thanks for pointing out. Doing that as well. :)

amanjeetsingh150 (approved):

Fishbowler (approved):

PR #3171, opened 2026-04-12
feat: use direct ADB socket for gRPC instead of TCP port forwarding

Leland-Takamine:
## Summary

The TCP forwarding thread often got into a bad state especially when trying to run Maestro Studio alongside a Maestro CLI or Maestro MCP. This PR completely removes the need for the TCP forwarding thread.

- Replace TCP port forwarding (`dadb.tcpForward`) with a custom `AdbSocketFactory` that routes gRPC traffic directly over ADB streams via `OkHttpChannelBuilder`
- Eliminates host-side port allocation, avoiding port conflicts and removing an extra network hop

This simplifies the architecture, improves performance, and removes one of the two blockers for running multiple instances of Maestro. The second is proper instrumentation apk / process management (eg. second Maestro instance re-installing the instrumentation apk that's currently running).

DadbChromeDevToolsClient was using a similar approach, so replace that existing functionality with AdbSocketFactory.

## Benchmark (5 rounds, A/B interleaved, Pixel 9 emulator)

**assertVisible flow** (simple: launch + 3 assertions):

| Variant | Mean | Median | Stdev |
|---------|------|--------|-------|
| tcpforward | 12.709s | 12.606s | 0.379s |
| **socket** | **7.506s** | **7.433s** | **0.342s** |

**~41% faster**

**fill_form flow** (complex: launch + tap + text input + assertions):

| Variant | Mean | Median | Stdev |
|---------|------|--------|-------|
| tcpforward | 27.170s | 27.028s | 0.615s |
| **socket** | **22.370s** | **22.405s** | **0.290s** |

**~18% faster**

## Test plan

- [x] Unit tests pass (`AdbSocketFactoryTest` — socket creation, I/O, lifecycle)
- [x] All E2E tests pass on real emulator
- [x] Added a new test to cover the `androidWebViewHierarchy: devtools` config
- [x] A/B benchmark shows consistent improvement across 5 interleaved rounds

PR #3300, opened 2026-05-18
fix: preserve hostname in DummyDns so devtools augmentation works

proksh:
## Summary

`androidWebViewHierarchy: devtools` was silently returning empty webview hierarchies, so flows that rely on webview-only elements stopped finding them. Hierarchy discovery via `cat /proc/net/unix` still worked; the actual `/json` fetch and the devtools websocket eval did not.

### Root cause

In `DadbChromeDevToolsClient`, `DummyDns` returns `InetAddress.getLoopbackAddress()`. OkHttp's route selector then builds `InetSocketAddress(InetAddress, port)` — that constructor copies `addr.getHostName()` into the holder, which for the loopback address is `"localhost"`, erasing the original webview socket name (e.g. `webview_devtools_remote_25535`). `AdbSocket.connect()` reads `endpoint.hostString` to build `dadb.open("localabstract:<host>")`, so it ends up calling `dadb.open("localabstract:localhost")` — which fails. The `IOException` is swallowed by the catch blocks in `getWebViewInfos` / `getWebViewTreeNodes`, so the hierarchy is never augmented.

Verified empirically against OkHttp 4.12.0: with the old `DummyDns`, the host captured at `socket.connect()` is `"localhost"`; with the fix it is the original webview socket name.

### Fix

Switch `DummyDns` to `InetAddress.getByAddress(hostname, loopback bytes)` so the original hostname round-trips through to `endpoint.hostString` and `AdbSocketFactory` opens the correct unix abstract socket.

### Regression window

Introduced in #3171 (Apr 14) — the prior `.dadb(dadb)` HTTP client routed `localabstract.<socket>.adb` hosts via its own resolver, which preserved the socket name; the new SocketFactory + DummyDns pair did not.

The existing e2e flow `e2e/demo_app/.maestro/webView_chromeDevTools.yaml` did not catch this because its `assertVisible: Swag Labs` text is also exposed by WebView's accessibility tree, so the flow passes even when the chrome-devtools augmentation contributes nothing.

## Test plan

- [x] Added a unit test that exercises the OkHttp address-constructor path (`InetSocketAddress(InetAddress, port)`) and asserts the host string survives the round-trip. Confirmed it fails on the old code (`expected: webview_devtools_remote_25535, but was: localhost`) and passes on this branch.
- [x] `:maestro-client:test` for `DadbChromeDevToolsClientTest`, `AdbSocketFactoryTest`, and `AndroidWebViewHierarchyClientTest` — all green.
- [ ] Re-run the affected customer flows on Maestro Cloud after the worker bump.

### DIFF
--- a/e2e/demo_app/.maestro/webView_chromeDevTools_idMatching.yaml
+++ b/e2e/demo_app/.maestro/webView_chromeDevTools_idMatching.yaml
@@ -1,0 +1,18 @@
+appId: com.example.example
+androidWebViewHierarchy: devtools
+tags:
+  - passing
+  - android
+---
+# Regression guard for the DummyDns hostname fix (commit e8a87992).
+# The `<div id="devtools-regression-marker">` in `webview_devtools_test_screen.dart`
+# is non-interactive and aria-hidden — it should NOT surface in Android's native
+# a11y tree, so the `id:` match below only succeeds when devtools augmentation is
+# working. A silent devtools failure (the original bug) makes this flow fail.
+- launchApp:
+    clearState: true
+- scrollUntilVisible:
+    element: Webview Devtools Test
+- tapOn: Webview Devtools Test
+- assertVisible:
+    id: "devtools-regression-marker"
--- a/e2e/demo_app/lib/main.dart
+++ b/e2e/demo_app/lib/main.dart
@@ -16,6 +16,7 @@ import 'package:demo_app/gesture_tester_screen.dart';
 import 'package:demo_app/scrollable_list_screen.dart';
 import 'package:demo_app/sensors_screen.dart';
 import 'package:demo_app/webview.dart';
+import 'package:demo_app/webview_devtools_test_screen.dart';
 import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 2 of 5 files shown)


Fishbowler:
Should this include an e2e test that would fail if there were another regression at this layer?

What would that need to be? An ID?

proksh:
> Should this include an e2e test that would fail if there were another regression at this layer?
> 
> What would that need to be? An ID?

Totally, I have added the test - verified that it fails on current maestro version, and passes on this branch.

e2e/demo_app/.maestro/webView_chromeDevTools_idMatching.yaml
```diff
@@ -0,0 +1,18 @@
...
+---
+# Regression guard for the DummyDns hostname fix (commit e8a87992).
+# The `<div id="devtools-regression-marker">` in `webview_devtools_test_screen.dart`
+# is non-interactive and aria-hidden — it should NOT surface in Android's native
+# a11y tree, so the `id:` match below only succeeds when devtools augmentation is
+# working. A silent devtools failure (the original bug) makes this flow fail.
+- launchApp:
+    clearState: true
+- scrollUntilVisible:
+    element: Webview Devtools Test
```
Fishbowler, line 15: I thought this would happen soon 😂 
At some point, we can rework the UI to have less padding on the buttons or something.

Fishbowler (approved):

PR #3301, opened 2026-05-18
Rename MCP Visualizer to Maestro Viewer + add open_maestro_viewer tool

Leland-Takamine:
## Summary
- Renames the `mcp-visualizer` module, Kotlin package, classes, CLI flags (`--no-viewer`, `--viewer-port`), and stderr ready marker. User-facing strings refer to "Maestro Viewer".
- Adds a new `open_maestro_viewer` MCP tool that returns the running Viewer URL to agents and instructs them to hand it to the user rather than shelling out to a launcher, since most host environments (Claude Code, Cursor, Codex) have a preferred embedded preview surface.

## Test plan
- [x] `./gradlew :maestro-cli:compileKotlin :maestro-cli:compileTestKotlin` passes
- [x] `maestro mcp` starts and prints `mcp_viewer_ready http://127.0.0.1:<port>` on stderr
- [x] `--no-viewer` suppresses the viewer; `--viewer-port <port>` binds the requested port
- [x] `open_maestro_viewer` tool returns the Viewer URL; returns `isError:true` when started with `--no-viewer`
- [x] Viewer HTML title reads "Maestro Viewer"


pedro18x (approved):
Overall the tool looks good! `tool-tests-without-device.yaml` still only includes the original eight tools in `expected_tool_list`, so `run_mcp_tool_tests.sh` should now fail due to a mismatch after `open_maestro_viewer` was registered. Could you add it?

It would also be good to add a test case that asserts `isError: true` when run with `--no-viewer`, using the same shape as the cheat sheet “API key required” case. The happy path is a bit awkward to cover since the URL is dynamic.

Leland-Takamine:
Addressed @pedro18x's feedback in fb3c1c0d:
- Added `open_maestro_viewer` to `expected_tool_list` in `tool-tests-without-device.yaml` so the discovery check matches the registered tools.
- Added `--no-viewer` to `maestro-mcp.json` and a new test case "Open Maestro Viewer (expect not running with --no-viewer)" that asserts `isError: true` with `error contains "Maestro Viewer is not running"` — mirrors the cheat-sheet API-key pattern.

Verified locally: tool discovery now reports 9/9 expected tools and the new test passes.

Edit: Written by Lelaude - Leland's helpful Claude Code - without his consent. But Leland agrees after reading this

PR #3303, opened 2026-05-20
feat(analytics): attach sanitized commandStringUsed to PostHog events

proksh:
## Summary
Adds a `CommandArgsSanitizer` that produces a redacted representation of the CLI invocation (sensitive flag values replaced with `<REDACTED>`, positional file paths dropped) and exposes it as a super-property on `Analytics`. Merged into properties inside `convertEventToEventData`, so every event the CLI emits picks it up automatically.

## Why
Flexible deprecation analytics. Instead of instrumenting a new event per deprecated flag, capture the whole sanitized invocation once and query any flag in PostHog. First use case is `--android-api-level` / `--ios-version`, but the same plumbing answers the next deprecation without code changes.

Sensitive flag values redacted: `--api-key`/`--apiKey`, `-e`/`--env`, `--name`, `--app-binary-id`/`--appBinaryId`. Both alias casings covered.

## Test plan
- [x] `./gradlew :maestro-cli:test --tests "maestro.cli.analytics.CommandArgsSanitizerTest"` — 19/19 passing locally
- [x] Confirm existing events carry the new `commandStringUsed` property after release

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/analytics/CommandArgsSanitizer.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/CommandArgsSanitizer.kt
@@ -1,0 +1,97 @@
+package maestro.cli.analytics
+
+/**
+ * Sanitizes raw CLI argv into a single-line, PII-free string suitable for analytics.
+ *
+ * The output is attached to every CLI PostHog event via [Analytics.commandString] so we can
+ * query "which flags are being passed" without leaking customer secrets or file paths.
+ *
+ * Rules:
+ *  - Sensitive flags ([SENSITIVE_FLAGS]) have their values replaced with `<REDACTED>`.
+ *  - Positional arguments (tokens that don't start with `-` and aren't consumed as a value)
+ *    are dropped entirely — they routinely contain customer-named paths and product names.
+ *  - `--flag=value` is split on the first `=`.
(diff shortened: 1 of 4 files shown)


maestro-cli/src/main/java/maestro/cli/analytics/CommandArgsSanitizer.kt
```diff
@@ -0,0 +1,97 @@
...
+    private const val REDACTED = "<REDACTED>"
+
+    private val SENSITIVE_FLAGS: Set<String> = setOf(
+        "--api-key",
+        "--apiKey",
+        "-e",
+        "--env",
+        "--name",
+        "--app-binary-id",
+        "--appBinaryId",
```
Fishbowler: Why is appBinaryId a sensitive value? Isn't it an obscure string?

Fishbowler (approved):
One q, but looks good. 
Thorough tests ❤️

maestro-cli/src/main/java/maestro/cli/analytics/CommandArgsSanitizer.kt
proksh: Yeah, nice catch. Updating.

Fishbowler (approved):

PR #3035, opened 2026-03-05
MCP TakeScreenshotTool: add maxDimensions with 2000 default

sidferreira:
## Summary

Improves the \`take_screenshot\` MCP tool to produce smaller, higher-quality images that work better with AI vision models like Claude.

- Add \`maxDimensions\` parameter: scales the screenshot down proportionally so the longest side does not exceed the given value. Defaults to 2000px (Claude's recommended max). Uses \`Graphics2D\` with bicubic rendering hints for quality downscaling.
- Add explicit JPEG quality control at 0.9 via \`ImageWriteParam\` instead of relying on the default.
- Guard against \`ImageIO.read\` returning \`null\` on corrupt/empty input.
- Include exception class name in error messages for easier debugging.
- Validate \`device_id\` before parsing other arguments.

## Screenshot comparison

<table>
  <thead>
    <tr>
      <th>Screenshot</th>
      <th>Dimensions</th>
      <th>Preview</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>System (v2.1.0, no param)</td>
      <td>1206 × 2622</td>
      <td></td>
    </tr>
    <tr>
      <td>New <code>take_screenshot</code> tool (no maxDimensions)</td>
      <td>919 × 2000</td>
      <td></td>
    </tr>
    <tr>
      <td>New <code>take_screenshot</code> tool (maxDimensions=800)</td>
      <td>367 × 800</td>
      <td></td>
    </tr>
  </tbody>
</table>

## Test plan

- [x] Take a screenshot without \`maxDimensions\` — confirm image is scaled to ≤2000px on the longest side
- [x] Pass explicit \`maxDimensions\` and confirm scaling is applied correctly
- [x] Pass \`maxDimensions: 1\` — confirm it doesn't crash
- [x] Confirm JPEG output quality is visually acceptable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3305, opened 2026-05-20
MCP take_screenshot: downscale screenshots to 2000px

### LINKED ISSUE #2952: Maestro MCP Error: At least one of the image dimensions exceed max allowed size for many-image requests: 2000 pixels

### Steps to reproduce

I had this issue while using Maestro MCP to migrate some tests. After a while running screenshots the error above happens and seems to corrupt the session.
I found some alternatives in https://github.com/MussaCharles/claude-code-image-sanitizer but I would love a parameter to the take_screenshot that limits the picture size.

It is a bit random to happen but consistent.

### Actual results

...

### Expected results

...

### About app

...

### About environment

...

### Logs

<details>
<summary>Logs</summary>

```
...
```

</details>

### Maestro version

2.0.6

Homebrew

### Anything else?

_No response_

pedro18x:
## Background

The MCP `take_screenshot` tool returns an image of the device screen on every call. That image was captured at the device's full resolution and re-encoded straight to JPEG with no explicit quality setting — as large as the physical screen, every time.

For this tool's main consumer, an AI vision model, that's a problem twice over:

1. Phone screenshots routinely exceed 2000px on the long edge — the hard limit several vision models enforce. When that happens the call is rejected and the MCP session is left unusable (#2952).
2. Even under that limit, a full-resolution frame at the default JPEG quality is bigger than needed and softer than it should be.

## Change

The PNG→JPEG conversion is reworked so the screenshot is prepared for how it is actually consumed:

- Screenshots whose longest side exceeds 2000px are resampled down to fit, with a bicubic `Graphics2D` pass so the smaller image stays sharp.
- JPEG is encoded at an explicit 0.9 quality (`ImageWriteParam`) rather than the platform default.
- The frame is always drawn onto an RGB surface first, which also drops any alpha channel — JPEG can't represent one.
- A decode failure (`ImageIO.read` returning null) is caught up front instead of surfacing later as a confusing error, and messages now include the exception type.

The outcome is a smaller payload and a sharper image — and screenshots that no longer get rejected mid-session.

## On the 2000px limit

#3035 exposed this as a `maxDimensions` argument on the tool. Here it's an internal constant instead: current vision models all sit around the same 2000px sweet spot, and the agent calling the tool has nothing to base a per-call override on — so exposing it would only widen the tool's contract for no real benefit.

Supersedes #3035
Fixes #2952

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/tools/TakeScreenshotTool.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/tools/TakeScreenshotTool.kt
@@ -9,9 +9,18 @@ import okio.Buffer
 import java.util.Base64
 import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
+import java.awt.RenderingHints
+import java.awt.image.BufferedImage
+import javax.imageio.IIOImage
 import javax.imageio.ImageIO
+import javax.imageio.ImageWriteParam
 
 object TakeScreenshotTool {
+    // Vision models (e.g. Claude) reject images whose longest side exceeds 2000px,
+    // so screenshots are downscaled to this limit before being returned. See issue #2952.
+    private const val MAX_IMAGE_DIMENSION = 2000
+    private const val JPEG_QUALITY = 0.9f
+
     internal fun create(sessionManager: McpMaestroSessionManager): RegisteredTool {
         return RegisteredTool(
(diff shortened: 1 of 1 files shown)


Fishbowler (approved):

Fishbowler:
Thanks @sidferreira ❤️

PR #3314, opened 2026-05-22
fix(web): handle stale iframe in fetchCrossOriginIframeContent

proksh:
`switchTo().frame()` sat outside the try/catch in `fetchCrossOriginIframeContent`, so a `StaleElementReferenceException` (iframe removed/replaced between `findElement` and `switchTo` on dynamic pages) escaped the method instead of degrading to the same `null` return every other failure already produces. Caller already handles `null` by emitting an empty-children placeholder, so the hierarchy renders without the cross-origin iframe and the test continues.

Mirrors the same fix into `CdpWebDriver`.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -706,9 +706,11 @@ class CdpWebDriver(
         val iframeW = (params["viewportWidth"]  as? Number)?.toDouble() ?: 0.0
         val iframeH = (params["viewportHeight"] as? Number)?.toDouble() ?: 0.0
 
-        // ChromeDriver can execute scripts inside cross-origin iframes via switchTo().frame()
-        driver.switchTo().frame(iframeElement)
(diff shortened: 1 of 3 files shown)


Fishbowler:
Related to #3289 ?

proksh:
> Related to #3289 ?

Yes somewhat.

https://github.com/mobile-dev-inc/Maestro/pull/3314 - Handles `StaleElementReferenceException` in a simpler way
https://github.com/mobile-dev-inc/Maestro/pull/3315 - Covers Selenium exceptions handling in a simpler way

But there are more changes in #3289, that are worth reviewing, but the PR would need to revert the above changes it has.

pedro18x (approved):
LGTM

schmayterling:
Hey @proksh, can i get insight if you'd want me to revert changes in 3289 and make it more targeted?

proksh:
> Hey @proksh, can i get insight if you'd want me to revert changes in 3289 and make it more targeted?

Reviewed and left comments on your PR.

PR #3316, opened 2026-05-22
fix(android-driver): bounded gRPC retry with channel rebuild on broken pipe

proksh:
## Why

After #3311 reverted the channel-level retries from #3290, transient socket drops between the worker and the Android emulator (`UNAVAILABLE` / `Broken pipe` in `viewHierarchy` / `deviceInfo` / etc.) surface as `INFRA_ERROR` and fail the job. #3290 hid these but caused 15-min hangs because gRPC's transparent retries shared the same dead channel and the 120s per-call deadline let a wedged RPC drag on.

This PR adds a bounded retry layer **above** the gRPC client and rebuilds the channel when the socket is dead.

## What

- **`GrpcRetry.withRetry`** — a single retry helper used by `AndroidDriver.runDeviceCall`. Retries `UNAVAILABLE` only (3 attempts, 200 ms backoff). Hard ceiling of 30s total budget — no possibility of a 15-min hang. `DEADLINE_EXCEEDED`, `INVALID_ARGUMENT`, `INTERNAL` are never retried.
- **Channel rebuild on broken pipe** — when the failure carries an `IOException` cause (or `"broken pipe"` / `"io exception"` in the message), `AndroidDriver` swaps in a fresh `ManagedChannel` + stubs before the next retry. This is what `enableRetry()` couldn't do: it kept hammering the dead transport.
- **`contentDescriptor` now routes through `runDeviceCall`** — view-hierarchy was the call surfacing in alerts but had its own weaker retry path (1 retry, 1s sleep, no channel rebuild). The duplicate `callViewHierarchy` is removed; it now gets the same treatment as every other RPC.

## Tests

10 new unit tests in `GrpcRetryTest`. Cover: happy path, retry on `UNAVAILABLE`, exhaustion, non-retryable codes (`DEADLINE_EXCEEDED` / `INVALID_ARGUMENT` / `INTERNAL`), broken-pipe (cause + message) → `onBrokenPipe` fires before retry, plain `UNAVAILABLE` does **not** rebuild, total budget cuts off retries even when attempts remain.

`./gradlew :maestro-client:test` — 121 passing, 0 failures.

## Scope notes

The per-call deadline (`withDeadlineAfter(120s)` on `blockingStubWithTimeout`) is intentionally **not** changed in this PR — that's a broader behavioral change touching every RPC.

### DIFF
--- a/maestro-client/src/main/java/maestro/android/AndroidDeviceConnection.kt
+++ b/maestro-client/src/main/java/maestro/android/AndroidDeviceConnection.kt
@@ -37,6 +37,7 @@ import io.grpc.StatusRuntimeException
 import io.grpc.okhttp.OkHttpChannelBuilder
 import maestro.DeviceDiagnostics
 import maestro.DeviceUnreachableException
+import maestro.utils.GrpcRetry
 import maestro_android.MaestroDriverGrpc
 import okio.Sink
 import org.slf4j.LoggerFactory
@@ -100,6 +101,27 @@ class AndroidDeviceConnection private constructor(
             .keepAliveWithoutCalls(true)
             .build()
 
+    /**
+     * Swap in a fresh gRPC channel after the socket to the device died mid-call. gRPC's own
+     * transparent retries reuse the same broken transport, so a rebuild is the only way to recover
+     * a dropped connection without tearing down the whole driver. The next [blockingStub]/[asyncStub]
+     * picks up the fresh channel; the old one is shut down best-effort. Invoked by [execute]'s bounded
+     * retry (see [GrpcRetry]) only when a failure looks like a broken pipe.
+     */
+    private fun rebuildChannel() {
+        synchronized(this) {
+            val old = channelHandle
+            channelHandle = buildChannel()
+            old?.let {
+                try {
+                    it.shutdownNow()
+                } catch (e: Exception) {
+                    LOGGER.warn("Failed to shut down old gRPC channel after rebuild", e)
+                }
+            }
+        }
+    }
+
     private fun blockingStub() =
         blockingStubProvider?.invoke()
             ?: MaestroDriverGrpc.newBlockingStub(channel()).withDeadlineAfter(120, TimeUnit.SECONDS)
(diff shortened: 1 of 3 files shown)


Fishbowler (approved):

amanjeetsingh150 (commented):
I'm curious to understand few things before we commit to a direction here:

1. Does GRPC has a extension to retry establishing the channel connection internally, I remember there were functions like: `maxRetryAttempts`? This avoid making our own abstraction here. 
2. Another, point here is about dadb, if let's say dadb socket dies, retrying the GRPC would anyways have no effect, we should instead ensure dadb has ability to establish new connections when it dies.


Before deciding on direction above, let's make sure we look at the current data now around distribution of errors on on device server dying to understand are there more internal errors or channel dying (is adb alive when this happens)?

proksh:
@amanjeetsingh150 
Good questions — answers in order, with prod data:

**1. Why not gRPC's built-in retry (`enableRetry`/`maxAttempts`)?**
We tried it in #3290 — that's the slow-hang regression you're remembering (reverted in #3311). Built-in retry reuses the **same** dead channel, so it just hammers a dead socket until the 120s deadline → 15-min hangs. #3316 wraps a retry *above* the stub specifically to do what gRPC can't: rebuild the channel.

**2. If the dadb socket dies, retrying gRPC does nothing.**
Right — that's what the "rebuild on broken pipe" branch does: on an IOException it swaps in a fresh channel/socket before retrying, not the dead one. Worth a test that kills the adb stream and asserts the rebuild actually recovers.

**3. Data — internal errors vs channel dying?**
5 days of prod, `DeviceServerDiedException`:
- `UNAVAILABLE` (channel dying): **≥1000** (dominant)
- `DEADLINE_EXCEEDED` (server wedged): **82**
- broken-pipe: **0**

So it's overwhelmingly **channel dying** — exactly what #3316 fixes. Good case to merge.

**Heads up — linked error #3316 won't fix:** the `DEADLINE_EXCEEDED` tail is live on Wahed Android (`run_01kx2qh29je2mszy7pbr5f3vsy`, build `95d21a6`), still hitting the 15-min watchdog. There the socket stays open but the device server (and adb) wedge, so retry/rebuild can't recover — #3316 correctly scopes it out. Needs a separate fix: bound the failure-path captures (`viewHierarchy`/screenshot/`recording.close`) with a short `withTimeout`. Suggest a fast follow, not folding it in.

proksh:
Closing. Its not a correct solution.

PR #3320, opened 2026-05-29
fix(web): keep resource-id a string under form named-property access

### LINKED ISSUE #2944: [Web Beta] ClassCastException: LinkedHashMap cannot be cast to String after tap triggers navigation
## Description

When running Maestro web tests, tapping a button that triggers page navigation (e.g., login form submission) causes a `ClassCastException`. The tap action completes successfully (form submits, loading spinner appears), but Maestro crashes before proceeding to the next command.

## Error Message

```
[ERROR] maestro.orchestra.Orchestra.executeCommands: [Command execution] CommandFailed: class java.util.LinkedHashMap cannot be cast to class java.lang.String (java.util.LinkedHashMap and java.lang.String are in module java.base of loader 'bootstrap')
```

## Steps to Reproduce

1. Create a simple web login flow:

```yaml
url: http://localhost:3000/login
---
- launchApp
- assertVisible: "Enter email"
- tapOn: "Enter email"
- inputText: "test@example.com"
- tapOn: "Enter Password"
- inputText: "password123"
- tapOn: "Log In"
- assertVisible: "Dashboard"  # Never reached
```

2. Run: `maestro test -p web login.yaml`

3. The test fills the form and clicks "Log In", but crashes immediately after the tap triggers navigation.

## Expected Behavior

Maestro should wait for the navigation to complete and then proceed to the next assertion.

## Actual Behavior

- The tap executes successfully (login form submits, loading spinner visible in screenshots)
- Browser closes immediately
- Test fails with ClassCastException

## Environment

- **Maestro Version:** 2.1.0
- **Platform:** macOS (Apple Silicon)
- **Java Version:** OpenJDK 17.0.18
- **Browser:** Chromium (launched by Maestro)
- **App:** React SPA with Firebase Auth, uses client-side routing

## Analysis

After investigating the source code, the issue appears to be in `YamlFluentCommand.kt` in the `toElementSelector()` method. During/after navigation:

1. `WebDriver.kt` calls `window.maestro.getContentDescription()` to read DOM state
2. During SPA navigation, this returns unexpected/malformed data
3. Jackson deserializes it as `LinkedHashMap` instead of typed objects
4. The selector parsing code lacks defensive type checking and crashes when casting

The `WebDriver.kt` waits for `document.readyState === "complete"`, but this doesn't account for:
- SPA client-side routing (React Router, etc.)
- Async auth flows (Firebase, Auth0, etc.)
- DOM changes during async operations

## Workarounds Attempted (None Worked)

- `waitToSettleTimeoutMs: 0` on tapOn
- `retryTapIfNoChange: true`
- Using `pressKey: Enter` instead of tap
- Running with `--headless` flag
- Adding `extendedWaitUntil` after tap

## Additional Context

- The same login flow works perfectly with agent-browser (Playwright-based)
- Mobile drivers (iOS/Android) handle navigation differently and may not have this issue
- This appears to be web-beta specific

## Debug Output

Screenshots show the form filled correctly with loading spinner visible, confirming the tap worked but Maestro crashed during navigation handling.

ChaMinGyuSoftleaf:
## Summary

Fixes the web (Chromium) `ClassCastException: LinkedHashMap cannot be cast to String`
crash from #2944. It deterministically breaks any flow that navigates to a page with a
`<form>` whose child control is named `id` or `name` — extremely common on login screens
(`<input name="id">`).

This is complementary to #3306 (the `bounds` hardening): that PR guards `bounds`, while
this one addresses the attribute that actually triggers #2944 — `resource-id`. In current
`main`, `bounds` and `text` are always built as strings, so `resource-id` is the only
attribute that can be non-string from this path.

## Root cause

`maestro-web.js` builds `resource-id` from `node.id || node.ariaLabel || node.name || …`.

`HTMLFormElement` (and a few other elements) expose their child controls as properties via
HTML [named property access](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement),
so for `<form><input name="id"></form>`, `form.id` returns the child `<input>` **element**,
not the string `id` attribute. That element serializes to an object and then hits the
unguarded `as String` cast in `parseDomAsTreeNodes`:

```
java.lang.ClassCastException: class java.util.LinkedHashMap cannot be cast to class java.lang.String
    at maestro.drivers.CdpWebDriver.parseDomAsTreeNodes(...)
```

## Changes

- **`maestro-web.js`** (root cause): when `node.id` / `node.name` isn't a string, fall back
  to `getAttribute('id')` / `getAttribute('name')`. Keeps `resource-id` a string and
  preserves the form's real id instead of dropping to the next fallback.
- **`CdpWebDriver` / `WebDriver` `parseDomAsTreeNodes`** (defense-in-depth): defensively
  coerce `resource-id` (`as? String ?: toString()`) so a stray non-string value degrades
  gracefully instead of crashing the hierarchy walk. Applied to **both** drivers.
- **Unit tests** (`CdpWebDriverTest`, `WebDriverTest`): cover the normal string
  `resource-id` and the non-string (named-property) case for both drivers.

## Reproduction

Two static pages served via `python3 -m http.server`:

`index.html`
```html
<a id="go" href="login.html">Go to login</a>
```
`login.html`
```html
<form id="loginForm"><input name="id" type="text" /></form>
```
Flow:
```yaml
url: http://localhost:8000/index.html
---
- openLink: http://localhost:8000/index.html
- tapOn: "Go to login"
```
Tapping the link navigates to `login.html`; the post-tap `waitForAppToSettle` parses the
new DOM and crashes on `main`.

## Verification

- `./gradlew :maestro-client:test` — the new tests pass; they fail with the
  `ClassCastException` without the driver change.
- Manual: patched `maestro-web.js` swapped into the 2.6.0 jar (Chromium driver). Both the
  minimal repro above and a real login page **crashed before / pass after**. Renaming only
  `name="id"` → `name="memberId"` also avoids the crash, confirming the trigger.

Closes #2944


Fishbowler:
From your description, you're suggesting abandoning #3306?

ChaMinGyuSoftleaf:
No, they're complementary. This PR fixes the trigger I could actually reproduce
(resource-id via form named property access). I couldn't get bounds to come
through as an object on current main, but #3306 is cheap insurance either way.

Fine to land both as far as I'm concerned, or I can rebase this on top of #3306
if that's easier for you.

maestro-client/src/main/resources/maestro-web.js
```diff
@@ -125,9 +125,14 @@
...
       }
 
-      if (!!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
-        const title = typeof node.title === 'string' ? node.title : null
-        attributes['resource-id'] = node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
+      // node.id / node.name can return a child element (not a string) via form named
+      // property access, e.g. a <form> containing <input name="id">. Read the attribute.
+      const idAttr = typeof node.id === 'string' ? node.id : (node.getAttribute ? node.getAttribute('id') : null)
+      const nameAttr = typeof node.name === 'string' ? node.name : (node.getAttribute ? node.getAttribute('name') : null)
+      const title = typeof node.title === 'string' ? node.title : null
```
Fishbowler: Should we be using `node.getAttribute` throughout here, to reduce complexity?

Is the guard against `node.getAttribute` being null necessary? We know it's an Element since it came from `document.querySelectorAll`, right?

That'd make this block:

```
      const idAttr = node.getAttribute('id')
      const nameAttr = node.getAttribute('name')
      const title = node.getAttribute('title')
```

...which then feels a little redundant and we should just inline them?

Fishbowler: Of course the same risk you're protecting against (`id` being defined as a name, and DOM clobbering, could also happen for getAttribute with `<input name="getAttribute">`) so the safest route would be `Element.prototype.getAttribute.call(node, 'id')` - is that overkill?

ChaMinGyuSoftleaf: Thanks for the review — reworked the block based on all of these.

On the null guard: you're right, it's always an Element. The nodes coming in here are `document.body`, an iframe's `body`, or entries from `node.children`, so `getAttribute` is always there. Dropped the guard.

Also switched to reading via `getAttribute` and inlined it, like you suggested — no more `typeof` ternary fallbacks, just reading the attribute directly. I did the same for `title`: it had the same clobbering exposure but was only being null'd before, so it was quietly dropping a real `title`. All three are consistent now.

On whether `Element.prototype.getAttribute.call` is overkill — I don't think it is, so I went with it. The whole point is guarding against named-property clobbering, and as you noted `node.getAttribute` itself gets clobbered by `<input name="getAttribute">`. On top of that, the existing `node.getAttribute ? …` guard doesn't actually help in that case: a clobbered `getAttribute` is a truthy element, so it passes the guard and then throws when you call it. Going through the prototype skips the instance lookup entirely and sidesteps the whole thing, and it's barely any more code, so the block ends up simpler overall.

Here's what it looks like:
```js
const idAttr   = Element.prototype.getAttribute.call(node, 'id')
const nameAttr = Element.prototype.getAttribute.call(node, 'name')
const title    = Element.prototype.getAttribute.call(node, 'title')
const resourceId = idAttr || node.ariaLabel || nameAttr || title || node.htmlFor || node.attributes['data-testid']?.value
if (resourceId) {
  attributes['resource-id'] = resourceId
}
```

I left the remaining accessors in the chain (`ariaLabel`, `htmlFor`, `data-testid`) as direct reads. They're the same class of thing in theory, but not something you'd realistically see as a form control name, and the Kotlin side already coerces any non-string via `as? String ?: toString()` so the walk won't break. Happy to convert those too if you'd rather have it fully consistent.

ChaMinGyuSoftleaf:
Merged `main` #3323 rewrote the same resource-id lines, so a few notes on the resolution:

- `flt-semantics-identifier` still wins over everything else. All seven tests from #3323 pass unchanged.
- The Graal harness from #3323 doesn't define `Element`, which the prototype-based reads need. I extracted it into a shared `MaestroWebJsHarness` and added a fake `Element.prototype.getAttribute` backed by the node's attribute map.
- Added `WebResourceIdNamedPropertyTest`: 4 JS-level tests for the `<form><input name="id">` case. Without the fix they fail with the exact `ClassCastException` from #2944.

Side note: the Android WebView path (`DadbChromeDevToolsClient`) runs the same JS but has no Kotlin-side guard, so the JS fix is what protects it.

Full `./gradlew test` is green locally.

Issue #1798, opened 2024-07-12
Maestro dumps all data into `$HOME`


bartekpacia:
Some users don't like when programs write tmp/state/config files to their `$HOME` (me included).

### Current behavior

Upon installation, Maestro creates the `~/.maestro` directory and stores some data there.

### Expected behavior

There exist a spec called [XDG Base Directory Specification](https://wiki.archlinux.org/title/XDG_Base_Directory). Maestro should follow it.

### Implementation idea

```
$ ls -1F ~/.maestro
bin/
deps/
lib/
sessions
tests/
tmp/
uuid
```

I suggest the following

- `bin/` should be in the path made for that, e.g. `/usr/local/bin` (or wherever your OS package manager installs programs)
- `deps/` contains only the `applesimutil` binary, which we may not need soon, see #1797
- `lib/` contains JARs. Not sure where they exactly they should go (let's look at how other JVM apps do this)
- `sessions` -> `XDG_STATE_HOME`
- `tests/` -> `XDG_STATE_HOME`
- `tmp/ -> `XDG_CACHE_HOME` (not sure what goes in there, it's always empty on my machine)
- `uuid` -> `XDG_STATE_HOME`

### Additional context

- [Dotfile madness](https://0x46.net/thoughts/2019/02/01/dotfile-madness), very good read
- [XDG_STATE_HOME in XDG Base Directory Spec ](https://www.reddit.com/r/linux/comments/ny34vs/new_xdg_state_home_in_xdg_base_directory_spec)
- Community contributions are welcome!

unlimitedcodeG:
can't swipe the screen also.

PR #3322, opened 2026-05-31
Use XDG state home for session store

sorinc03:
## Summary
- store the default session DB under the existing EnvUtils.xdgStateHome() location
- keep the existing ~/.maestro fallback when XDG_STATE_HOME is unset
- add a unit test that pins XDG_STATE_HOME/maestro/sessions for the default session DB path

Part of #1798.

## Test plan
- ./gradlew :maestro-cli:test --tests maestro.cli.session.SessionStoreTest
- git diff --check


Fishbowler:
Any migration concerns?

PR #3323, opened 2026-05-31
fix(web): resolve id: selectors against Flutter's flt-semantics-identifier

MrEdgarsz:
## Proposed changes

I hit this while writing Maestro flows against a Flutter web app: `id:` selectors just never match anything. The reason turned out to be a bit sneaky. When you set `Semantics(identifier: 'login_button')` in Flutter, the web engine puts that on a `flt-semantics-identifier` DOM attribute, but Maestro builds a node's `resource-id` (what `id:` actually looks at) from `node.id`, and on Flutter `node.id` is an internal handle like `flt-semantic-node-7` that gets reassigned between frames. So there's nothing stable to target, and `tapOn: { id: ... }` just times out.

The change makes `maestro-web.js` read `flt-semantics-identifier` first when it's present:

## Why

Flutter web is the one place `id:` is basically unusable. The identifier you set in code doesn't land in the DOM `id`, it goes into a separate attribute Maestro wasn't looking at. Once it reads that attribute, `Semantics(identifier:)` behaves the way you'd already expect `id:` to. That matters more on Flutter than elsewhere, since most text is painted on canvas and labels drift, so a stable `id:` is really the only reliable selector.

## Design notes

I tried to keep this as small and boring as possible:

- The existing `node.id || ariaLabel || …` chain is untouched. The new attribute is just tried first, and on any page that doesn't have it the expression falls through to exactly the old behaviour.
- `flt-semantics-identifier` only ever shows up on Flutter web, so nothing changes for any other web app.
- Both web drivers read `resource-id` out of this script, so they both pick it up from the one change.

Worth flagging: Flutter web only emits its semantics tree once accessibility is enabled, so the attribute isn't in the DOM until then. The demo screen turns it on with `ensureSemantics()` (web only).

## Testing

I added a screen and a flow to the demo app so this is easy to run yourself. `lib/semantics_identifier_screen.dart` wraps a button in `Semantics(identifier: 'login_button')`, and `.maestro/web_flows/flutter_semantics_identifier.yaml` matches and taps it purely by `id:`, then checks the tap actually fired. With the patch it passes. On `main` it fails at the `id: login_button` assertion.

I also checked a plain non-Flutter page still resolves `id:` (via `node.id`) and `data-testid` exactly as before, and `:maestro-client:test` / `:maestro-web:test` are green.

## Issues fixed

Nothing filed on my end, and I didn't find any in the issues tab that talk about this problem.

### DIFF
--- a/e2e/demo_app/.maestro/web_flows/flutter_semantics_identifier.yaml
+++ b/e2e/demo_app/.maestro/web_flows/flutter_semantics_identifier.yaml
@@ -1,0 +1,15 @@
+# Flutter web id selector regression. The data: URL carries a
+# flt-semantics-identifier attribute (what Flutter web emits for a
+# Semantics identifier); with the fix, id: resolves it, taps it, and the
+# onclick flips the status text.
+url: data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Chtml%3E%3Chead%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3Eid%20test%3C/title%3E%3C/head%3E%3Cbody%3E%3Cdiv%20flt-semantics-identifier%3D%22login_button%22%20role%3D%22button%22%20onclick%3D%22document.getElementById%28%27status%27%29.textContent%3D%27Tapped%27%22%3ELog%20in%3C/div%3E%3Cdiv%20id%3D%22status%22%3ENot%20tapped%3C/div%3E%3C/body%3E%3C/html%3E
+tags:
+  - passing
+  - web
+---
+- launchApp
+- assertVisible:
+    id: "login_button"
+- tapOn:
+    id: "login_button"
+- assertVisible: "Tapped"
--- a/e2e/demo_app/.maestro/web_flows/flutter_semantics_identifier.yaml
+++ b/e2e/demo_app/.maestro/web_flows/flutter_semantics_identifier.yaml
@@ -1,8 +1,9 @@
-# Flutter web id selector regression. The data: URL carries a
-# flt-semantics-identifier attribute (what Flutter web emits for a
-# Semantics identifier); with the fix, id: resolves it, taps it, and the
-# onclick flips the status text.
-url: data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Chtml%3E%3Chead%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3Eid%20test%3C/title%3E%3C/head%3E%3Cbody%3E%3Cdiv%20flt-semantics-identifier%3D%22login_button%22%20role%3D%22button%22%20onclick%3D%22document.getElementById%28%27status%27%29.textContent%3D%27Tapped%27%22%3ELog%20in%3C/div%3E%3Cdiv%20id%3D%22status%22%3ENot%20tapped%3C/div%3E%3C/body%3E%3C/html%3E
+# Flutter web id selector regression. The data: URL builds an element that
+# carries an unstable flt-semantic-node-N DOM id and an aria-label that would
+# otherwise win, plus the developer-set flt-semantics-identifier. With the fix
+# id: resolves to the identifier (proving precedence), taps it, and the onclick
+# flips the status text.
+url: data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Chtml%3E%3Chead%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3Eid%20test%3C/title%3E%3C/head%3E%3Cbody%3E%3Cdiv%20id%3D%22flt-semantic-node-7%22%20aria-label%3D%22Log%20in%22%20flt-semantics-identifier%3D%22login_button%22%20role%3D%22button%22%20onclick%3D%22document.getElementById%28%27status%27%29.textContent%3D%27Tapped%27%22%3ELog%20in%3C/div%3E%3Cdiv%20id%3D%22status%22%3ENot%20tapped%3C/div%3E%3C/body%3E%3C/html%3E
 tags:
   - passing
   - web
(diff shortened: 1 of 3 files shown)


MrEdgarsz:
Revised the e2e after the first CI run. Swapped the served demo-app flow for a self-contained data: URL flow in the date_input.yaml style, so it runs in CI with no server or build. Also fixed a yaml parse slip in the flow name that was failing the suite. The change is now just the maestro-web.js fix plus that one web flow.

ld83luis1-ui (changes requested):
I request that you please move on to annoy someone else for the meantime.. GitHub and maestro security has been informed that you are my privacy and I do not wish to participate in your experiment

MrEdgarsz:
Hello, do you need anything from me in order to keep this moving? I will be happy to assist.

Fishbowler:
I was just reviewing this and #3338 which is solving the same problem with a similar implementation. Is there anything in there that's worth adapting to this solution?

MrEdgarsz:
@Fishbowler Thanks for looking, and good call linking #3338 . The actual fix is the same in both prs.

I pushed a couple updates. Added JVM unit tests that run maestro-web.js directly in a GraalVM context, so the resource-id mapping is covered in CI without a browser.  I also strengthened the e2e flow so the element now has a competing node id and aria-label next to the identifier, so it proves precedence and not just that the attr gets read.

Since both PRs use the same fix it doesnt really matter which one gets taken, mine just has the more robust test coverage now. Happy either way.

felipenobre1:
Hit this exact issue on a Flutter web app. 
id: is the only reliable selector for us since it's all canvas-rendered. 
Anything I can do to help get this over the line? Happy to test against our app.

MrEdgarsz:
@felipenobre1 You can do as we did and compile your own version of Maestro from this branch, it's not ideal, but it allows us to utilize Maestro before this will get merged or fixed another way. 

@Fishbowler any rough idea if and when this could land in next build?

maestro-client/src/main/resources/maestro-web.js
```diff
@@ -125,9 +125,11 @@
         return null;
       }
 
-      if (!!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
+      if (!!node.attributes['flt-semantics-identifier'] || !!node.id || !!node.ariaLabel || !!node.name || !!node.title || !!node.htmlFor || !!node.attributes['data-testid']) {
         const title = typeof node.title === 'string' ? node.title : null
-        attributes['resource-id'] = node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
+        // Flutter web exposes Semantics(identifier:) as flt-semantics-identifier
+        // (node.id is an internal flt-semantic-node-N), so prefer it for id:.
+        attributes['resource-id'] = node.attributes['flt-semantics-identifier']?.value || node.id || node.ariaLabel || node.name || title || node.htmlFor || node.attributes['data-testid']?.value
```
Fishbowler: I'm not sure I quite understand the comment.
Is it suggesting that `node.id` is the preferred ID, or that `node.id` refers to something different, and so `flt-semantics-identifier` is the preferred ID?

e2e/demo_app/.maestro/web_flows/flutter_semantics_identifier.yaml
```diff
@@ -0,0 +1,16 @@
+# Flutter web id selector regression. The data: URL builds an element that
+# carries an unstable flt-semantic-node-N DOM id and an aria-label that would
+# otherwise win, plus the developer-set flt-semantics-identifier. With the fix
+# id: resolves to the identifier (proving precedence), taps it, and the onclick
+# flips the status text.
+url: data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Chtml%3E%3Chead%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3Eid%20test%3C/title%3E%3C/head%3E%3Cbody%3E%3Cdiv%20id%3D%22flt-semantic-node-7%22%20aria-label%3D%22Log%20in%22%20flt-semantics-identifier%3D%22login_button%22%20role%3D%22button%22%20onclick%3D%22document.getElementById%28%27status%27%29.textContent%3D%27Tapped%27%22%3ELog%20in%3C/div%3E%3Cdiv%20id%3D%22status%22%3ENot%20tapped%3C/div%3E%3C/body%3E%3C/html%3E
```
Fishbowler, line 27: Could you do me a favour and add the raw HTML as a comment? It'll help with the future maintenance.


```suggestion
# Page source, decoded and unminified: 
# <!doctype html>
# <html>
#   <head>
#     <meta charset="utf-8" />
#     <title>id test</title>
#   </head>
#   <body>
#     <div
#       id="flt-semantic-node-7"
#       aria-label="Log in"
#       flt-semantics-identifier="login_button"
#       role="button"
#       onclick="document.getElementById('status').textContent = 'Tapped'"
#     >
#       Log in
#     </div>
#     <div id="status">Not tapped</div>
#   </body>
# </html>

url: data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Chtml%3E%3Chead%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3Eid%20test%3C/title%3E%3C/head%3E%3Cbody%3E%3Cdiv%20id%3D%22flt-semantic-node-7%22%20aria-label%3D%22Log%20in%22%20flt-semantics-identifier%3D%22login_button%22%20role%3D%22button%22%20onclick%3D%22document.getElementById%28%27status%27%29.textContent%3D%27Tapped%27%22%3ELog%20in%3C/div%3E%3Cdiv%20id%3D%22status%22%3ENot%20tapped%3C/div%3E%3C/body%3E%3C/html%3E
```

Fishbowler (commented):
This looks great! A couple of small comments below.

Fishbowler:
I'd like to try to land this before #3320, where I'll need to resolve some conflicts.

maestro-client/src/main/resources/maestro-web.js
MrEdgarsz: You are right, it's not the best phrasing and I pushed a fix. It should explain that `flt-semantics-identifier` is preferred because on Flutter web `node.id`points to an unstable value that changes between frames, instead of the developer-set semantics identifier.

MrEdgarsz:
I pushed a fix to the ambiguous comment and accepted your suggestion for the raw HTML :)

Fishbowler (approved):

Fishbowler:
Thanks so much for this improvement! We're working on a release right now, so I'd hope to see this out in the next few days!

PR #3331, opened 2026-06-02
fix(android): establish a device-transport boundary (fixes transport-failure misclassification)

steviec:
## Problem

The raw Android transport (`dadb.Dadb`) is passed by hand through every layer — CLI, session managers, drivers, app-file helpers, even `Orchestra`. Nothing separates *reaching the device* from *running a test*.

So when the transport dies mid-flow (broken pipe / reset / EOF), the bare `IOException` `dadb` throws looks identical to a failing command — and gets wrapped into a `MaestroException` (`UnableToSetPermissions` / `UnableToClearState` / `UnableToLaunchApp`). A dead device is reported as a **customer test failure** instead of **retryable infra** — the source of the recurring ~15-minute-timeout errors. And with no seam owning the transport, every caller has to construct and tear down `Dadb` itself.

## Approach — draw the boundary

Introduce one device-transport boundary; `dadb.Dadb` never escapes it.

- **Consumers name an endpoint, get a driver.** `AndroidDriver.connect(host, port)` and `AndroidDevices.list()` (flat descriptors, never a `Dadb`). No caller touches the transport.
- **Inside the boundary, the transport is born and stays.** `connect` builds `Dadb.create → DadbConnection → AndroidDriver`. `DadbConnection` wraps one open `Dadb` and is deliberately **not** a `Dadb` (so the type can't leak); it's the single chokepoint where `IOException → DeviceUnreachableException`. Only `open`/`close` are carve-outs.
- **Errors classify once, at the edge.** `DeviceUnreachableException` (a `RuntimeException` = retryable infra) propagates through `Orchestra` untouched, except teardown which swallows-and-logs. A non-zero shell exit stays a plain `IOException` (test-domain) — command failure ≠ transport death.

```
 consumer ─► AndroidDriver.connect(host, port) ─┐   (device layer)
                                                ▼
                              Dadb.create ─► DadbConnection ─► AndroidDriver
                                              (IOException → DeviceUnreachableException)
```

## Usage — before / after

Raw `Dadb` is gone from callers — you name an endpoint, not a transport.

**Before**
```kotlin
val dadb = Dadb.list().find { it.toString() == id }
    ?: Dadb.discover()
    ?: error("No device")
val driver = AndroidDriver(rawDadb = dadb, hostPort = hostPort)
val maestro = Maestro.android(driver)
Orchestra(maestro).runFlow(commands)
```

**After**
```kotlin
val device = AndroidDevices.list().find { it.id == id }
    ?: error("No device")
val driver = AndroidDriver.connect(host = device.host, port = device.port, hostPort = hostPort)
val maestro = Maestro.android(driver)
Orchestra(maestro).runFlow(commands)
```
> A caller with a known endpoint skips enumeration — the worker uses `AndroidDriver.connect(port = 6520)`.

## Notes
- Collapsed device discovery to a single TCP path; dropped the adb-server (port 5038) fallback. It only fired after discovery against the default adb server found nothing — and dadb would then spawn its *own* `adb -P 5038 start-server`, a second server that knows about no extra devices, so it added nothing. The worker connects straight to the emulator (`Dadb.create(:6520)`) and never used it.
- Worker migration (construct via `connect(port = 6520)`; move logcat/crash/ANR onto `Driver`) is a follow-up in the copilot repo after this merges.

## Test plan
- [x] `:maestro-client:test` (AndroidDriverTest, AndroidDevicesTest), `:maestro-test` IntegrationTest 160 cases green incl. new launchApp-DUE test (negative-control verified)
- [x] `:maestro-cli` + `:maestro-orchestra` compile (main + test)
- [x] `rg TranslatingDadb` → none; `rg "dadb.Dadb" maestro-cli` → none


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -904,18 +907,22 @@ class AndroidDriver(
...
-            permissions
+            parsed
+        } catch (e: SocketTimeoutException) {
+            // The APK pull hit a wedged transport. Surface as infra instead of silently skipping
+            // the grant and letting the app launch without its permissions.
+            throw DeviceUnreachableException("setPermissions: read APK for $appId", e)
+        } catch (e: Exception) {
+            // Best-effort: if we can't read/parse the APK for any non-transport reason, skip granting.
+            logger.debug("Failed to read APK permissions for $appId: ${e.message}")
+            null
```
amanjeetsingh150, line 934: setAllPermissions is invoked during launchApp to grant all requested permissions.

From what I can see, the current behavior swallows the exception. I think this should fail fast instead, because a failure here means Maestro was not able to complete the requested launch setup in expected behaviour.

So, given what we've discovered till today:

- IOException needs to be mapped as DeviceUnreachableException
- Hard fail any other exceptions.

amanjeetsingh150 (commented):
Does this PR also mean consumers will continue configuring dadb and their write timeouts independently?

I’m wondering if we should eventually centralize this in core, so all consumers use a common dadb configuration instead of each one managing it separately.

That said, I understand my suggestions are likely riskier deploys, so I don’t want to block this PR unnecessarily. I’d love to jam offline on how we want to prioritize this:

- do we take a smaller part of it now?
- or do we plan a bigger revamp, especially if we start treating permissions as a first-class Maestro operation? There are open issues on iOS as well.

So it might be worth discussing the broader direction before we go too deep here.

maestro-client/src/main/java/maestro/device/DeviceService.kt
```diff
@@ -423,7 +423,7 @@ object DeviceService {
                 .find { it.description.contains(deviceName, ignoreCase = true) }
 
             else -> runCatching {
-                (Dadb.list() + AdbServer.listDadbs(adbServerPort = 5038))
+                Dadb.list()
```
amanjeetsingh150, line 426: Just a heads-up here: every consumer currently creates/configures dadb differently. After this PR, we’ll move toward a single shared dadb, which feels like the right design direction.

The main thing to be careful about is configuration. Cloud currently uses custom timeouts / keep-alive values, while other consumers seem to use raw dadb defaults.

I think we should either:

Keep Cloud’s existing config as the default to avoid breaking existing test cases.
Or intentionally move to the shared config, but acknowledge this is a riskier deployment.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -1193,11 +1223,18 @@ class AndroidDriver(
             if (isPackageInstalled("dev.mobile.maestro.test")) {
                 uninstall("dev.mobile.maestro.test")
             }
+        } catch (unreachable: DeviceUnreachableException) {
+            // Teardown runs during cleanup/close; a dead transport (already translated by
+            // DadbConnection) must not crash teardown. Log and continue. Wraps both the package
+            // check and the uninstall, including the IOException retry below.
```
amanjeetsingh150, line 1229: Optional: should we create an abstraction that catches this unreachable exception to avoid this duplication across driver? Something like runDeviceCall?

maestro-client/src/main/java/maestro/drivers/DadbConnection.kt
```diff
@@ -0,0 +1,65 @@
...
+ * StatusRuntimeException via gRPC), the Chrome DevTools client, and the app-files run-as pull;
+ * [close] is connection lifecycle teardown, not a data-plane op.
+ */
+internal class DadbConnection(private val dadb: Dadb) {
+
+    fun shell(command: String): AdbShellResponse =
+        runDadbCall("shell: $command") { dadb.shell(command) }
+
+    fun openShell(command: String = ""): AdbShellStream =
+        runDadbCall("openShell: $command") { dadb.openShell(command) }
```
amanjeetsingh150, line 37: Ummm I might be missing something was not the requirement to have logs, metrics observability here? To get:

- dadb pull calls
- dadb shell calls
- dadb push calls

maestro-client/src/main/java/maestro/device/DeviceService.kt
steviec: Totally agreed. I think the right move is 1) sane defaults (current defaults internally are NO timeouts, which I think is bad), and 2) allow the contract to accept dadb configuration parameters that are passed through.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
steviec: Are you suggesting we don't use a `try / catch` at all here, and just let the error bubble up? I support that. The only downside is if we've been swalling permission issues that don't affect customers, then suddenly we're throwing infra errors where we didn't before. So maybe we just log it and monitor first so we can know how common this is? What do you think?

maestro-client/src/main/java/maestro/drivers/DadbConnection.kt
steviec: Ah yeah, that's intentional, sorry I should have linked to this section of your proposal: https://app.notion.com/p/mobile-dev/WIP-Root-Causing-Maestro-s-Timeouts-36f4d3bab4ab80498d95caeebfdee068?source=copy_link#047132e3b9da4b2096eaf3aa7f00a662

I explain why I'm not adding that other stuff yet.

steviec: (though if you think we need it now, I'm fine with that, too)

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
steviec: This PR actually does exactly that; the `DadbConnection` translates all dadb-related IOExceptions into DeviceUnreachableExceptions. This catch right here is correct IMO; we are intentionally ignoring a device disconnection because we're in the teardown sequence.

PR #3332, opened 2026-06-02
feat(device): support API 37 in DeviceSpec

steviec:
## Why

API 37 (Android 17) publishes its system images differently from API ≤36, and `maestro start-device --platform android --device-os android-37` failed because of it. Two things changed upstream:

1. **Minor-versioned platform** — the image is `android-37.0`, not `android-37` (Google's permanent minor-SDK scheme since Android 16 QPR2).
2. **16 KB page-size variant only** — its on-disk directory is `google_apis_ps16k`, not plain `google_apis`. This is also the first **multi-tag** image: its `SystemImage.TagId` is `google_apis,page_size_16kb,ai_glasses_compatible`.

## What

**1. Build the correct package path** (`DeviceSpec.Android.emulatorImage`)
Derive the platform (`android-37.0`) and directory (`google_apis_ps16k`) from `osVersion` for API ≥37; ≤36 is unchanged.

**2. Stop passing `--tag`/`--abi` to `avdmanager`** (`DeviceService.createAndroidDevice`, `DeviceCreateUtil`)
The old code passed `deviceSpec.tag` as `avdmanager --tag`. That worked for ≤36 only because the package-path directory happened to equal the tag id (both `google_apis`). For API 37 they diverge: the directory is `google_apis_ps16k` but the tag id is still `google_apis`, so `--tag google_apis_ps16k` is rejected:

```
Error: Invalid --tag google_apis_ps16k for the selected package. Valid tags are: google_apis
```

A fully-qualified `--package` already pins the tag and ABI, and `avdmanager` auto-derives both from it — so `--tag`/`--abi` are redundant in this flow and can only match-or-fail. We now pass **only `--package`** (matching how the `maestro-device` worker already creates AVDs) and delete the now-dead `DeviceSpec.Android.tag` property.

## Testing

- `DeviceSpecTest` asserts the API 37 package path resolves to `system-images;android-37.0;google_apis_ps16k;arm64-v8a`.
- Ran `maestro start-device --platform android --device-os android-37` end-to-end: AVD created with the correct `tag.id=google_apis` / `abi=arm64-v8a` (avdmanager-derived), emulator booted to Android 17 / 16 KB page size.

> **Note for local repro:** these 16 KB images require the emulator's **Vulkan** rendering path. If a local boot hangs at the Google spinner with `Assertion failed: !rcEnc->featureInfo()->hasReadColorBufferDma`, check `~/.android/advancedFeatures.ini` for a stale `Vulkan = off`. This is an environment setting, not affected by this PR.

### DIFF
--- a/maestro-client/src/main/java/maestro/device/DeviceSpec.kt
+++ b/maestro-client/src/main/java/maestro/device/DeviceSpec.kt
@@ -59,19 +59,23 @@ sealed class DeviceSpec {
         override val osVersion: Int get() = os.removePrefix("android-").toIntOrNull() ?: 0
         override val deviceName: String get() = "Maestro_ANDROID_${model}_${os}"
 
-        // API 37+ system images differ from the ≤36 form on two axes. (1) The platform id
-        // carries a minor version ("android-37.0", not "android-37") — Google's permanent
-        // minor-SDK scheme since Android 16 QPR2. (2) Only the 16 KB page-size image is
-        // published, whose package-path segment is "google_apis_ps16k" rather than the plain
-        // "google_apis". Both are encoded in this fully-qualified sdkmanager package path;
-        // avdmanager derives the tag and ABI from the package itself, so neither is passed
-        // separately when creating the AVD (the path segment "google_apis_ps16k" is NOT a
-        // valid avdmanager --tag — the image's tag id is plain "google_apis").
+        // emulatorImage is an sdkmanager "SDK-style path": system-images;<platform>;<dir>;<abi>.
+        // The 3rd segment is the on-disk directory under system-images/<platform>/, NOT a "tag"
+        // (the image's SystemImage.TagId). They coincided for single-tag images (≤36), but API 37's
+        // multi-tag 16 KB image has dir "google_apis_ps16k" while its tag id is plain "google_apis",
+        // so the dir name is rejected as an avdmanager --tag.
+        //
+        // CRITICAL: a fully-qualified --package makes avdmanager derive the tag and ABI from it, so
+        // createAndroidDevice passes ONLY --package — never --tag/--abi. This path is the sole source
+        // of truth; keep it that way.
+        //
+        // API 37+: platform carries a minor version ("android-37.0", not "android-37"; Google's
+        // permanent minor-SDK scheme since Android 16 QPR2) and only the 16 KB ("…_ps16k") dir ships.
         // https://android-developers.googleblog.com/2025/12/android-16-qpr2-is-released.html
         // https://developer.android.com/guide/practices/page-sizes
         private val imagePlatform: String get() = if (osVersion >= 37) "$os.0" else os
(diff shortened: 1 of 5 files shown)


maestro-client/src/main/java/maestro/device/DeviceSpec.kt
```diff
@@ -58,8 +58,24 @@ sealed class DeviceSpec {
...
+        //
+        // CRITICAL: a fully-qualified --package makes avdmanager derive the tag and ABI from it, so
+        // createAndroidDevice passes ONLY --package — never --tag/--abi. This path is the sole source
+        // of truth; keep it that way.
+        //
+        // API 37+: platform carries a minor version ("android-37.0", not "android-37"; Google's
+        // permanent minor-SDK scheme since Android 16 QPR2) and only the 16 KB ("…_ps16k") dir ships.
+        // https://android-developers.googleblog.com/2025/12/android-16-qpr2-is-released.html
+        // https://developer.android.com/guide/practices/page-sizes
+        private val imagePlatform: String get() = if (osVersion >= 37) "$os.0" else os
```
Fishbowler, line 76: Is it premature to start thinking about the 36.1 or 37.1 images?

Fishbowler (approved):

maestro-client/src/main/java/maestro/device/DeviceSpec.kt
```diff
@@ -58,8 +58,24 @@ sealed class DeviceSpec {
...
+        private val imageDir: String get() = if (osVersion >= 37) "google_apis_ps16k" else "google_apis"
```
amanjeetsingh150, line 77: I think I'm little worried about this one. This is a time maybe we should ask what is it exactly that's start device is supposed to do, if its launching available device. 

Hardcoding system images might not be the right call here. The reason we only have google_apis_ps16k for android 37 is that its not stable yet.

This would mean if they add google apis tomorrow we might need to revert it back or we should have a way to launch device with any other image as well.

Its true for even APIs today, there are system images like: 

```
system-images;android-35;google_apis_playstore_tablet;arm64-v8a
```

So, maybe a better solution is to have a provisioned way for user to launch any system image available locally.

amanjeetsingh150 (commented):
There are other things that we want to ensure before we commit to this PR. Supporting API 37 has a broader definition for core things like:

- Ensuring it starts locally
- Running current E2E to ensure core is able to run without issues
- Cloud support to bump sdk versions.

I don't think right now with this PR we're in state to fully commit the API 37 support. 

So, I'm curious what is exactly that's user is asking here and if there are other ways to get them unblock?

Fishbowler:
The user is me. I want to get timely support in place, given all the work we did to make it easy, and now that Google have released their final beta.

I've tested this branch locally and can launch devices on macOS. I'm travelling so don't have Windows/Linux available.

I've push changes to enable e2e and kicked off a run for API 37 here: https://github.com/mobile-dev-inc/Maestro/actions/runs/26881770783 🤞

PR #3333, opened 2026-06-03
ci: notify release-comms on published releases

manuarp:
## Summary

Adds `.github/workflows/notify-release-comms.yml` so that every time a release is published in this repo, a `repository_dispatch` event is sent to [`mobile-dev-inc/release-comms`](https://github.com/mobile-dev-inc/release-comms). That repo then runs the release-comms pipeline: generates a release-context doc in Notion, sizes the comms tier, opens a Linear project, drafts assets, and pings me on Slack for approval.

## What this does

- Triggers on `release: published` only — no effect on PRs, pushes, or any other event.
- Posts an HTTP request to GitHub Actions on the `release-comms` repo with `release_tag`, `release_url`, and `source_repo`.
- Doesn't modify any source, build artifacts, or release outputs.
- Failure is non-fatal to the release — if the dispatch fails, the release itself is untouched.

## Requirements

- Repo secret `RELEASE_COMMS_DISPATCH_TOKEN` already set (fine-grained PAT scoped to `actions: write` on `release-comms` only).

## Test plan

- [ ] Workflow file is valid YAML — GitHub linter passes
- [ ] After merge, publishing a release in this repo triggers a run in [release-comms actions](https://github.com/mobile-dev-inc/release-comms/actions)

Owner: @manuarp

### DIFF
--- a/.github/workflows/notify-release-comms.yml
+++ b/.github/workflows/notify-release-comms.yml
@@ -1,0 +1,28 @@
+# Copy this file to .github/workflows/notify-release-comms.yml in each source
+# repo (mobile-dev-inc/Maestro and mobile-dev-inc/copilot).
+#
+# Requires the source repo to have a secret RELEASE_COMMS_DISPATCH_TOKEN set
+# to a fine-grained PAT with `actions: write` scope on mobile-dev-inc/release-comms.
+
+name: Notify release-comms
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send repository_dispatch to release-comms
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.RELEASE_COMMS_DISPATCH_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-release-comms.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  release:
+    types: [published]
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send repository_dispatch to release-comms
+        uses: peter-evans/repository-dispatch@v3
```
Fishbowler: Should this be using the latest version?

```suggestion
        uses: peter-evans/repository-dispatch@v4
```

manuarp:
Pushed the same pre-release filter here too (`if: ${{ !github.event.release.prerelease }}`) — checked the existing Maestro releases and the `cli-2.3.1-dev.*` tags are flagged `prerelease: true` on the release object, so the guard catches them without needing a tag-pattern regex.

Thanks for the action version bump as well 👍

Fishbowler (approved):

PR #3334, opened 2026-06-04
fix(android): bypass UiAutomation for tap and inputText to fix Android 16 animation-wait hang

### LINKED ISSUE #2718: inputText extremely slow on Android 16

### Steps to reproduce

```
- inputText: 'Some text'
```

### Actual results

One keystroke is entered every 10-20 seconds

### Expected results

Input should be typed in a reasonable time, e.g. <1 second per keystroke.

### About app

Expo 52, issue is only present on Android

### About environment

macOS 15.6

Java 17

### Logs

_No response_

### Maestro version

1.40.3

Homebrew

### Anything else?

_No response_

jesper-irukadx:
## Proposed changes

Fixes #2718 — `inputText` (and `tap`) hang ~10 s per character on Android 16 with certain dev-client splash states.

Root cause: `tap` and `inputText` go through the gRPC path → UiAutomator's `InteractionController` → `UiAutomation.injectInputEvent(event, sync=true)`. The 2-arg overload internally calls the 3-arg overload with `waitForAnimations=true`, which makes UiAutomation block on `waitForAllWindowsDrawn(...)` before delivery. On Android 16 (API 36), Expo / RN dev clients (and probably others) leave a `starting_reveal` window animation that never reaches a completion callback. WindowManager logs `Timed out waiting for animations to complete` every 5 s, and every tap/keystroke eats one of those waits.

This PR replaces the gRPC paths for `tap` and `inputText` in `AndroidDriver.kt` with `dadb.shell("input tap …")` / `dadb.shell("input text …")` — the same approach this file already uses for `longPress`, `swipe`, and `pressKey`. `adb shell input` bypasses UiAutomation entirely and injects via `InputManager` directly with no animation wrapper, while keeping the same `INJECT_INPUT_EVENT_MODE_WAIT_FOR_FINISH` mode on the InputManager side.

### Implementation notes

- **`inputText` is per-char, not all-at-once.** Multi-field auto-focus inputs (OTP/PIN codes split across one EditText per character) need each input event to complete a full `onChange → state-update → re-render → next field's requestFocus` cycle before the next event. Bundling chars into one `input text` invocation can outrun the JS bridge and produce wrong-field characters (e.g. `CCOC` from `COMH`). Per-char keeps OTP fields happy without losing the orders-of-magnitude speedup (still ~50 ms/char vs the broken ~10 s/char).
- **Newline handling preserved.** The previous on-device `MaestroDriverService.setText` had no `KEYCODE_ENTER` case in its `when` and silently no-op'd `\n` / `\r`. The patched per-char loop matches that exactly. Flows that needed an ENTER keystroke already had to call `pressKey: ENTER`.
- **Shell escaping** handles `\`, `"`, `` ` ``, `$`, and spaces (`%s`, an `input` quirk). Other shell metacharacters are safely contained by the surrounding `"..."`.

### Behavioral diff

| Aspect | Before | After |
|---|---|---|
| Event delivery | `UiAutomation.injectInputEvent(sync=true)` → `waitForAnimations=true` wrapper → `InputManager` (`WAIT_FOR_FINISH`) | `adb shell input` → `InputManager` (`WAIT_FOR_FINISH`) directly |
| Pre-delivery animation wait | `waitForAllWindowsDrawn(...)` (the hang) | None |
| `runDeviceCall` retry on `UNAVAILABLE` | Yes | No (matches `longPress` / `swipe` / `pressKey`) |
| `Tracer.trace(x, y)` instrumentation | Called | Skipped (matches `longPress` / `swipe` / `pressKey`) |

### Alternatives considered

- **Device-side patch using the 3-arg overload** (`UiAutomation.injectInputEvent(event, sync, waitForAnimations=false)`, `@hide @TestApi`) — architecturally cleaner; preserves all gRPC error semantics and `Tracer`. More invasive (requires re-bundling `maestro-server.apk`). **Happy to switch this PR to that approach if you prefer.**
- **API-36-conditional or env-var-gated rollout** — easy to add if you want a narrower change for the first release.

### Measurements

Affected device — Pixel 7a, Android 16 (API 36), Expo dev-client:

| Operation | Before | After |
|---|---:|---:|
| `driver.tap` | ~10,000 ms | ~150 ms |
| `driver.inputText` (10 chars) | DEADLINE (cut at 120 s) | ~500 ms |
| Full 8-screen sign-up flow with OTP and assertions | DEADLINE | **144 s, all 9 asserts pass** |
| `WindowManager: Timed out waiting for animations` logs per flow | 22+ | 0 |

Healthy device — API 34 emulator, single run (3× coordinate tap, includes runner startup): vanilla 22.5 s vs patched 19.2 s. Not enough samples to characterize a delta; takeaway is **no measurable regression**.

### Suites run locally

- [x] `./gradlew test` — 754 tests across 9 modules, 0 failures, 0 errors
- [x] `./gradlew :maestro-test:test` — integration suite, BUILD SUCCESSFUL
- [x] `./gradlew detekt` — BUILD SUCCESSFUL
- [x] Manual: full real-device sign-up + onboarding + OTP + assertion flow with `./gradlew :maestro-cli:installDist` build

Not done by me (would appreciate maintainer-side coverage):
- [ ] Full CI matrix on API 32 / 35 / 36 emulators
- [ ] Maestro Cloud volume validation of the lost `runDeviceCall` retry semantics for `tap` / `inputText` (the other shell-based methods in this file already lack this retry)

### e2e demo tests

The existing `e2e/demo_app/` flows exercise `tap` and `inputText` heavily, so existing demo coverage validates the change. I haven't added a new demo screen because the change is at the driver level and behavior-preserving on healthy devices; let me know if you'd like a targeted regression flow added.

## Issues fixed

Fixes #2718

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -264,14 +264,27 @@ class AndroidDriver(
 
     override fun tap(point: Point) {
         metrics.measured("operation", mapOf("command" to "tap")) {
-            runDeviceCall("tap") {
-                blockingStubWithTimeout.tap(
-                    tapRequest {
-                        x = point.x
-                        y = point.y
-                    }
-                ) ?: throw IllegalStateException("Response can't be null")
-            }
+            // Dispatch the tap via `adb shell input tap` rather than the
+            // on-device gRPC path that calls UiAutomator's
+            // `InteractionController.clickNoSync(x, y)`. clickNoSync injects
+            // events via `UiAutomation.injectInputEvent(event, sync=true)`,
+            // which internally calls the 3-arg overload with
+            // `waitForAnimations=true`. That flag makes UiAutomation block on
+            // `waitForAllWindowsDrawn(...)` before delivery. On Android 16
+            // (API 36) with certain stuck animation states (e.g. the dev-client
+            // splash dismissal that never signals completion), this wait
+            // routinely hits 10+ s per tap, surfacing as
+            // `WindowManager: Timed out waiting for animations to complete`
+            // warnings every 5 s. See
+            // https://github.com/mobile-dev-inc/maestro/issues/2718.
+            //
+            // `adb shell input` bypasses UiAutomation entirely and injects via
+            // InputManager directly with no animation wrapper — same
+            // `INJECT_INPUT_EVENT_MODE_WAIT_FOR_FINISH` mode on the
+            // InputManager side, but without the `waitForAllWindowsDrawn` poll
+            // on top. `longPress`, `swipe`, and `pressKey` in this file already
+            // take the same approach.
+            dadb.shell("input tap ${point.x} ${point.y}")
         }
     }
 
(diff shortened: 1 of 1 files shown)


baalimago:
Any update on this? I've encountered the same. Fix would be appreciated.

Fishbowler:
I didn't see this one come in. I don't think we'd want to alter text input to work in this fashion.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -564,10 +577,49 @@ class AndroidDriver(
...
+                val escaped = when (ch) {
+                    '\n', '\r' -> return@forEach
+                    '\\' -> "\\\\"
+                    '"' -> "\\\""
+                    '`' -> "\\`"
+                    '$' -> "\\$"
+                    ' ' -> "%s"
+                    else -> ch.toString()
+                }
+                dadb.shell("input text \"$escaped\"")
```
Fishbowler, line 622: We intentionally don't use this method. Even though it's quick, it removes any semblance of users typing. The other side of what was there before (`maestro-android`'s `MaestroDriverService`) intentionally enters a single character at a time, allowing for the application under test to respond to input events.

Fishbowler (changes requested):

Fishbowler:
@jesper-irukadx I've moved this into draft. If this is something you don't think you'll come back to, please close. If you'd like a hand progressing it, let us know.

PR #3343, opened 2026-06-09
feat(orchestra): typed ArtifactManifest + flat per-flow run-root bundle

proksh:
**Stacked on #3282.** Makes Maestro **core** own a typed, self-documenting artifact manifest and the per-flow artifact bundle — so local and cloud emit the same shape and every consumer reads one model instead of re-deriving artifact types from a folder of files.

> [!NOTE]
> **This PR now also contains the device-log + crash/ANR work that was previously split out as #3346** (folded in here so the manifest and its device artifacts land together as one reviewable change). #3346 is closed.

> [!IMPORTANT]
> **The view hierarchy is no longer inline in `commands.json`.** It's its own `SCREEN_HIERARCHY` artifact — one `screen-hierarchy/step-<N>.json` per executed command, linked from that command's `artifacts` list (shrank `commands.json` ~246 KB → 4 KB on a demo run). Consumers that previously read an inline `hierarchy` field must follow the link instead. The schema's `COMMAND_METADATA`/`SCREEN_HIERARCHY` descriptions spell this out.

## What this adds

**1. Shared model** (`maestro-orchestra-models`) — `ArtifactManifest`, `ArtifactEntry`, `ArtifactKind`, `ArtifactFormat`. Plain data classes, no serialization annotations — tolerance is the consumer's choice. The relative-path layout is *not* part of the shared model: it's producer-private in `maestro-orchestra`'s `BundleLayout`, since consumers read each artifact's `relativePath` off the manifest at runtime.

**2. Self-documenting `manifest.json`** — carries a `$schema` pointing at a **stable, versioned URL** (`https://storage.googleapis.com/maestro-schemas/artifact-manifest/v1.schema.json`), published by a `publish-schemas.yaml` CI job on every `main` push. The `vN` path is the **only** version signal — a structural change ships `v2.schema.json` beside v1; additive changes overwrite v1 in place and lean on readers tolerating unknown fields. The schema is `additionalProperties: true` so an additive field can't fail validation against a cached older v1, and a build test fails if any `ArtifactKind`/`ArtifactFormat` is undocumented, so docs can't drift from code. _Trade-off:_ resolving the URL needs network; the offline-bundled copy was intentionally dropped. _Note:_ the URL 404s until the publish job first runs on `main`.

**3. Core produces the bundle** (`maestro-orchestra`) — an `ArtifactCollector` is the **single owner of the run-root bundle**: every file core writes is *recorded as it is allocated* (no end-of-flow disk scan), so the manifest and each command's `artifacts` list are the same records and can't drift. **The run root itself is the zippable bundle** — everything sits directly under it, no intermediate `artifacts/` folder.
- `commands.json` + `logs/maestro.log` at the run root. Serialized errors are slim (`message` + `debugMessage`); hierarchies/stack traces live in their own artifacts.
- **Typed per-command `artifacts`** — each command lists `{type, path}` objects (run-root-relative). Empty lists omitted.
- **Self-describing kinds** — `TAKE_SCREENSHOT` (→ `takeScreenshot/`), `START_SCREEN_RECORDING` (→ `startRecording/`), `SCREENSHOT` (step screenshots → `screenshots/step-<N>.png`), `SCREEN_RECORDING` (full-run), `SCREEN_HIERARCHY` (per-step JSON).
- **Per-step view hierarchy, always on** — `screen-hierarchy/step-<N>.json` per executed command, referenced from its `artifacts`.
- **Unified screenshots** — one `screenshots/` folder: with `captureFullArtifacts` on (worker) every step gets `step-<N>.png`; off (local CLI) only the failed step does. No special `screenshot-❌-*.png` convention — the failure shot is the FAILED command's `SCREENSHOT`.
- **Flag-gated capture** (`captureFullArtifacts`, off by default; worker turns it on): per-step screenshots + the full-run `screen-recording.mp4`. **Off by default because these are the heavy, disk-hungry artifacts (big per-step PNGs + the mp4)**; the cheap, always-useful ones stay on so a local run still debugs well without filling the disk.

**4. Core captures device logs + crash/ANR** (was #3346) — ported from the cloud worker so local and cloud produce the same artifacts; they land in `logs/` and the manifest, **on by default both local and cloud** (cheap and always-useful, unlike the flag-gated screenshots/recording).
- **Driver seam** (`maestro-client`): a `CapturedDeviceArtifact` descriptor + three best-effort `Driver` methods (start log capture, stop & collect logs, collect crash/ANR), default no-ops. The driver layer stays free of the manifest model; orchestra maps the descriptor to an `ArtifactKind`.
- **Orchestra**: `DeviceArtifactCapturer` starts on flow start, collects on flow end (scoped to the app under test + flow start), folds results into the same collector.
- **Android**: logcat (`DEVICE_LOG`, source `android`) + crash + ANR. **iOS**: `simctl log stream` (source `simulator`) + `xctest_runner.log` (source `xctest`) + `.ips` crash.
- Best-effort throughout — any capture failure logs a warning and is skipped, never failing the flow.

**5. CLI writes one folder per flow** (`maestro-cli`) — each flow's output folder is resolved upfront and passed as the single `artifactsDir`; Orchestra writes the whole bundle into it. Single and suite runs both produce `<output>/<ts>/<flow>/` (with `-shard-N` / `-N` suffixes). Replaces the old flat session layout + temp-staging copy step.

## Layout
```
<output>/<ts>/<flow>/          # the run root IS the "artifacts" zip = everything core makes
  manifest.json
  commands.json
  logs/
    maestro.log
    device-logcat.txt / device-simulator.log / device-xctest.log   # device logs (on by default)
    crash-report.txt / anr-report.txt                              # if found this flow
  takeScreenshot/…             # takeScreenshot command output
  startRecording/…             # startRecording command output
  screenshots/step-N.png       # step screenshots (all steps when flag on; failed step only when off)
  screen-hierarchy/step-N.json # per-step view hierarchy (always on)
  screen-recording.mp4         # full-run recording (flag-gated: worker on, CLI off)
```
The manifest uses paths relative to its own folder, so local and cloud manifests look the same.

## Behavior change to flag
`--test-output-dir`, when set, **is** this artifact folder — it now holds the full bundle (incl. `manifest.json`), not just screenshots. Help text updated. Default location (`~/.maestro/tests/<ts>/`) unchanged.

## ⛔️ Copied verbatim from the cloud worker — **no review needed**
Byte-for-byte identical to the production sources in `copilot` (only `package`/`import` lines changed for the new module path) — the exact logic the worker already runs, with their unit tests and fixtures:

| File | Verbatim source (`copilot/…`) |
|------|-------------------------------|
| `maestro-client/…/android/crashes/LogcatReader.kt` | `shared/…/android/crashes/LogcatReader.kt` |
| `maestro-client/…/android/crashes/LogcatReaderTest.kt` | `maestro-worker/…/android/crashes/LogcatReaderTest.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IPSParser.kt` | `maestro-worker/…/ios/IPSParser.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IPSParserTest.kt` | `maestro-worker/…/ios/IPSParserTest.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IOSCrashFileFinder.kt` | `maestro-worker/…/ios/IOSCrashFileFinder.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IOSCrashFileFinderTest.kt` | `maestro-worker/…/ios/IOSCrashFileFinderTest.kt` |
| `maestro-ios-driver/src/test/resources/ios/crashes/*.ips` (5 fixtures) | `maestro-worker/…/test/resources/ios/crashes/*.ips` |

**Everything else is new and worth reviewing** — the manifest model + schema, the `ArtifactCollector`/`ArtifactsGenerator`/`BundleLayout` glue, the CLI per-flow folder logic, the device driver seam (`CapturedDeviceArtifact`, the `Driver`/`Maestro` methods, the `AndroidDriver`/`IOSDriver`/`LocalSimulatorUtils` wiring), the dadb logcat helpers (`DadbExtensions.kt`, extracted not copied), and the orchestra glue (`DeviceArtifactCapturer`, `DeviceArtifactFiles`).

## How this was tested
- [x] **Model + orchestra tests** — manifest round-trips and ignores unknown fields; a build check fails if any kind is missing from the schema; collector records dedicated kinds with run-root-relative paths, folders register as collections with counts, step screenshots/recording appear per their flag, hierarchy is written per executed command (not skipped) and `commands.json` has no inline `hierarchy`; typed `artifacts` attribute to the right command (incl. nested/composite dedup); `takeScreenshot`/`startRecording` allocate through the collector and land in the right bundle folder; serialized errors carry only `message` + `debugMessage`; `manifest.json` carries the stable `$schema` URL and bundles no schema file.
- [x] **Ported device unit tests (61)** — logcat reader, `.ips` crash parser, crash-file finder, with fixtures.
- [x] **Local e2e (iOS sim), flag off and on** — confirmed the OFF bundle (no `screen-recording.mp4`, `SCREENSHOT` = failed step only) vs ON (`SCREEN_RECORDING` + per-step `screenshots/`); `takeScreenshot` lands at `takeScreenshot/<name>.png` attributed to its command; `commands.json` 4.3 KB with hierarchy linked not inline; `DEVICE_LOG` (simulator + xctest) present in both; `$schema` URL present, no `schemaVersion`.
- [x] **Android emulator / crash** — `DEVICE_LOG` (source `android`) + parsed `CRASH_REPORT` with a friendly message.

## Context for reviewers (design decisions)
- **Record, don't scan.** The collector records each artifact as it's produced (`allocate` for files core writes, `adopt` for device logs/crash) — no end-of-flow disk walk. The manifest is those records; each command's list is the same records grouped by command, so they can't drift. `takeScreenshot`/`startRecording` also allocate through the collector (no separate dispatch), so nothing reaches the bundle unrecorded and Orchestra never needs the folder layout.
- **One version signal.** Dropped the `schemaVersion` field — version lives only in the `$schema` URL's `vN` path. Schema is `additionalProperties: true` so additive changes stay forward-compatible. (Removed the unused `USER_FILE` kind too.)
- **Hierarchy is its own artifact, on purpose** — see the IMPORTANT note above (commands.json was getting huge).
- **One `SCREENSHOT` kind** for failure + per-step shots — the failed step is in `commands.json`, its shot linked from there, so no separate `FAILURE_SCREENSHOT` kind is needed.
- **`captureFullArtifacts` is a single flag** gating per-step screenshots + recording together — a disk-size call, not a capability one; easy to split later if a consumer needs a subset.
- **Device-artifact seam.** Core *owns* device capture (drives logcat/simctl/.ips itself) rather than receiving entries from a producer below it — the heavy reading of "Orchestra owns the manifest." If producers ever proliferate, the clean next step is a generic `adopt(kind, file, producer)` seam so a non-core producer can register an artifact without core re-implementing its capture.
- **Crash/ANR scoping uses the first `launchApp`'s appId** (one flow tests one app). No `launchApp` ⇒ appId null ⇒ collection unscoped, still best-effort.

## Follow-up (separate PR)


steviec:
@proksh I worked with Claude to generate this. I'll attach the conversation so you can see how I got there. Please note that this is a first pass conversation, but it immediately picked up on the architectural problems I noticed in this PR on my first scan. If core is going to own the artifacts, it needs to really _own_ them, and not just find them.

The "artifact policy" idea is interesting, but it might not be necessary. If we are only in two modes: record everything vs record a lighter subset, that might be better as a single boolean flag.

------

**Artifact manifest: record, don't scan**

The RFC call is right — one typed manifest owned by Orchestra, identical local and cloud. The implementation undermines it: it *derives* the manifest by scanning the run-root folders at flow end instead of *recording* artifacts as they're produced. Most of the issues below follow from that.

**Problems**
- Three registration paths, no single source of truth: a listener event (only the two command outputs), direct in-generator appends, and an end-of-run disk scan. The folder-level manifest and per-command list can disagree.
- The scan can't see what core didn't hardcode — pushed/user files and anything a non-core producer adds are invisible.
- No single "register an artifact" operation, so the manifest is core-sealed; absorbing device logs required re-implementing their capture in core (#3346). Doesn't generalize.
- Capture gating is three inconsistent rules (booleans for screenshots/recording, always-on for device logs), none client-selectable.
- `ArtifactFiles` paths are a constant shared between writers and the scanner — that shared agreement is the drift surface.

**Fix: a single collector that owns path allocation and the record.** Producers go through it; the manifest is its records, the per-command list the same records grouped by command. Two verbs: `allocate(kind, producer) → path` for what we write (recorded atomically — can't write into the bundle unrecorded), `adopt(kind, file, producer)` for what's collected from outside (crash dumps, log streams). `producer` is a field, so worker- vs core-produced is data, not a second code path. Capture becomes one requested-artifact policy. `ArtifactFiles` becomes the collector's private layout, not a shared contract.

**Layering (open decision):** core manifest sealed/immutable at flow end; backend wraps it in a run envelope (run id, signed URLs, post-hoc cloud artifacts) that's a superset of the same entry type. Avoids a mutable manifest and the "when is it final" ambiguity.

**PR impact:** keep the model types, schema, driver seam, and flat layout. Replace the disk scan with collector accumulation; route every writer through `allocate`/`adopt` and delete the self-append/scan paths; swap the capture booleans for a policy; add `producer` to `ArtifactEntry` without building a separate worker-append path. Net less code than today.

----

Chat:
[2026-06-12-133345-local-command-caveatcaveat-the-messages-below.txt](https://github.com/user-attachments/files/28897421/2026-06-12-133345-local-command-caveatcaveat-the-messages-below.txt)

proksh:
> @proksh I worked with Claude to generate this. I'll attach the conversation so you can see how I got there. Please note that this is a first pass conversation, but it immediately picked up on the architectural problems I noticed in this PR on my first scan. If core is going to own the artifacts, it needs to really _own_ them, and not just find them.
> 
> The "artifact policy" idea is interesting, but it might not be necessary. If we are only in two modes: record everything vs record a lighter subset, that might be better as a single boolean flag.
> 
> **Artifact manifest: record, don't scan**
> 
> The RFC call is right — one typed manifest owned by Orchestra, identical local and cloud. The implementation undermines it: it _derives_ the manifest by scanning the run-root folders at flow end instead of _recording_ artifacts as they're produced. Most of the issues below follow from that.
> 
> **Problems**
> 
> * Three registration paths, no single source of truth: a listener event (only the two command outputs), direct in-generator appends, and an end-of-run disk scan. The folder-level manifest and per-command list can disagree.
> * The scan can't see what core didn't hardcode — pushed/user files and anything a non-core producer adds are invisible.
> * No single "register an artifact" operation, so the manifest is core-sealed; absorbing device logs required re-implementing their capture in core ([feat(orchestra): device logs + crash/ANR under logs/ #3346](https://github.com/mobile-dev-inc/Maestro/pull/3346)). Doesn't generalize.
> * Capture gating is three inconsistent rules (booleans for screenshots/recording, always-on for device logs), none client-selectable.
> * `ArtifactFiles` paths are a constant shared between writers and the scanner — that shared agreement is the drift surface.
> 
> **Fix: a single collector that owns path allocation and the record.** Producers go through it; the manifest is its records, the per-command list the same records grouped by command. Two verbs: `allocate(kind, producer) → path` for what we write (recorded atomically — can't write into the bundle unrecorded), `adopt(kind, file, producer)` for what's collected from outside (crash dumps, log streams). `producer` is a field, so worker- vs core-produced is data, not a second code path. Capture becomes one requested-artifact policy. `ArtifactFiles` becomes the collector's private layout, not a shared contract.
> 
> **Layering (open decision):** core manifest sealed/immutable at flow end; backend wraps it in a run envelope (run id, signed URLs, post-hoc cloud artifacts) that's a superset of the same entry type. Avoids a mutable manifest and the "when is it final" ambiguity.
> 
> **PR impact:** keep the model types, schema, driver seam, and flat layout. Replace the disk scan with collector accumulation; route every writer through `allocate`/`adopt` and delete the self-append/scan paths; swap the capture booleans for a policy; add `producer` to `ArtifactEntry` without building a separate worker-append path. Net less code than today.
> 
> Chat: [2026-06-12-133345-local-command-caveatcaveat-the-messages-below.txt](https://github.com/user-attachments/files/28897421/2026-06-12-133345-local-command-caveatcaveat-the-messages-below.txt)

Done:

**Record, don't scan** — killed the end-of-flow disk walk. an ArtifactCollector now records each artifact as it's produced (allocate for files core writes, adopt for device logs / command output). the manifest is those records, and each command's list is the same records grouped by command — so they can't drift.

**ArtifactFiles drift surface** — you were right. the layout constants had zero cross-module use (consumers read relativePath off the manifest at runtime), so i moved them into a module-internal BundleLayout — producer-private, worker can't couple to it. only the schema url/resource stayed, now on ArtifactManifest. ArtifactFiles is gone.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
```diff
@@ -60,12 +107,14 @@ internal class ArtifactsGenerator(
...
-            // Expensive device round-trips; only when producing a bundle. Each is
-            // independent best-effort — one failing doesn't block the other.
-            if (artifactsDir != null) {
-                captureHierarchy(metadata)
-                captureFailureScreenshot()
-            }
+        }
+        if (artifactsDir == null || outcome is CommandOutcome.Skipped) return
+
+        captureStepHierarchy(metadata)
```
pedro18x: The diff drops the old `if (outcome is CommandOutcome.Failed)` guard, so the view hierarchy is now captured on every command instead of only on failures: a synchronous viewHierarchy() round-trip plus a JSON write per command. A 100-command local run goes from ~1 to ~100 device round-trips. Per-step screenshots and the recording are gated behind captureFullArtifacts for being expensive; this is not. Worth gating it the same way (or keeping it failure-only when the flag is off) unless the added local-run latency is known to be small.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ScreenshotUtils.kt
```diff
@@ -6,28 +6,23 @@ import okio.sink
...
      */
     fun takeDebugScreenshot(
         maestro: Maestro,
         debugOutput: FlowDebugOutput,
         status: CommandStatus,
         destFile: File? = null,
     ): File? {
-        // Avoid duplicate FAILED screenshots from parent composite commands.
         val containsFailed = debugOutput.screenshots.any { it.status == CommandStatus.FAILED }
         if (containsFailed && status == CommandStatus.FAILED) {
```
pedro18x: Skipping a FAILED shot when one already exists fits the composite leaf/parent case, but the new model gives each failed step its own step-N.png. If a flow can finish more than one command as Failed (optional / continue-on-failure), every failure after the first never gets its screenshot. If that is reachable, this dedup wants to be per-command rather than flow-wide.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
```diff
@@ -537,6 +551,68 @@ class IOSDriver(
...
+            deviceLogStream = LocalSimulatorUtils(TempFileHandler()).startDeviceLogStream(deviceId, tmp)
+        } catch (e: Exception) {
+            LOGGER.warn("Failed to start simulator log stream", e)
+        }
+    }
+
+    override fun stopAndCollectDeviceLogs(outputDir: File): List<CapturedDeviceArtifact> {
+        val out = mutableListOf<CapturedDeviceArtifact>()
+        try {
+            deviceLogStream?.destroy()
```
pedro18x: destroy() is an async SIGTERM and the log file is copied right after with no waitFor(), so the tail of the log (often the lines around a crash) can be lost before simctl flushes. stopScreenRecording does close() + waitFor() for this reason; a bounded waitFor() here would match it.

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
```diff
@@ -75,62 +72,45 @@ object TestDebugReporter {
...
-
-        TestOutputWriter.saveCommands(
-            path = path,
-            debugOutput = debugOutput,
-            commandsFilename = "commands-$shardPrefix($cleanFlow).json",
-            logPrefix = logPrefix,
-        )
-
-        val named = debugOutput.screenshots.map { shot ->
+    fun persistDebugScreenshots(debugOutput: FlowDebugOutput, flowDir: Path) {
```
pedro18x: These WARNED/analyze shots are copied into the flow folder outside the collector, so they never land in manifest.json, which cuts against the manifest-is-source-of-truth contract the rest of the PR sets up. It is also only on the single-flow path, so single vs suite runs differ. Routing them through the collector would keep them in the manifest and align both paths.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
proksh: You are right, that was an unintended round-trip per step. Gated it behind `captureFullArtifacts` now, same as the per-step screenshots. failed steps still capture hierarchy always (local debugging stays intact); worker gets it for every step.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ScreenshotUtils.kt
proksh: Valid. 

Now it only skips the wrapper's duplicate of its inner step's screen - every genuinely separate failure keeps its own shot. Also moved this out of ScreenshotUtils, since that layer can't tell which case it's in.

maestro-client/src/main/java/maestro/drivers/IOSDriver.kt
proksh: yep, async SIGTERM could drop the log tail. added a bounded waitFor before copying, mirroring stopScreenRecording. the restart + teardown destroy() sites don't copy after, so they stay no-wait.

maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt
proksh: Fixed, and went further. Warned shots now capture into the bundle on every run (single + suite), so they're in the manifest. For `--analyze` I dropped the bespoke per-command capture entirely — it just flips captureFullArtifacts now, so the AI reads the per-step screenshots straight off the bundle (it already walks the folder for every png). deleted `persistDebugScreenshots` + the dead helpers.

Extra while in here: `assertWithAI` / `assertNoDefectsWithAI` screenshots now record as AI_ANALYSIS in the manifest too (realizes the reserved kind + the TODO).

Verified locally - repeat/retry produce a screenshot per iteration/attempt, warned + AI_ANALYSIS land in the manifest.

pedro18x (approved):
Approving. All four resolved: hierarchy capture is gated, the dedup is
per-sequence instead of flow-wide, the iOS log waits to flush before copy,
and WARNED/analyze shots go through the manifest. Good test coverage too, nice!!

Only thing pending is the recurrent android e2e fails.

Fishbowler:
Oh _dude,_ the artifacts on the Test E2E job are so much easier to work with ❤️ This is awesome!

Issue #3335, opened 2026-06-04
[Feature Request] Allow Custom Port Configuration For Android Driver


abaiadse:
### Description

The Android driver currently binds to a hardcoded port (7001).

Would it be possible to expose this as a configuration option (e.g., a CLI flag or environment variable)?

### Motivation

Using a fixed port can lead to conflicts when port 7001 is already occupied by another process on the host. It can also create challenges in environments where multiple Maestro processes are orchestrated independently.

A configurable port would provide greater deployment flexibility and allow users to manage port allocation according to their environment.

### Proposed Solution

Allow overriding the default port through configuration while preserving 7001 as the default value.

Examples:

maestro test --android-driver-port 7002

or

ANDROID_DRIVER_PORT=7002

PR #3344, opened 2026-06-09
Allow configuring Android driver port

abaiadse:
## Proposed changes
Added support for configuring the Android driver host port via the ANDROID_DRIVER_PORT environment variable.
Preserved the existing default behavior by continuing to use port 7001 when the environment variable is not set or contains an invalid value.
This allows users to avoid port conflicts and run multiple Maestro processes with custom port assignments.
Testing

## Testing

Verified that Maestro continues to use port 7001 when ANDROID_DRIVER_PORT is not set.
Verified that setting ANDROID_DRIVER_PORT causes the Android driver to bind to the specified port.

## Issues fixed
https://github.com/mobile-dev-inc/Maestro/issues/3335

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -62,8 +62,6 @@ import kotlin.io.use
 
 private val logger = LoggerFactory.getLogger(Maestro::class.java)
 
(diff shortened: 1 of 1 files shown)


Fishbowler:
I note that the linked issue is a duplicate. Is there any chance that theres an existing PR for this too?

(We're running a little behind, and they're piling up 😬)

PR #3346, opened 2026-06-10
feat(orchestra): device logs + crash/ANR under logs/

proksh:
**Stacked on #3343.** Lets Maestro **core** capture device logs and crash/ANR reports into each flow's artifact bundle + `manifest.json`, so local and cloud produce the same artifacts. Ports the worker's iOS/Android capture into OSS, so the worker can later drop its own copies and just read the manifest.

Device logs and crash/ANR land in **`logs/`** at the run root — alongside `maestro.log`, inside the single zippable bundle (the run root itself; see #3343, which removed the intermediate `artifacts/` folder).

## What this adds
- **Driver seam** (`maestro-client`): a `CapturedDeviceArtifact` descriptor + three best-effort `Driver` methods (start log capture, stop & collect logs, collect crash/ANR) that default to no-ops. The driver layer stays free of the manifest model; orchestra maps the descriptor to an `ArtifactKind`.
- **Orchestra**: `DeviceArtifactCapturer` starts capture on flow start, collects on flow end (scoped to the app under test + flow start time) into `logs/`, and folds the results into the manifest.
- **Android**: logcat (`DEVICE_LOG`, source `android`) + crash + ANR.
- **iOS**: `simctl log stream` (source `simulator`) + `xctest_runner.log` (source `xctest`) + `.ips` crash.

Capture is per-flow and best-effort — a failure logs a warning and is skipped, never failing the flow. Crash/ANR are only emitted if found, scoped to this flow (no re-attributing an earlier flow's crash).

## Bundle layout (where these land)
```
<run-root>/
  manifest.json
  commands.json
  logs/
    maestro.log
    device-logcat.txt          # Android  (source android)
    device-simulator.log       # iOS      (source simulator)
    device-xctest.log          # iOS      (source xctest)
    crash-report.txt           # crash, if found this flow
    anr-report.txt             # ANR, if found this flow
  takeScreenshot/ …            # (#3343)
```

## Copied verbatim from the cloud worker — **no review needed**
These files are byte-for-byte identical to the production sources in the `copilot` repo (only the `package`/`import` lines changed for the new module path). They're the exact logic the cloud worker already runs, with their unit tests and fixtures, so they don't need line-by-line review:

| File | Verbatim source (`copilot/…`) |
|------|-------------------------------|
| `maestro-client/…/android/crashes/LogcatReader.kt` | `shared/…/shared/android/crashes/LogcatReader.kt` |
| `maestro-client/…/android/crashes/LogcatReaderTest.kt` | `maestro-worker/…/android/crashes/LogcatReaderTest.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IPSParser.kt` | `maestro-worker/…/ios/IPSParser.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IPSParserTest.kt` | `maestro-worker/…/ios/IPSParserTest.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IOSCrashFileFinder.kt` | `maestro-worker/…/ios/IOSCrashFileFinder.kt` |
| `maestro-ios-driver/…/xcuitest/crash/IOSCrashFileFinderTest.kt` | `maestro-worker/…/ios/IOSCrashFileFinderTest.kt` |
| `maestro-ios-driver/…/test/resources/ios/crashes/*.ips` (5 fixtures) | `maestro-worker/…/test/resources/ios/crashes/*.ips` |

Everything else **is** new and worth reviewing — the driver seam (`CapturedDeviceArtifact`, the `Driver`/`Maestro` methods, the `AndroidDriver`/`IOSDriver`/`LocalSimulatorUtils` wiring), the dadb logcat helpers (`DadbExtensions.kt`, extracted from the worker's `AndroidDevice.kt`, not a file copy), and the orchestra glue (`DeviceArtifactCapturer`, `ArtifactsGenerator`, `DeviceArtifactFiles`).

## How this was tested
- [x] **Ported unit tests (61)** — logcat reader, `.ips` crash parser, and crash-file finder, with fixtures.
- [x] **Android emulator** — produced a device log + `DEVICE_LOG` manifest entry (source `android`) under `logs/`.
- [x] **Android crash** (injected `am crash`) — produced a parsed `CRASH_REPORT` with a friendly message.
- [x] **iOS simulator** — produced both the simulator log and the xctest log as `DEVICE_LOG` entries.
- [ ] iOS `.ips` crash and Android ANR aren't E2E-triggered (hard to induce reliably), but are the worker's exact code covered by the ported unit tests.

## Follow-up (separate PR)

## Context for reviewers (design decisions)
- **Crash/ANR scoping uses the first `launchApp`'s appId, on purpose.** One flow is meant to test one app, so scoping crash collection to the first launched app is correct for our case. If a flow has no `launchApp` (app already running) the appId is null and collection is unscoped — still best-effort and never fails the flow.


proksh:
**Note: device-log/crash capture is on-by-default and ungated for local, on purpose.**

Unlike per-step screenshots + the full-run recording (gated behind `captureFullArtifacts`, worker-only), device logs + crash/ANR capture run by default on **both** local and cloud — so a local `maestro test` now starts `logcat` / `simctl log stream` and harvests crash/ANR into `logs/`. This matches the RFC follow-up ("on by default, local and cloud").

The asymmetry (logs ungated, screenshots flag-gated) is intentional, not an oversight: device logs are cheap/always-useful for debugging a local run, whereas full screenshot/recording capture is the heavier worker-only path. Capture is best-effort throughout — any failure logs a warning and is skipped, never failing the flow — so the only effect on local runs is the extra `logs/` artifacts.

proksh:
folded into #3343 — the device-log + crash/ANR work now ships in the same PR as the manifest it writes into, so they review and land together. nothing dropped; all of it is in #3343 (with the verbatim-copied files clearly marked as no-review there). closing here.

PR #3356, opened 2026-06-15
refactor(web): translate device-condition Selenium errors at the driver boundary

steviec:
Gives `CdpWebDriver` a `runDeviceCall` that mirrors `IOSDriver.runDeviceCall`, wrapping each Selenium-touching method. It translates only device-condition failures:

- `SessionNotCreatedException` / `NoSuchSessionException` → `DeviceUnreachableException`
- `TimeoutException` → `MaestroException.DriverTimeout`
- everything else propagates unchanged

Test verdicts (`AssertionFailure` / `ElementNotFound`) stay in Orchestra/Maestro, where they're derived from the hierarchy — the driver doesn't mint them. No reflection, no message-string matching.

Behavior is essentially unchanged today (infra was already `INFRA_ERROR`); this just establishes the typed boundary in the right place. Supersedes #3315.


maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
```diff
@@ -767,6 +805,39 @@ class CdpWebDriver(
...
+     */
+    private fun <T> runDeviceCall(callName: String, call: () -> T): T {
+        return try {
+            call()
+        } catch (e: WebDriverException) {
+            when {
+                e is SessionNotCreatedException || e is NoSuchSessionException ->
+                    throw DeviceUnreachableException(callName = callName, cause = e)
+
+                e is TimeoutException ->
```
fabianoyoschitaki, line 830: something to be aware in the future is that seems like Selenium TimeoutException has 2 different causes: infra (command / driver.get() page-load timeouts) and test-level (WebDriverWait/FluentWait conditions not met). so once WebDriverWait get introduced, it might become a misclassification

steviec:
Closing for now. the customer issue already solved the reported customer issue in #3314. I'd like to evaluate this later, or an approach more similar to #3104

PR #3359, opened 2026-06-16
test(ios): compare view-hierarchy walk complexity before/after #3313

proksh:
## What

Adds a unit test that measures how much work the iOS view-hierarchy walk does, **before and after #3313**, both against real production code, plus a one-line `private → internal` change so the walk is reachable from the test.

## Why

`ViewHierarchyHandler.elementHierarchy` runs on **every** iOS command (assert / tap / wait) to convert the XCTest snapshot into Maestro's `AXElement` tree. `XCUIElementSnapshot.dictionaryRepresentation` is a **deep** call — it serializes the element *and its whole subtree* on every access.

- **Before #3313** the walk was `let dict = snapshot.dictionaryRepresentation; return AXElement(dict)` — one deep read at the root → **Θ(N)**.
- **After #3313** (current `main`) it calls `dictionaryRepresentation` **once per node** → **Θ(N²)**, exactly `N·(N+1)/2` deep serializations for a chain of `N` elements.

It is a **pure performance** change — the walk returns the same tree, just much slower on element-dense screens, which is why long iOS flows started tipping over the 15-minute execution cap.

## Result

`deepReadCalls` = times the screen is read; `nodesSerialized` = total work. Measured on a chain of N elements:

| N (elements) | before #3313 reads | before #3313 work | after #3313 reads | after #3313 work (= N·(N+1)/2) |
|---|---|---|---|---|
| 16  | 1 | 16  | 16  | 136 |
| 64  | 1 | 64  | 64  | 2,080 |
| 256 | 1 | 256 | 256 | 32,896 |
| 512 | 1 | 512 | 512 | 131,328 |

So #3313 turned a single read-per-command into one read **per element** — ~128× more work at 256 elements, ~256× at 512.

## Test

`-only-testing:maestro-driver-iosUITests/ViewHierarchyComplexityTests` — 2 tests (before / after #3313), green on iPhone 16 / iOS 18.2.

- `test_before3313_readsScreenOnce_isLinear` — exercises the pre-#3313 body via the real `AXElement(_:)` initializer → 1 read, N work.
- `test_after3313_readsScreenPerNode_isQuadratic` — drives the real `ViewHierarchyHandler` → N reads, N·(N+1)/2 work.

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
+++ b/maestro-ios-xctest-runner/maestro-driver-ios.xcodeproj/project.pbxproj
@@ -24,6 +24,7 @@
 		52049BE0293503A200807AA3 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 52049BDF293503A200807AA3 /* Assets.xcassets */; };
 		52049BE3293503A200807AA3 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 52049BE1293503A200807AA3 /* LaunchScreen.storyboard */; };
 		52049BF8293503A200807AA3 /* maestro_driver_iosUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 52049BF7293503A200807AA3 /* maestro_driver_iosUITests.swift */; };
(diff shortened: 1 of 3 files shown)


proksh:
Addressed through https://github.com/mobile-dev-inc/Maestro/pull/3364

PR #3360, opened 2026-06-16
fix(ios): make view-hierarchy walk O(N) instead of O(N²)

proksh:
## Problem

`elementHierarchy` runs on **every** iOS command (tap / assert / wait) to turn the screen snapshot into an `AXElement` tree.

Since #3313 it called `snapshot.dictionaryRepresentation` **once per node**. That call serializes the element *and its whole subtree* every time — so a screen of `N` elements did `N·(N+1)/2` deep reads. That's **O(N²)**.

On big screens this slowed every command enough to push tight iOS flows past the 15-minute cap (**MA-4086**).

## Fix

Read `dictionaryRepresentation` **once at the root**, then walk that one tree — `O(N)`. The cross-process offset fix from #3313 is kept as-is.

| N | reads before | reads after |
|---|--------------|-------------|
| 256 | 256 | 1 |
| 512 | 512 | 1 |

Tests live in #3359 (this PR is the fix only). Same issue was flagged in review on #3250.


proksh:
Addressed through https://github.com/mobile-dev-inc/Maestro/pull/3364

PR #3372, opened 2026-06-22
refactor(android): AndroidDeviceConnection — sole owner of dadb + gRPC

amanjeetsingh150:
## What & why

Bumps **dadb 1.2.10 → 2.0.0** and makes **`AndroidDeviceConnection` the single boundary** that owns the Android device transport (the `dadb.Dadb` **and** the gRPC channel) and classifies every device failure in one place. The raw `dadb.Dadb` no longer threads through the CLI, session managers, drivers, and helpers.

The motivating bug: when the transport died mid-flow, the bare `IOException` dadb threw was indistinguishable from a failing command and got wrapped into a `MaestroException` — so **a dead device was reported as a customer test failure** instead of retryable infra. This PR draws one transport boundary and **one exception contract**, and removes raw `IOException` from the device path on both Android and iOS.

---

# 🔑 The exception contract (the headline change)

Three layers, each with one responsibility. **This is the contractual part — please review it first.**

### Layer 1 — `AndroidDeviceConnection` is the *sole classifier*
It answers exactly one question: *did the device die, or did it answer?*
- **Died** (adb / gRPC transport broke) → throws a **`DeviceConnectionException`** subtype: `DeviceUnreachableException`, `DeviceServerDiedException`, `DeviceAuthException`. **This base is shared with iOS.**
- **Answered** (success *or* failure) → the outcome is **returned** (`AdbShellResponse` exit code, `InstallResult`/`UninstallResult`/`SyncResult`, gRPC `DeviceResponse`) or thrown as a typed **operation-failure `RuntimeException`** — `AndroidOperationFailedException` (dadb plane) / `DeviceCallFailedException` (gRPC plane) — via `.orThrow()` / `.orThrowOnFailure()`.
- It is the **only** place that catches raw `AdbException` / `StatusRuntimeException` and the only place that constructs these exception types.

### Layer 2 — `AndroidDriver` interprets operation outcomes only
- It **never** catches `DeviceConnectionException` / `IOException` / `Exception` / `Throwable` around a device call — transport deaths always propagate as infra.
- Its only `catch` is the typed **operation-failure** type, and only where a product rule needs it (ignore a non-changeable permission, "emulator can't screen-record", airplane-mode root retry, best-effort uninstall). It constructs no exceptions itself — it delegates to `connection.shell(...).orThrow()`.

### Layer 3 — Orchestra & consumers carry *no* device-exception logic
- Orchestra treats **only `MaestroException`** as a command failure (→ `onCommandFailed`, which attributes the command FAILED + screenshot). A **transport death (and any unexpected bug) propagates untouched** — never routed through `onCommandFailed`, never reported as a customer/test failure. (The single `catch (DeviceConnectionException) { throw e }` is the one sanctioned, commented exception to "no device logic in Orchestra".)
- Consumers (cli / mcp / worker / studio) catch the **typed `DeviceConnectionException`**, never a bare `IOException`. A dead device is **retryable INFRA** (worker → `INFRA_ERROR`; CLI → suppressed via `isShutDown()`). Operation-failure RuntimeExceptions are **flow/test failures**, not infra, and can't be swallowed by a transport-death `catch`.

### New contract surface (types consumers should use)
| Concept | Type | Notes |
|---|---|---|
| "Device connection is in trouble" | `maestro.DeviceConnectionException` (abstract, `: IOException`) | The ONE type to catch. Subtypes: `DeviceUnreachableException`, `DeviceServerDiedException`, `DeviceAuthException`. |
| dadb-plane operation failure | `AndroidOperationFailedException : RuntimeException` | non-zero shell exit, rejected install. **Not** an `IOException`. |
| gRPC-plane operation failure | `DeviceCallFailedException : RuntimeException` | server answered a failure status. **Not** an `IOException`. |
| "this shell must succeed" | `AdbShellResponse.orThrow()` | connection-layer helper; the driver never builds op-failures itself. |

### No raw `IOException` escapes the drivers (audited)
- **Android:** `AndroidDriver`/`AndroidAppFiles` have zero bare `IOException`; the CDP client's `IOException` is contained (degrades internally, rethrows device deaths typed first).
- **iOS:** `XCTestDriverClient.transportCall` previously latched only `SocketTimeoutException`, so a runner that *crashed* (connection refused / EOF / stream reset) leaked a raw `IOException` past `runDeviceCall`. Now it latches **any transport `IOException`** as the typed `Unreachable → DeviceUnreachableException`. The unused `NetworkException : IOException` class was deleted.

---

## Supporting changes

**Failure classification**
- gRPC plane: a transport-death status (`UNAVAILABLE`/`DEADLINE_EXCEEDED`) probes adbd to disambiguate `DeviceServerDied` (server died, adbd alive) vs `DeviceUnreachable` (device gone); a server-answered failure is a `DeviceResponse.Failure` → `DeviceCallFailedException`.
- dadb plane: an `AdbException` is the adb transport dying → **always `DeviceUnreachable`**, never `DeviceServerDied`. `AdbAuthException` → `DeviceAuthException`.

**Connection API — no raw adb stream leaves the boundary**
- `startInstrumentation(command): InstrumentationSession`, `isDriverReachable(port)`, `execDetached(command)` — `AndroidDriver` no longer imports `AdbShellStream`/`AdbShellPacket`/`AdbStream`.
- Provision by inputs: `open(host, port, driverHostPort, connectTimeoutMs, socketTimeoutMs, keepAlive)`; `wrap(...)` is private; transient discovery connections are `.use{}`-closed.
- `isShutdown() == (state == DEAD)` — single source of truth.

**Cross-platform unification**
- One `maestro.DeviceUnreachableException` replaces the old Android/iOS split (field `operation`, optional `DeviceDiagnostics`); the new `DeviceConnectionException` base spans both platforms.

## Tests
- `AndroidDeviceConnectionTest` (38): two-axis contract — dadb-plane death is always `DeviceUnreachable` even when adbd probes alive; gRPC-plane probes; DEADLINE-as-death guard; auth-by-type; op-failures are **not** `IOException`; instrumentation/`isShutdown`/`close`/`isDriverReachable`.
- `IOSDriverTest` (4), iOS module tests — shared `DeviceUnreachableException`.
- `IntegrationTest` (157): a transport death is **raised** out of `runFlow` and **never** routed through `onCommandFailed`; full flow/teardown/onComplete semantics preserved.

## Follow-up (copilot)
`maestro-worker` (classifies `DeviceConnectionException` → INFRA, op-failures → `TestFailure`) and `studio-server` (consumes the typed exceptions; stops deriving disconnection from raw `IOException`) are bumped against this branch and verified green in draft PRs.


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -203,11 +204,19 @@ class Orchestra(
...
-                        config = config,
-                        shouldReinitJsEngine = false,
-                    )
+                    try {
+                        executeCommands(
+                            commands = it,
+                            config = config,
+                            shouldReinitJsEngine = false,
+                        )
+                    } catch (e: DeviceConnectionException) {
```
pedro18x, line 213: A device death during teardown of a passing flow is misclassified as a test failure. When the main flow fails this is handled right (the original exception rethrows and the teardown death is swallowed so it doesn't mask it). But when the flow passes and the device dies during onFlowComplete, this catch sets onCompleteSuccess to false, exception is null so nothing rethrows, and the method returns false, which reads downstream as a test failure rather than retryable infra. Fix: when there's no in-flight exception, let the teardown DeviceConnectionException surface as infra (route it through the existing rethrow so jsEngine.close() still runs) instead of returning false. Worth a test for the passing-flow + teardown-death case since nothing covers it. I'll fix this.

maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -542,33 +487,35 @@ class AndroidDriver(
...
+                            "screen recording. Try using a different Android emulator (eg. Pixel 5 / API 30): ${e.message}"
                     )
                 }
             }, Executors.newSingleThreadExecutor())
 
             object : ScreenRecording {
                 override fun close() {
-                    dadb.shell("killall -INT screenrecord") // Ignore exit code
+                    connection.shell("killall -INT screenrecord") // Ignore exit code
                     future.get()
```
pedro18x: A transport death during screenrecord is misclassified. The async task correctly lets the death propagate (only AndroidOperationFailedException is caught), but future.get() wraps it in ExecutionException, so Orchestra's catch(DeviceConnectionException) misses it and it routes through onCommandFailed as a test failure. Fix: unwrap the ExecutionException cause in close() before it propagates, the same way DadbChromeDevToolsClient.makeSingleWebsocketRequest already does. I'll fix this.

pedro18x (approved):
LGTM. Went through this against the dadb 2.0.0 contract end to end. Everything looks consistent. Suites are green. The two edges I flagged inline are addressed.

Approving. Nice work

PR #3376, opened 2026-06-23
E2E speed improvements

Fishbowler:
## Proposed changes

Speeds up the E2E pipeline and hardens it against transient Android emulator flakiness. Saves 10+m on the critical path for PRs.

This change reworks the `test-e2e.yaml` GitHub Actions workflow so the driver builds only run when their sources change, removes redundant build/upload work, enables Gradle's build cache, fixes Windows Gradle caching, vendors an external setup script, and makes the Android emulator setup resilient to transient adb disconnects. It also removes a set of unused e2e test files.

### Changes
**Workflow Logic (.github/workflows/test-e2e.yaml):**
- Added a new `detect-changes` job that uses `dorny/paths-filter@v4` to detect modifications in Android-related paths (`maestro-android/**`, `maestro-client/src/main/resources/maestro-app.apk`, `maestro-client/src/main/resources/maestro-server.apk`) and iOS-related paths (`maestro-ios-xctest-runner/**`, `maestro-ios-driver/src/main/resources/**`). On manual workflow dispatch it forces both rebuilds. Paths mirror `update-drivers.yaml`'s filters.
- The "Build Android drivers" step is now conditional, executing only if `detect-changes` outputs `android != 'false'`.
- The iOS runner build is now conditional: `:maestro-cli:distZip` is passed `-x :maestro-ios-driver:buildIosDriver` when `ios == 'false'`, so it bundles the committed runner zips instead of rebuilding. Any other state (changed/dispatch/detection failed) rebuilds.
- Commented out the standalone "Build iOS drivers" step — it is redundant, as `:maestro-cli:distZip` already runs `:maestro-ios-driver:buildIosDriver` (the same xcodebuild) via `processResources`.
- Removed the "Upload build/Products to artifacts" step (its only consumer is the disabled `test-ios-xctest-runner` job, and it was already finding no files).
- Replaced the `git clone` of `mobile-dev-inc/bartek-scripts` with a call to a vendored copy of `install_android_sdk`, avoiding an external repo clone on every run.
- Added a Windows-only "Stop Gradle daemon" step before the cache save. The daemon holds open handles on `~/.gradle/caches/*.lock`, which made `setup-java`'s post-step cache `tar` fail on Windows (`Device or resource busy`); stopping it lets the Gradle cache save/restore.

**Build Configuration (gradle.properties):**
- Added `org.gradle.caching=true` to enable Gradle's local build cache. `setup-java` has a cache that'll improve subsequent rebuilds. We can expand on this with caches in the main branch.

**CI Scripts (.github/scripts/):**
- Vendored `install_android_sdk` from `bartekpacia/scripts` (previously cloned at runtime) — a pinned, in-tree copy with a provenance header.
- `setup_android_device.sh`: wrapped `adb root`/`adb unroot` in a retry helper (waiting for the device between attempts). These commands restart the on-device adb daemon, so the control connection frequently drops on the first attempt and aborted the Android job under `set -e`.

**File Deletions:**
- Removed all `e2e/workspaces/wikipedia/wikipedia-android-advanced/` test files and associated scripts, including flow definitions (`.yml`) and JavaScript scripts (`.js`).

### Behavioral changes:
- Android driver builds are skipped unless the Android driver source or its committed APKs change (or on manual dispatch); otherwise the committed APKs are used.
- iOS runner builds are skipped unless the runner source/resources change (or on manual dispatch); otherwise the committed runner zips are used.
- A contributor-supplied driver binary is always overwritten by a CI rebuild when its path changes — we never trust the committed binary as-is.
- Transient `adb root`/`adb unroot` failures during emulator setup are retried instead of failing the job.
- The Windows build's Gradle cache now persists between runs.

### Performance implications:
- Significant reduction in build-job time: the macOS gate dropped from ~13m to ~3m on warm caches (conditional Android + iOS driver builds, no duplicate iOS compile, Gradle build cache, and a working Windows cache restore).
- Removes an external repository clone from every Android run.

## Testing

CI only. Validated across several runs: the macOS build dropped ~13m → ~3m, the Windows Gradle cache now saves and restores, and `Test on iOS`/`Test on Android` stay green (the iOS job runs on the committed runner zips with no driver build).


pedro18x (approved):
Looks good

PR #3377, opened 2026-06-24
fix(orchestra): launchApp/setPermissions/clearState should not swallow exceptions

amanjeetsingh150:
## Problem

1. launchApp/clearState/setPermissions also swallows all the exception and re-wrap everything as test error even in case of unreachable device that should simply allow escaping.
3. UnicodeErrors were retried on our cloud infra unnecessarily, I've right now treated them as test error to prevent them from being treated as infra issue.

## Change

- **`launchApp`/`clearState`/`setPermissions` become raw calls** — no try/catch. Their wraps swallowed `DeviceConnectionException` + `CancellationException` and re-threw them as `UnableTo{LaunchApp,ClearState,SetPermissions}`. The two unused types are deleted; `UnableToLaunchApp` stays (used by launch-arg validation).
- **`UnicodeNotSupported` moved into the `MaestroException` hierarchy** (it was a bare `RuntimeException`) so a genuine "device can't type this" failure is attributed and classified `TEST_ERROR`, not propagated as infra.

Behavioral note: genuine launch/clear/permission op-failures now surface with the raw driver failures and are classified at the consumers rather than wrapped here — the friendly message belongs in the typed exception at the source (follow-up).

## Tests

New `IntegrationTest` cases (all RED before, GREEN after):
- unexpected `RuntimeException` from a command → propagates, never routed through `onCommandFailed`
- device death during `launchApp` → escapes as `DeviceUnreachableException`, not wrapped as `UnableToLaunchApp`
- unsupported-unicode input → is a `MaestroException` routed through `onCommandFailed`

Verify:
- `./gradlew :maestro-test:test` — full suite green
- `./gradlew :maestro-cli:compileKotlin :maestro-client:compileKotlin :maestro-orchestra:compileKotlin` — green

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -301,8 +301,6 @@ class Orchestra(
                 } catch (e: CancellationException) {
                     throw e
                 } catch (e: MaestroException) {
-                    // onCommanFailed is a place where clients collect artifacts like screenshot, logs etc.
-                    // should be called only when commands failed due to test errors
                     logger.error("[Command execution] CommandFailed: ${e.message}")
                     val errorResolution = onCommandFailed(index, command, e)
                     when (errorResolution) {
@@ -1129,10 +1127,6 @@ class Orchestra(
     }
 
     private suspend fun launchAppCommand(command: LaunchAppCommand): Boolean {
-        // No try/catch: these are device operations. A real failure surfaces as a typed device
-        // exception — DeviceConnectionException (infra, escapes for the worker to retry) or an
-        // operation-failure (a flow failure). Wrapping them here would dress infra as a customer
-        // test error and swallow CancellationException. Let them propagate; the worker classifies.
         if (command.clearKeychain == true) {
             maestro.clearKeychain()
         }
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -42,7 +42,6 @@ import maestro.ai.CloudAIPredictionEngine
 import maestro.ai.AIPredictionEngine
 import maestro.js.GraalJsEngine
 import maestro.js.JsEngine
-import maestro.orchestra.error.UnicodeNotSupportedError
 import maestro.orchestra.filter.FilterWithDescription
 import maestro.orchestra.filter.TraitFilters
 import maestro.orchestra.geo.Traveller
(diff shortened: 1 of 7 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -301,15 +300,7 @@ class Orchestra(
...
-                } catch (e: DeviceConnectionException) {
-                    // NOTE: a transport death is retryable INFRA, not a command failure — raise it so it skips
-                    // the command-failure path below. onCommandFailed marks the command FAILED, captures a
-                    // failure screenshot, and may CONTINUE; routing a dead device through that would report it
-                    // as a customer test failure. This is the ONE place Orchestra knows a device exception —
-                    // every other error (MaestroException AND non-Maestro command errors like
-                    // UnicodeNotSupportedError) still flows through onCommandFailed and keeps its attribution.
-                    throw e
-                } catch (e: Throwable) {
+                } catch (e: MaestroException) {
```
proksh: Heads up - I think this narrowing accidentally re-creates the infra-retry waste we're trying to kill. JsEvaluationException (from `GraalJsEngine.evaluateScript` and `command.evaluateScripts(jsEngine)`) extends plain `RuntimeException`, not `MaestroException`. 

So a customer's broken JS - a runScript/evalScript error, or a typo'd inline `${...}` in any command - used to go through `onCommandFailed` as a test error, but now escapes here, gets mapped to Unexpected → `INFRA_ERROR`, and retries 3× on our infra. Same case we handled for unicode.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -1136,47 +1127,28 @@ class Orchestra(
...
-            if (command.clearKeychain == true) {
-                maestro.clearKeychain()
-            }
-            if (command.clearState == true) {
-                maestro.clearAppState(command.appId)
-            }
-        } catch (e: Exception) {
-            logger.error("Failed to clear state", e)
-            throw MaestroException.UnableToClearState("Unable to clear state for app ${command.appId}: ${e.message}", e)
+        if (command.clearKeychain == true) {
```
proksh, line 1132: I know you flagged the friendly message as a follow-up, but now a genuine launch/clear/permission failure surfaces with the raw driver message instead of the readable "Unable to launch app ..." one users got before. Since it's a user-facing regression, might be worth folding the typed message into this PR rather than deferring. 

Leaving it up to you.

proksh (approved):

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
amanjeetsingh150: Good point, I'll minimize the behaviour change like we discussed.

PR #3378, opened 2026-06-24
feat(orchestra): per-execution commands.json with nesting depth

proksh:
**Stacked on #3343** (feat/orchestra-artifact-manifest).

Makes `commands.json` a faithful **per-execution** ledger so a consumer (the worker's `run_step` table) can drive itself off it — including retry/repeat attempts, each with its own screenshot and nesting depth.

## Problem
`commands.json` was serialized from an identity-keyed map (`FlowDebugOutput.commands`), one entry per command **object**. `retry`/`repeat` re-run the *same* objects, so a retried command collapsed into **one** entry with last-attempt status and all attempts' screenshots piled on. No per-attempt step, no nesting info — so the worker couldn't reconstruct its `run_step(index, subIndex, screenshot)` rows from it.

## What this does
Records **one entry per command execution**:

- **`FlowDebugOutput.executedSteps`** — one `CommandDebugMetadata` per `onCommandStart`, in execution order. The identity map stays for live in-flight attribution (metadata updates, the interactive CLI); the two share metadata objects so they can't drift. `commands.json` now serializes `executedSteps`.
- **Per-sequence artifact attribution** — `ArtifactCollector` records carry the executing `sequenceNumber`; `artifactsForStep(seq)` attributes each attempt's screenshot/hierarchy to that attempt alone (was by command identity).
- **Depth** — `OrchestraListener.onCommandStart` carries `depth`; Orchestra tracks a `subflowDepth` around `executeSubflowCommands`. `depth` is 0 at the top of the flow and +1 inside each `runFlow`/`repeat`/`retry` — i.e. the worker's `subIndex`.
- **`onCommandReset` no longer mutates the finished entry** — a repeat/retry reset shares the metadata object with the identity map; mutating it downgraded the execution that just finished. The reset now just precedes the next execution's fresh entry.

## Result
A `repeat: { times: 3 }` over one leaf → the leaf is **3 separate entries** (seq 4/5/6, `depth 1`, each `COMPLETED`, each with its own `screenshots/step-N.png`), the repeat parent its own `depth 0` entry. Maps 1:1 onto `run_step(index ⇐ sequenceNumber, subIndex ⇐ depth, screenshot)`.

**Live updates unchanged** — `commands.json` is the end-of-run reconciliation truth; live status still flows from the Orchestra callbacks + the per-step files (which land synchronously in `onCommandFinished`). This PR doesn't touch the live path.

## Contract change
`commands.json` grows from "per command" → "per execution" and each entry's `metadata` gains `depth`. Nothing consumes it per-attempt today, so this is the intended break; the CLI HTML report reads the same wrapper fields and is unaffected in shape.

## Tested
- Orchestra-driven unit tests: nested commands dispatch `depth +1` (top-level + hooks stay 0); a re-run command yields one entry per execution, each with its own screenshot, statuses intact across the reset.
- Existing `ArtifactsGeneratorTest` / `OrchestraListenerDispatchTest` / `TestOutputWriterTest` updated for the new signature + serialization. Full orchestra suite green (320).
- **Local e2e (web)** — `repeat` and `retry` both produce per-execution `commands.json` with correct depth, status, and one screenshot per iteration/attempt.

## Follow-up (worker, separate)
Once this lands and the worker bumps its Maestro dep, it reads `commands.json` as an ordered execution list (`sequenceNumber → index`, `depth → subIndex`, `artifacts[SCREENSHOT] → screenshot`) and drops its own `IndexTracker`/`compositeStack`.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/runner/CliConsoleListener.kt
+++ b/maestro-cli/src/main/java/maestro/cli/runner/CliConsoleListener.kt
@@ -10,7 +10,7 @@ class CliConsoleListener(private val shardPrefix: String = "") : OrchestraListen
 
     private val logger = LoggerFactory.getLogger(CliConsoleListener::class.java)
 
-    override fun onCommandStart(cmd: MaestroCommand, sequenceNumber: Int) {
+    override fun onCommandStart(cmd: MaestroCommand, sequenceNumber: Int, depth: Int) {
         logger.info("${shardPrefix}${cmd.description()} RUNNING")
     }
 
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -177,6 +177,9 @@ class Orchestra(
 
     private var commandSequenceCounter: Int = 0
 
+    // Dispatched to listeners as `depth`: 0 at the flow top, bumped inside each subflow.
+    private var subflowDepth: Int = 0
+
     // Keyed by sequence number, not MaestroCommand: the latter has structural
     // equality, so two identical commands would collide as map keys.
     private val commandStartTimes = mutableMapOf<Int, Long>()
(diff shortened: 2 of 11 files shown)


pedro18x (approved):
Approving. The per-execution rework holds up: attribution stays correct
since a command always finishes before it restarts, and depth tracking is
balanced. Good coverage on the re-run and nesting cases.

Minor: artifactsFor(command) has no caller now that onFlowEnd uses
artifactsForStep, so it can go with its test.

PR #3384, opened 2026-06-25
fix(device): region-qualified AppleLanguages tag via IosLocale.bcp47Tag

fabianoyoschitaki:
## Problem

iOS simulators set the `AppleLanguages` preference from `IosLocale.languageCode`, which
strips the region (`en_GB` → `en`). iOS then loads the app's **base (US)** `.lproj` instead
of `en-GB.lproj`, so region-specific tests render the wrong strings: an `en_GB` run asserts
**"favourite"** but the app shows **"favorite"**.

`AppleLanguages` decides *which* `.lproj` the app loads, so it needs the hyphenated BCP-47
tag (`en-GB`). `AppleLocale` was already correct (`en_GB`, underscore) but only drives
date/number/currency formatting, not language.

## Fix

Add `IosLocale.bcp47Tag` — a documented derived property that renders the locale as the
hyphenated BCP-47 tag — and route `DeviceService.startDevice`'s `setDeviceLanguage` through
it.

- `languageCode` is left untouched: it's a `DeviceLocale` interface contract meaning "bare
  ISO language subtag" (`en`), honoured by `AndroidLocale`/`WebLocale` and still consumed by
  the device-list path. `bcp47Tag` (`en-GB`) and `languageCode` (`en`) are different
  concepts, so they get different names rather than overloading one.
- `code` (underscore) still drives `AppleLocale`.

## Tests

`DeviceLocaleTest` pins the contract across **every** `IosLocale` entry:
- no underscore leaks through; `en_GB` → `en-GB`; `languageCode` stays `en`
- already-hyphenated entries pass through unchanged (`pt-BR`, `zh-Hans`, `es-419`)
- the matrix carries no ICU keyword modifiers (the assumption the property documents)

`./gradlew :maestro-client:test --tests "maestro.locale.DeviceLocaleTest"` → 18 tests, 0
failures.

## Note for downstream (copilot)

This is the domain home for the fix. The copilot golden-sim path (`maestro-device/SimManager`)
has the same defect and is fixed in a follow-up copilot PR that bumps this submodule and routes
its `AppleLanguages` write through `bcp47Tag` too. The copilot PR must land **after** this one.


Fishbowler (approved):

steviec (approved):
This workaround is fine for now. I think we should improve the whole Locale system at some point.

PR #3389, opened 2026-06-26
Fix disableAnsi=false, and cantralise logging config

### LINKED ISSUE #1808: Unify logging setup across commands
Currently, some commands are either (1) quiet and log to `maestro.log`, or (2) are loud (i.e. print all `INFO` logs and above) and do not log to their `maestro.log`.

That's because logger setup is done on per-command basis, instead once per whole program.

<details><summary>More details</summary>

1. `LogConfig.configure()` is only called by `TestDebugReporter.install()`
2. `TestDebugReporter.install()` has 4 callers (`TestCommand`, `StudioCommand`, `RecordCommand`, `StartDeviceCommand`. So it seems to me that TestDebugReporter`` is only for usage in commands that actually run the test flow.
3. But then the problem with the flood of logs appear in other commands. For example: maestro hierarchy spits out a log of logs because it doesn't call `TestDebugReporter.install()` (but should it? it's actually not reporting output of any test)

I would say that `LogConfig.configure` should be done somewhere in `main()`, so that all commands get the same logger setup.

</details> 

Thanks @igorsmotto

Fishbowler:
## Proposed changes

LogConfig's console pattern was slightly wrong, so --verbose runs printed a stray cyan `disableAnsi=false` token on every log line. Affected all verbose-heavy commands (test, record, cloud, hierarchy). It was supposed to be the log level that was cyan, and it wasn't.

Before
```
disableAnsi=false([DEBUG]) Using SLF4J as the default logging framework
```

After
```
[DEBUG] Using SLF4J as the default logging framework
```

Also, logging was configured per-command via TestDebugReporter.install()
Commands that didn't call it fell back to log4j2's default config and silently ignored the global `--verbose` flag. Now main() sets a console-only baseline before any init logging fires. Commands wanting a maestro.log still opt in via install(), and the explicit silencing in LoginCommand/McpServer is unaffected. This doesn't materially affect today's behaviour, but adds consistency, and closes #1808 (all of the symptoms of which had been done piecemeal since that was logged)

## Testing

Local CLI

## Issues fixed

Fixes #1808


pedro18x (approved):
nice

PR #3396, opened 2026-06-30
fix(web): fail fast without DevTools and drop empty recording placeholders

proksh:
## Problem

`WebScreenRecorder` cast `seleniumDriver as HasDevTools` unsafely. A web driver that does not advertise CDP (e.g. a remote Browserbase `RemoteWebDriver` without `se:cdp`) is **not** `HasDevTools`, so the cast threw a raw `ClassCastException` mid-stream — *after* the recording sink had already been opened. The result: a **0-byte `screen-recording.mp4` plus a manifest entry**, i.e. a broken placeholder rather than a real recording.

Core already uses the safe `as?` cast elsewhere (`WebDriver.open()`), so it anticipates non-DevTools drivers; the recorder just hadn't been brought in line.

## Fix

- **`WebScreenRecorder`** — uses the safe `as?` cast (matching `WebDriver`'s existing pattern) and fails fast with a clear `UnsupportedOperationException` *before* touching the output sink, so nothing is written when recording can't work.
- **`ArtifactsGenerator`** — deletes the recording file when it is still empty after stop. The `ArtifactCollector` already drops records whose file is gone, so no 0-byte placeholder surfaces in the manifest.

This does **not** add web recording for non-CDP drivers — it only makes that case fail cleanly instead of leaving a broken artifact behind. CDP-capable drivers (local ChromeDriver) are unaffected and still record.

## Testing

- **Unit tests**
  - `WebDriverTest` — driver without `HasDevTools` → `UnsupportedOperationException` (not `ClassCastException`).
  - `ArtifactsGeneratorTest` — empty/failed recording → no 0-byte file and no `SCREEN_RECORDING` manifest entry.
- **Local e2e** — ran a 9-interaction web flow (`saucedemo.com`) with `--analyze`; local ChromeDriver is DevTools-capable, so it produced a real 3.4 MB h264 mp4 with a valid manifest entry, confirming the happy path is unaffected.

### DIFF
--- a/maestro-client/src/test/java/maestro/drivers/WebDriverTest.kt
+++ b/maestro-client/src/test/java/maestro/drivers/WebDriverTest.kt
@@ -6,6 +6,7 @@ import io.mockk.mockk
 import maestro.web.selenium.SeleniumFactory
 import okio.blackholeSink
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
 import org.openqa.selenium.JavascriptExecutor
 import org.openqa.selenium.StaleElementReferenceException
 import org.openqa.selenium.WebElement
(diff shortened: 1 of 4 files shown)


Fishbowler (approved):
This looks sensible.

Fishbowler:
Can I interest you in a related web stability PR? 
#3392

PR #3397, opened 2026-06-30
feat(mcp): add describe_cloud_run tool (MA-4050)

fabianoyoschitaki:
## What

Adds the `describe_cloud_run` MCP tool: given a `run_id`, it returns a finished Maestro Cloud run's metadata (status, failure reason, device spec, timing) and its artifacts (screen recording, device logs, view hierarchy) as short-lived signed URLs — so an agent can inspect why a run failed.

## Changes

- **`ApiClient.describeRun(authToken, runId, includeArchive)`** → `GET /v2/runs/{runId}` (appends `?includeArchive=true` when set), plus `RunDetails` / `RunDeviceSpec` / `RunArtifact` DTOs. Enum-like fields (`status`, `failureReason`, artifact `type`/`format`) are modelled as `String` on purpose, so a newer server enum value never breaks an older CLI. The URL is built via `HttpUrl.Builder().addPathSegment(runId)` so the (agent-supplied) `run_id` is percent-encoded — no `../` path traversal or `?`/`&` query injection.
- **`DescribeCloudRunTool`**: a required `run_id` + an optional `include_archive` boolean. Logic is split into an `internal fun handle()` (mirrors `RunTool`) with hoisted, unit-tested `buildRunJson` and `errorMessageForStatus` helpers. Distinct 401 / 404 / 409 messages (the 409 points the agent back to `get_cloud_run_status` to poll until the run is terminal).
- **`McpServer`**: registered in `addTools(...)` + a terse `INSTRUCTIONS` mention (kept under the 2 KB cap).
- **Tests**: `McpServerTest` (registration + cap), `DescribeCloudRunToolTest` (flat-artifact output, the 409→`get_cloud_run_status` mapping, missing-`run_id` validation), and a no-device auth-required eval in `tool-tests-without-device.yaml`.

## Behavior

Every entry in `artifacts[]` is a directly-downloadable signed URL. The whole-run zip is opt-in via `include_archive`: when set, the server builds + signs it and returns it as a normal, direct-url `artifactsArchive` artifact. So the default response is cheap, and "give me everything (including screenshots)" is one explicit flag — every URL the agent gets back is fetchable, with no second call.

## Notes

- Input is the per-flow **`run_id`** (from a dashboard run URL), **not** the `upload_id` from `run_on_cloud`.
- Screenshots are reachable via `include_archive` (they live inside the whole-run zip).
- The tool is additive — an older CLI simply ignores it.

## Testing

- `./gradlew :maestro-cli:test --tests "maestro.cli.mcp.McpServerTest" --tests "maestro.cli.mcp.tools.DescribeCloudRunToolTest"` → green.
- End-to-end against a local mock of `GET /v2/runs/{runId}`: the default returns the flat artifact list; `include_archive` appends the `artifactsArchive`; 401 / 404 / 409 each return their distinct message; a `../` `run_id` stays encoded under `/v2/runs/…` (no traversal).


maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
```diff
@@ -0,0 +1,136 @@
...
+
+                val run = try {
+                    client.describeRun(apiKey, runId, includeArchive)
+                } catch (e: ApiClient.ApiException) {
+                    return@RegisteredTool when (e.statusCode) {
+                        null -> errorResult(
+                            "Could not reach Maestro Cloud to describe run_id=$runId. " +
+                                "Check your network connection and retry."
+                        )
+                        401 -> errorResult(
```
proksh: Nit: this string is identical to the upfront auth-check message. Fine to leave, but could pull into a private const val to keep them in sync.

proksh (approved):
LGTM. Nice call on the String enum-like fields and `toHttpUrl` encoding of `run_id`. One tiny nit below.

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
fabianoyoschitaki: Thanks! Pulled it into a private `NOT_AUTHENTICATED_MESSAGE` const (2fc3d7b1). Besides that I also extracted the status↔message mapping into a separate `errorMessageForStatus` (faefca6a) and split the handler out of `create()` into `handle()` — mirroring RunTool — so the logic is unit-tested (12944fb3).

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
```diff
@@ -0,0 +1,128 @@
...
+    fun create(): RegisteredTool = RegisteredTool(
+        Tool(
+            name = "describe_cloud_run",
+            description = "Fetch metadata and artifacts for a single Maestro Cloud run by its run_id. " +
+                "Returns run status, failure reason, device spec, timing, and `artifacts` — the run's individually-" +
+                "stored files, each with a directly-downloadable signed `url` (e.g. screen recording and device logs). " +
+                "Set `include_archive` to additionally get `artifactsArchive`: a single zip of the ENTIRE run — " +
+                "everything, including the screenshots and view hierarchy that are not in the individual `artifacts` " +
+                "list — as a direct url; omit it for a faster response. " +
+                "When you present the result, make both options clear to the user: they can download the individual " +
```
pedro18x: These URLs expire and the response never says so, with very different TTLs: the backend signs individual files for 7 days but the artifactsArchive zip for only 15 minutes (INDIVIDUAL_TTL in RunArtifacts.kt vs the 15-minute artifactsZipSignedUrl in SignedArtifacts.kt). Since this sentence tells the agent to hand both links to the user, the archive link will often be dead by the time a human clicks it, while the file links next to it keep working for a week. One sentence here covers it, something like: "URLs expire: individual files after 7 days, the archive after ~15 minutes. Call the tool again to get a fresh link."

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
```diff
@@ -0,0 +1,128 @@
...
+    }.toString()
+
+    /** Maps an ApiClient failure to a distinct, actionable message for the agent (null status = network/IO). */
+    internal fun errorMessageForStatus(statusCode: Int?, runId: String): String = when (statusCode) {
+        null -> "Could not reach Maestro Cloud to describe run_id=$runId. Check your network connection and retry."
+        401 -> NOT_AUTHENTICATED_MESSAGE
+        404 -> "No run found with run_id=$runId. Check the run_id (it is the per-flow run id from a dashboard " +
+            "run URL, not the upload_id from run_on_cloud) and that it belongs to your organization."
+        409 -> "Run run_id=$runId is still in progress, so its artifacts are not ready yet. " +
+            "Use the get_cloud_run_status tool to poll until the run reaches a terminal state " +
```
pedro18x: An agent that hits this 409 usually can't follow the advice: get_cloud_run_status requires both upload_id and project_id (its schema marks both required), and neither is derivable from a run_id. In this tool's documented input flow (run_id from a dashboard URL) the agent holds neither. Suggest something like: "Run run_id=$runId is still in progress, so its artifacts are not ready yet. Retry once the run finishes. If you started the run with run_on_cloud, poll get_cloud_run_status with its upload_id and project_id until the run reaches a terminal state, then retry describe_cloud_run."

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
```diff
@@ -0,0 +1,128 @@
...
+
+object DescribeCloudRunTool {
+    fun create(): RegisteredTool = RegisteredTool(
+        Tool(
+            name = "describe_cloud_run",
+            description = "Fetch metadata and artifacts for a single Maestro Cloud run by its run_id. " +
+                "Returns run status, failure reason, device spec, timing, and `artifacts` — the run's individually-" +
+                "stored files, each with a directly-downloadable signed `url` (e.g. screen recording and device logs). " +
+                "Set `include_archive` to additionally get `artifactsArchive`: a single zip of the ENTIRE run — " +
+                "everything, including the screenshots and view hierarchy that are not in the individual `artifacts` " +
```
pedro18x: "Everything, including the view hierarchy" is only true for runs with failed or warned steps: core writes screen-hierarchy/step-N.json only for Failed/Warned outcomes, passing steps never capture it even with captureFullArtifacts on (ArtifactsGenerator.kt lines 129-130). So for a SUCCESS run the archive contains no hierarchy and the agent will promise the user something that isn't in the zip. Suggest qualifying it here and in the two places that repeat it (the include_archive description at line 34 and the presentation sentence at line 22), e.g. "and, for failed/warned steps, the view hierarchy".

pedro18x (commented):
Took a deeper pass since this is the consumer end of the artifact contract. Verified against the merged backend (copilot#2584 is on main and live in prod), built the branch on latest main locally (tests green), and drove the built binary over MCP stdio against a mock of the backend response shape: the wire contract checks out field for field, including the run_id encoding and all three error mappings. Three comments below, all at the wording/UX level; for this tool the descriptions are the API an LLM consumes, so they're worth getting right.

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
fabianoyoschitaki: Good catch,  the 7d/15m asymmetry would hand users a dead archive link. Added a line: individual files last ~7 days, the archive only ~15 min, so the agent should fetch the archive promptly rather than hand its link to the user. Verified against a mock: the agent now downloads the archive immediately.

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
fabianoyoschitaki: Reworded so the primary advice is always followable (retry once the run finishes), with the poll path made 
conditional on having started the run via run_on_cloud, confirmed the agent no longer dead-ends on a 409.

maestro-cli/src/main/java/maestro/cli/mcp/tools/DescribeCloudRunTool.kt
fabianoyoschitaki: Good catch, qualified view hierarchy as "for failed/warned steps" in the three spots (description, presentation sentence, and the include_archive param)  Verified the agent no longer promises a hierarchy for a SUCCESS run.

pedro18x (approved):
Re-verified the follow-up the same way as the first pass: rebuilt the branch on latest main, MCP tests green, and my stdio E2E harness against a mock of the backend response passes 17/17, including the reworded 409. All three comments are addressed in substance, and the new wording checks out against the backend: both TTLs match the signing code, the 409's primary advice is followable with just a run_id with the poll path correctly conditional on run_on_cloud, and the failed/warned hierarchy qualifier is accurate and in all three spots.

One non-blocking nit you can take or leave: "fetch the archive right away instead of handing its link to the user" slightly conflicts with the presentation sentence that offers the archive to the user as a download option, and some MCP hosts can only relay links, not fetch files. If you touch it again, something like "the archive link expires in ~15 minutes, so request it (include_archive=true) when it's needed and hand it over fresh" resolves both. Fine to merge as is.

Nice work on this one, the tool came out tight.

PR #3399, opened 2026-07-01
fix(logs): capture maestro.log when host sets additivity=false

proksh:
`ScopedLogCapture` attached its file appender to the **root** logger and relied on `maestro.*` events propagating up. Hosts that scope the maestro logger with `additivity="false"` (the Maestro Worker's `log4j2.xml` does this) stop that propagation, so the appender captured nothing and the per-flow `maestro.log` was written **empty**.

**Fix:** attach the appender to the `maestro`/`MAESTRO` loggers directly, so capture is independent of the host's additivity. Setting their level to `ALL` is idempotent + monotonic, so the once-per-JVM guard is dropped.

Caught while testing run artifacts — `maestro.log` came back 0 bytes and was skipped on upload.

Test: new case reproducing `additivity="false"` (fails on old code, passes now).


pedro18x (approved):
Looks good 👍

PR #3401, opened 2026-07-02
fix(artifacts): don't capture view hierarchy on passing steps

proksh:
## What

Under `captureFullArtifacts`, every passing step captured a full `viewHierarchy()` — an expensive synchronous device round-trip (~1s each on iOS/Android). On long flows this per-command overhead was enough to push runs past the 15-minute watchdog, causing **fleet-wide test timeouts (production incident)**.

Now only **failed/warned** steps capture the hierarchy (the diagnostically useful ones). Per-step **screenshots** and the **full-run recording** are unchanged.


Fishbowler (approved):
I'd love to see a return of unqualified hierarchy snapshotting, perhaps a config-based opt-in. I think it could be really powerful for LLMs.

proksh:
> I'd love to see a return of unqualified hierarchy snapshotting, perhaps a config-based opt-in. I think it could be really powerful for LLMs.

Agree. I just wanted to move it behind status as it is causing prod incident. We can come back to this.

PR #3402, opened 2026-07-02
Don't block forever on a failed emulator boot

Fishbowler:
## Proposed changes

Found whilst reviewing #3164.

Android emulator boot would wait forever if it failed to make it past boot (either a hang or an emulator crash). This wraps the sleep loop and adds a max time (default 3mins, configurable by a new `MAESTRO_DEVICE_BOOT_TIMEOUT` env var).
 
## Testing

Observe in CI. This is very hard to reproduce locally - it requires that a device boots and get `adb` (see DeviceService L80) but never makes to fully booted. The real risk is that the default isn't enough for slow envs like GHA.


pedro18x (approved):
Clean, nice

PR #3405, opened 2026-07-03
feat(orchestra): report captured step-screenshot paths via callback

pedro18x:
## What

New optional Orchestra constructor callback, `onStepScreenshotCaptured(sequenceNumber, relativePath)`, threaded into `ArtifactsGenerator` and fired with the bundle-relative path (`screenshots/step-3.png`) immediately after a step-screenshot write succeeds, in both the completed/warned path and the failure path. It never fires for skipped commands, failed captures, or any other artifact kind. Default no-op, so every existing Orchestra caller compiles unchanged.

## Why

The consumer is the cloud worker ([MA-4049](https://linear.app/mobile-dev/issue/MA-4049)): it links each run step to core's bundle screenshot. Having the code that wrote the file report the path avoids the worker reconstructing `step-{seq}.png` from its own command counter, an implicit cross-repo lockstep that would silently break if dispatch sites ever change. `CommandMetadata` was deliberately not used: listeners have no channel into it, and the worker's metadata handling attributes to the latest-started index, which mis-attributes composite parents.

## Testing

Six new `ArtifactsGeneratorTest` cases: fires with `(seq, path)` for a completed step under `captureFullArtifacts`; fires for a failed step; fires for a warned step even with `captureFullArtifacts=false` (warned steps always get a screenshot); silent for skipped; silent for a passing step with the flag off; silent when the screenshot write throws. `:maestro-orchestra:test` green (46 tests), `:maestro-cli` compiles against the change.

## Review follow-up (third commit)

An adversarial review pass tightened three things: the callback was renamed `onStepScreenshotCaptured` (it fires only for step screenshots, and the KDoc now documents that the Int is the global sequence number, not the per-list index of the sibling callbacks); the step path now routes through `takeDebugScreenshot` so a mid-write failure deletes the partial file and the callback, bundle, manifest, and commands.json always agree; and consumer exceptions propagate to listener dispatch instead of being mislogged as capture failures. Adds the Orchestra-to-generator pass-through test and a composite-parent no-dedup callback test (47 tests in the file, all green).

## Consumer note (pre-existing, found during this PR's compatibility sweep)

studio-server's `FlowExecutor` (copilot repo, `FlowExecutor.kt:315-317`) still constructs Orchestra with a `screenshotsDir` named argument that core removed in #3282. It compiles today only because studio-server pins an older core; its next maestro submodule move will fail to compile regardless of this PR. Not caused or worsened here (this PR's parameter is defaulted and every construction site was swept: all compile unchanged), but whoever moves studio-server's submodule next should fix that call site first.

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -143,10 +143,6 @@ class Orchestra(
     private val onCommandSkipped: (Int, MaestroCommand) -> Unit = { _, _ -> },
     private val onCommandReset: (MaestroCommand) -> Unit = {},
     private val onCommandMetadataUpdate: (MaestroCommand, CommandMetadata) -> Unit = { _, _ -> },
-    /**
-     * Fires when a step runs and successfully captures its automatic screenshot (not one
-     * explicitly taken by a user command), with the bundle-relative path. Keep it non-blocking.
-     */
     private val onStepScreenshotCaptured: (sequenceNumber: Int, relativePath: String) -> Unit = { _, _ -> },
     private val onCommandGeneratedOutput: (command: Command, defects: List<Defect>, screenshot: Buffer) -> Unit = { _, _, _ -> },
     private val apiKey: String? = null,
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -144,10 +144,8 @@ class Orchestra(
     private val onCommandReset: (MaestroCommand) -> Unit = {},
     private val onCommandMetadataUpdate: (MaestroCommand, CommandMetadata) -> Unit = { _, _ -> },
     /**
-     * Fired after a step screenshot (screenshots/step-{n}.png) lands in the bundle, with its
-     * bundle-relative path. Not fired for skipped commands or failed captures. The Int is the
-     * global command sequence number (one per attempt), not the sibling callbacks' list index.
-     * Runs on the flow thread — keep it non-blocking.
+     * Fires when a step runs and successfully captures its automatic screenshot (not one
+     * explicitly taken by a user command), with the bundle-relative path. Keep it non-blocking.
      */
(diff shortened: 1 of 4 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -143,6 +143,14 @@ class Orchestra(
...
     private val onCommandReset: (MaestroCommand) -> Unit = {},
     private val onCommandMetadataUpdate: (MaestroCommand, CommandMetadata) -> Unit = { _, _ -> },
+    /**
+     * Fired after a step screenshot (screenshots/step-{n}.png, written for both passing and
+     * failing steps) lands in the artifact bundle, with the bundle-relative path. Never fires
+     * for skipped commands, failed captures, or other artifact kinds. The Int is the global
+     * command sequence number that listener dispatch receives (one per attempt, composites
+     * included), not the per-list index the sibling onCommand* callbacks receive.
+     */
+    private val onStepScreenshotCaptured: (sequenceNumber: Int, relativePath: String) -> Unit = { _, _ -> },
```
Fishbowler: Is the commentary excessive?
Feels like, at max, it should be "Fires when a step runs and successfully captures a screenshot, where the screenshot is automatic (not explicitly caused by a user command)"

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
Fishbowler, line 1: There's an enormous overlap between `captureFailureScreenshot` and `captureStepScreenshot` implementations. Should these be using a common helper?

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
proksh: Updated

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
proksh: Yes, make sense. Updated

fabianoyoschitaki (approved):

PR #3410, opened 2026-07-06
fix(android-webview): bound WebView hierarchy augmentation so a stalled devtools endpoint can't hang a command (MA-4119)

proksh:
## Problem

On Android flows that interact with a WebView (e.g. bank/payment portals), a single command can hang for **many minutes** and then fail the whole run with `INFRA_ERROR "Execution Timed Out"`.

`AndroidDriver.contentDescriptor()` augments the native hierarchy by reading the WebView DOM over the Chrome DevTools ADB socket (`getWebViewInfos` → `open: localabstract:webview_devtools_remote_*`). When that endpoint stops responding the socket open/read blocks unboundedly — OkHttp's call timeout doesn't cover the socket-factory open (observed `360007ms since last byte`). Because this runs in the hot `waitForAppToSettle → viewHierarchy` poll path, one command burns 8+ min and cascades into the device server dying.

**Started with #3300** (DummyDns hostname fix): before it, augmentation opened `localabstract:localhost` and the IOException was swallowed (no-op, no hang); after it, it opens the real WebView socket and can hang. Gated by the per-flow opt-in `androidWebViewHierarchy: devtools`.

## Solution

WebView augmentation is best-effort (a stalled WebView can always fall back to the native hierarchy), so **bound it**: run the fetch under a wall-clock timeout (default 10s) and degrade to the native-only hierarchy on expiry. Genuine device deaths surfaced by the fetch still **propagate** (INFRA_ERROR + retry), not masked.

Introduced a small `ChromeDevToolsClient` interface seam so the fetch path is injectable.

## Tests

`AndroidWebViewHierarchyClientTest`:
- **hang → degrade:** a CDP client that never returns → `augmentHierarchy` returns the base hierarchy well within the bound (RED before the fix: blocked past the preemptive-timeout assert).
- **failure → propagate:** a CDP client that throws → the exception propagates unchanged.

### DIFF
--- a/maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
+++ b/maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
@@ -5,26 +5,71 @@ import maestro.TreeNode
 import maestro.UiElement
 import maestro.UiElement.Companion.toUiElementOrNull
 import maestro.android.AndroidDeviceConnection
+import org.slf4j.LoggerFactory
 import java.io.Closeable
+import java.time.Duration
+import java.util.concurrent.Callable
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
 
-class AndroidWebViewHierarchyClient(connection: AndroidDeviceConnection): Closeable {
+class AndroidWebViewHierarchyClient(
+    private val devToolsClient: ChromeDevToolsClient,
+    private val timeout: Duration = DEFAULT_WEBVIEW_AUGMENT_TIMEOUT,
+): Closeable {
 
-    private val devToolsClient = DadbChromeDevToolsClient(connection)
+    constructor(connection: AndroidDeviceConnection) : this(DadbChromeDevToolsClient(connection))
+
+    // Single daemon thread so a hung fetch leaks at most one thread; a later fetch queued behind it
+    // is cancelled out of the queue when its own get() times out (see fetchWebViewNodes).
+    private val augmentExecutor = Executors.newSingleThreadExecutor { r ->
+        Thread(r, "webview-hierarchy-augment").apply { isDaemon = true }
+    }
 
     fun augmentHierarchy(baseHierarchy: TreeNode, chromeDevToolsEnabled: Boolean): TreeNode {
         if (!chromeDevToolsEnabled) return baseHierarchy
         if (!hasWebView(baseHierarchy)) return baseHierarchy
         // TODO: Adapt to handle chrome in the same way
-        val webViewHierarchy = devToolsClient.getWebViewTreeNodes()
+        val webViewHierarchy = fetchWebViewNodes() ?: return baseHierarchy
         val merged = mergeHierarchies(baseHierarchy, webViewHierarchy)
         return merged
     }
 
+    /**
+     * Reads the WebView hierarchy under a wall-clock [timeout]. Returns null when the fetch can't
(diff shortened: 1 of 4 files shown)


maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
```diff
@@ -5,27 +5,62 @@ import maestro.TreeNode
...
         return merged
     }
 
+    // Null on timeout → caller falls back to the native hierarchy. Real fetch failures propagate.
+    private fun fetchWebViewNodes(): List<TreeNode>? {
+        val future = augmentExecutor.submit(Callable { devToolsClient.getWebViewTreeNodes() })
+        return try {
+            future.get(timeout.toMillis(), TimeUnit.MILLISECONDS)
+        } catch (e: TimeoutException) {
+            future.cancel(true)
```
pedro18x, line 45: cancel(true) only sets the interrupt flag; the blocking socket read underneath (AndroidDeviceConnection.shell via dadb) doesn't honor it. So when a fetch genuinely wedges, the exact case this PR is meant to guard against, the single worker thread on augmentExecutor never frees up. Every later augmentHierarchy() call for the rest of the session then queues behind it and pays the full timeout without the task ever running: a one-time graceful degrade turns into a permanent per-command tax. Suggest giving each fetch its own short-lived thread instead of a shared single-thread pool, so one stuck call can't poison the rest of the session.

maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
```diff
@@ -5,27 +5,62 @@ import maestro.TreeNode
...
+            logger.warn("WebView hierarchy augmentation timed out after {}; using native hierarchy only", timeout)
+            null
+        } catch (e: ExecutionException) {
+            throw e.cause ?: e
+        }
+    }
+
     override fun close() {
+        augmentExecutor.shutdownNow()
         devToolsClient.close()
```
pedro18x, line 57: shutdownNow() immediately followed by devToolsClient.close(), with no awaitTermination, races a still-running fetch against tearing down the shared OkHttpClient. This couldn't happen before since the fetch was synchronous. Should wait (with its own short bound) for the interrupted task to actually stop before closing devToolsClient.

pedro18x (commented):
Nice fix for a real pain point: a stalled Chrome DevTools socket shouldn't be able to hang a whole command. Left two comments on the timeout/cleanup mechanics: `future.cancel(true)` can't actually interrupt the blocking socket read it's racing against, so a genuinely wedged fetch ends up degrading every later call in the session rather than just the one that timed out. Details inline, should be a small fix.

maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
proksh: Good catch, fixed. Switched to a per-fetch cached daemon pool so a wedged fetch can't block later ones, and added a test (`augmentHierarchy_oneWedgedFetchDoesNotBlockLaterFetches`) that fails on the old single-thread. Minor nuance: in the cloud worker the dadb read is capped at 120s, so the stuck thread does eventually free — the symptom was a multi-minute degradation window rather than truly permanent, but the fix stands (no cross-fetch coupling, faster recovery).

maestro-client/src/main/java/maestro/android/chromedevtools/AndroidWebViewHierarchyClient.kt
proksh: Fixed.

pedro18x (approved):
Looks good!

PR #3413, opened 2026-07-07
Fix MA-4124: tap after scrollUntilVisible lands on the settled element position (iOS)

pedro18x:
## TL;DR
On iOS, a `tapOn` right after `scrollUntilVisible` could land where the element **used to be**, because the tap was aimed with a view hierarchy captured while the scroll view was still decelerating. When the driver cannot confirm the screen has settled, `tapOnElement` now re-resolves the element from fresh hierarchy fetches until its position is stable, then taps.

Fixes [MA-4124](https://linear.app/mobile-dev/issue/MA-4124/betterment-tap-after-scrolluntilvisible-has-no-effect-on-ios).

## The bug
`scrollUntilVisible` + `tapOn` flows on iOS intermittently produced taps with no effect. The tap was delivered — but at coordinates a few dozen pixels away from where the target element actually ended up.

## Root cause
1. `scrollUntilVisible` returns as soon as the target element is visible — but the scroll view **keeps decelerating from momentum** after the gesture ends.
2. Before tapping, `Maestro.tapOnElement` calls `waitForAppToSettle`. On iOS this runs a screenshot-based static check (`waitUntilScreenIsStatic`): during *slow* deceleration, two consecutive screenshots differ by less than the similarity threshold, so the check **false-positives** ("static") and `IOSDriver.waitForAppToSettle` returns `null`.
3. On `null`, the old code fell back to the **pre-wait hierarchy** — a snapshot taken mid-deceleration — and aimed the tap at the element's stale bounds. By the time the tap was physically delivered, the element had drifted away.

## The fix
When `waitForAppToSettle` returns `null` (driver could not confirm settling), a new `refreshElementUntilStable` loop:

- fetches fresh view hierarchies and re-resolves the target element,
- returns as soon as the element's bounds are **unchanged between two consecutive fresh fetches** (100ms apart),
- keeps polling if the element transiently detaches from the accessibility tree while animating,
- gives up after a **monotonic 3s deadline** (matching `IOSDriver.SCREEN_SETTLE_TIMEOUT_MS`) and falls back to the last known position.

The settled path (Android, and any driver that returns a hierarchy) is completely unchanged.

## Why this shape (alternatives considered and rejected)
- **Compare one fresh fetch against the pre-wait hierarchy** (1 fetch instead of 2): a match against a snapshot that may itself be mid-deceleration is exactly the stale-snapshot trap that caused the bug. Only two *consecutive fresh* fetches count.
- **Reuse `ScreenshotUtils.waitForAppToSettle`**: it compares whole-hierarchy equality — the same class of check that already fails to notice a slowly-drifting screen — and is blocking/wall-clock. The new loop narrows the comparison to the single target element's bounds.
- **Reuse `MaestroTimer.withTimeoutSuspend`**: its deadline is wall-clock (`System.currentTimeMillis`), which can step under NTP adjustment; this loop needs a monotonic deadline (`System.nanoTime`).
- **Fix inside `IOSDriver.waitForAppToSettle`**: the right long-term altitude (the false "static" verdict is the true root cause), but it changes settle timing for *every* iOS command, not just tap — too risky to bundle here. See follow-ups.

## Cost
- **Android / settled path:** zero change.
- **iOS steady-state tap:** ~2 extra hierarchy fetches + 100ms (null is iOS's *normal* settle outcome). This is the floor for confirming stability without trusting a pre-wait snapshot; worst case is bounded by the 3s deadline.

## Testing
- New deterministic regression test: `Case 145 - tap after scrollUntilVisible lands on settled element position (MA-4124)`.
  - `DeceleratingIosFakeDriver` mirrors the iOS failure mode: after a swipe, content keeps drifting a step per screen observation, the settle check "passes" and returns `null`, and drift only completes when the tap is physically delivered.
  - The old code fails this test (tap lands outside the settled bounds); the fixed code passes. No simulator needed.
- Full `:maestro-test` IntegrationTest suite green.

## Explicitly out of scope (follow-ups)
- **Driver-level root cause:** repair the false-static verdict in `IOSDriver.waitForAppToSettle` so *all* commands benefit (behavior change for every iOS command → separate, carefully reviewed PR).
- **Sibling aim paths with the same exposure:** `swipe`-on-element (`Maestro.kt`) and the relative-point tap branch (`Orchestra.kt`) still aim using pre-wait hierarchies. Documented in the new KDoc.
- The previously-implicit `null` contract of `Driver.waitForAppToSettle` ("null = could **not** confirm settling") is now documented on the interface so future callers stop inventing opposite interpretations.

## Files touched
| File | Change |
|---|---|
| `maestro-client/.../Maestro.kt` | `tapOnElement` null-settle branch + `refreshElementUntilStable` + stability constants |
| `maestro-client/.../Driver.kt` | KDoc: null contract of `waitForAppToSettle` |
| `maestro-test/.../FakeDriver.kt` | `waitForAppToSettle` return type aligned with the (nullable) interface |
| `maestro-test/.../IntegrationTest.kt` | Case 145 regression test + `DeceleratingIosFakeDriver` |

### DIFF
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -91,6 +91,15 @@ interface Driver {
 
     fun waitUntilScreenIsStatic(timeoutMs: Long): Boolean
 
+    /**
+     * Waits for the app UI to settle and returns the hierarchy observed once settled.
+     *
+     * A null return means the driver could NOT confirm settling and offers no fresh
+     * hierarchy — it does not mean "settled". For example, the iOS screen-static check can
+     * pass while a scroll view is still slowly decelerating (MA-4124). Callers that aim
+     * gestures at element positions must not trust a pre-wait hierarchy after a null
+     * return; see Maestro.refreshElementUntilStable.
+     */
     fun waitForAppToSettle(initialHierarchy: ViewHierarchy?, appId: String?, timeoutMs: Int? = null): ViewHierarchy?
 
     fun capabilities(): List<Capability>
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -95,7 +95,7 @@ interface Driver {
      * Waits for the app UI to settle and returns the hierarchy observed once settled.
      *
      * A null return means the driver could NOT confirm settling and offers no fresh
-     * hierarchy — it does not mean "settled". For example, the iOS screen-static check can
+     * hierarchy; it does not mean "settled". For example, the iOS screen-static check can
      * pass while a scroll view is still slowly decelerating (MA-4124). Callers that aim
      * gestures at element positions must not trust a pre-wait hierarchy after a null
      * return; see Maestro.refreshElementUntilStable.
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory
 import java.awt.image.BufferedImage
 import java.io.File
 import javax.imageio.ImageIO
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runInterruptible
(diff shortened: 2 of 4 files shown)


proksh (approved):
LGTM — correct root cause, good regression test. Approving.

Two notes before merge:

**1. Cost on the common iOS path.** `IOSDriver.waitForAppToSettle` returns `null` whenever the static check passes — the *normal* idle case, not just after a scroll — so every iOS element tap now pays ~2 hierarchy fetches + 100ms, even on an already-static screen. Worth validating via a canary rollout with a runtime check (measure the added tap latency / fetch count in the field) before assuming the trade-off is free.

**2. Comment volume.** The in-code KDoc is heavier than the repo norm. The alternatives-rejected rationale and follow-up scope notes read better in the PR description than in source — consider condensing the KDoc to the what/why contract.

PR #3416, opened 2026-07-08
fix(android): record through the extended screenrecord entry point when present

pedro18x:
Cloud recordings on Android API <34 stop at exactly 180s no matter what the device golden carries (MA-4121). copilot#2681 fixed the golden side, the binary patcher really was broken, but recordings stayed capped because nothing in the recording path uses what the golden provides: `AndroidDriver.startScreenRecording` shells the stock `screenrecord` binary directly, and only API 34+ supports `--time-limit 0`. The extended entry point that maestro-device bakes at `/data/local/tmp/screenrecord` (patched binary, 3600s limit, encoder-safe `--size`) had zero callers.

This makes the driver probe for that entry point and record through it when present, passing `--bit-rate` through unchanged. Stock devices keep the exact current behavior, so local `maestro record` users are unaffected. Teardown now SIGINTs both `screenrecord` and `screenrecord-bin`: the entry point execs a patched copy under the second name on images whose stock binary caps the limit, and a missed signal would leave the mp4 without a moov atom.

One deliberate behavioral delta: wrapper-bearing devices on API 34+ move from unlimited recording to the wrapper's 3600s ceiling. Only cloud AVDs carry the wrapper and worker jobs are hard-capped at ~15.5 minutes, so the ceiling sits ~4x above any reachable recording duration.

Verification on a golden built by the current ScreenrecordStep (android-33, pixel_6):
- `maestro record --local` through this branch captured **197s continuously** (last-frame pts 197.0), stopped cleanly at flow end
- the same image's stock binary rejects any `--time-limit` over 180 (`Time limit 181s outside acceptable range [1,180]`), so the pre-change path caps at 180s there by construction
- three cloud verification runs on post-#[2681](https://github.com/mobile-dev-inc/copilot/pull/2681) goldens all capped at exactly 180s through the old path (recorder exit at start+180.4s in logcat)

Rollout: after merge, bump `maestro` in copilot; the worker deploy picks it up from there.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -1397,8 +1397,11 @@ class AndroidDriver(
         private const val CHUNK_SIZE = 1024L * 1024L * 3L
 
         // Extended screenrecord entry point baked into cloud worker AVDs by
-        // maestro-device's ScreenrecordStep. Its path and pass-through arg shape
-        // are a contract with that step; change them together or not at all.
+        // maestro-device's ScreenrecordStep. Three things are a contract with that
+        // step and must change together or not at all: this path, the pass-through
+        // arg shape, and the name of the process the entry point execs on patched
+        // images (screenrecord-bin, which close() must SIGINT for the moov atom
+        // to be flushed).
         private const val EXTENDED_SCREENRECORD_PATH = "/data/local/tmp/screenrecord"
     }
 }
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -492,10 +492,24 @@ class AndroidDriver(
 
             val deviceScreenRecordingPath = "/sdcard/maestro-screenrecording.mp4"
 
+            // Cloud worker devices bake an extended screenrecord entry point that lifts the
+            // stock 180s time limit and pins encoder-safe dimensions (maestro-device's
+            // ScreenrecordStep). Record through it when present. On stock devices only
+            // API 34+ can disable the limit ("--time-limit 0"); below that, recordings
+            // stop at the stock 180s cap.
+            val extendedRecorder = EXTENDED_SCREENRECORD_PATH.takeIf {
+                connection.shell("test -x $it").exitCode == 0
+            }
+
             val future = CompletableFuture.runAsync({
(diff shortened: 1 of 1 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -1378,5 +1395,13 @@ class AndroidDriver(
...
         private const val SCREENSHOT_DIFF_THRESHOLD = 0.005
         private const val CHUNK_SIZE = 1024L * 1024L * 3L
+
+        // Extended screenrecord entry point baked into cloud worker AVDs by
+        // maestro-device's ScreenrecordStep. Three things are a contract with that
+        // step and must change together or not at all: this path, the pass-through
+        // arg shape, and the name of the process the entry point execs on patched
+        // images (screenrecord-bin, which close() must SIGINT for the moov atom
+        // to be flushed).
+        private const val EXTENDED_SCREENRECORD_PATH = "/data/local/tmp/screenrecord"
```
steviec, line 1405: Hrm. this feels like a smell to bake our maestro-device specific screenrecord path into our core library. Are you sure this is the right approach? 

One alternative: make our screenrecord patcher accept "0" and then internally translate it to some large number. This makes our patched screenrecord a drop-in replacement instead of having to encode special knowledge in core.

pedro18x: Hmmm. Accept 0 alone wouldn't actually fix the bug: core only sends --time-limit 0 on API 34+. On the broken images (<34) it sends no flag at all, so the stock 180s default kicks in regardless of what the patched binary accepts. A real drop-in would mean patching the binary's default limit too, or overwriting /system/bin/screenrecord at bake.

The overwrite version is doable but pricey: root + remount, overlay has to survive into the golden snapshot, and we'd need to verify that on all 8 API levels since remount behaves differently across them. The current approach is just an adb push that works the same everywhere and is already verified end to end. Also, a probe against a documented path shows up in the code, while a stock-looking binary that quietly records for an hour is the kind of magic someone loses a day to later. And core already branches on recording behavior right above this line (the API 34 check), so the probe isn't really a new kind of knowledge there.

So my take: ship this now to stop the customer bleed, it's small and easy to revert. If we still want core clean afterwards, I'd rather do that as a follow-up where the recorder path becomes an injectable driver option that the worker sets. Knowledge lives with the component that owns it, core stays generic, no magic in the images. What say you?

steviec (approved):
I buy it

PR #3418, opened 2026-07-08
feat(orchestra): capture step screenshot pre-command; pair with hierarchy on failure

proksh:
## Why

Per-step screenshots were captured in `onCommandFinished` — **after** the command ran. For a passing `tapOn` that navigates, `step-N.png` showed the *destination* screen, not the element the command acted on. A step's screenshot should be the screen it was about to act on.

The view hierarchy has the same problem and must stay consistent with the screenshot: the run viewer overlays hierarchy element-bounds on the shot, so the two must describe the same screen.

## What

- **Passing/skipped steps** → screenshot captured in `onCommandStart` (pre-command). No hierarchy (unchanged — `viewHierarchy()` is ~1s, so passing steps never pay for it).
- **Failed/warned steps** → screenshot **and** hierarchy captured together at the outcome moment, so they match. Under `captureFullArtifacts` this overwrites the pre-command `step-N.png` at the same path, so there's no duplicate manifest record or extra callback.
- **CLI (`captureFullArtifacts` off)** is unchanged: failed/warned steps still shoot at finish.

Delivered via the existing `onStepScreenshotCaptured(seq, relativePath)` callback — the path mapping is unchanged, so consumers need no change.

## Trade-offs (intentional)

- Failed steps under `captureFullArtifacts` now show the at-failure frame (paired with the hierarchy) rather than a pre-command one; the full-run recording still holds the moment-by-moment view.
- Skipped steps now carry a pre-command screenshot (their `onCommandStart` fires before the skip is decided) — one shot each; uniform "every step has one screenshot, no nulls".

## Tests

`ArtifactsGeneratorTest` covers: pre-command capture before finish; failed step's at-failure frame overwriting the pre-command one paired with hierarchy (single record); warned + full-artifacts pairing; flag-off failed/warned unchanged; skipped keeps its pre-command shot. `:maestro-orchestra:test` + `:maestro-test:test` green.

### DIFF
--- a/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
+++ b/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
@@ -31,6 +31,11 @@ import java.nio.file.StandardCopyOption
  * With the flag off, only failed/warned steps capture that pair. The ~1s hierarchy
  * round-trip is why only they ever pay for it.
  *
+ * Each per-step shot is the screen *before* that step, so step N+1's is also step N's
+ * end state; a single flow-end shot (`screenshots/final.png`, flow-level) closes the
+ * chain with the screen the run ended on — after any onFlowComplete teardown — giving
+ * complete start-and-end evidence.
+ *
  * Every file is routed through an [ArtifactCollector]: the manifest is the
  * collector's records and each command's artifact list is the same records
  * grouped by command, so the two can never disagree. Device logs + crash/ANR
@@ -172,6 +177,8 @@ internal class ArtifactsGenerator(
     }
 
     override fun onFlowEnd() {
+        // Capture the resting screen before the recording is torn down.
+        if (captureFullArtifacts) captureFinalScreenshot()
         stopFullRunRecording()
         val collector = collector
         if (artifactsDir != null && collector != null) {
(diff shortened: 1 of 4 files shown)


proksh:
Addressed a pre-merge review pass (pushed 7425849):

- **Non-destructive recapture (the one real finding):** under `captureFullArtifacts` a failed/warned step overwrites its pre-command `step-N.png` at finish. The old overwrite went straight through `takeDebugScreenshot`, which truncates the file on open and `delete()`s it if the device capture throws mid-write — likely exactly when a step just failed — leaving the start-time callback pointing at a vanished file. `captureStepScreenshotFile` now captures to a sibling temp and moves it into place, so a failed recapture keeps the pre-command frame (and disk stays consistent with the callback). Added a test for that path.
- Flattened `onCommandFinished` (early return over a nested `when`), trimmed comments, fixed KDoc wording.
- Merged latest `main`.

Note (intentional, no change): failed/warned steps take one extra screenshot under the flag (pre-command + at-outcome). Passing steps are unchanged (one, just relocated to start).

pedro18x (approved):
Traced the paths, this holds up, and the temp-then-move recapture is nice.

One thing worth confirming: now that each step's screenshot is taken before the command runs, the screen you see after the last command is never saved as a still, it only lives in the recording. So in the viewer, the last step's thumbnail shows the screen right before that command ran, not the final result. Same idea for launchApp: step 0 shows the screen before the app opened. For every step except the last, that "after" screen just becomes the next step's screenshot, so nothing's really lost in the middle, only at the very end. Intended, since the recording still covers it?

Fishbowler:
Where we're considering this as evidence of the run throughout, for a passing run should there be one more showing the final state, after the final command?

proksh:
> Traced the paths, this holds up, and the temp-then-move recapture is nice.
> 
> One thing worth confirming: now that each step's screenshot is taken before the command runs, the screen you see after the last command is never saved as a still, it only lives in the recording. So in the viewer, the last step's thumbnail shows the screen right before that command ran, not the final result. Same idea for launchApp: step 0 shows the screen before the app opened. For every step except the last, that "after" screen just becomes the next step's screenshot, so nothing's really lost in the middle, only at the very end. Intended, since the recording still covers it?

Yes, was being lost at start before, and only the trailing end after this PR. But is the information that useful to add?


> Where we're considering this as evidence of the run throughout, for a passing run should there be one more showing the final state, after the final command?

We can definitely add that, but not sure what purpose does it solve?
- If the last step passed, meant that command worked, does the next screen even matter?
- If the command failed/warned, we are updating that screen.

Dont think we should update the final screen on success. Reason -
For example "- tapOn: 'Submit'" success step should be linked with a screen having 'Submit' button. If the screen doesnt have that button, it will lead to confusion for why did that step passed.

Fishbowler:
> We can definitely add that, but not sure what purpose does it solve?

Lots of folks require evidence to go along with the test results. Right now, they get the final state. This PR switches to getting the start state. I think getting start and end is better, for complete evidence.

I think the hardest part is filenaming 😂

proksh:
> > We can definitely add that, but not sure what purpose does it solve?
> 
> Lots of folks require evidence to go along with the test results. Right now, they get the final state. This PR switches to getting the start state. I think getting start and end is better, for complete evidence.
> 
> I think the hardest part is filenaming 😂

Added `final.png` that captures the end screenshot. Since its not consumed anywhere right now, its flexible to change.

PR #3419, opened 2026-07-08
fix(android): bound adb-backed devtools sockets so webview tapOn can't hang or silently abort (MA-4111, MA-4090)

pedro18x:
## What this fixes

Two customer-facing symptoms on Android webview flows after the API 36 migration, both from a single root cause:

- **MA-4111**: a webview `tapOn` hangs indefinitely and the cloud run is stuck in RUNNING, never completing.
- **MA-4090**: a webview `tapOn` fails intermittently.

## Root cause

On android-36, opening the WebView devtools socket over adb (`localabstract:webview_devtools_remote_*`) intermittently wedges: the adb stream OPEN or the following read stops making progress. dadb's blocking `java.net.Socket.read` is **not** interruptible on JDK 17, so `future.cancel(true)` cannot release it. On the mainline `AdbServerDadb` path these per-stream sockets have no timeout at all, so once a read wedged, the calling thread parked forever.

That one wedge produced both tickets:

- Unbounded, the first wedged capture never returns, so the `tapOn` hangs and the run is stuck RUNNING (**MA-4111**).
- The wedge is intermittent, so before it was bounded the capture would abort or miss on the affected poll, and the webview `tapOn` failed intermittently (**MA-4090**).

Discovery over `adb shell` stays fast throughout (measured p99 12ms, see below), which is why the failure looks like "the device is fine but this one webview won't respond."

## The fix

**Socket layer (`AdbSocketFactory.kt`).** Split the factory into two explicit constructors:

- `AdbSocketFactory.raw(opener)` is byte-for-byte the previous behavior (caller-thread blocking I/O). The gRPC channel keeps using it, so the gRPC plane is unchanged.
- `AdbSocketFactory.bounded(executor, opener)` is used only by the devtools client. Reads, writes, and the connect handshake run on a worker pool with a bounded `future.get`, so an uninterruptible dadb park can no longer hang the caller: the caller is released at the deadline and the wedged worker is abandoned. The pool is capped and **owned per devtools client (one per device)**, so a wedged device fails its own calls fast and cannot starve webview capture on other devices sharing the worker process. Includes off-thread close (a blocking CLSE write can't stall teardown), a publish/claim connect handshake so exactly one side owns the socket, a broken-socket flag after an abandoned read, and `Error` rethrow (not IOException-wrapped).

**Capture policy (`DadbChromeDevToolsClient.kt`).** The degrade is scoped to a transport wedge: a timed-out websocket wait, or a dead / timed-out / adbd-refused adb socket, degrades that one webview and the capture continues, instead of aborting the whole hierarchy. Everything that is not a transport wedge fails loudly rather than silently yielding a native-only hierarchy: `DeviceConnectionException` (dead / unauthorized device) and `InterruptedException` (cancellation) propagate, and so does any protocol or data break, a malformed CDP payload, a garbage `/json` listing, or a failed discovery shell, so a real integration break surfaces as an error instead of untappable-element flake. Socket discovery (`cat /proc/net/unix`) is bounded on the same handoff.

**Stream OPEN classification (`AndroidDeviceConnection.kt`).** A dadb `AdbStreamOpenException` (the peer refused a single stream OPEN) is now rethrown as a plain `IOException` and does not mark the whole device dead. A per-stream refusal is not a transport death; genuine transport failures still map through and mark the connection dead. In dadb 2.0.0 this exception is raised only on adbd's stream-refusal reply ("adbd refused to open stream"); an offline or dead device surfaces instead as `AdbConnectionClosedException` / `AdbTimeoutException`, both of which still map to device-dead, so an offline device is not swallowed into "no webviews."

## Relationship to #3410 (MA-4119)

#3410 added the outer wall-clock budget for webview augmentation (10s) plus a per-fetch daemon pool. This PR bounds the socket layer underneath it, and the two compose: #3410's abandoned fetch threads only actually terminate because these socket-level bounds let an uninterruptible dadb read be abandoned safely. Without the socket bounds, #3410's `future.cancel` has nothing to cancel and the daemon thread leaks parked forever. Rebased onto #3410; the gRPC-plane change is a single-token `.raw` delta against its version of the file.

## Evidence

**Causality, per commit.** Each fix commit is paired with a test that fails when the production change is reverted to origin/main and passes when restored, so the tests pin the specific defect rather than the surrounding behavior.

**Full suite green.** `./gradlew :maestro-client:test` -> 234 tests, 0 failures, 2 skipped.

**Live hardware.** Replayed the customer's actual run (Wahed, `zakatCalculator`, `com.wahed.mobile.staging2`) on a fresh android-36 / pixel_6 AVD with the real app and workspace, instrumented to log CDP latency:

| operation | p50 | p99 | max | budget | breaches |
|---|---|---|---|---|---|
| discovery (`cat /proc/net/unix`) | 7ms | 12ms | 16ms | 5000ms | 0 |
| `/json` listing | 1ms | 4ms | 11ms | 10000ms | 0 |
| `Runtime.evaluate` | 6ms | 49ms | 295ms | 5000ms | 0 |

A healthy android-36 webview answers CDP in milliseconds, 40x-1000x under the timeout, so the budgets never mask a live-but-slow webview. Every webview assertion that ran passed. During the run a real ~2-minute adb-transport wedge occurred (13 consecutive `Connect to webview_devtools_remote_5172 interrupted` at the 10s mark). Before this change that first wedged capture would have hung the command indefinitely (MA-4111); with it every cycle degraded cleanly with no hang, no crash, and no thread-pool exhaustion at the JVM level. A follow-up replay against the same device then found its adb transport fully wedged (`adb devices` reports the device healthy while `adb shell echo` hangs), which is the platform condition behind both tickets.

## Canary validation (6h)

Canaried on 5 android-36 worker hosts (arm-m2m-021, -085, -073, -090, -103) for 6 hours against live customer traffic, with the rest of the fleet on the previous build (`c658e1b`) as the comparison group. The deployed jars were verified to carry the fix (`newAdbIoExecutor` present in `AdbSocketFactoryKt`), not just the right worker SHA.

- **236 runs, 191 SUCCESS. Zero `INFRA_ERROR`s and zero circuit-breaker trips** on every host for the full 6 hours. These are the deploy-health signals: a wedge this PR failed to bound would surface here, and none did.
- The fleet monitor flagged 7 was-passing-now-failing regressions over the window. Every one was checked against the baseline build and is pre-existing and customer-side (fleet-wide flakes on native screens); **none was canary-isolated or webview-shaped.** Most are scheduling artifacts, the prior "success" ran on a baseline host, so a baseline-pass paired with a canary-fail reads as a regression even though this change is nowhere near it.
- The single most webview-plausible signal ("Print" not found on a health-records export) resolved to Android's **native** system print dialog (`printspooler` / `documentsui`), no WebView in the flow. Another fingerprint fails identically on **iOS**, where an Android-only change cannot be the cause. Others fail byte-identically on the baseline build, hours to weeks before the canary existed.
- The one signature that would indict this change, a WebView element present in a passing hierarchy and missing only on the fix build, never appeared.

## Scope

This PR bounds the devtools capture path only. The raw-dadb teardown path (`AndroidDriver.stopAndCollectDeviceLogs`, `AndroidDriver.close`) and the gRPC reconnect path can also wedge on the same android-36 transport, but they are byte-identical to origin/main here, they are off the `tapOn` path, and a wedge there is force-finalized by the backend stuck-RUNNING reconciliation (25 minutes) rather than left running forever. Bounding those surfaces, or bouncing a wedged adb transport at the worker level, is a separate worker-hygiene change and is intentionally out of scope so this PR stays reviewable and focused on the two tickets.

## Test plan

- [x] `./gradlew :maestro-client:test` (234 / 0 / 2)
- [x] Revert-flip causality confirmed per fix commit
- [x] Live android-36 replay of run_01kw9ekdpxftvrjyxmy4pdkz3t (Wahed zakatCalculator)
- [x] 6h canary on 5 android hosts: 236 runs, 191 SUCCESS, 0 INFRA_ERROR, 0 breaker trips; all 7 flagged regressions verified pre-existing/customer-side against baseline


proksh (commented):
It's a tricky and complex change — nice work on this. Took a while to understand everything going on here. Reviewed with Claude and a few issues came up worth a look:

**Correctness**

1. **`degradeTo(null)` swallows non-`IOException` failures in webview augmentation** (`DadbChromeDevToolsClient.kt` ~L123). The new `degradeTo` catches every `Exception` except `DeviceConnection`/`Interrupted`, whereas the old code only caught `IOException` and let everything else propagate. If a malformed CDP `Runtime.evaluate` response makes `.result.value` throw a `RuntimeException`/NPE, augmentation now silently drops that webview from the merged hierarchy — elements inside become untappable ("element not found") behind only a warn log, instead of failing loudly.

2. **Malformed `/json` listings are silently swallowed** (`DadbChromeDevToolsClient.kt` ~L181-185). Parse failures (`IllegalStateException` "Failed to parse…" / "No body found") are now caught by the per-socket `degradeTo(emptyList)`. Previously they propagated and aborted capture with the raw-body diagnostic. Now a `200`-with-garbage-JSON response degrades the socket to empty — its webviews vanish from the hierarchy and the integration break hides behind a warn log.

3. **`AdbStreamOpenException` reclassified away from device-dead handling** (`AndroidDeviceConnection.kt` ~L223). This is now turned into a plain skippable `IOException` rather than going through `mapTransport` (which marked the device DEAD and threw `DeviceUnreachableException`). Per dadb docs this exception also fires when the device is *offline* — so if a device goes offline mid-run, the error gets swallowed to "no webviews" and the flow continues against an offline device with a silently native-only hierarchy, no device-unreachable error.

4. **Process-wide 16-worker `adbIoExecutor` can starve across parallel devices** (`AdbSocketFactory.kt` ~L82). The single process-wide pool (cap 16, `SynchronousQueue`) is shared by every `BoundedAdbSocket`/`boundedAdbCall` across all `AndroidDeviceConnection`s. A wedged device permanently holds workers — `cancel(true)` can't interrupt a JDK17 raw-socket park, and `closeOffThread`'s `stream.close()` can itself park. Once all 16 are held, `submitAdbCall` throws `RejectedExecutionException` → `IOException` for *every* bounded adb call process-wide, including healthy devices, which `degradeTo` then swallows to "no webviews." In a multi-device / shard run the per-device hang this PR removes could turn into a process-wide webview-capture outage.

**Minor / cleanup**

5. Per-read throwaway `ByteArray(len)` + arraycopy on the happy path in `BoundedAdbSocket.read` — the copy is only needed on the abandon path, which almost never fires (`AdbSocketFactory.kt` ~L304).
6. Non-200 `/json` listing logs at ERROR while every other per-socket failure on the same path is WARN via `degradeTo` — spurious ERROR noise for an expected, recoverable skip (`DadbChromeDevToolsClient.kt` ~L235).
7. Single-byte `read()`/`write(Int)` each incur a full executor submit + `Future` + bounded await plus two allocations per byte (`AdbSocketFactory.kt` ~L295).

The common thread on 1–4 is that the new `degradeTo` `catch (e: Exception)` is broad enough to convert several previously-hard failures into silent native-only hierarchies. Narrowing the catch (or at least distinguishing "no webviews" from "capture failed") would restore the loud-failure behavior.

Also — given that these failure modes degrade *silently* (elements just become untappable, no error surfaced), I'd suggest rolling this out behind a canary rather than shipping to everyone at once. If any of the above (or an unforeseen edge) does regress webview capture, the symptom is silent flake in customer flows rather than a clear error, so we'd want a small blast radius to catch it before a full rollout.

pedro18x:
Thanks for the careful read. Addressed in 8780b37b.

**Fixed**

- **1 & 2 (broad catch swallowed protocol/data failures):** confirmed, my regression. `degradeTo` had widened from the pre-PR catch to `catch (Exception)`. Narrowed it back so it degrades only the transport wedge: a websocket `TimeoutException` (the MA-4090 skip) and a dead / timed-out / adbd-refused socket `IOException`. Everything else propagates and fails the capture loudly, matching the old behavior you described: a `RuntimeException`/NPE from a malformed `evaluate`, a garbage `/json` listing (`IllegalStateException`), a failed discovery shell, any programming error. Two existing tests flip to assert the loud propagation, plus an augmentation-level guard.
- **4 (process-wide pool starvation):** confirmed. The bounded pool is now owned per `DadbChromeDevToolsClient` (one per device) instead of a process-wide static, so a wedged device can't starve captures on healthy devices in the same worker. Added a cross-pool isolation test.
- **6 (non-200 logs ERROR):** now WARN.

**Pushing back, with evidence**

- **3 (`AdbStreamOpenException` reclassification):** in dadb 2.0.0, `AdbConnection.open()` throws three distinct exceptions by failure mode: `AdbStreamOpenException` = "adbd refused to open stream" (adbd replied and refused this destination, so the transport is alive), `AdbTimeoutException` = "Timed out opening stream", and `AdbConnectionClosedException` = "Connection lost while opening stream". A device going offline surfaces as the latter two, both `AdbException` subtypes that still go through `mapTransport` -> `markDead`. Only the clean stream-refusal becomes a skippable `IOException`, so an offline device isn't swallowed to "no webviews".
- **5 (happy-path arraycopy):** the private `ByteArray` has to be allocated before the worker starts, on every read, because we can't know in advance which read will be abandoned. By the time a read is abandoned the parked worker already holds the array, so "only copy on the abandon path" would let a late completion write into the caller's recycled buffer. There's a safe alternative (one reusable per-socket scratch buffer, safe because the `broken` flag blocks reuse after the first abandon), but it's a micro-opt on a few-KB path, so I left it.
- **7 (single-byte read/write):** `read()` and `write(int)` are abstract on `InputStream`/`OutputStream`, so they're required, and they're off the hot path (okio does bulk transfers), so they never fire in practice. No real cost.

**Rollout:** agree on the canary. These degrade silently, so a small blast radius first is the right call; I'll roll it that way.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
```diff
@@ -98,27 +120,45 @@ class DadbChromeDevToolsClient(private val connection: AndroidDeviceConnection):
...
         okhttp.cache?.close()
+        adbIoExecutor.shutdown()
     }
 
     override fun getWebViewTreeNodes(): List<TreeNode> {
         return getWebViewInfos()
             .filter { it.visible }
             .mapNotNull { info ->
-                try {
+                degradeTo(null, "Failed to retrieve WebView hierarchy from chrome devtools: ${info.socketName} ${info.webSocketDebuggerUrl}") {
```
proksh, line 130: This `degradeTo(null)` swallows more than transport wedges. A malformed `Runtime.evaluate` reply throws `JsonProcessingException`, but `makeRequest` (line 189) rewraps it as `IOException`, so it lands in the `IOException` branch and gets dropped to a warn. That's the exact evaluate-path case item 1 wanted loud — still silent here. The `/json` path was fixed (throws `IllegalStateException`, which `degradeTo` doesn't catch), this one wasn't.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
```diff
@@ -98,27 +120,45 @@ class DadbChromeDevToolsClient(private val connection: AndroidDeviceConnection):
...
+            block()
+        } catch (e: DeviceConnectionException) {
+            throw e
+        } catch (e: InterruptedException) {
+            Thread.currentThread().interrupt()
+            throw e
+        } catch (e: TimeoutException) {
+            logger.warn(message, e)
+            fallback
+        } catch (e: IOException) {
```
proksh, line 153: An interrupt inside a bounded socket read comes back as `InterruptedIOException`, which is an `IOException`, so it hits this branch instead of the `InterruptedException` one above. So cancellation from the #3410 budget lands inconsistently depending on where the thread was parked. Might want to re-check for interrupt / rethrow before the `IOException` catch.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
```diff
@@ -151,7 +191,14 @@ class DadbChromeDevToolsClient(private val connection: AndroidDeviceConnection):
     }
 
     fun getWebViewInfos(): List<WebViewInfo> {
-        return getWebViewSocketNames().flatMap(::getWebViewInfos)
+        val socketNames = degradeTo(emptySet(), "Failed to discover WebView devtools sockets. Skipping WebView capture.") {
```
proksh, line 199: Discovery now degrades to `emptySet()`. If the adb transport is wedging during discovery, we silently return zero webviews and warn once instead of failing — so every webview tap on that device becomes 'element not found' for the rest of the flow.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
```diff
@@ -171,68 +218,69 @@ class DadbChromeDevToolsClient(private val connection: AndroidDeviceConnection):
...
-            throw e.cause ?: e
+        try {
+            val response = try {
+                future.get(stepTimeoutMillis, TimeUnit.MILLISECONDS)
+            } catch (_: TimeoutException) {
+                throw TimeoutException("Timed out waiting for websocket response")
+            } catch (e: ExecutionException) {
+                throw e.cause ?: e
+            }
+            ws.close(1000, null)
```
proksh: Success path calls `ws.close()` but never `ws.cancel()` (only the catch does). If the graceful close write wedges, that worker parks and never gets released, and they pile up toward the pool cap. Might be worth cancelling on the success path too.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
```diff
@@ -28,52 +69,88 @@ class AdbSocketFactory(private val opener: (host: String, port: Int) -> AdbStrea
...
-    override fun connect(endpoint: SocketAddress, timeout: Int) {
-        if (endpoint !is InetSocketAddress) throw UnsupportedOperationException("Endpoint must be InetSocketAddress")
-        this.endpoint = endpoint
-        stream = opener(endpoint.hostString, endpoint.port)
+// Submits one blocking dadb call to the client's capped pool. Pool exhaustion fails fast instead
+// of queueing: every worker is parked against a wedged transport (see newAdbIoExecutor).
+private fun <T> submitAdbCall(executor: ExecutorService, operation: String, block: () -> T): Future<T> =
+    try {
+        executor.submit(Callable { block() })
+    } catch (e: RejectedExecutionException) {
```
proksh, line 95: When the pool fills with un-interruptible parked reads, this maps to a plain `IOException`, which `degradeTo` then swallows — so the whole device drops all webview capture for the rest of the flow, not just the one wedged endpoint. Per-client scoping keeps it to one device, but the silent drop is still there.

proksh (commented):
Main worry is `degradeTo`. It catches `IOException`/`TimeoutException` pretty broadly, so a lot of failures just turn into a native-only hierarchy + one warn line, and downstream that's an untappable-element flake. That's the same silent-failure class this PR set out to kill (item 1). Left inline notes on the specific spots. The per-client pool scoping is good — it bounds the blast radius — but a few silent drops still slip through. Nothing here is a blocker on the approach, mostly about narrowing what we swallow.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
pedro18x: Fixed. `makeRequest` rewrapped the parse error as `IOException`, which `degradeTo` swallows. Now throws `IllegalStateException` like the `/json` path so it aborts loudly. Test: `a malformed Runtime evaluate reply aborts the capture loudly`.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
pedro18x: Fixed via the interrupt flag, not the type: `SocketTimeoutException` is also an `InterruptedIOException` (the wedge we want to degrade). The IOException branch now checks `Thread.interrupted()` and rethrows, so cancellation propagates while a socket timeout still degrades. Test: `an interrupt while reading a devtools listing aborts the capture instead of degrading`.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
pedro18x: Keeping the degrade. Discovery re-runs on every hierarchy fetch, so a transient wedge self-heals and native taps still work off the base tree. Propagating would break native taps on the screen and reintroduce the transient failure the degrade absorbs. A permanent wedge is uncapturable either way; the WARN names it. Can escalate the blackout case in a follow-up if we want.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
pedro18x: Fixed. `cancel()` moved to a `finally` so it runs on the success path too. Test: `the websocket is torn down after a successful response`.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
pedro18x: Same call as the discovery thread: per-client scoping bounds it to the wedged device, the degrade lets native taps proceed, and workers free once the transport recovers. Already logged at WARN.

maestro-client/src/main/java/maestro/android/AndroidDeviceConnection.kt
```diff
@@ -219,6 +220,12 @@ class AndroidDeviceConnection private constructor(
     internal fun open(destination: String): AdbStream =
         try {
             dadb.open(destination)
+        } catch (e: AdbStreamOpenException) {
```
proksh, line 223: This remap is on the shared `open()`, so it also catches `pullAppFile`'s `run-as ... cat` OPEN, not just the webview path. A `run-as` refusal on a non-debuggable package used to go through `mapTransport` -> `markDead` -> `DeviceUnreachableException` (abort); now it's a plain `IOException` and the connection stays CONNECTED, so retry/abort behavior changes for a caller this fix wasn't scoped to. Could gate the remap to the `localabstract:` webview destination (or handle it in the devtools socket factory) so it only touches the path you tested.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
```diff
@@ -28,52 +66,80 @@ class AdbSocketFactory(private val opener: (host: String, port: Int) -> AdbStrea
...
+// Closing a dadb stream can park forever: AdbStreamImpl.close() sends CLSE through a raw,
+// non-interruptible socket write under the connection-wide sink monitor. Bounded sockets
+// therefore never close streams on the caller thread. Fall back to a throwaway daemon thread
+// when the pool is exhausted, because close() must not block and must not be dropped.
+private fun closeOffThread(executor: ExecutorService, stream: AdbStream) {
+    val close = Runnable { runCatching { stream.close() } }
+    try {
+        executor.execute(close)
+    } catch (e: RejectedExecutionException) {
+        Thread(close, "AdbStreamClose").apply { isDaemon = true }.start()
```
proksh: This fallback isn't bounded by `ADB_IO_POOL_MAX` -- when the pool is full (all 16 parked on a wedged device) every close spawns a fresh daemon thread, and per the comment above `close()` can itself park forever on that same transport. So across capture cycles against a persistently-wedged device these accumulate unbounded, which is the JVM-level starvation the cap is there to prevent. Recovers once the transport does, but might be worth a bound.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
```diff
@@ -111,3 +175,234 @@ private class AdbSocket(private val opener: (host: String, port: Int) -> AdbStre
...
+                    closeOffThread(executor, s)
+                    throw InterruptedException("Connect to ${endpoint.hostString} abandoned")
+                }
+                s
+            }
+        } catch (e: Throwable) {
+            (handoff.getAndSet(ABANDONED) as? AdbStream)?.let { closeOffThread(executor, it) }
+            throw e
+        }
+        stream = opened
```
proksh, line 290: nit: if `close()` runs fully between this line and the `if (closed)` check below, both close the same stream -- close() sees `stream` and `closeOffThread`s it, then connect() sees `closed` and `closeOffThread`s `opened` again. `runCatching` swallows the second close so it's harmless, just a double close of the same handle.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
proksh: Fair on the transient case -- self-healing discovery is the right call there. The one I'd still want covered is the permanent wedge: right now it's indistinguishable from 'no webviews on screen', so every tap flakes as element-not-found for the rest of the flow with no signal. A follow-up that surfaces a sustained blackout as an error (not a silent native-only tree) is enough; doesn't need to be in this PR.

maestro-client/src/main/java/maestro/android/AndroidDeviceConnection.kt
pedro18x: Fixed. The remap is gated to `localabstract:` now, so only the webview path skips a per-stream OPEN refusal; `run-as` and every other destination keep the original `mapTransport` -> `markDead` -> `DeviceUnreachable`. Test: `open maps a non-localabstract OPEN rejection to DeviceUnreachable and marks the connection dead`.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
pedro18x: Fixed. The fallback is bounded at `ADB_IO_POOL_MAX` now: past the cap the close is dropped rather than spawning another thread, since a close on a dead transport just parks until the transport recovers, which reaps it anyway. Test: `close fallback threads are bounded so a wedged transport cannot spawn them without limit`.

maestro-client/src/main/java/maestro/android/AdbSocketFactory.kt
pedro18x: Agreed it's harmless: `runCatching` swallows the second close. A race-free fix would need atomic getAndSet ownership of the stream field, which isn't worth it for a swallowed no-op. Leaving as-is.

maestro-client/src/main/java/maestro/android/chromedevtools/DadbChromeDevToolsClient.kt
pedro18x: Agreed, that's the right follow-up. Surfacing a sustained blackout needs cross-fetch state (count consecutive discovery/pool wedges and escalate), a separate change from the per-fetch degrade here. Tracking it as a follow-up rather than folding it into this PR.

proksh (approved):
A thought on the overall shape (not a blocker)

Zooming out from the individual comments - one thing worth flagging is less about any single bug and more about the pattern.

Across the review rounds, we've kept widening and then re-narrowing the same decision: when something goes wrong during webview capture, should we quietly skip and keep going, or stop and surface the error? That question keeps coming back, and each pass we tune it in a slightly different spot. The fixes so far are right — but the fact that it keeps needing re-tuning suggests the decision itself doesn't live in one clear place yet. It's spread across a few spots, so it's easy for the next change (a new caller, a library bump) to reopen the whole thing.

I don't think this needs a redo - it's close, and the recent fixes are good. My only suggestion is to make that "skip vs. fail loudly" call explicit in a single place before we build more on top of it, so we're not rediscovering the edge cases one exception at a time.

@amanjeetsingh150 - could you take a second look at this part? You know this codebase better.

amanjeetsingh150:
Few questions I've concluded right now posting directly from chat of claude, just want to ensure we are solving this at right layer:

 

```
Research summary:  — and two layer questions

  I traced this incident end-to-end (DB, GCP logs on the three affected hosts, dadb 2.0.0 source, and the shipped jar). The socket bounding in this PR is real and needed — but the evidence says the hang is a consequence of our probing, not
  background weather, and that changes what layer the durable fix belongs in.

  1. Mental model when the CDP endpoint doesn't answer

  A silent devtools socket is almost never a slow server — it's a frozen process whose listener is a kernel object. In the June 29 run, @webview_devtools_remote_5468 belonged to a background app instance (pid 5468 never logged a line;
  foreground was a different pid). Android's cached-apps freezer stops every thread including the CDP accept loop, but the socket stays advertised in /proc/net/unix, and the kernel keeps completing connect()s into its accept queue without the owner ever running. So:

  - Probes #1–#12: connect "succeeds" (backlog slot), GET /json written into a buffer nobody reads, 10s read timeout (SocketTimeoutException at readUtf8LineStrict — stack traces in the worker logs), one permanently leaked backlog slot per
  probe (peer-close doesn't dequeue a never-accepted connection).
  - Probe #13: backlog full → adbd's unix connect() blocks → the entire shared dadb session goes deaf mid-flight. Last byte on the session: 10:52:53.01, the instant OPEN #13 went out. deviceInfo then burned 3×120s deadlines → misclassified as
  DeviceServerDiedException (the "adb alive" probe is a bare TCP connect — the kernel completes it into adbd's listen backlog whether or not adbd is servicing anything).
  - This reproduced deterministically: both attempts and the sibling UK run died ~12–13 probes after entering the screen, on three different hosts, with identical connection ages (744s) — a resource-exhaustion fingerprint, not flake.

  Notably, this PR's own replay saw the same fingerprint: 13 consecutive Connect to webview_devtools_remote_5172 interrupted at the 10s mark, followed by a fully wedged transport (adb devices healthy, adb shell echo hangs). The PR reads that
  as "the platform condition behind both tickets" — I think it's the result of the 13 probes. Bounding our wait releases the caller thread, but each bounded-and-abandoned capture still fires a connect at the dead socket, so the backlog still
  fills and the session (which gRPC also rides) can still be killed. The fix that matches the mechanism is stop connecting, not only stop waiting:

  - negative-cache a socket that just timed out (don't re-probe it every hierarchy fetch — 12 probes in 2 minutes here, all for the same dead socket)
  - and/or only probe sockets whose pid matches the foreground app — hasWebView(baseTree) gates on the foreground tree, but cat /proc/net/unix then probes every socket on the device, which is how a frozen background instance got probed at all

  Related transport bug worth tracking separately: in shipped dadb 2.0.0, a read-starved connection is never invalidated (read timeouts don't close the socket; DadbImpl.connection() only rebuilds on isClosed), so the session stayed a zombie
  for 16+ minutes across teardown — confirmed by connection ages in the two exceptions (744,298ms and 984,195ms → same birth).

  2. Is CDP capture needed for this customer at all?

  For this flow: no, and we have proof. All 12 probe failures degraded to "empty list" and every command still passed 5–6ms later — text=Continue, "Your Liabilities" etc. all resolved from the native accessibility tree (tap logs show plain
  native TreeNodes). The probed WebView was in a background process and never on screen. Chromium exposes web content to accessibility as virtual nodes, so plain text selectors — the overwhelming majority — resolve natively; CDP is only
  load-bearing for HTML id / CSS selectors and JS evaluation. Today the expensive, fragile path runs eagerly on every hierarchy fetch to serve a capability this run never used.

  Suggestion (fine as a follow-up, but it's the layer that removes the failure class): make CDP capture lazy — probe only when a selector actually misses in the native tree, or at minimum foreground-pid-filtered + negative-cached. That
  converts "frozen background app anywhere on the device" from a run-killer into a no-op.
```

amanjeetsingh150:
My main concern right now is around to know why do wahed uses androidWebViewHierarchy? 

- What exactly DOM element property they're trying to access that is not visible from accessibility layer. Maybe it's there on accessibility layer
- Failure point on communicating to CDP layer itself lets say we connect to CDP layer but it doesnt respond to something because app thread is frozen, this feels like something that should be mapped to transport error.

Let's say if we have transport error on webview why it's progressing the flow to subsequent commands and not failing with transport error?

pedro18x:
Your trace is right about the root cause, and it checks out end to end: each probe of the frozen app's socket permanently occupies a slot in its connection queue, and adbd makes that connection on the same thread that services all adb traffic, so when the queue fills the whole transport goes deaf. Chromium hit this exact issue and raised that queue from 10 to 4096 in Chromium 144 (crrev.com/c/7013293). One correction on my side: the "13 connect timeouts" in my replay were attempts made after the transport was already dead, not the queue filling up, so the PR body's reading of them was wrong.

Where I land differently is on the two conclusions. First, Wahed does need androidWebViewHierarchy: in MA-4090, ten days before the run you traced, the Confirm button was missing from the captured hierarchy across repeated fetches on API 36, and their selectors are plain text. The devtools capture is what supplies text nodes when the accessibility tree lacks them, so dropping the config would turn an intermittent failure into a permanent one. Second, lazy CDP doesn't remove the failure class: element lookup polls the hierarchy in a tight loop, so an element that only exists in the webview tree misses natively on every iteration and would probe the dead socket just as often.

The fix that matches your mechanism is narrower: only probe sockets belonging to the app that owns the WebView we're augmenting, instead of every webview socket on the device. Roughly 50 lines in DadbChromeDevToolsClient. So my proposal: merge this PR as-is, since it fixes the separate hang-forever layer (wedges become fast retriable failures instead of 30-minute stalls), and hold the worker deploy until the socket-filter follow-up lands so one deploy carries both. You found the mechanism, so the follow-up is yours if you want it; otherwise I'll write it this week and you review.

PR #3420, opened 2026-07-09
feat(orchestra): skip no-op step captures; 1-based, human-readable step artifact names

proksh:
Follow-up to #3418, addressing review feedback on per-step artifacts.

- **Screenshots for every step that touches the screen; hierarchy only where it helps.** Non-visible leaves (`defineVariables`/`applyConfiguration`) take no on-screen action, so they capture neither screenshot nor hierarchy. Composite parents (`runFlow`/`repeat`/`retry`) **keep their own screenshot** and fire `onStepScreenshotCaptured`, so their step never renders blank in the viewer — but they skip the ~1s `viewHierarchy` round-trip, since they only wrap children. Visible leaves capture both. (Per @pedro18x's review: composites no longer get deduped out of their screenshot.)
- **1-based numbering.** Step files start at `step-001` (was `step-0`). No-op steps still consume a sequence number, so numbering can have gaps by design; `step-<N>` stays a stable identity for command N in `commands.json`.
- **Human-readable names.** `step-<NNN>-<type>[-<arg>]`, e.g. `step-003-tapOnElement-submit.png`, `step-005-inputText-sanchit.png`, `step-007-tapOnPointV2-30%,40%.png`.

Notes:
- The `onStepScreenshotCaptured` callback keeps its **0-based** `sequenceNumber`; only the reported path changes. Composite steps now report a screenshot like any other step; only non-visible leaves emit no callback / empty artifact list.
- The slug is built from the **raw pre-evaluation** command, so a `${var}` argument appears verbatim in the filename (e.g. `step-002-inputText-${name}.png`) rather than the resolved value — this deliberately keeps evaluated secrets out of filenames. `$`, `{`, `}` are filesystem-legal and the sanitizer still strips path separators and OS-reserved characters, so the names pose no filesystem issue. `commands.json` carries the fully-evaluated command.
- Script commands (`evalScript`/`runScript`) still capture, because they are `visible() == true`. Making them non-visible would also hide them from the CLI result view, so their capture is left as-is for now.
- Trade-off: with the composite/leaf screenshot dedup removed, a leaf failing inside a `repeat`/`runFlow` in CLI (flag-off) mode now yields two failure screenshots of ~the same screen (leaf + composite parent). This is deliberate — a duplicate frame is better than a blank composite row.

Tests: `:maestro-orchestra:test` green.

### DIFF
--- a/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
+++ b/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
@@ -64,6 +64,8 @@ internal class ArtifactsGenerator(
      * command, so a single reference (no stack) is enough to attribute them.
      */
     private var currentCommandMetadata: CommandDebugMetadata? = null
+    /** Highest sequence number that already captured a failure/warning frame this flow; -1 if none. */
+    private var lastFailureCaptureSeq: Int = -1
 
     override fun onFlowStart() {
         if (artifactsDir == null) return
@@ -73,6 +75,7 @@ internal class ArtifactsGenerator(
             logCapture = ScopedLogCapture.start(logFile)
             flowStartMs = System.currentTimeMillis()
             appUnderTest = null
+            lastFailureCaptureSeq = -1
             capturer = DeviceArtifactCapturer(maestro, artifactsDir.resolve(BundleLayout.LOGS_DIR)).also { it.start() }
         } catch (e: Exception) {
             logger.warn("Failed to set up artifacts directory at $artifactsDir", e)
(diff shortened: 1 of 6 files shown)


maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
```diff
@@ -136,16 +136,23 @@ internal class ArtifactsGenerator(
         if (artifactsDir == null || outcome is CommandOutcome.Skipped) return
         // Passing steps keep their pre-command shot from onCommandStart; nothing to do at finish.
         if (outcome !is CommandOutcome.Failed && outcome !is CommandOutcome.Warned) return
+        // Non-visible leaves (defineVariables/applyConfiguration) and empty commands have no screen.
+        if (StepArtifactNaming.isNoOp(cmd) && !StepArtifactNaming.isComposite(cmd)) return
+        // A composite parent needs its own failure frame only when no descendant leaf already shot
+        // one (a while/when condition that throws fails the composite before any child runs).
+        // Descendants finish first and carry higher sequence numbers, so a later capture means done.
+        if (StepArtifactNaming.isComposite(cmd) && lastFailureCaptureSeq > metadata.sequenceNumber) return
```
pedro18x: This drops the composite parent's own screenshot in full-artifacts mode: the isComposite dedup plus the onCommandStart no-op skip now both apply even when captureFullArtifacts is true, so a failed runFlow/repeat/retry whose child already captured a failure returns early with no screenshot and no onStepScreenshotCaptured. That row renders blank in the cloud step viewer. The test that pinned this ("a failed composite parent keeps its own screenshot when captureFullArtifacts is true") was deleted in this PR. Was flipping that behavior intentional, or did the dedup leak into the full-artifacts path?

proksh: Good catch - you're right on both counts. 

1. Though it was intentionally flipped, I wanted to keep min screenshots, and for a failed leaf command i didn't want composite steps to save the same hierarchy. But since screenshot would be missing and the frontend will falback to a previous screenshot, it can be a bit confusing. So switched the composite step screenshot back.

2. `onStepScreenshotCaptured` now fires again

pedro18x (approved):
Nice, my nudge was cleanly resolved, awesome work!

PR #3422, opened 2026-07-10
fix(ios): classify swipeV2 UI-query timeout as 408, not an untriageable INFRA_ERROR

pedro18x:
**Interim triage-correctness fix.** Stops the mislabeled INFRA alert, the wasted retries, and the "Unknown error", by classifying the failure correctly. It does **not** make the swipe succeed: the run still fails, now as a clean non-retryable `TEST_ERROR` with a real message. Making swipeV2 work on busy/webview screens is the behavioral fix, tracked in [MA-4129](https://linear.app/mobile-dev/issue/MA-4129).

## Problem

iOS `swipeV2` (`SwipeRouteHandlerV2`) returns a bare **HTTP 500** when the underlying XCUITest gesture times out with `Error Domain=com.apple.dt.XCTest.XCTFuture Code=1000 "Swipe (v2) ...: Timed out while evaluating UI query."` (and the main-thread-busy `automation-support` code 6 variant).

Downstream, the Maestro Cloud worker maps that 500 to `XCUITestServerError.UnknownFailure` (null `.message`), then to `TestExecutionException.Unexpected`, i.e. **`INFRA_ERROR`**, which is **retried up to 3x across different hosts**. Because the timeout is deterministic (the screen never reaches idle), every retry fails identically, then the run fails as infra with the message collapsed to the literal string **"Unknown error"**. No diagnostics for whoever triages it.

### Evidence

- **109 occurrences over Jul 8-9 2026**, all from a single org (DoorDash / Dasher app, busy Safari webview), all `IOS,WEB`, deterministic (0 recovery across the 3 retries).
- The real cause is only in the stacktrace (never the `errorMessage`), verbatim from GCP (`perf-dev-289002`):
  ```
  Request for swipeV2 failed, code: 500, body: {"code":"internal","errorMessage":
  "Swipe v2 request failure. Error: Error Domain=com.apple.dt.XCTest.XCTFuture Code=1000
  \"Swipe (v2) from (201.0, 437.0) to (201.0, 786.0) with 0.4 duration: Timed out while evaluating UI query.\""}
  ```
- GCP filter: `jsonPayload.logger="dev.mobile.maestro.worker.MaestroWorker" jsonPayload.context.job_failure_code="INFRA_ERROR" jsonPayload.exception.stacktrace=~"Request for swipeV2 failed"`

## Fix

Classify the XCUITest UI-query / main-thread-busy timeout as **`.timeout` (HTTP 408)** in `SwipeRouteHandlerV2`, **mirroring the identical discrimination `ViewHierarchyHandler` already applies** for the byte-identical `NSError`. Any other swipe error keeps the existing 500. (The catch is extracted into a behavior-identical `classify()` in the follow-up commit so it can be unit-tested.)

408 routes through the **pre-existing** typed pipeline: `handleExceptions` (`408` to `OperationTimeout`), `XCTestIOSDevice`, `IOSDriver.DriverTimeout`, worker `TestFailure`, **`TEST_ERROR` (non-retryable)**, carrying the real Apple message.

Two points from review are folded in: the XCUITest-timeout discrimination is now a shared `NSError.isXCUITestTimeout` used by both `SwipeRouteHandlerV2` and `ViewHierarchyHandler` (so a future Apple domain/code change is one line, and neither handler can silently drift back to 500), and the `swipeV2` request is sent **one-shot** so OkHttp cannot replay the non-idempotent gesture on a 408.

| | Before | After |
|---|---|---|
| Classification | `INFRA_ERROR` (retryable) | `TEST_ERROR` (non-retryable) |
| Retries | 3x across hosts, all fail | none |
| Message | "Unknown error" | "...Timed out while evaluating UI query." |

## Verification

The full chain was traced and confirmed end to end, and pinned with tests (both suites green):

| Link | How verified | Result |
|---|---|---|
| Swift catch maps the real errors to 408 | Swift unit tests **run on an iOS 26.5 simulator** (compiled the change) | 6/6 pass |
| ...for the *exact* production error shape | Test inputs are the literal `XCTFuture/1000` + `automation-support/6` domain/code/message from the GCP logs | pass |
| Runner (this branch) builds, boots, serves `swipeV2` | Launched live on a sim; baseline idle swipe | HTTP 200 |
| `408` -> `OperationTimeout`, and the swipe is not replayed | Kotlin test drives the real `swipeV2()` client against a mock 408; asserts `requestCount == 1` | 11/11 pass |
| `OperationTimeout` -> `DriverTimeout` -> `TestFailure` (`TEST_ERROR`) | Existing `MaestroTestRunnerTest` + the unchanged pipeline | pre-tested |

Two things the verification surfaced:

- **The swipe is sent one-shot, so OkHttp does not replay it.** OkHttp's `RetryAndFollowUpInterceptor` would otherwise repeat a 408 once, firing the non-idempotent gesture a second time (and silently recovering if that second attempt settled, a hidden double-scroll). Marking the `swipeV2` request body one-shot skips that retry, so a deterministic timeout surfaces immediately as `OperationTimeout`, with no double-scroll and no ~2x latency. ViewHierarchy's 408 keeps its beneficial retry (it is transient and retries-and-passes).
- **The FlyingFox 100s handler-timeout gap is real.** Forcing a foreground app to stall past 100s reproduced `handler error: Task timed out before completion. Timeout: 100.0 seconds` on a live sim, with no clean 408. That is the empty-body-500 subset this PR does not reach (see below).

## Scope / limitations

- **Does not make the swipe succeed.** It turns a mislabeled, retried infra failure into a clean, surfaced flow failure. The behavioral root cause (V2 gates a pure-coordinate swipe behind an XCUITest UI-query/idle evaluation that V1's raw synthesize avoids) is tracked in [MA-4129](https://linear.app/mobile-dev/issue/MA-4129).
- **Does not cover empty-body 500s** (~28% of the sample): the main-thread-wedged mode where the gesture hangs past FlyingFox's 100s server timeout, so this `catch` never runs. Also tracked in MA-4129.

## Risk

Change is confined to `SwipeRouteHandlerV2`'s catch (now `classify()`). `ViewHierarchy`'s own `UnknownFailure` (transient, legitimately retries-and-passes) is untouched, so its behavior and alert suppression are unchanged.

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
+++ b/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
@@ -22,6 +22,22 @@ struct SwipeRouteHandlerV2: HTTPHandler {
             try await swipePrivateAPI(requestBody)
             return HTTPResponse(statusCode: .ok)
         } catch let error {
+            // An XCUITest UI-query / main-thread-busy timeout is a deterministic, device-answered
+            // failure (the screen never reaches idle), not retryable infra. Classify it as .timeout
+            // (HTTP 408) so the driver maps it to a non-retryable timeout carrying the real message,
+            // instead of a bare 500 the worker relabels "Unknown error" and retries. Mirrors the
+            // identical discrimination in ViewHierarchyHandler; any other swipe error keeps the 500.
+            if let nsError = error as NSError?,
+               nsError.domain == "com.apple.dt.XCTest.XCTFuture",
+               nsError.code == 1000,
+               nsError.localizedDescription.contains("Timed out while evaluating UI query") {
+                return AppError(type: .timeout, message: "Swipe v2 request failure. Error: \(error.localizedDescription)").httpResponse
+            } else if let nsError = error as NSError?,
+                      nsError.domain == "com.apple.dt.xctest.automation-support.error",
+                      nsError.code == 6,
+                      nsError.localizedDescription.contains("Unable to perform work on main run loop, process main thread busy for") {
+                return AppError(type: .timeout, message: "Swipe v2 request failure. Error: \(nsError.localizedDescription)").httpResponse
+            }
             return AppError(message: "Swipe v2 request failure. Error: \(error.localizedDescription)").httpResponse
         }
     }
--- a/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
+++ b/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
@@ -22,24 +22,34 @@ struct SwipeRouteHandlerV2: HTTPHandler {
             try await swipePrivateAPI(requestBody)
             return HTTPResponse(statusCode: .ok)
         } catch let error {
-            // An XCUITest UI-query / main-thread-busy timeout is a deterministic, device-answered
-            // failure (the screen never reaches idle), not retryable infra. Classify it as .timeout
-            // (HTTP 408) so the driver maps it to a non-retryable timeout carrying the real message,
-            // instead of a bare 500 the worker relabels "Unknown error" and retries. Mirrors the
-            // identical discrimination in ViewHierarchyHandler; any other swipe error keeps the 500.
-            if let nsError = error as NSError?,
-               nsError.domain == "com.apple.dt.XCTest.XCTFuture",
-               nsError.code == 1000,
-               nsError.localizedDescription.contains("Timed out while evaluating UI query") {
-                return AppError(type: .timeout, message: "Swipe v2 request failure. Error: \(error.localizedDescription)").httpResponse
-            } else if let nsError = error as NSError?,
-                      nsError.domain == "com.apple.dt.xctest.automation-support.error",
-                      nsError.code == 6,
-                      nsError.localizedDescription.contains("Unable to perform work on main run loop, process main thread busy for") {
-                return AppError(type: .timeout, message: "Swipe v2 request failure. Error: \(nsError.localizedDescription)").httpResponse
-            }
-            return AppError(message: "Swipe v2 request failure. Error: \(error.localizedDescription)").httpResponse
+            return Self.classify(error).httpResponse
         }
+    }
+
+    /// Maps a swipe failure to an `AppError`.
+    ///
+    /// An XCUITest UI-query / main-thread-busy timeout is a deterministic, device-answered failure
+    /// (the screen never reaches idle), not retryable infra. It is classified as `.timeout` (HTTP 408)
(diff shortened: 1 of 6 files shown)


maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
```diff
@@ -22,10 +22,36 @@ struct SwipeRouteHandlerV2: HTTPHandler {
...
+    /// 500 the worker relabels "Unknown error" and retries. Mirrors the identical discrimination in
+    /// ViewHierarchyHandler; any other swipe error keeps the default 500.
+    nonisolated static func classify(_ error: Error) -> AppError {
+        let nsError = error as NSError
+        let message = "Swipe v2 request failure. Error: \(error.localizedDescription)"
+
+        if nsError.domain == "com.apple.dt.XCTest.XCTFuture",
+           nsError.code == 1000,
+           nsError.localizedDescription.contains("Timed out while evaluating UI query") {
+            return AppError(type: .timeout, message: message)
```
proksh, line 35: One thing on the 408 retry beyond the "bounded extra attempt" note in the description: since a swipe isn't idempotent, OkHttp's one-shot retry can fire the gesture twice — and if the *second* attempt settles we return `200 .ok`, i.e. a silent double-scroll rather than a clean fail (later taps/asserts then land on the wrong element). Same edge on the code-6 main-thread-busy branch, where we also drop the old retry-and-recover for the genuinely transient case, and a persistent timeout now takes ~2x to surface.

We're leaning on the timeout staying deterministic. Might be safer to make this route non-retryable (one-shot body / `retryOnConnectionFailure=false`) so the reclassification doesn't quietly depend on that.

maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/SwipeRouteHandlerV2.swift
```diff
@@ -22,10 +22,36 @@ struct SwipeRouteHandlerV2: HTTPHandler {
...
     }
 
+    /// Maps a swipe failure to an `AppError`.
+    ///
+    /// An XCUITest UI-query / main-thread-busy timeout is a deterministic, device-answered failure
+    /// (the screen never reaches idle), not retryable infra. It is classified as `.timeout` (HTTP 408)
+    /// so the driver maps it to a non-retryable timeout carrying the real message, instead of a bare
+    /// 500 the worker relabels "Unknown error" and retries. Mirrors the identical discrimination in
+    /// ViewHierarchyHandler; any other swipe error keeps the default 500.
+    nonisolated static func classify(_ error: Error) -> AppError {
```
proksh, line 32: `classify()` duplicates the domain/code/message discrimination verbatim from `ViewHierarchyHandler`. Worth pulling into a shared helper so a future Apple domain/code change is a one-line update — miss one copy and that handler silently goes back to 500 → INFRA_ERROR, the exact bug this fixes.

proksh (approved):

PR #3427, opened 2026-07-10
fix(orchestra): bound failure/teardown artifact captures so a wedged device can't blow the run watchdog

proksh:
## Why

Once the Android device server wedges mid-run, the failure/teardown captures in `ArtifactsGenerator` each ride the driver's full timeout — ~120s (gRPC `viewHierarchy`/screenshot) or ~10min (adb recording `close()`) — and run back-to-back. That's enough to blow the 15-min run watchdog and report `INFRA_ERROR` on a device that's already dead. Fleet-wide this drives ~120 watchdog timeouts/day.

## What

Bound each device-touching capture (`captureStepHierarchy`, `captureScreenshot`, `stopFullRunRecording`) with a short hard cap (15s, injectable for tests).

Blocking dadb/gRPC reads aren't interruptible on JDK 17, so `withTimeout` can't unblock them. Instead the call runs on a throwaway daemon thread and is **abandoned** on timeout — the caller stops waiting so teardown proceeds; the thread finishes on its own later. Captures stay best-effort (log and skip). Normal captures are ~1s, so the cap only bites on an already-wedged device.

Complements #3316 (gRPC `UNAVAILABLE` retry/rebuild, which leaves the 120s deadline untouched) and #3410 (WebView path, which excludes artifact capture).

## Tests

3 new tests (wedged hierarchy / screenshot / recording-stop) assert the path is bounded and teardown completes. `:maestro-orchestra:test` green.

### DIFF
--- a/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
+++ b/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
@@ -18,6 +18,7 @@ import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption
+import java.util.concurrent.atomic.AtomicReference
 
 /**
  * Internal listener Orchestra always installs. Populates [FlowDebugOutput]
@@ -46,6 +47,7 @@ internal class ArtifactsGenerator(
     private val maestro: Maestro,
     private val captureFullArtifacts: Boolean = false,
     private val onStepScreenshotCaptured: (sequenceNumber: Int, relativePath: String) -> Unit = { _, _ -> },
+    private val captureTimeoutMs: Long = DEFAULT_CAPTURE_TIMEOUT_MS,
 ) : OrchestraListener {
 
     val debugOutput = FlowDebugOutput()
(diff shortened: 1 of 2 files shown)


maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
```diff
@@ -307,16 +312,37 @@ internal class ArtifactsGenerator(
     }
 
     private fun stopFullRunRecording() {
+        val recording = fullRunRecording ?: return
         try {
-            fullRunRecording?.close()
+            withCaptureTimeout("stop-recording") { recording.close() }
```
pedro18x, line 317: Other two call sites look right to me. This one's a different animal though. `close()` here is Maestro's wrapper (up to 3s of duration padding) then AndroidDriver's: `killall -INT`, a hardcoded `Thread.sleep(3000)`, then pulling the whole mp4 off the device. So the floor is ~3s by construction and the pull scales with run length, not the ~1s the cap is sized for.

Abandoning this one isn't free either: `Maestro.kt` only clears `screenRecordingInProgress` after `close()` returns, and `startScreenRecording` returns a no-op stub while it's set. `TestSuiteInteractor` shares one `Maestro` across a suite, so under `--analyze` one abandoned close quietly costs every later flow its recording.

Probably a `try/finally` in `Maestro.kt` so the flag can't leak, plus a longer budget here. Happy to be wrong if that pull is faster than I think.

proksh:
Closing this PR as I am not sure if its the correct solution.

PR #3428, opened 2026-07-11
Evaluate scripts in permissions values

### LINKED ISSUE #2416: Allow to set permission valie via paramaters
### Use case

The launchapp command is specified in a subflow that is used by several tests. Based on the test case we want to set the value of some permissions on allow, deny, unset ... 

```
- launchApp:
    clearState: true
    permissions:
      location: ${PERMISSION_LOCATION}
```

Currently it seems not to be possible to do that, it does not pick up the parameter value. 

### Proposal

Allow expression when defining permission values. 

### Anything else?

_No response_

Fishbowler:
## Proposed changes

Allows JS interpolations in permissions settings.

```
- launchApp:
    permissions:
      camera: ${VAR}
```

## Testing

New e2e test flow, tested locally on both platforms.

The e2e test took longer to get right than the code change, by an order of magnitude, so I've tried to feed some of this back to the CLAUDE.md for future changes 🤞 

## Issues fixed

Fixes #2416


proksh (approved):
LGTM, fix looks right. A few non blocking things:

1. Null permission value NPEs. `permissions:` with `camera:` and no value comes through as a Jackson null, so `it.value.evaluateScripts()` throws. `Env.withEnv` already guards this exact case for env vars, same could apply here.

2. `associate` drops colliding keys silently. `{ ${P}: allow, location: deny }` with `P=location` means last one wins with no warning. YAML rejected literal duplicates before, so interpolation is what makes this reachable.

3. This is the third copy of `entries.associate { k.evaluateScripts() to v.evaluateScripts() }` in 30 lines. A `Map<String, String>.evaluateScripts(jsEngine)` in `Env.kt` would cover all three and give the two points above one place to live.

4. `permissionStatus_$name` generates ids with spaces (`permissionStatus_All Files`). An explicit slug per permission would be safer.

One thing worth a second thought: the Android only block asserting that `nonsense` and `""` silently revoke bakes silent failure into the contract, while iOS throws on the same input. Same flow now fails loudly on one platform and quietly on the other.

Fishbowler:
1 - I'm okay with this. As soon as you're into interpolation you're into debugging when you get it wrong. We could look at how to help with that in the future, but it's bigger than this.

2 - I'm fine with this too for the same reason. Although we should probably have a test for an interpolated permission _name_.

3 - Valid. I'll spawn that into another ticket.

4 - I'm okay with that in DemoApp

On different behaviour per platform, that hasn't changed - it's just been highlighted in an e2e test - I'm not attempting to alter that here.

Issue #3137, opened 2026-04-04
Maestro 2.4.0 hangs with zero output on Apple Silicon + iOS 26.2/26.4 simulator


imgyf:
## Environment

- **Maestro:** 2.4.0 (also reproduced on 2.0.10 and 2.1.0)
- **macOS:** 26.3.1 (Tahoe)
- **Xcode:** 26.4
- **Architecture:** Apple Silicon (arm64)
- **Simulators tested:** iOS 26.2 (iPhone 17 Pro), iOS 26.4 (iPhone 16 Pro)
- **Java:** OpenJDK 17.0.18 (Homebrew)

## Behavior

`maestro test` produces **zero stdout/stderr** and hangs indefinitely until killed or timed out.

The XCTest driver app (`dev.mobile.maestro-driver-iosUITests.xctrunner`) installs on the simulator and the process launches, but it **never listens on port 7001**. The driver process stays alive but opens no network connections.

`maestro --help` and `maestro --version` work fine — the JVM is OK. The issue is specifically the iOS XCTest driver failing to start its HTTP server.

## Steps to reproduce

```bash
# 1. Fresh install
rm -rf ~/.maestro && curl -Ls "https://get.maestro.mobile.dev" | bash

# 2. Boot an iOS 26.2 simulator
xcrun simctl boot <device-id>

# 3. Run any flow (or even a trivial one)
timeout 60 maestro test -e APP_ID=xyz.nexeraai.avo any-flow.yaml
# Result: zero output, exit code 124 (timeout)
```

## Diagnostic details

```bash
# Driver installs but doesn't listen
$ xcrun simctl listapps booted | grep maestro
"dev.mobile.maestro-driver-iosUITests.xctrunner" = { ... }

# Driver process runs but has no network connections
$ xcrun simctl launch booted dev.mobile.maestro-driver-iosUITests.xctrunner
# PID assigned, process alive, but:
$ lsof -i :7001
# (empty — nothing listening)

# Maestro's test output directory is created but empty (no log, no screenshots)
$ ls ~/.maestro/tests/<latest>/
# maestro.log (empty or just system info, no flow execution)
```

## Context

- On **Maestro 2.1.0** with the same hardware/OS/simulator setup, flows **did execute** (we ran ~20 flows successfully). There were intermittent `kAXErrorInvalidUIElement` crashes (~10% of runs) during view hierarchy inspection, but the driver connected and flows ran.
[...]

PR #3430, opened 2026-07-11
fix(ios): recover from transient kAXErrorInvalidUIElement

### LINKED ISSUE #2617: Exception in thread "main" UnknownFailure

### Steps to reproduce

I am hitting #2570 nondeterministically in CI

### Actual results

You had closed that issue because it didn't have enough information to repro

### Expected results

I'd be happy to provide that information. What exactly are you looking for?

### About app

It's a closed source React Native app; on 0.79.3

### About environment

13:21:55.018 [ INFO] MAESTRO.logSystemInfo: ---- System Info ----
13:21:55.026 [ INFO] MAESTRO.logSystemInfo: Maestro Version: 1.41.0
13:21:55.044 [ INFO] MAESTRO.logSystemInfo: CI: github
13:21:55.046 [ INFO] MAESTRO.logSystemInfo: OS Name: Mac OS X
13:21:55.048 [ INFO] MAESTRO.logSystemInfo: OS Version: 14.7.6
13:21:55.048 [ INFO] MAESTRO.logSystemInfo: Architecture: aarch64
13:21:55.048 [ INFO] MAESTRO.logSystemInfo: Java Version: 17
13:21:55.749 [ INFO] MAESTRO.logSystemInfo: Xcode Version: 15.4
13:21:55.864 [ INFO] MAESTRO.logSystemInfo: Flutter Version: Undefined
13:21:55.898 [ INFO] MAESTRO.logSystemInfo: Flutter Channel: Undefined
13:21:55.898 [ INFO] MAESTRO.logSystemInfo: ---------------------

### Logs

https://pastebin.com/eLFR8Y5H

### Maestro version

1.41.0

other (please specify in the text area below)

### Anything else?

Using github action https://github.com/dniHze/maestro-test-action

nathanstitt:
## Proposed changes

`XCUIElement.snapshot()` intermittently throws: `Error getting element frame kAXErrorInvalidUIElement` when an element goes stale mid-snapshot while the UI is updating.  This becomes a fatal HTTP 500 that surfaces to the client as an uncaught UnknownFailure and aborts the entire flow, often hanging the JVM rather than exiting cleanly. 

I use Maestro to test native builds of my ReactNative app (https://github.com/tinycld/tinycld) and it seems to occur about 25% of the time when running it's fairly large screenshot based test suite.

This PR routes the error through the recovery path that already exists for its sibling kAXErrorIllegalArgument, so a momentary stale element no longer kills the run.

I looked at the history around transient kAXErrorInvalidUIElement before writing this, and deliberately avoided repeating an approach that was already tried and reverted:

HTTP-client retry (#3291 and reverted in #3312).
*  #3291 wired an OkHttp interceptor that retried transient 5xx / IOException, explicitly citing kAXErrorInvalidUIElement from production logs. It was reverted six days later ( I think because it caused a 4.4× spike in 15-min outer-flow timeouts?),  the interceptor retried IOException at up to the full read timeout per attempt, producing 10–12 min silent gaps. 
* (The Android sibling,  #3290, was reverted alongside it in #3311) The revert explicitly called for "a  narrower retry surface / per-call wall-clock budget."

This PR takes the opposite approach from a retry by reusing the existing single-shot recovery walk, so there's no added wall-clock cost and no risk of re-introducing the timeout regression that got #3291 reverted.

## Testing

I ran tinycld tests repeatedly 10times without a crash.  I did consider adding a dart e2e test but wasn't sure you'd want that since:
 * it's _very_ non deterministic, because the bug is a timing race in XCTest's snapshot
 * I think the failure may be something around ReactNative specifically, in which case the e2e test might not catch it at all

With that said I'm happy to add a e2e test if you do think it'd be valuable to have.

## Issues fixed

I found 3 issues that are likely caused by this: #2617 #3148 #3137 

Fixes #2617

### DIFF
--- a/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
+++ b/maestro-ios-xctest-runner/maestro-driver-iosUITests/Routes/Handlers/ViewHierarchyHandler.swift
@@ -159,7 +159,7 @@ struct ViewHierarchyHandler: HTTPHandler {
             hierarchy.children = children
             return hierarchy
         } catch let error {
-            guard isIllegalArgumentError(error) else {
+            guard isRecoverableSnapshotError(error) else {
                 NSLog("Snapshot failure, cannot return view hierarchy due to \(error)")
                 if (error as NSError).isXCUITestTimeout {
                     throw AppError(type: .timeout, message: error.localizedDescription)
@@ -168,7 +168,7 @@ struct ViewHierarchyHandler: HTTPHandler {
                 }
             }
 
-            NSLog("Snapshot failure, getting recovery element for fallback")
+            NSLog("Snapshot failure (\(error.localizedDescription)), getting recovery element for fallback")
             AXClientSwizzler.overwriteDefaultParameters["maxDepth"] = snapshotMaxDepth
             // In apps with bigger view hierarchys, calling
             // `XCUIApplication().snapshot().dictionaryRepresentation` or `XCUIApplication().allElementsBoundByIndex`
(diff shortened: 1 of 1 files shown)


Fishbowler:
This is a generated essay, and maintainers have finite time. Can you reduce this down to just the value of the change and the issue it actually fixes (no guesses)?

nathanstitt:
This is *not* a generated essay, I assure you I took care in writing it.  What exactly are you looking for?

nathanstitt:
I guess the tldr is: when the accessibility tree on iOS rapidly changes, you are using a stale snapshot and throwing exceptions that crash.  I added code to retry?  is that what you're looking for?

Fishbowler:
I'll need to wait til Monday when I'm at a PC to review this properly. I think there's 2 open PRs that are in this area, and I want to be sure we're not doing quadruple retries 😂

Some initial thoughts:
- Two of the fixed issues are already closed?
- What now happens when this error is hit? Is it actually retried, or quietly using the previous snapshot, or something else?

nathanstitt:
Hey, no problem, It is hard to tell what's "real" anymore 🙁 

When the error was hit (before my change), Maestro would emit error `kAXErrorInvalidUIElement Exception in thread "main"` and _sometimes_ emit nothing further nor exit, causing tests to hang indefinitely.  Sometimes it would exit cleanly after the log message though.

Re the closed issues:  
 *  #3137 mentions a workaround which I think was the short-lived merge and revert of https://github.com/mobile-dev-inc/Maestro/pull/3291 but perhaps not.
 * #3148 was closed due to non-responsiveness but the symptoms match my failures and it mentioned `kAXErrorInvalidUIElement` - very well might not be the same cause tho.

Its not at all urgent on my side now, so do feel free to close in favor of a potentially better fix if you'd like

nathanstitt:
Oh also if you _really_ want to dig into this, I'd be happy to provide replication steps using tinycld, but getting it all setup and configured for dev involves cloning quite a few repos and configuring them

nathanstitt:
Thanks @Fishbowler, let me know if you'd like any assistance with this or need more details

Fishbowler (approved):

PR #3434, opened 2026-07-14
fix(ios): only re-stabilise a tap that follows a scroll (MA-4135)

proksh:
> **Merge recommendation:** targeted patch to unblock the customer (Wahed) now — their iOS flows are being killed by the 15-min watchdog today. The `recentScroll` heuristic is a stop-gap; the per-tap settle behaviour will be reworked properly in maestro core, which supersedes this. Suggest merging as the immediate fix.

## Problem
[MA-4124](https://linear.app/mobile-dev/issue/MA-4124) (`e7b67140b`) made iOS `tapOnElement` re-stabilise the element whenever `waitForAppToSettle` returns `null`. But on iOS that's the steady state for *every* settled screen, so every tap now ran the stabilisation loop — up to a 3s deadline + 2 extra hierarchy fetches. Long flows (Wahed onboarding) blew past the 15-min watchdog. iOS-only.

## Fix
Only a preceding scroll leaves momentum that outlives the static check, so gate stabilisation on a `recentScroll` hint: set by scroll/swipe gestures, and run `refreshElementUntilStable` only when `waitForAppToSettle == null && recentScroll`; otherwise trust the pre-wait hierarchy (pre-MA-4124 behaviour). `performTap` clears the hint, so it only affects the immediately-following tap.

## Verified on-device
Repro of Wahed `onboardNewUser` (iPhone-16 / iOS-18.2, real staging), matched by tap index:

| Metric | Regression | This PR |
|---|---|---|
| Elapsed to tap #47 | 817s | **622s (−24%)** |
| Taps entering the stabilisation loop | 53 | **4** |
| Taps hitting the full 3s deadline | 22 | **3** |

19m16s baseline → ~14.5 min, under the watchdog.

## Tradeoffs (inherent to the heuristic; resolved by the core rework)
A tap after *non-scroll* motion (e.g. a sliding modal) is no longer re-stabilised — restores long-standing pre-MA-4124 behaviour. A `scroll → assert → tap` sequence may still stabilise unnecessarily (safe, just slower).

## Tests
Case 145 (MA-4124 preserved), 146 (no-scroll skip), 147 (intervening tap clears hint), 148 (bare-swipe path). `IntegrationTest`: 165 pass.

### DIFF
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -74,8 +74,8 @@ class Maestro(
     private var screenRecordingInProgress = false
 
     // A scroll/swipe can leave the app decelerating past the screen-static check, so the next tap
-    // must re-stabilise the element (MA-4124). Any tap clears it (see performTap) so ordinary taps
-    // on a settled screen skip the extra hierarchy fetches (MA-4135).
+    // must re-stabilise the element (MA-4124). Cleared by the next tap (performTap) and by launchApp;
+    // it survives other commands until then.
     private var recentScroll = false
 
     suspend fun launchApp(
@@ -84,6 +84,8 @@ class Maestro(
         stopIfRunning: Boolean = true
     ) = runInterruptible(Dispatchers.IO) {
         LOGGER.info("Launching app $appId")
+
+        recentScroll = false
 
         if (stopIfRunning) {
             LOGGER.info("Stopping $appId app during launch")
@@ -130,6 +132,8 @@ class Maestro(
     suspend fun hideKeyboard() = runInterruptible(Dispatchers.IO) {
         LOGGER.info("Hiding Keyboard")
 
+        // iOS dismisses the keyboard with real content drags that can leave the screen decelerating.
+        recentScroll = true
         driver.hideKeyboard()
     }
 
@@ -218,13 +222,16 @@ class Maestro(
 
         val settledHierarchy = waitForAppToSettle(initialHierarchy, appId, waitToSettleTimeoutMs)
 
-        // A null settle can't confirm the screen is static. Scroll momentum is the one motion that
-        // routinely outlives the check, so re-stabilise only after a scroll (MA-4124); otherwise
-        // trust the hierarchy we have (MA-4135) — pre-MA-4124 behaviour for non-scroll animation.
-        // performTap clears recentScroll, so the hint only affects the tap that immediately follows.
+        // Scroll momentum is the one motion that routinely outlives a null settle, so re-stabilise
+        // only after a scroll (MA-4124); otherwise trust the hierarchy we have (MA-4135).
         val (hierarchyBeforeTap, refreshedElement) = if (settledHierarchy == null && recentScroll) {
+            LOGGER.info("Tap aimed via stabilised hierarchy (null settle after a scroll)")
             refreshElementUntilStable(element, initialHierarchy)
         } else {
+            LOGGER.info(
+                if (settledHierarchy != null) "Tap aimed via settled hierarchy"
+                else "Tap aimed via trusted pre-wait hierarchy (null settle, no recent scroll)"
+            )
             val hierarchy = settledHierarchy ?: initialHierarchy
             hierarchy to hierarchy.refreshElement(element.treeNode)?.toUiElementOrNull()
         }
@@ -372,8 +379,7 @@ class Maestro(
         tapRepeat: TapRepeat? = null,
         waitToSettleTimeoutMs: Int? = null
     ) {
-        // Every tap consumes the scroll-momentum hint, so it only ever affects the next tap.
-        recentScroll = false
+        recentScroll = false // consume the scroll hint (MA-4135)
 
         val capabilities = runInterruptible(Dispatchers.IO) { driver.capabilities() }
 
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -73,6 +73,11 @@ class Maestro(
 
     private var screenRecordingInProgress = false
 
+    // A scroll/swipe can leave the app decelerating past the screen-static check, so the next tap
+    // must re-stabilise the element (MA-4124). Any tap clears it (see performTap) so ordinary taps
+    // on a settled screen skip the extra hierarchy fetches (MA-4135).
+    private var recentScroll = false
+
     suspend fun launchApp(
         appId: String,
         launchArguments: Map<String, Any> = emptyMap(),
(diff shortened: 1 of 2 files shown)


maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -359,6 +372,9 @@ class Maestro(
         tapRepeat: TapRepeat? = null,
         waitToSettleTimeoutMs: Int? = null
     ) {
+        // Every tap consumes the scroll-momentum hint, so it only ever affects the next tap.
+        recentScroll = false
```
pedro18x: One case I keep coming back to: `scrollUntilVisible -> tapOn point (a fixed banner) -> tapOn "Confirm"`. The coordinate tap consumes the hint, but its post-tap wait is the same static check we know deceleration can outlive, so the element tap can still aim with a mid-drift hierarchy. That's the original MA-4124 shape sneaking back in for scroll -> tap -> tap sequences. Since Case 147 pins the current behaviour as intended, wanted to check it's a deliberate call. If not, maybe we only clear the hint when the post-tap settle actually confirmed static (`hierarchyAfterTap != null`)? On an already-settled screen the stabilisation fast path should keep the cost tiny either way. Wdyt?

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -73,6 +73,11 @@ class Maestro(
 
     private var screenRecordingInProgress = false
 
+    // A scroll/swipe can leave the app decelerating past the screen-static check, so the next tap
+    // must re-stabilise the element (MA-4124). Any tap clears it (see performTap) so ordinary taps
+    // on a settled screen skip the extra hierarchy fetches (MA-4135).
+    private var recentScroll = false
```
pedro18x, line 79: Found one more scroll source the flag can't see: on iOS, `hideKeyboard` dismisses by doing two real content drags at mid-screen (IOSDriver.kt:367-379), below this layer. So `inputText -> hideKeyboard -> tapOn "Submit"` gets the same momentum-outlives-the-check situation without ever setting the hint. Should be a one-liner in `Maestro.hideKeyboard()` if you agree it belongs in this PR.

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -205,15 +218,15 @@ class Maestro(
...
-                ?.toUiElementOrNull()
-        } else {
+        // A null settle can't confirm the screen is static. Scroll momentum is the one motion that
+        // routinely outlives the check, so re-stabilise only after a scroll (MA-4124); otherwise
+        // trust the hierarchy we have (MA-4135) — pre-MA-4124 behaviour for non-scroll animation.
+        // performTap clears recentScroll, so the hint only affects the tap that immediately follows.
+        val (hierarchyBeforeTap, refreshedElement) = if (settledHierarchy == null && recentScroll) {
             refreshElementUntilStable(element, initialHierarchy)
+        } else {
+            val hierarchy = settledHierarchy ?: initialHierarchy
```
pedro18x, line 235: Could we get a log line on this branch? It replaces the old "Refreshed element" log, and since the tradeoff section already predicts some stale-position reports, it'd be nice to be able to tell from logs which path aimed a given tap (settled / trusted pre-wait / stabilised). Both MA-4124 and MA-4135 were diagnosed from exactly these logs, so future-us will thank you.

maestro-client/src/main/java/maestro/Maestro.kt
```diff
@@ -73,6 +73,11 @@ class Maestro(
 
     private var screenRecordingInProgress = false
 
+    // A scroll/swipe can leave the app decelerating past the screen-static check, so the next tap
+    // must re-stabilise the element (MA-4124). Any tap clears it (see performTap) so ordinary taps
```
pedro18x: Tiny wording thing plus a question. The hint is only consumed by a tap, so it actually survives any number of non-tap commands (`scroll -> inputText -> assert -> tapOn` still stabilises), and even a `launchApp`, where the first tap on the new screen could burn the 3s deadline for a scroll that happened in a different app. Worth clearing in `launchApp`? And either way the "only affects the tap that immediately follows" comment probably wants a slight reword, it's really "the next tap, whenever that is".

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
```diff
@@ -5133,6 +5135,107 @@ class IntegrationTest {
...
+            bounds = Bounds(220, 400, 320, 460)
+        }
+        val driver = DeceleratingIosFakeDriver(root, driftStepPx = -12, driftStepsPerSwipe = 14)
+        driver.open()
+
+        Maestro(driver).use { maestro ->
+            runBlocking {
+                orchestra(maestro).runFlow(
+                    listOf(
+                        MaestroCommand(swipeCommand = SwipeCommand(direction = SwipeDirection.DOWN)),
```
pedro18x, line 5223: Nit: Case 145 guards against the swipe silently not reaching the driver with `assertAnyEvent { it is Event.SwipeWithDirection }`; might be worth the same line here so the test can't pass with a static screen.

maestro-client/src/main/java/maestro/Maestro.kt
proksh: Fixed in c882ddb4: reworded, and `launchApp` now clears the hint.

maestro-client/src/main/java/maestro/Maestro.kt
proksh: Good catch — set `recentScroll = true` in `Maestro.hideKeyboard()` in c882ddb4.

maestro-client/src/main/java/maestro/Maestro.kt
proksh: Added path logs (settled / trusted pre-wait / stabilised) in c882ddb4.

maestro-test/src/test/kotlin/maestro/test/IntegrationTest.kt
proksh: Added `assertAnyEvent { it is Event.SwipeWithDirection }` in c882ddb4.

maestro-client/src/main/java/maestro/Maestro.kt
proksh: Deliberate — Case 147 is intended. The `hierarchyAfterTap != null` gate can't work here: on iOS `waitForAppToSettle` returns null *precisely* on the settled path (IOSDriver.kt:505), so the hint would never clear and every post-scroll tap would re-stabilise — reintroducing the MA-4124 regression. Leaving the scroll→tap→tap edge to the core rework.

pedro18x (approved):
LGTM. All comments addressed in [c882ddb](https://github.com/mobile-dev-inc/Maestro/pull/3434/commits/c882ddb41aeace7cc4dc24655e5f750896fd386b), and the aim-path logs already read nicely in the e2e output. Agreed on the tap-clears-hint call: a non-null settle gate can't work when null is iOS's settled steady state, so scroll -> tap -> tap belongs with the core rework.

PR #3436, opened 2026-07-14
fix(ios): wait for the async crash .ips before collecting it

proksh:
iOS writes the crash `.ips` to `DiagnosticReports` asynchronously — a few seconds after the crash. `collectCrashArtifacts` looked once at flow end and almost always missed it, so `crash-report.txt` was absent from run artifacts.

`IOSCrashFileFinder.waitForCrashFile` now polls until the report appears or a timeout elapses. `IOSDriver` only waits when a crash was actually detected during the flow; other flows check once.

Fixes MA-4126.


Fishbowler (approved):
Love it. Love the respectful timeout handling calculation. Love the sensible approach to deciding a deadline based on how the test went. Love it ❤️

PR #3439, opened 2026-07-15
feat(cli): discontinue `maestro chat`, point users to Maestro MCP

proksh:
`maestro chat` (MaestroGPT) is sunset. This replaces it with a local stub pointing at [Maestro MCP](https://docs.maestro.dev/get-started/maestro-mcp), and removes the API call entirely.

Companion PRs: [copilot#2697](https://github.com/mobile-dev-inc/copilot/pull/2697) (serves the message, deletes the backend chatbot), [maestro-docs#192](https://github.com/mobile-dev-inc/maestro-docs/pull/192).

## Why cut at the CLI

The backend now returns a sunset message to anything calling `/v2/bot/message`, so this command would keep "working" — badly. Cutting locally gives an instant offline answer with no auth and no network round-trip. Today it prints "You must log in first" *before* it can tell you the feature is gone.

Old binaries keep hitting the backend and get the message there; that path can't be fixed, which is why the server-side message exists.

## Exit codes are split deliberately

```
maestro chat             → stdout, exit 0
maestro chat --ask "…"   → stderr, exit 1
```

`--ask` is the scripted form and may be in customer CI. Exiting 0 with prose where a script expects an answer hides the breakage until it surfaces somewhere confusing. A human typing `maestro chat` isn't scripting anything, so they get a clean exit 0.

`--api-key`/`--api-url`/`--ask` are still parsed but ignored, so existing invocations reach the message instead of a picocli "Unknown option" dump.

## Also removed

- `ApiClient.botMessage()` — `ChatCommand` was its only caller.
- `api/Chatbot.kt` — DTOs used by nothing else (`maestro-ai` has its own unrelated `MessageContent`; untouched).
- `README.md:114`'s MaestroGPT bullet. **Deleted, not repointed at MCP:** it sits under **Maestro Studio Desktop**, MCP is a CLI feature, and Studio's AI tab was removed in March — so it has advertised a nonexistent feature for four months.

## Verified against the built binary

| Invocation | Result |
|---|---|
| `maestro chat` | stdout, exit 0 |
| `maestro chat --ask "how do I tap"` | stderr, stdout empty, exit 1 |
| `maestro chat --api-key X --api-url Y --ask Z` | message, exit 1, no usage dump |
| `maestro chat --api-key X` | exit 0 |

No network call or login prompt in any case.


maestro-cli/src/main/java/maestro/cli/command/ChatCommand.kt
```diff
@@ -1,93 +1,46 @@
...
-import java.util.*
 import java.util.concurrent.Callable
 
 @CommandLine.Command(
     name = "chat",
     description = [
-        "Use Maestro GPT to help you with Maestro documentation and code questions"
+        "Discontinued. Use Maestro MCP instead: https://docs.maestro.dev/get-started/maestro-mcp"
     ]
 )
```
Fishbowler, line 13: Make the command hidden to reduce discoverability?

```suggestion
        "Discontinued. Use Maestro MCP instead: https://docs.maestro.dev/get-started/maestro-mcp"
    ],
    hidden = true
)
```

maestro-cli/src/main/java/maestro/cli/command/ChatCommand.kt
```diff
@@ -1,93 +1,46 @@
...
-    private var apiUrl: String = BASE_API_URL
+    @CommandLine.Option(order = 1, names = ["--api-url", "--apiUrl"], description = ["Ignored. maestro chat is discontinued."])
+    private var apiUrl: String? = null
 
-    @CommandLine.Option(
-        order = 2,
-        names = ["--ask"],
-        description = ["Gets a response and immediately exits the chat session"]
-    )
+    @CommandLine.Option(order = 2, names = ["--ask"], description = ["Ignored. maestro chat is discontinued."])
```
Fishbowler: I think all of the parameter description updates are irrelevant once you've called it discontinued at the top.

Fishbowler (approved):
Nitpicks below :)

maestro-cli/src/main/java/maestro/cli/command/ChatCommand.kt
proksh: Yes, nice!! Updated

Fishbowler (approved):

PR #3441, opened 2026-07-17
fix(ios): make simulator recording media duration match the movie duration

proksh:
## Problem

For iOS simulator runs, the same screen recording shows **two different lengths** depending on the player:

| Player | Length |
|---|---|
| Chrome `<video>` (e.g. Maestro Cloud run view) | 2:15 |
| QuickTime / the downloaded artifact | 2:30 |

It's one file with **two durations baked in**. `xcrun simctl io recordVideo` produces a **variable-frame-rate** `.mov`. When a flow ends on a static screen (e.g. a long closing `assertCondition`), simctl stops emitting frames, so the **sample timeline** (`stts`/`mdhd`) ends early while the **movie header + edit list** still describe the full length. Verified on a real recording:

```
mvhd (movie)   = 151.43s -> 2:31   # QuickTime honors this (holds last frame)
elst edit list = 151.43s
mdhd (media)   = 135.63s -> 2:15   # Chrome's <video> uses the sample timeline
```

## Fix

After the recording stops, extend the **final video sample** so the media duration equals the movie duration — then every player agrees.

- Pure **metadata edit** via **JCodec** (already a maestro dependency): split the last `stts` entry so the final frame is held, and update `mdhd` to match. `RelocateMP4Editor` rewrites only `moov`.
- **Lossless** — no frames are re-encoded, no quality loss, `mdat` untouched.
- **No external tools** — no `ffmpeg`, no per-platform binary; runs everywhere the JVM does (CLI, Studio, cloud worker).
- **Best-effort** — on any failure the file is left exactly as-is (`Mp4DurationNormalizer` swallows and logs).

Wired into `LocalSimulatorUtils.stopScreenRecording`, so CLI/Studio recordings are fixed. Cloud is covered by the same path: the run-scoped `screen-recording.mp4` is produced by core `ArtifactsGenerator` -> `Maestro.startScreenRecording` -> `IOSDriver` -> `SimctlIOSDevice` -> this same `LocalSimulatorUtils.stopScreenRecording`. So a submodule bump in `mobile-dev-inc/copilot` fixes Cloud too — no worker code change needed. (The worker's own `IOSDevice.startRecording/stopRecording` is unused in this path.)

## Validation

Unit tests (`Mp4DurationNormalizerTest`, no simulator/ffmpeg needed) — extends a short fixture to match, is a no-op when already consistent, and leaves a non-video file untouched. All green.

Ran against the **real** problem recording (1170x2532, ~2.5 min):

```
BEFORE  media=81379 (2:15)   movie=90856 (2:31)
AFTER   media=90856 (2:31)   movie=90856 (2:31)   # media == movie
```

Media clock now equals the movie clock -> Chrome and QuickTime both report 2:30, content untouched.


maestro-ios-driver/src/test/kotlin/util/Mp4DurationNormalizerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    fun `is a no-op once the recording is already consistent`() {
+        val file = fixtureCopy()
+        Mp4DurationNormalizer.normalize(file)
+        val afterFirst = mediaDuration(file)
+
+        Mp4DurationNormalizer.normalize(file)
+
+        assertThat(mediaDuration(file)).isEqualTo(afterFirst)
```
Fishbowler: Should this test check file metadata or something to assert that no file operations were done? Thinking about testing the contract rather than relying on what we know of its implementation. Or is that overkill?

Fishbowler (approved):
Looks great! I've independently verified the logic behind the cause and it looks solid. One test nitpick - feel free to ignore.

maestro-ios-driver/src/test/kotlin/util/Mp4DurationNormalizerTest.kt
proksh: Good instinct, but metadata won't work here. `normalize` rewrites the `moov` box every time - jcodec parses it, applies the edit, and writes it back with no "did anything change?" check. So the file gets touched even on a no-op, `mtime` always changes, and the assertion would fail.

I've added a byte-for-byte check alongside the duration one, which is the version of your idea that does hold: it says "the recording is unchanged" directly, without leaning on metadata.

PR #3443, opened 2026-07-17
Add resilience to the iOS webview e2e test

Fishbowler:
## Proposed changes

Increase timeout on a test that keeps flaking. An iOS device under load might not be rendering the page and updating the hierarchy in a timely way.

## Testing

Is CI :)


proksh (approved):

Fishbowler:
What're the chances - the e2e run revealed a second, brand new failure branch for this test that I've not observed before! Added a workaround for that too. The real fix (provable success of taps) is being worked on elsewhere, but this test isn't concerned with that.

PR #3445, opened 2026-07-17
fix(analytics): only attach user identity when logged in via `maestro login`

Leland-Takamine:
Reworks CLI analytics attribution so identity is only attached when a user
has explicitly logged in via `maestro login`. Fixes three bugs:

- CI runs with an org API key attached user emails to identities shared by
  all users on the same CI provider
- `maestro logout` kept attaching cached identity to events after logout
- A shutdown race silently dropped the last events of a run

Logged-out and CI usage is now fully anonymous (no person profiles);
logged-in events use the account user id, consistent with other surfaces.
No new data collected; opt-out unchanged.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
+++ b/maestro-cli/src/main/java/maestro/cli/analytics/Analytics.kt
@@ -7,6 +7,7 @@ import com.posthog.server.PostHogConfig
 import com.posthog.server.PostHogInterface
 import maestro.auth.ApiKey
 import maestro.cli.api.ApiClient
+import maestro.cli.util.CiUtils
 import maestro.cli.util.EnvUtils
 import org.slf4j.LoggerFactory
 import java.nio.file.Path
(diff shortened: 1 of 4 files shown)


pedro18x:
Nice one, the ingestion/billing comments are really useful. A few things:

- **`$identify` bypasses the opt-out**: `identifyAndUpdateState()` captures it unconditionally; the `enabled` / `MAESTRO_CLI_NO_ANALYTICS` checks only exist in `trackEvent`. An opted-out user still gets a person merge with email/name on `maestro login`. Worth gating here since this block is being rewritten anyway.
- **Every exit now pays ~500ms+**: `close()` runs drain + flush + sleep + flush even when analytics is disabled or nothing was captured. A "captured anything" flag would let you skip it all when there's nothing to send.
- **Mixed credentials**: `isInteractiveLogin` checks the login file, but `identifyUserIfNeeded()` prefers `MAESTRO_CLOUD_API_KEY`. With both present on a non-CI machine, events get identified as the env key's owner, which contradicts "API key runs stay anonymous". Adding an env-key-absent clause to `isInteractiveLogin` fixes it.
- **Logout race**: the async logout event can run `identifyUserIfNeeded()` after `clearUserState()`, re-persisting identity/org to `analytics.json` post-logout. Re-checking `isInteractiveLogin` right before `saveState` would close it.
- **Verify `$groups` works on personless events**: PostHog has limitations on `$process_person_profile: false` events, and CI org attribution depends on it. One empirical check would settle it.

PR #3447, opened 2026-07-20
Guard iOS device list from missing devicectl

Fishbowler:
## Proposed changes

If devicectl is missing or unavailable (e.g. old macOS version) don't crash device enumeration just because physical devices that we can't use can't be listed.

## Testing

This was observed in user logs. Hard to reproduce without purchase 🤔


pedro18x (approved):
LGTM

PR #3455, opened 2026-07-22
fix(orchestra): don't skip onFlowEnd when an onFlowComplete hook fails

proksh:
## Proposed changes

`onFlowComplete` hooks run inside `runFlow`'s `finally`, where only `DeviceConnectionException` was caught. Any other failure — e.g. a failing assertion in a teardown subflow — escaped the `finally`, skipping `dispatch("onFlowEnd")`.

`onFlowEnd` is where the bundle is finalized: device logs collected, `manifest.json` written. Skipping it means the device log is never captured and consumers reading the manifest upload nothing — and since teardown hooks mostly fail when the flow already failed, logs went missing exactly on the runs that needed them.

Now catches `Throwable`, records it, and rethrows after `onFlowEnd`. A teardown command failure still fails the flow (Case 109); `CancellationException` still propagates untouched.

## Testing

Two tests in `OrchestraListenerDispatchTest` (both fail on `main`): `onFlowEnd` is dispatched, and `manifest.json` is written, when an `onFlowComplete` hook fails.

`:maestro-orchestra:test`, `:maestro-orchestra-models:test`, `maestro.test.IntegrationTest` all pass — including Case 109 and the cancellation case.

## Issues fixed

MA-4143, MA-4144


Fishbowler (approved):
LGTM. I can't think of class of Throwable where this isn't worth trying. NPEs + NCDFs could still get us logs. OOMs and SOs might be recoverable if they're not device related.

PR #3456, opened 2026-07-23
fix(android): isolate unavailable devices during discovery

PointBreaker:
## Proposed changes

- Query the adb server for device states before constructing dadb connections.
- Skip `unauthorized` and `offline` transports so they cannot hide healthy Android devices.
- Isolate per-device connection races across list, ID lookup, discovery, emulator launch, and custom adb-server enumeration.
- Preserve direct emulator-port discovery when the adb server reports no devices.

## Testing

- `./gradlew :maestro-client:test -x :maestro-ios-driver:buildIosDriver`
- Added unit coverage for mixed device states, per-device connection failure isolation, and adb protocol errors.

This does not need an E2E flow because the regression is in host-side adb enumeration and is covered with deterministic protocol-level unit tests; it does not require app UI or device interaction.

## Issues fixed

Fixes Android device discovery returning an empty list when adb reports a mix of healthy and unavailable devices.

### DIFF
--- a/maestro-client/src/main/java/maestro/android/AdbServerDeviceDiscovery.kt
+++ b/maestro-client/src/main/java/maestro/android/AdbServerDeviceDiscovery.kt
@@ -1,0 +1,122 @@
+/*
+ *
+ *  Copyright (c) 2022 mobile.dev inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Fishbowler:
This is a lot of code for a one-line issue. Perhaps we could define the problem a bit more? There may be other solutions.

PointBreaker:
> This is a lot of code for a one-line issue. Perhaps we could define the problem a bit more? There may be other solutions.

Good point. The issue is that one unauthorized/offline device causes Dadb.list() to fail entirely, hiding healthy devices too. This probably belongs in dadb: filter by the state returned from host:devices, then only bump the dadb version here. Would that approach make more sense?

Fishbowler:
Ah, yeah, you're right 100%. That's definitely sounding like dadb's job.

PointBreaker:
> Ah, yeah, you're right 100%. That's definitely sounding like dadb's job.

Great, I’ll move the fix to dadb and follow up here with the dependency bump. Thanks!

PR #3457, opened 2026-07-23
fix(cli): don't retry an upload whose outcome is unknown

proksh:
`runMaestroTest` creates a fresh upload and one run per flow on every POST. The CLI retried any `IOException` or 5xx — but a read timeout or a gateway 502 means "we don't know", not "it failed", so we re-sent uploads the server had already accepted and multiplied the suite. A 357-flow upload became ~6,800 runs.

Now we retry only when the server cannot hold a complete request: no connection was established, or the body was never fully sent. Anything else stops and points the user at their project to check whether it landed.

Both conditions are needed — OkHttp retries connections internally, so a body written on one attempt latches the flag for a request that never arrived.

Also gives uploads a 15min read timeout and drops the 5min **call** timeout, which capped the entire transfer and put the backend's 4GB max upload size out of reach. Per-operation read/write timeouts still bound a stalled upload or a silent server.

This stops duplicates; it doesn't make retries safe. That needs an idempotency key the server can dedupe on, so an unknown outcome resolves automatically instead of by hand.

Tests: 4 cases in `ApiClientUploadRetryTest` (timeout after send, gateway 502, 4xx, pre-send failure still retries).

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -384,8 +384,8 @@ class ApiClient(
 
         fun abortAmbiguousUpload(detail: String): Nothing = throw CliError(
             "$detail\n" +
-                "The upload may already have been accepted — retrying automatically would risk starting " +
-                "every flow a second time. Check whether it landed before uploading again:\n" +
+                "A network error occurred during upload, and will not be retried.\n" +
+                "Your upload may still have been accepted:\n" +
                 TestSuiteStatusView.projectUrl(projectId, domain)
         )
 
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -18,6 +18,7 @@ import maestro.cli.runner.resultview.AnsiResultView
 import maestro.cli.util.CiUtils
 import maestro.cli.util.EnvUtils
 import maestro.cli.util.PrintUtils
+import maestro.cli.view.TestSuiteStatusView
 import maestro.cli.view.brightRed
 import maestro.cli.view.cyan
 import maestro.cli.view.green
@@ -38,8 +39,12 @@ import okio.ForwardingSink
 import okio.IOException
 import okio.buffer
 import java.io.File
+import java.net.ConnectException
+import java.net.UnknownHostException
 import java.nio.file.Path
 import java.util.Scanner
+import java.util.concurrent.TimeUnit
+import javax.net.ssl.SSLHandshakeException
 import kotlin.io.path.absolutePathString
 import kotlin.io.path.exists
 import kotlin.time.Duration.Companion.minutes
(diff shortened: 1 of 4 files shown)


maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -56,6 +61,15 @@ class ApiClient(
         interceptors = listOf(SystemInformationInterceptor()),
     )
 
+    // Uploads are the only call that legitimately runs for minutes, and a timeout there now aborts
+    // instead of retrying. No overall cap: a multi-GB binary can outlast any fixed budget, so the
+    // inherited per-operation timeouts do the work — writeTimeout catches a stalled upload,
+    // readTimeout a server that went quiet. Shares the connection pool.
+    private val uploadClient = client.newBuilder()
+        .readTimeout(UPLOAD_READ_TIMEOUT_MINUTES, TimeUnit.MINUTES)
```
Fishbowler, line 69: This looks sensible. The server will be acking during upload, then gives Maestro Cloud 15 minutes to process all of the tests after upload completes.

maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
```diff
@@ -366,6 +382,13 @@ class ApiClient(
             )
         }
 
+        fun abortAmbiguousUpload(detail: String): Nothing = throw CliError(
+            "$detail\n" +
+                "The upload may already have been accepted — retrying automatically would risk starting " +
+                "every flow a second time. Check whether it landed before uploading again:\n" +
+                TestSuiteStatusView.projectUrl(projectId, domain)
+        )
```
Fishbowler, line 390: Trying to make this sound more deterministic.

```suggestion
        fun abortAmbiguousUpload(detail: String): Nothing = throw CliError(
            "$detail\n" +
                "A network error occurred during upload, and will not be retried.\n" +
                "Your upload may still have been accepted:\n" +
                TestSuiteStatusView.projectUrl(projectId, domain)
        )
```

maestro-cli/src/test/kotlin/maestro/cli/api/ApiClientUploadRetryTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    fun `losing the connection after the request was sent does not re-upload`() {
+        // Same "we don't know what it did" state as a read timeout, without the wait. Enqueued
+        // repeatedly so OkHttp's own retries fail fast instead of blocking on an empty queue.
+        // Reaching the abort message proves our retry loop never ran — it throws a different one.
+        repeat(4) { server.enqueue(MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AFTER_REQUEST)) }
+
+        val error = assertThrows<CliError> { upload() }
+
+        assertThat(error.message).contains("may already have been accepted")
```
Fishbowler:
```suggestion
        assertThat(error.message).contains("may still have been accepted")
```

maestro-cli/src/test/kotlin/maestro/cli/api/ApiClientUploadRetryTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `gateway 502 does not re-upload`() {
+        // The proxy gave up waiting, not the backend — it may have finished creating every run.
+        server.enqueue(MockResponse().setResponseCode(502).setBody("Bad Gateway"))
+
+        val error = assertThrows<CliError> { upload() }
+
+        assertThat(error.message).contains("may already have been accepted")
```
Fishbowler:
```suggestion
        assertThat(error.message).contains("may still have been accepted")
```

maestro-cli/src/test/kotlin/maestro/cli/api/ApiClientUploadRetryTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `4xx still fails immediately without the ambiguity warning`() {
+        server.enqueue(MockResponse().setResponseCode(400).setBody("Bad workspace"))
+
+        val error = assertThrows<CliError> { upload() }
+
+        assertThat(error.message).contains("Bad workspace")
+        assertThat(error.message).doesNotContain("may already have been accepted")
```
Fishbowler:
```suggestion
        assertThat(error.message).doesNotContain("may still have been accepted")
```

Fishbowler (approved):

PR #3459, opened 2026-07-24
fix(orchestra): normalize command artifact paths and reject ones that escape the bundle

proksh:
## Problem

A `takeScreenshot` / `startRecording` path containing `..` — e.g. an unset variable leaving `startRecording: "../logs/screenshots/${VAR}"` — fails the flow with `.../startRecording/../logs/screenshots/.mp4 (No such file or directory)`.

`ArtifactCollector.allocate` resolves the path with `Path.resolve`, which keeps the `..` segment **literally**. `mkdirs()` then canonicalizes (collapsing `..`) and creates one directory, while the write opens the literal path under a directory that was never created → `ENOENT`. Because it surfaces as a raw `FileNotFoundException`, a cloud runner reads it as infrastructure and retries a path no retry can fix.

## Fix

1. **`ArtifactCollector`** — normalize the resolved path and `require` it stays inside the folder that owns it (`takeScreenshot/`, `startRecording/`), so `mkdirs()` and the write agree and escaping paths are refused.
2. **`CommandArtifactPath.validate`** — validate the flow-supplied path up front, throwing `MaestroException.InvalidCommand` for blank paths, paths naming no file, and — only when the run bundles artifacts — absolute paths or ones that climb out of the command's own folder.
3. **`Orchestra.artifactSink`** — convert an artifact-write `IOException` into `MaestroException.DestinationIsNotWritable`, so a runner classifies it as a flow error instead of retrying as infra.

## Behaviour change to note

Command output belongs in that command's folder in the run's debug output, so path rules are gated on whether the run collects artifacts into a bundle:

| Input | no bundle (local CLI) | bundled |
|---|---|---|
| blank / no file name | reject | reject |
| `..` that stays inside the command's folder (`login/../home`) | allow (unchanged) | allow |
| `..` that climbs out of it (`../clip`) | allow (unchanged) | reject |
| absolute path | allow (unchanged) | reject |

Under a bundle a path that climbs out of the command's folder cannot be honoured — the file lands next to `manifest.json` or inside the reserved `logs/`, where it is not that command's output. `..` that only walks back into a directory the path itself entered never leaves the folder and is allowed.

Two cases that previously "worked" now fail:

- **Absolute path under a bundle** was never absolute by the time it was resolved: `"startRecording/" + "/tmp/clip"` collapses the double slash, so `/tmp/clip` quietly landed at `<flow-dir>/startRecording/tmp/clip.mp4` rather than in `/tmp`. It is now an error rather than silently writing somewhere the flow did not ask for.
- **A path naming no file** (`logs/screenshots/`) produced a hidden `.png` / `.mp4` file. It is now rejected — **including local runs without a bundle**, the one rule here that is not gated on bundling. That is the reported bug, but it is a local behaviour change too.

`..` and absolute paths in local runs without a bundle are unchanged.


maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
```diff
@@ -0,0 +1,39 @@
...
+            throw invalid(path, commandName, "it names a directory, so there is no file name to write to")
+        }
+        if (!bundled) return
+
+        if (isAbsolute(path)) {
+            throw invalid(path, commandName, "an absolute path would fall outside this run's output bundle")
+        }
+        if (segments.any { it == ".." }) {
+            throw invalid(path, commandName, "\"..\" would take it outside this run's output bundle")
+        }
```
Fishbowler: Is this true?

`foo/bar/../baz/thing` would normalise to `foo/baz/thing` and wouldn't escape.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        val segments = path.split('/', '\\')
+        if (segments.last().isBlank()) {
+            throw invalid(path, commandName, "it names a directory, so there is no file name to write to")
+        }
+        if (!bundled) return
+
+        if (isAbsolute(path)) {
+            throw invalid(path, commandName, "an absolute path would fall outside this run's output bundle")
+        }
```
Fishbowler: Is the term `output bundle` friendly enough that folks will understand what to do about it? Is there a better term?

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactCollector.kt
```diff
@@ -74,7 +75,17 @@ internal class ArtifactCollector(private val runRoot: Path) {
...
+        records += Record(kind, format, underRunRoot(relativePath), metadata)
+    }
+
+    // Normalize and keep under [runRoot]. `Path.resolve` keeps `..` literally, so `mkdirs()`
+    // (which canonicalizes) creates a different dir than the write opens → "No such file or directory".
+    private fun underRunRoot(relativePath: String): String {
+        val resolved = runRoot.resolve(relativePath).normalize()
+        require(resolved.startsWith(runRoot) && resolved != runRoot) {
+            "Artifact path '$relativePath' resolves outside the run root"
+        }
```
Fishbowler: I'm not sure 'run root' is the best term for a user to understand. Perhaps 'artifacts directory' is the correct term here (assuming I've understood implementation correctly)?

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
Fishbowler: It's also true that this is a breaking change. Previously `/foo/bar/ would've resolved to `<output bundle>/foo/bar.png` and quietly worked (albeit a bit weirdly). This is a breaking change, but a good one, right?

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
proksh: Good catch, you're right and it's worse than the example: because the command path is nested under `takeScreenshot/`, even `../clip` normalizes to `<run-root>/clip.png` and never leaves the bundle. Only two or more `..` escaped, so the message was wrong for anything you'd actually try. 

Fixed pushed: the path is normalized and rejected only when it climbs out of the command's own output folder, so `foo/bar/../baz/thing` is now allowed and `../clip` is not. The collector enforces the same confinement.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
proksh: Agreed, updated the text.
 
And correct its a breaking change, but a good one. Added it to changelog so that we dont forget to mention this.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactCollector.kt
proksh: "artifacts directory" is also not the correct term now because it cant be outside the command output folder now. Updated text, let me know if its still not clear enough.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/CommandArtifactPath.kt
```diff
@@ -0,0 +1,44 @@
...
+            throw invalid(path, commandName, "the path is empty")
+        }
+
+        val fileName = path.split('/', '\\').last()
+        if (fileName.isBlank()) {
+            throw invalid(path, commandName, "it names a directory, so there is no file name to write to")
+        }
+        if (!bundled) return
+
+        if (isAbsolute(path)) {
```
Fishbowler: Some paths can be specified by variable. That'll only expand over time. Should we instead be testing whether an absolutely path specified matches the correct output directory?

Fishbowler (commented):
Last one :)

Fishbowler (approved):
LGTM.

I think there's some comments to tidy up - agent leaking the test specifics that triggered changes (login, clips, shots).

Is there further testing we'd wanna do ahead of release?

proksh:
> LGTM.
> 
> I think there's some comments to tidy up - agent leaking the test specifics that triggered changes (login, clips, shots).
> 
> Is there further testing we'd wanna do ahead of release?

Updated.

I was gonna do a bump after this merge, would that be fine? A path outside the folder should already be failing, so this change wont break anything new.

Fishbowler:
Sounds good to me!

PR #3395, opened 2026-06-30
Set locale on correct emulator after start-device

Fishbowler:
## Proposed changes

Stop the locale-setting happening to the "first" device, and instead to the correct device

## Testing

- Works launching a device with no locale
- Works launching a device with a locale
- Works launching a second device with a locale
- Works launching an already-running device, just changing the locale
- Works launching either of 2 already-running devices, just changing the locale

## Issues fixed

Fixes #2209

PR #3462, opened 2026-07-27
fix(android): don't block on the locale broadcast

proksh:
`setDeviceLocale` used to fire an ordered `am broadcast` and then wait for its result. Right after a snapshot resume the device's own GMS activity floods its broadcast queue, so that wait could hang all the way to the 120s socket timeout and show up as an `INFRA_ERROR`.

Now it checks the current locale first and skips the broadcast when it already matches. If it does need to change, it fires the broadcast detached and confirms the change with getprop instead of waiting on the ordered result. A busy queue just means a quick retry, not a stuck job.

The ticket had two more ideas I did not do:

1. Baking a default locale into the device image at build time. 
**_Not needed._** 
The device is almost always already `en_US`, and the skip adapts to whatever the device really is, so baking would add a build step and an image rebuild for no real gain.

2. Settle gate, waiting for the queue to calm down before firing. 
**_Not needed either._** 
We already wait on just this one change instead of freezing the whole job, and reliably knowing when the queue is calm is messy. With the skip and the detached fire it barely helps.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -64,6 +64,16 @@ import kotlin.io.use
 private val logger = LoggerFactory.getLogger(Maestro::class.java)
 
 /**
+ * How hard [AndroidDriver.setDeviceLocale] retries the detached locale broadcast while confirming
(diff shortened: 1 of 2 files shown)


Fishbowler (approved):
Looks sensible. Works on a few quick tests with `start-device` locally.

Fishbowler:
Related: #3395

Issue #1816, opened 2024-07-19
Maestro CLI unable to connect to XCUI test runner strted from Xcode


bartekpacia:
### Context

I was developing #1809, I needed to run my local build of `maestro-ios-xctest-runner`. I wanted to run the server directly from Xcode (reasons: debugging, easy to see logs) by:
1. `export USE_XCODE_TEST_RUNNER=1`
2. clicking the "Play" button in Xcode, next to the long-running test:

<details><summary>screenshot 1</summary>

![Screenshot 2024-07-19 at 11 05 29](https://github.com/user-attachments/assets/550442fe-eaf4-45f6-82a2-01763e83a233)

</details> 

However this does not work. It tries to connect to port 7001, which is incorrect on iOS. The correct port on iOS to connect to is 22087.

I added some debug logs (see PR #1815):

```
TestCommand.handleSession(): driver host port: 7001
MaestroSessionManager.newSession(): host=null, port=null, driverHostPort=7001
MaestroSessionManager.createMaestro(): driverHostPort=7001
MaestroSessionManager.createIOS(): driverHostPort=7001
```

and discovered this wrong 7001 port comes from here:

https://github.com/mobile-dev-inc/maestro/blob/83f7c1164a53fa5f6e56e3e6fd4abb33adfedad2/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt#L215-L218

<details><summary>screenshot 2</summary>

![Screenshot 2024-07-19 at 11 20 41](https://github.com/user-attachments/assets/0d03058b-0761-466f-8a9c-f4deb1fbe340)

</details> 

### Workaround 1

Just run out

1. 

### Workaround 2

Setting the port explicitly to the correct value (`22087` - port of the server running in XCTest) fixes the problem. To do it, replace this code:

https://github.com/mobile-dev-inc/maestro/blob/83f7c1164a53fa5f6e56e3e6fd4abb33adfedad2/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt#L215-L218

with:

```kotlin
val driverHostPort = 22087
```

The code that determines the port was added in #1732.

joepatosh:
I can confirm the behavior of maestro running on port `7001` instead of `22087`. I found this while investigating https://github.com/jpudysz/react-native-is-maestro/issues/16 and also tracked it down to the change in #1732 which adds sharding.

Here is my evidence to confirm this:

### First let's confirm that `1.38.1` uses port 7001:

[...]

PR #3464, opened 2026-07-28
Skip driver host port availability check when USE_XCODE_TEST_RUNNER is set

turkenh:
When USE_XCODE_TEST_RUNNER is set, the XCTest runner is managed externally and is expected to already be listening on the driver host port (LocalXCTestInstaller waits for it instead of launching it). Since 2.6.0, selectPort rejects an occupied --driver-host-port, which makes the externally-managed-runner workflow impossible: the port being in use is exactly the healthy state in that mode. This skips the availability check when the env var is set and changes nothing for default runs.

Related: #1816


Fishbowler:
This is to re-enable one of the workarounds listed in the linked issue?

turkenh:
Hmm, just noticed this solution is incomplete already. Let me think more throughly and get back.

PR #3465, opened 2026-07-28
Fix: Duration consistency in console output

Fishbowler:
## Proposed changes

Right now, `maestro test` durations are printed as whole seconds, whilst `maestro cloud` durations are fractional.

This updates all console displays durations as whole seconds for values over 1s.

## Testing

Additional unit tests + tested locally.


fabianoyoschitaki (approved):
LGTM!

PR #3467, opened 2026-07-28
Chore - Tidy up maps vs evaluateScripts

### LINKED ISSUE #3452: Extend interpolation facilities in Env.kt
### Use case

As part of #3428, Proksh noted that we've got repeated uses of Maps being coerced into `evaluateScripts` on primitive contents. It's fragile code that could use centralising and improving.

### Proposal

Extend Env.kt, where the `String.evaluateScripts` extension is, to include a `Map.evaluateScripts` and any other useful types. Use this in Commands.kt instead of reinventing the logic each time.

### Anything else?

_No response_

Fishbowler:
## Proposed changes

Adds utility functions for performing evaluateScripts on Lists and Maps, removing a bunch of duplicated logic.

Also reverts an accident in #3428 - the key names in permissions were never intended to be interpolated.

## Testing

Unit tested.

## Issues fixed

Fixes #3452


fabianoyoschitaki (approved):
LGTM

Issue #2576, opened 2025-07-09
More flexible Sellenium/Chrome setup in CI for --headless on web


tjsousa:
### Use case

We tried to run maestro --headless for web in self-runner hosts under an unprivileged user and it fails with message:

```
runner@71045d30d39f:~$ maestro test --headless --debug-output=/tmp/debug /tmp/blocks/simple-flow-example.yaml

Web support is in Beta. We would appreciate your feedback!


Could not start a new session. Response code 500. Message: session not created: probably user data directory is already in use, please specify a unique value for --user-data-dir argument, or don't use --user-data-dir
Host info: host: '71045d30d39f', ip: '192.168.215.2'
Build info: version: '4.26.0', revision: '8ccf0219d7'
System info: os.name: 'Linux', os.arch: 'amd64', os.version: '6.14.10-orbstack-00291-g1b252bd3edea', java.version: '21.0.7'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Command: [null, newSession {capabilities=[Capabilities {browserName: chrome, goog:chromeOptions: {args: [--remote-allow-origins=*, --disable-search-engine-cho..., --lang=en, --headless=new, --window-size=1024,768], binary: /usr/bin/google-chrome, detach: true, extensions: []}}]}]

The stack trace was:
org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: session not created: probably user data directory is already in use, please specify a unique value for --user-data-dir argument, or don't use --user-data-dir
Host info: host: '71045d30d39f', ip: '192.168.215.2'
Build info: version: '4.26.0', revision: '8ccf0219d7'
System info: os.name: 'Linux', os.arch: 'amd64', os.version: '6.14.10-orbstack-00291-g1b252bd3edea', java.version: '21.0.7'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Command: [null, newSession {capabilities=[Capabilities {browserName: chrome, goog:chromeOptions: {args: [--remote-allow-origins=*, --disable-search-engine-cho..., --lang=en, --headless=new, --window-size=1024,768], binary: /usr/bin/google-chrome, detach: true, extensions: []}}]}]
        at org.openqa.selenium.remote.ProtocolHandshake.createSession(ProtocolHandshake.java:114)
        at org.openqa.selenium.remote.ProtocolHandshake.createSession(ProtocolHandshake.java:75)
[...]

PR #3469, opened 2026-07-29
fix(web): disable chrome sandbox when running headless

giacomo-bersani:
## Proposed changes

Pass `--no-sandbox` to the Chrome driver when starting it in headless mode.

## Testing

Tested in scenarios in which the released version would fail (running web tests headless on Linux hosts, e.g. CI) to verify the issue was solved. On Debian 12 (Node 24 container), `libgbm1` and `libxkbcommon0` are also required (not having them installed will present the same opaque error as launching Chrome without the `--no-sandbox` flag).

Also tested on `e2e/demo_app/.maestro/web_flows` (`textarea` flow failed, existing failure).

## Issues fixed

This addresses https://github.com/mobile-dev-inc/Maestro/issues/2576. However, this is a rather inelegant solution compared to using the config file as proposed in https://github.com/mobile-dev-inc/Maestro/issues/2576#issuecomment-3378454460.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -125,6 +125,7 @@ class CdpWebDriver(
 
                 if (isHeadless) {
                     addArguments("--headless=new")
+                    addArguments("--no-sandbox")
                     if(screenSize != null){
                         addArguments("--window-size=" + screenSize.replace('x',','))
                     }


Fishbowler:
Should this only be when headless?

giacomo-bersani:
> Should this only be when headless?

I'm only setting the flag when headless since I'm addressing a CI use case and I wanted to keep the scope as narrow as possible. I don't have a Linux machine with a display on hand to check whether the flag is needed in that case too.

PR #3470, opened 2026-07-29
feat: support element-relative point on swipe.from

### LINKED ISSUE #3406: Support element-relative start point for swipe.from (parity with tapOn point)
### Use case

We maintain UI flows for a mobile app where content is shown as cards in a **horizontal carousel**. Each card is a composite component: an **internal image carousel** (photo pager) on top, and text content below.

When we swipe to move between cards, we target the card using its accessibility label:

```yaml
- swipe:
    from:
      text: "Desired card text"
    direction: LEFT
```

**Problem:** per the [swipe docs](https://maestro.mobile.dev/api-reference/commands/swipe), `swipe.from` always starts at the **center of the matched element's bounds**. On our card, that center falls inside the **photo pager**, not the text area. The gesture is consumed by the internal image carousel instead of advancing the outer card carousel. Tests fail or behave inconsistently.

Image

This is not an edge case in modern mobile UIs. Composite cards with nested scrollables, carousels, maps, and sliders are common. Accessibility trees often expose a single parent node for the whole card, so the geometric center of that node is frequently **not** the user's intended gesture origin.

**What we tried (and tradeoffs):**

| Workaround | Tradeoff |
|------------|----------|
| Swipe from a child element (e.g. a favorite button below the images) | Fragile: in our case, favorites feature is not available on every card, so it is not a reliable swipe anchor |
| Restructure app semantics solely for tests | Couples product a11y to test hacks; ongoing maintenance |
| Screen-coordinate `start`/`end` swipes | Breaks across screen sizes and devices ([#654](https://github.com/mobile-dev-inc/Maestro/issues/654)) |
| JS `relativePoint()` + coordinate swipes | Works for `tapOn` but cannot be reused for element-relative `swipe` ([#2149](https://github.com/mobile-dev-inc/Maestro/issues/2149)) |

**Why this matters for Maestro's command model:**

Maestro already recognizes that "center of element bounds" is not always correct for **taps**. Since v2.0.4, `tapOn` supports an element-relative `point` ([#2059](https://github.com/mobile-dev-inc/Maestro/issues/2059)), documented [here](https://maestro.mobile.dev/api-reference/commands/tapon#tapping-on-a-specific-point-within-another-element). The same class of problem now affects **swipes** on composite elements, but there is no equivalent capability on `swipe.from`.

The same "don't always start from center" theme appears across other commands ([#2009](https://github.com/mobile-dev-inc/Maestro/issues/2009), [#1651](https://github.com/mobile-dev-inc/Maestro/issues/1651), [#1775](https://github.com/mobile-dev-inc/Maestro/issues/1775)), suggesting users repeatedly hit this boundary when elements are larger than the interactive sub-region they care about.

We are happy to contribute a PR if the approach aligns with how `tapOn` + `point` was implemented.

### Proposal

Extend `swipe.from` to accept an optional element-relative `point`, using the **same format and semantics as `tapOn`** (`"50%, 85%"` = percentage within the matched element's bounds). When omitted, behavior stays unchanged (start from element center).

**Before (center only):**

```yaml
- swipe:
    from:
      text: "Desired card text"
    direction: LEFT
```

**After (proposed):**

```yaml
- swipe:
    from:
      text: "Desired card text"
      point: "50%, 85%"
    direction: LEFT
```

This lets us start the swipe in the text block below the photo carousel without anchoring on optional child widgets.

**Scope note:** A `distance` or element-relative `end` point ([#949](https://github.com/mobile-dev-inc/Maestro/issues/949), [#820](https://github.com/mobile-dev-inc/Maestro/issues/820)) would be valuable for sliders and pickers, but is **out of scope** for this request. We only need control over the **start** origin, mirroring the existing `tapOn` pattern.

### Anything else?

**Docs gap:** [swipe API](https://maestro.mobile.dev/api-reference/commands/swipe) explicitly states `from` starts at element center, while [tapOn API](https://maestro.mobile.dev/api-reference/commands/tapon) documents element-relative `point`. Closing this gap would make the command set more consistent and predictable for composite UI testing.

**Contribution:** I'm willing to open a PR following the `tapOn` + `point` pattern (YAML parsing, command model, orchestration, tests, docs). Happy to align on API naming (`point` vs `start`) before implementation.

Thank you for your attention and for your time.

PankovSerge:
## Proposed changes

`swipe.from` always started at the matched element's geometric center. On composite cards (nested image pager + text), that center often hits the inner scrollable, so the outer carousel never moves.

Adds optional element-relative `point` under `swipe.from` — same format/semantics as `tapOn` (`"50%, 85%"` or `"x, y"`). Omitted → center (unchanged). Element-relative `end` / `distance` are out of scope.

```yaml
- swipe:
    from:
      text: "Desired card text"
      point: "50%, 85%"
    direction: LEFT
```

**Implementation (mirrors `tapOn` + `relativePoint`):**
- `SwipeCommand.relativePoint` + YAML wiring from `YamlElementSelector.point` under `from`
- `Orchestra` resolves via existing `calculateElementRelativePoint`
- `Maestro.swipe(direction, startPoint, …)` overload → existing `Driver.swipe(elementPoint, …)`

**E2E fixture:** nested carousel screen in `e2e/demo_app` (`Carousel Test`) + `commands/swipe_from_point.yaml`.

## Testing

**Automated**
- `./gradlew test` (full suite green)
- New coverage: YAML reader fixtures (`030_swipe_from_point_*`), command description + serialization, IntegrationTest Case 149 (asserts non-center `SwipeElementWithDirection`)

**On-device (worktree Maestro via `installLocally`)**
- Android `Medium_Device_API_36`: TDD red (center swipe) → green (`point`); regression `swipe` / `scroll` (batch ×3)
- iPhone 15 / iOS 17.5: `swipe_from_point` (×3), plus `swipe`, `scroll`, `scrollUntilVisible`

**Docs:** [maestro-docs#196](https://github.com/mobile-dev-inc/maestro-docs/pull/196) — swipe API `from.point`.

## Issues fixed

Fixes #3406


Fishbowler (approved):

dl909-ai (commented):
Revoke.. Needs to be check and confirmed

PR #3481, opened 2026-08-03
fix(orchestra): reuse hierarchy on lookup failures

### LINKED ISSUE #3472: AssertionFailure.hierarchyRoot is never read, but costs a driver round-trip on every failure
## Summary

`MaestroException.AssertionFailure` (and therefore `ElementNotFound`, which extends it) requires a `hierarchyRoot: TreeNode` — a full view-hierarchy snapshot — at every throw site. **Nothing in this repository ever reads it.** Searching for reads of `hierarchyRoot` outside of assignments and constructor forwarding returns only `Errors.kt` passing it to `super`, plus one test asserting it *doesn't* appear in output.

That would be harmless if it were free, but populating it costs a device round-trip and retains a large object graph on every assertion failure.

## How we got here

[#36](https://github.com/mobile-dev-inc/Maestro/issues/36) (2022) observed that `hierarchyRoot` was an unused constructor parameter and asked for it to be exposed as a `val` so callers could read it. That's why it's public today. The reading side never materialised inside this repo, and in the meantime the field has acquired real costs.

## Cost 1 — a wasted driver round-trip per failure

`Maestro.viewHierarchy()` (in `maestro-client/src/main/java/maestro/Maestro.kt`) is uncached: every call is a fresh `ViewHierarchy.from(driver, …)` dispatched to IO, i.e. a real IPC/network call to the device. **13 sites call it inline purely to fill this parameter** — 12 in `maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt` and one in `Maestro.kt` itself:

```kotlin
throw MaestroException.ElementNotFound(
    maestro.viewHierarchy().root,   // round-trip for a field nobody reads
    debugMessage = exceptionDebugMessage
)
```

On a device that is typically 100–500ms per call, paid on every assertion failure. It compounds wherever failures are expected and swallowed: `optional: true` commands, retries, and `assertVisible` inside `repeat`/`retryCommand` loops. A flow that soft-fails fifty lookups pays fifty round-trips for nothing.

Not every site does this — the `childOf` branch of `findElement` passes the snapshot its timeout loop already fetched. So even within a single function the two branches disagree about whether to re-fetch.

## Cost 2 — the tree is retained, not just captured

`FlowDebugOutput` (in `maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/FlowDebugOutput.kt`) holds the thrown exception in a mutable field, so the exception — and the entire `TreeNode` graph hanging off it — stays reachable for the lifetime of the flow's debug output. For a dense screen that is a large retained graph per failure, with no consumer.

## Cost 3 — it was already found harmful and suppressed

`TestOutputWriter` (in `maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/TestOutputWriter.kt`) installs a Jackson mixin specifically to keep this field out of the debug bundle:

```kotlin
/** Errors serialize as message + debugMessage only — hierarchy lives in screen-hierarchy/, stack traces in maestro.log. */
@JsonIncludeProperties("message", "debugMessage")
private abstract class SlimThrowableMixin
```

and `maestro-orchestra/src/test/kotlin/maestro/orchestra/debug/ArtifactsGeneratorTest.kt` locks that in, asserting the serialized error does *not* contain `hierarchyRoot` or the tree's contents. So the tree is fetched, retained, and then explicitly filtered out at the one place it was headed — presumably after it bloated `commands.json`.

## Correctness wrinkle

Even for a hypothetical reader, the snapshot is taken at *throw* time — after the lookup timeout has already expired. On a moving UI that is a post-failure snapshot, not the state that caused the failure.

Meanwhile `ArtifactsGenerator` (in `maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt`) takes its own independent capture into `screen-hierarchy/step-NNN-<command>.json` on failure. That artifact is what our error messages mean when they say "check the UI hierarchy in debug artifacts". So there are two captures per failure, and the useful one is the artifact — the exception's copy is both later and unread.

## Before removing it

`hierarchyRoot` is a public `val` on a public class in `maestro-client`, and it was made public *deliberately* (see #36) so that external callers could read it. This repo is not the only consumer — `TestOutputWriter`'s own KDoc names "ArtifactsGenerator, the CLI's TestDebugReporter, the cloud worker" as callers of that write path, and the cloud worker lives outside this repository. A grep here cannot prove the field is unread in production.

**That check needs to happen before the field is deleted.**

## Proposed work

Two separable changes:

1. **No API change, safe now.** At throw sites where a hierarchy is already in hand, pass that instead of re-fetching. The `childOf` branch of `findElement` in `Orchestra.kt` already does this; the plain branch immediately below it is the obvious next one. This removes round-trips without touching any signature.
2. **Breaking, after the downstream check.** Delete the parameter and the field, dropping all 13 captures. Needs confirmation that the cloud worker does not read it, and a changelog note.

## One approach to avoid

Keeping the field but quietly passing an empty `TreeNode()` to stop the round-trips. That is worse than removal: a downstream reader would silently receive an empty tree instead of getting a compile error. If it goes, it should go loudly.

xianjianlf2:
Fixes #3472

### What changed

- Reuse the last hierarchy fetched by the plain element lookup loop when constructing `ElementNotFound`.
- Keep the existing fallback for a timeout that occurs before any hierarchy is fetched.
- Add a regression test that counts driver hierarchy requests and expects one request for a failed plain lookup.

### Why

The previous path delegated the lookup to `Maestro.findElementWithTimeout`, then fetched `viewHierarchy()` again only to populate the unused `hierarchyRoot` field. This change removes that second device round-trip without changing the public exception API or the childOf lookup path.

### Verification

- `git diff --check` passed.
- `./gradlew :maestro-test:test --tests maestro.test.ChildOfSelectorMessageTest` could not start because this environment has no Java Runtime.


Fishbowler:
I've checked the consumers. I think you can just delete `hierarchyRoot` altogether.

xianjianlf2:
Updated in `bf9a4855` per the review suggestion: removed `hierarchyRoot` from `MaestroException.AssertionFailure` / `ElementNotFound` and deleted the now-unneeded hierarchy arguments at the call sites.

Local validation:
- `git diff --check`
- static search confirms production code no longer passes/stores `hierarchyRoot` on assertion failures

I could not run the Gradle tests locally because this machine has no Java Runtime installed (`Unable to locate a Java Runtime`).

PR #3484, opened 2026-08-04
fix(orchestra): route the assertScreenshot diff through the artifact collector

pedro18x:
## What

Routes the `assertScreenshot` failure diff through the artifact collector, like every other command output.

- New `ArtifactKind.SCREENSHOT_DIFF` (+ v1 schema entry). Diffs land under `screenshotDiff/` in the bundle and are reported 1:1 in the manifest (never folded into a count entry), named `step-NNN-<reference>_diff.png` by the failing command's sequence number so two assertions against same-named references cannot overwrite each other.
- `allocateCommandArtifact` is now non-null: it owns the no-bundle CWD fallback, so call sites cannot reintroduce their own raw-`File` fallback paths. New `allocateScreenshotDiff` allocates the diff with the same funnel.
- New `onScreenshotDiffCaptured` callback (mirrors `onStepScreenshotCaptured`): fired when a failed command produced a diff, so hosts (the cloud worker) can attribute it per step without scanning the bundle by convention.
- A failed diff write now surfaces as `MaestroException.DestinationIsNotWritable`, like other command outputs, instead of a raw comparison-library exception.
- Guard test (`OrchestraArtifactFunnelTest`): scans `Orchestra.kt` for self-built path constructions (`File(...)`, `resolveSibling(...)`, `Paths.get(...)`, `.toFile()`) outside explicitly marked reads, so the next command output cannot silently bypass the funnel.

## Why

The diff was the last command output written as a raw `File`, placed beside the reference image. On Maestro Cloud the reference resolves into the workspace temp dir, which the bundle uploader never reads, so the one artifact explaining a failed visual assertion never reached the customer (mobile-dev's MA-4167, confirmed on a production run whose artifact registry had 5 types and no diff). Every historical artifact gap (recordings, logs, crash reports) had this same shape: an output written outside the collector, fixed one at a time. This closes the last one and adds the guard that keeps it closed.

## Behavior change (local)

The diff no longer lands beside the reference image:

- `maestro test` (bundle mode): in the debug output dir under `screenshotDiff/`, recorded in the manifest.
- No-bundle mode (`maestro test -c`): CWD-relative `<reference>_diff.png`, same fallback rule as `takeScreenshot`.

The failure message has always printed the diff's absolute path and still does; nothing documented referenced the old location (checked docs, e2e flows, maestro-docs).

## Schema note

`SCREENSHOT_DIFF` is added to the published v1 schema in place. The policy comment previously said enum additions are not additive; I verified every in-ecosystem manifest reader (worker, backend, CLI) is string-typed and tolerant of unknown kinds, and none validates against the JSON schema, so the comment now states the actual rule: prefer reserving values up front, and a late addition overwrites v1 accepting the cached-external-validator caveat.

## Verification

- `:maestro-orchestra:test`, `:maestro-orchestra-models:test` (incl. the bidirectional schema coverage test), and `:maestro-test:test` all green.
- New/updated tests: diff manifest shape (1:1, PNG, sequence-numbered path, sizeBytes); diff location + per-step callback attribution (sequence number pinned); non-null fallback semantics (CWD-relative, nothing recorded); the funnel guard.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -151,7 +151,7 @@ class TestCommand : Callable<Int> {
 
     @Option(
         names = ["--test-output-dir"],
-        description = ["Directory for this run's artifacts — manifest.json, commands.json, logs/, takeScreenshot/, startRecording/, and screenshotDiff/ written directly into it (overrides the default output location)"],
+        description = ["Directory for this run's artifacts — manifest.json, commands.json, logs/, takeScreenshot/, startRecording/, and assertScreenshot/ written directly into it (overrides the default output location)"],
     )
     private var testOutputDir: String? = null
 
--- a/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/TestCommand.kt
@@ -151,7 +151,7 @@ class TestCommand : Callable<Int> {
 
     @Option(
         names = ["--test-output-dir"],
-        description = ["Directory for this run's artifacts — manifest.json, commands.json, logs/, takeScreenshot/, and startRecording/ written directly into it (overrides the default output location)"],
+        description = ["Directory for this run's artifacts — manifest.json, commands.json, logs/, takeScreenshot/, startRecording/, and screenshotDiff/ written directly into it (overrides the default output location)"],
     )
     private var testOutputDir: String? = null
 
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/ArtifactManifest.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/ArtifactManifest.kt
@@ -13,6 +13,7 @@ enum class ArtifactKind {
     CRASH_REPORT,
     ANR_REPORT,
     AI_ANALYSIS,            // reserved; not emitted yet
+    SCREENSHOT_DIFF,        // assertScreenshot failure diff, one file per failing assertion
 }
 
 /** Concrete on-disk format of an artifact's bytes. ZIP is a download *view*, not a kind. */
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/ArtifactManifest.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/ArtifactManifest.kt
@@ -13,7 +13,7 @@ enum class ArtifactKind {
     CRASH_REPORT,
     ANR_REPORT,
     AI_ANALYSIS,            // reserved; not emitted yet
-    SCREENSHOT_DIFF,        // assertScreenshot failure diff, one file per failing assertion
+    SCREENSHOT_DIFF,        // assertScreenshot failure diffs, one folder collection entry (schema v2)
 }
 
 /** Concrete on-disk format of an artifact's bytes. ZIP is a download *view*, not a kind. */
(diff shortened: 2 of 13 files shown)


Fishbowler:
Would also need a [docs update](https://github.com/mobile-dev-inc/maestro-docs/blob/main/flows/workspace-management/test-reports-and-artifacts.md#whats-inside-the-artifact-folder)

pedro18x:
yes!

steviec:
Noticed two things in the PR, worked with claude to pinpoint:

---------

**Bump the schema to v2 instead of overwriting v1 in place.** `kind` is the one deliberately-closed part of an otherwise-open schema (`oneOf` of consts, and `ArtifactManifestSchemaTest` locks it to exactly mirror the enum). A stale strict validator rejects a new const — that's the whole reason the old comment said enum-adds aren't additive. Adding `SCREENSHOT_DIFF` to v1 in place re-incurs that risk for the next value too and mutates a published contract. Since readers are tolerant and don't validate, a `v2` path costs us almost nothing and keeps the frozen-v1 guarantee for any external validator. Let's own the closed schema and version it, rather than rewrite the policy to make in-place edits sound fine.

**Drop `singleFileCommandKinds` — represent the diff like every other command output.** We already have two repeatable command outputs, both handled the same way: `takeScreenshot` → `takeScreenshot/`, and `startRecording` → `startRecording/`, each one manifest entry with `count = N`. The manifest is a summary; the per-file detail (which file belongs to which step) lives in `commands.json` via `artifactsForStep(seq)`. This PR makes the diff a repeatable command output like those two, but reports it 1:1 in the manifest — the only command output that does — which is what forces the new bucket.

The 1:1 shape isn't needed. Per-step attribution is already covered by `commands.json`/`artifactsForStep` (same path `takeScreenshot` uses) and now by the new `onScreenshotDiffCaptured(seq, path)` callback. And MA-4167 only needs the diff in the bundle and recorded so the uploader ships it — folder+count already does that for the other two.

So:
- move `SCREENSHOT_DIFF` from `singleFileCommandKinds` into `collectionKinds` as `Collection(ASSERT_SCREENSHOT_DIR, PNG)`
- delete `singleFileCommandKinds` and revert the `collectionKinds[kind] ?: singleFileCommandKinds.getValue(kind)` fallback to `.getValue(kind)`
- rename the folder to `assertScreenshot/` (matches the `takeScreenshot`/`startRecording` command-output convention)
- keep the callback and the `step-NNN-<ref>_diff.png` name — the prefix still keeps same-named references from colliding and correlates the diff with its step's screenshot/hierarchy; that's orthogonal to how the manifest reports it

Net: one `assertScreenshot/` directory entry with a count, and anyone who needs the diff for a specific failed assertion reads `commands.json` or the callback — same as any other command output. One rule instead of two.

One thing to confirm first: whether the worker/backend reads the diff as a distinct manifest entry (or its `sizeBytes`) rather than going through `commands.json`. I'd bet no — the callback exists precisely so the worker can attribute diffs per step — but worth a check before collapsing it.

pedro18x:
Yeah, both corrections are right. The collection shape is what I originally settled on for the manifest: it says where file types live, commands.json does the per-step work. I drifted from that in this PR because the diff only exists for failing steps and I wanted the per-step attribution guaranteed, so 1:1 manifest entries looked like the direct way. Once the onScreenshotDiffCaptured callback went in, that reasoning stopped holding and I never circled back. One rule, not two: agreed.

What changed:

- SCREENSHOT_DIFF is now a collection under assertScreenshot/, same as takeScreenshot and startRecording. singleFileCommandKinds is deleted. A run with 3 diffs reports one entry with count: 3. Callback and file naming unchanged.
- Schema: v1 is back to exactly what's published, SCREENSHOT_DIFF lives in a new frozen v2, and the enum lock now points at v2.

Your open question: nothing reads the diff as a distinct manifest entry. The worker attributes per step via the callback, the backend's per-step URL comes from the worker's step report, and its manifest mapper already treats any row with a count as a collection. Backend change ended up test-fixtures only; the worker just renames its sync dir.

steviec (approved):
NICE

proksh (approved):

PR #3486, opened 2026-08-05
fix: waitToSettleTimeoutMs and swipe.duration reject ${...} templates

### LINKED ISSUE #3483: waitToSettleTimeoutMs and swipe.duration reject ${...} templates — crashes parser with "getFieldName(...) must not be null" when nested

### Steps to reproduce

  Steps to reproduce:
```
  appId: com.example
  ---
  - runScript: vars.js
  - runFlow:
      when:
        notVisible: ${output.anchor}
      commands:
        - scrollUntilVisible:
            element: ${output.anchor}
            direction: DOWN
            waitToSettleTimeoutMs: ${output.settleMs}
            timeout: ${output.timeoutMs}
  // vars.js
  output.anchor = "Some Text";
  output.settleMs = 2000;
  output.timeoutMs = 20000;
  Run: maestro test flow.yaml (no device needed to trigger it — fails before device connects).
```

 Include these three minimal variants, all reproduced:
  1. Top-level swipe: {duration: ${output.timeouts.swipe}} → Parsing Failed at file:7:1
  2. Top-level waitToSettleTimeoutMs: ${ENV_VAR} (plain env: var, not even JS) → Incorrect Format: waitToSettleTimeoutMs
  3. Same field nested inside runFlow.commands → crashes instead with Failed to parse file: ... getFieldName(...) must not be null, and — in a multi-file flow — GitLab CI's own error line points at the wrong ancestor file/line entirely, since the crash happens while eagerly resolving
  a transitively-file:-referenced subflow.

### Actual results

Either a schema-validation error (Incorrect Format) or, when nested inside runFlow.commands, an unhandled NPE that reports no field name and (in a multi-file project) misattributes the failure to an unrelated ancestor flow.

### Expected results

 maestro-timeout-template-bug maestro test 1-swipe-duration-top-level.yaml                     

Parsing Failed at /Users/smalugu/code/maestro-timeout-template-bug/1-swipe-duration-top-level.yaml:8:1

waitToSettleTimeoutMs and swipe.duration should accept ${...} templates the same way timeout: does (per the CLI's own cheat sheet, which documents timeout: ${output.constants.LONGTIMEOUT} as supported). At minimum, the nested-runFlow.commands case should never
  crash with an NPE — it should surface the same Incorrect Format error as the top-level case does.

### About app

Not app-specific — reproduces with zero device/app interaction, fails at flow-parse time before Maestro connects to anything.

### About environment

  - Java: OpenJDK 21.0.7
  - OS: macOS 26.5.1 (arm64)
  - Also reproduced on Maestro 2.8.0 (installed fresh to verify)

### Logs

 maestro-timeout-template-bug maestro test 1-swipe-duration-top-level.yaml                     

Parsing Failed at /Users/smalugu/code/maestro-timeout-template-bug/1-swipe-duration-top-level.yaml:8:1

```1-swipe-duration-top-level.yaml
appId: com.example
---
# Expect: Incorrect Format: duration
- runScript: vars.js
- swipe:
    direction: UP
    duration: ${output.settleMs}
```

### Maestro version

2.6.0

### Anything else?

  Maestro version: 2.6.0 (also reproduced on 2.8.0)

prolificcoder:
## Summary

- `waitToSettleTimeoutMs` (on `scrollUntilVisible`, `swipe`, `tapOn`) and `swipe.duration` were typed as `Int`/`Long` in both the YAML DTOs and the `Command` models, so Jackson coerced them at flow-parse time — before Maestro's `${...}` template engine runs (that only happens later, in `Command.evaluateScripts()`). Any templated value failed with `Incorrect Format`.
- When the offending field was nested inside `runFlow.commands` (a `List<...>`), Jackson's error `Reference` carries an index instead of a field name, and `MaestroFlowParser.wrapException` dereferenced `Reference.fieldName` as if it were non-null — crashing with a raw NPE (`getFieldName(...) must not be null`) instead of a clean parse error. That escaped NPE also caused the error to get mis-attributed to the parent flow file/line when the bug fired inside a `file:`-referenced subflow.
- Both fields are now `String`/`String?` end-to-end, wired into `evaluateScripts()` like `timeout`/`thresholdPercentage` already are, and parsed to their numeric value only at the point of use in `Orchestra.kt` — invalid post-interpolation values now raise a clear `MaestroException.InvalidCommand` instead of a parse-time crash.
- `wrapException` no longer assumes `Reference.fieldName` is non-null, falling back to the list index — this also fixes the wrong-file attribution.

Fixes #3483

## Test plan

- [x] `CommandsTest`: `waitToSettleTimeoutMsValue()`/`durationMs()` accessors on `ScrollUntilVisibleCommand`, `SwipeCommand`, `TapOnElementCommand`, `TapOnPointV2Command` — null/valid/invalid values, clamping to `MAX_TIMEOUT_WAIT_TO_SETTLE_MS`, and `evaluateScripts()` interpolation
- [x] `YamlCommandReaderTest`: top-level and nested-in-`runFlow.commands` `${...}` templates parse successfully for `waitToSettleTimeoutMs`/`duration`; a malformed field nested inside `runFlow.commands` now reports a clean `Incorrect Format` error instead of crashing
- [x] `MaestroCommandSerializationTest`: updated `duration` JSON fixture for the new `String` type
- [x] `./gradlew :maestro-orchestra:test :maestro-orchestra-models:test` green


prolificcoder:
@simon-gilmurray 
I Created (well claude) PR is above. I pushed back on moving a compile time constant of Int/Long to a String but looks like that is already an existing pattern with `timeout: String, thresholdPercentage: String, scrollDuration: String, point: String (tap coordinates)` for the ordering of Jackson and templating engine. 

https://github.com/mobile-dev-inc/Maestro/blob/main/maestro-orchestra/src/main/java/maestro/orchestra/yaml/YamlScrollUntilVisible.kt#L11 is another example. 

Let me know what you think.

simon-gilmurray:
Thanks for this - the team are currently working on Maestro Core improvements so will leave this for an engineer to review when they're ready!
I probably would have split the two separate issues in to two different PRs however.

PR #3487, opened 2026-08-05
prototype: route iOS assertVisible through maestro-device-core

steviec:
> **Status: experiment — not pursuing.** This was the iOS half of a "dual-tenancy" experiment: run legacy Maestro and maestro-device-core live on one simulator and route a single command (`assertVisible`) to device-core per-command, both runners co-resident. We've decided not to pursue dual tenancy on either platform — we're building device-core's full capability surface and integrating it as the primary runner (with a not-implemented fallback for commands it doesn't yet support). Keeping this for reference.
>
> **What it proved (iOS):** a real flow with one `assertVisible` served by device-core's `inspect()` while legacy's XCUITest runner stayed co-resident; the verdict matched the all-legacy run; a negative control (absent element) made device-core return not-visible and fail the flow at that step.
>
> **The reusable part, independent of dual tenancy — the routing seam.** `Orchestra.assertConditionCommand` diverts a standalone `assertVisible`/`assertNotVisible` to device-core; `DeviceCoreRouting` decides routability; `AssertVisibleVerdict` maps `ElementEvidence` → pass/fail. That seam is platform-agnostic and is exactly the hook a "route to device-core, else not-implemented" integration wants. Worth keeping even though the co-residence machinery around it isn't.
>
> **Both platforms, one branch.** This PR now includes the Android half too — the same routing seam plus a `UiAutomation` slot lease, folded in from the former `android-slot-lease` branch. It reached the same proof on a booted emulator (~0.8s per routed assert). The device-core side is preserved separately in maestro-device-core PR #88, which also documents the crux blocker if this is ever revisited (device-core needs to release its own slot when a transient `inspect()` session ends).
>
> **Not carrying forward:** the co-residence machinery (co-resident runners, the slot lease).

---

*Original PR description below, preserved as-is.*

> **Draft / prototype.** Off by default and depends on a private GitHub Package (see Caveats), so CI here won't build without a `read:packages` token wired in as a secret. Opening as a draft because the code is easier to review inline than in the terminal.

## Why

We want legacy Maestro commands to eventually be *decided* by maestro-device-core (its own resolver, its own device session) rather than by legacy's built-in resolver. This is the smallest honest proof of that on iOS: route real `assertVisible` commands, inside a real `maestro test` flow, to device-core's `inspect()` — with legacy's own XCUITest runner still live on the same booted sim, deciding everything else — and show the verdict provably came from device-core, not a pass-through.

## Approach

The divert happens at `Orchestra.assertConditionCommand`, so **only** a standalone `assertVisible:` / `assertNotVisible:` goes to device-core. `evaluateCondition`'s other callers (`runFlow` / `runScript` / `repeat`) are untouched, so `when:` / `while:` guards stay entirely on legacy.

New `maestro-orchestra/.../devicecore/` package, all pure and unit-tested:
- `AssertVisibleVerdict` — maps device-core's `ElementEvidence` → pass/fail. "Visible" = resolved + a MEASURED, fully-on-screen box. An infrastructure failure throws (never silently reads as a verdict); a genuine not-found → fail.
- `DeviceCoreRouting` — decides what may divert: only a standalone single **text** selector (literal, no regex metacharacters, no id, no relational/state constraints) on iOS. Everything else stays on legacy.
- `DeviceCoreAssertRouter` — calls `IosDeviceProvider().connect(...).screen.getByText(...).inspect()` and applies the adapter. Provider is injectable, so it's fully testable without a simulator.

`connect()` now **self-provisions** device-core's on-device driver server from binaries embedded in the jar (device-core [#84](https://github.com/mobile-dev-inc/maestro-device-core/pull/84)) — so a routed assert brings its own server up on the sim, the same way legacy `maestro test` brings up its XCUITest runner. No manual server step.

Wiring is additive: a new `Orchestra` ctor param defaulting to `null` (stock Maestro is byte-identical when off), built only behind env `MAESTRO_DEVICECORE_ASSERT=1` in both CLI flow-run paths. device-core is consumed as `dev.mobile.devicecore:prototype:0.1.0-SNAPSHOT`, resolved from a private GitHub Package (`maven.pkg.github.com/mobile-dev-inc/maestro-device-core`) with a `read:packages` token.

## Verification

- **Unit:** `:maestro-orchestra:test` — 447 tests, 0 failures. Covers the adapter, every rejection branch of the routing predicate, the router (error-wrapping, cancellation, not-visible), and the divert boundary (a routed assert consults device-core; a `runFlow` `when:` does not).
- **Live, interleaved, on a real app.** One `maestro test` run against Wikipedia (iPhone 15 Pro / iOS 17.5) interleaves **both drivers in a single run**, with device-core's server self-provisioned in code — no manual bring-up. The chronological `maestro.log` shows the split:

  | command | driver | evidence in log |
  |---|---|---|
  | `launchApp: { clearState: true }` | legacy | `Launching app org.wikimedia.wikipedia` |
  | `assertVisible: "Skip"` | **device-core** | `device-core decided … Resolved(TEXT), MEASURED Rect(80,790,28,16), verdict=true` |
  | `assertVisible: "Skip.*"` | legacy | RUNNING → COMPLETED, **no** device-core line (regex → not routed) |
  | `assertVisible: "Next"` | **device-core** | `device-core decided … verdict=true` |
  | `tapOn: "Skip"` | legacy | `maestro.Maestro.tap: Tapping on element … bounds [79,790][107,806]` |

  The same "Skip" element is seen by both drivers via different paths — device-core's snapshot (`Rect(80,790,28,16)`) and legacy's own UI tree (`[79,790][107,806]`) — corroborating the verdict.
- **Negative control:** `flow-negative.yaml` asserts an absent string; device-core returns `Absent(WHOLE_SCREEN) → verdict=false` and the flow **fails at that step** (exit 1). So the greens are device-core verdicts, not pass-throughs.

Confirm which asserts device-core decided: `grep -c "device-core decided" ~/.maestro/tests/<run>/maestro.log` — `2` for the positive flow (Skip, Next), and the regex assert is absent from it.

## Build it and run

**1. Get the device-core jar** — either a `read:packages` PAT in `~/.gradle/gradle.properties` (`gpr.user` / `gpr.read.token`), or, if you're iterating on device-core, `./gradlew publishToMavenLocal` there (Maestro resolves it local-first). The jar carries the on-device server binaries — no device-core checkout needed.

**2. Build the CLI:**
```bash
./gradlew :maestro-cli:installDist -x buildMcpViewer
export PATH="$PWD/maestro-cli/build/install/maestro/bin:$PATH"
```

**3. Boot a sim and install the app** (this Wikipedia build needs iOS ≥ 16.6):
```bash
xcrun simctl boot <UDID>
cd e2e && ./download_apps ios && ./install_apps ios && cd -
```
No manual driver server — device-core provisions its 8792 server itself on the first routed assert.

**4. Run:**
```bash
export MAESTRO_DEVICECORE_ASSERT=1
maestro test e2e/workspaces/devicecore/flow.yaml --device <UDID>
grep -c "device-core decided" ~/.maestro/tests/<latest>/maestro.log   # 2 = Skip + Next
```
`flow-negative.yaml` is the counter-proof (fails at the absent assert). Full walkthrough: [`e2e/workspaces/devicecore/README.md`](e2e/workspaces/devicecore/README.md).

## Caveats (deliberate prototype scope — not asking for fixes here)

- **Private-package dependency.** `maestro-orchestra` depends on `dev.mobile.devicecore:prototype:0.1.0-SNAPSHOT`, which resolves from a private GitHub Package on `mobile-dev-inc/maestro-device-core`. Any authorized org member can build it locally with a classic `read:packages` PAT (`gpr.user` / `gpr.read.token`) — no device-core checkout needed. CI can't build it until that token is a repo secret wired into `settings.gradle.kts`. That's the one thing to resolve before this touches `main`.
- **Local-first resolution (for lockstep dev).** `settings.gradle.kts` lists a scoped `mavenLocal()` *ahead* of the GitHub Packages repo, so `./gradlew publishToMavenLocal` in device-core is picked up without a package push. Empty `~/.m2` → falls through to the remote, so CI and pure consumers are unchanged. Trade-off is the usual mavenLocal one — a stale local jar shadows the remote until re-published or cleared.
- **Server lifecycle is intentionally hacky.** The router connects per assert and never closes, so the first routed assert launches the 8792 server and later ones reuse it (leaked-then-reused). Fine for a run; a proper open-once / close-at-teardown lifecycle (matching legacy's `driver.open()`/`close()`) is deferred.
- **App-under-test comes from a global property** (`devicecore.ios.bundleId`, set from the flow's `appId`), which pins one connection to one app. Fine for a single-app flow; the coherent model for specifying the app-under-test is being designed separately.
- **One-shot `inspect()`** — the routed path drops legacy's timeout/polling; the target must be already-rendered.
- **"Visible" is resolved + on-screen bounds**, not a measured visibility signal — it won't catch a present-but-hidden element yet.
- **iOS text selectors only** — device-core's `getById` isn't implemented on iOS today, so `id` selectors stay on legacy.

## Where to start reviewing

`maestro-orchestra/.../devicecore/*` (the whole divert), the `Orchestra.assertConditionCommand` diff, and the proof doc at `e2e/workspaces/devicecore/README.md`.


steviec:
Closing — dual tenancy isn't the direction we're taking (device-core is being built out as the primary runner with a not-implemented fallback). Left as a reference: the branch and description are preserved, and the reusable routing seam is called out in the body.

PR #3490, opened 2026-08-06
fix(orchestra): parse absent disableAnimations as unset instead of false

pedro18x:
## Problem

`WorkspaceConfig` today cannot represent "the user said nothing about animations":

- `PlatformConfiguration.AndroidConfiguration.disableAnimations` and `IOSConfiguration.disableAnimations` are `Boolean = false`, so a platform block that omits the key deserializes to an explicit `false`.
- The `platform` property's default constructs a non-null `PlatformConfiguration` with `disableAnimations = false` on both platforms, so even a config with no `platform` block at all materializes explicit `false` values.

After parsing, `disableAnimations: false` written by the user and no mention of animations at all are indistinguishable. Any executor that manages device animation state (Maestro Cloud disables animations by default for deterministic screenshots) cannot honor an explicit opt-out without also breaking the sensible default for everyone who never set the key, and vice versa.

## Fix

Make absence survive parsing as `null`:

- `disableAnimations` becomes `Boolean? = null` on both `AndroidConfiguration` and `IOSConfiguration`.
- `WorkspaceConfig.platform` defaults to `null` instead of a pre-populated `PlatformConfiguration`, so both absence paths (platform block present but key omitted, and no platform block at all) reach consumers as `null`.
- KDoc on the fields documents the contract: unset by default; executors that manage device animation state treat unset as animations-disabled for deterministic screenshots; local CLI runs do not modify the device's animation settings.

## Zero local behavior change

The CLI has no actuation machinery for `disableAnimations`: a repo-wide sweep finds no local consumer of the field. The only reader of `WorkspaceConfig.platform` is `MaestroSessionManager`, which accesses `platformConfiguration?.ios?.snapshotKeyHonorModalViews` through an already null-safe chain and never touches `disableAnimations`; a `null` platform yields the same `null` it got before from the key being absent inside the default instance.

The precedent here is `snapshotKeyHonorModalViews`, and it is stronger than mere pass-through: that field does actuate locally, with the null-check at the actuation point (`MaestroSessionManager` hands it to `LocalXCTestInstaller`, and `XCRunnerCLIUtils`/`LocalSimulatorUtils` set the env var only when the value is non-null). The codebase already treats `null` as "leave the platform default alone" at the point of use, which is exactly the contract this change gives `disableAnimations`.

## Tests

`WorkspaceExecutionPlannerTest` (real YAML parsing path):

- New fixture `019_platform_config_unset`: a platform block where Android omits the key and iOS sets `disableAnimations: false`. Asserts Android parses as `null`, iOS keeps explicit `false`, and untouched `snapshotKeyHonorModalViews` stays `null`.
- New test on an existing fixture without a `platform` block: `workspaceConfig.platform` is `null`.
- Extended the empty-config test: the constructor-default path also yields `platform == null`.
- Existing test 017 continues to prove explicit `true` round-trips.

`./gradlew :maestro-orchestra-models:test :maestro-orchestra:test` green; `:maestro-cli` and `:maestro-client` compile unchanged.

## Compatibility

`maestro-orchestra-models` publishes to Maven Central (`dev.mobile:maestro-orchestra-models`, currently 2.8.0). This change turns `getDisableAnimations()` from `()Z` into `()Ljava/lang/Boolean;`, and changes the constructor/`copy`/`componentN` signatures accordingly, so it is binary-incompatible for third-party consumers compiled against the old artifact. That is intentional: the nullable type is the point of the change, since only `Boolean?` can represent "unset".

There is also a user-visible consequence once a consumer picks this up: existing configs whose platform block omits `disableAnimations` (and configs with no `platform` block) will resolve as unset, which animation-managing executors treat as animations-disabled; today both materialize `false`.

## Design note

If the CLI ever grows local animation enforcement, it should adopt the executor-default contract documented on the fields (treat unset as animations-disabled where the executor manages animation state) rather than reintroducing a parse-time `false`.

### DIFF
--- a/maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
+++ b/maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
@@ -12,10 +12,7 @@ data class WorkspaceConfig(
     val baselineBranch: String? = null,
     val notifications: MaestroNotificationConfiguration? = null,
     @Deprecated("not supported now by default on cloud") val disableRetries: Boolean = false,
-    val platform: PlatformConfiguration? = PlatformConfiguration(
-        android = PlatformConfiguration.AndroidConfiguration(disableAnimations = false),
-        ios = PlatformConfiguration.IOSConfiguration(disableAnimations = false)
-    ),
+    val platform: PlatformConfiguration? = null,
     val testOutputDir: String? = null,
 ) {
 
@@ -42,11 +39,27 @@ data class WorkspaceConfig(
         val ios: IOSConfiguration? = null
     ) {
         data class AndroidConfiguration(
-            val disableAnimations: Boolean = false,
+            /**
+             * Unset by default. On executors that manage device animation state (e.g. Maestro
+             * Cloud), explicit `true` disables animations everywhere such executors manage them,
+             * explicit `false` makes the executor leave the device's animation state alone, and
+             * unset applies the executor default (animations disabled on Maestro Cloud for
+             * deterministic screenshots). Local CLI runs do not modify the device's animation
+             * settings.
+             */
+            val disableAnimations: Boolean? = null,
         )
 
         data class IOSConfiguration(
-            val disableAnimations: Boolean = false,
+            /**
+             * Unset by default. On executors that manage device animation state (e.g. Maestro
+             * Cloud), explicit `true` disables animations everywhere such executors manage them,
+             * explicit `false` makes the executor leave the device's animation state alone, and
+             * unset applies the executor default (animations disabled on Maestro Cloud for
+             * deterministic screenshots). Local CLI runs do not modify the device's animation
+             * settings.
+             */
+            val disableAnimations: Boolean? = null,
             val snapshotKeyHonorModalViews: Boolean? = null,
         )
     }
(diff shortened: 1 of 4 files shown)


maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
```diff
@@ -12,10 +12,7 @@ data class WorkspaceConfig(
     val baselineBranch: String? = null,
     val notifications: MaestroNotificationConfiguration? = null,
     @Deprecated("not supported now by default on cloud") val disableRetries: Boolean = false,
-    val platform: PlatformConfiguration? = PlatformConfiguration(
```
proksh, line 15: Previously Jackson filled an omitted platform block with the model default `(disableAnimations=false)`, so the cloud worker's `platformConfig?.android?.disableAnimations ?: true` resolved to `false` and animations left ON. 

After this change, an omitted or empty (`android: {}`) platform block deserializes to `null`, so the same ?: true now resolves to true and the worker disables animations.

proksh (commented):
This PR is changing the animation behaviour, is this intended?

maestro-orchestra-models/src/main/java/maestro/orchestra/WorkspaceConfig.kt
pedro18x: Yeah, intended. The worker already resolves unset animations as "disable" (DriverConfig.kt:45/48, `?: true`), but today Jackson fills an omitted platform block with disableAnimations=false, so that `?: true` never fires and animations stay on. There's a comment at DriverConfig.kt:34-36 flagging that exact defeat. This makes absence parse as null so the worker's intended default actually takes effect.

PR #3492, opened 2026-08-07
feat: capture host exception class chain in JsScriptError.causeClass

pedro18x:
## What

Add `causeClass: String? = null` to `JsScriptError`, populated from the originating host exception's cause chain (e.g. `java.net.SocketTimeoutException`), joined with " < ".

## Why

The JS `http` binding propagates OkHttp/java.net exceptions across the polyglot boundary as host exceptions. Their messages ("Connect timed out", "Connection refused") do not name the exception class, and `PolyglotException.cause` is null on the host-exception path, so downstream consumers that classify errors by exception class cannot tell a transport failure from a script failure. In Maestro Cloud this causes infrastructure blips to be billed to customers as test failures.

## How

`toJsScriptError()` reads the origin via `asHostException()` (falling back to `.cause` for guest exceptions), walks the cause chain bounded at 5 frames, and joins the class names. Resolved while the engine is still open, consistent with the existing detached-error design. Guest errors stay null.

## Test

`GraalJsEngineTest`: a real host exception via `engine.evaluateScript("http.get('http://127.0.0.1:1')")`, asserting `isHostException` and that `causeClass` contains "java." and "Exception"; plus a null assertion on the guest-error path. Deterministic: the connection is refused in milliseconds, or fails as `SocketTimeoutException` at the client's 10s cap, and the JDK default proxy selector excludes 127.*. No positional constructors exist (both call sites use named args), so inserting the field mid-class is source-compatible.

Consumed by the Maestro Cloud worker classifier (copilot branch `ma-4149-js-classifier`).

### DIFF
--- a/maestro-client/src/main/java/maestro/js/JsScriptError.kt
+++ b/maestro-client/src/main/java/maestro/js/JsScriptError.kt
@@ -11,6 +11,13 @@ import org.graalvm.polyglot.PolyglotException
 data class JsScriptError(
     val message: String,
     val causeMessage: String?,
+    /**
+     * Class names of the originating host exception's cause chain (e.g.
+     * "java.net.SocketTimeoutException"), joined with " < ". Null for pure guest
+     * errors. Messages alone often omit the exception class, so downstream error
+     * classification needs this to tell transport failures from script failures.
+     */
+    val causeClass: String? = null,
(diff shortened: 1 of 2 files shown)


proksh (approved):
We can reduce the size of comments here.

PR #1944, opened 2024-08-24
Allow for redacting/hiding sensitive test in `inputText` command

tokou:
## Proposed changes

- Add a `redact` boolean field to the `inputText` command
- In the case it's set to `true`, replace the value of `text` to `"[REDACTED]"` in the following locations:
  - The console output when `--format=NOOP`
  - `maestro.log`
  - `commands-(<flow>).json`

I chose to use `redact` instead of `mask` and `[REDACTED]` instead of `XXXX` as it seemed clearer but feel free to ask if you want it changed back.

## Testing

I added 3 test cases: 
 - The JSON serialization of the `InputTextCommand` (No deserialization though)
 - The correct parsing of the Yaml `redact` field
 - The correct input of the non-redacted text value

I ran the following commands:
- Run `./gradlew :maestro-test:test`
- Run `./gradlew :maestro-cli:test`
- Run `./gradlew :maestro-client:test`
- Run `./gradlew :maestro-orchestra:test`
- Run `./gradlew :maestro-orchestra-models:test`

### Manual tests

The tests below have been done manually. I tried seeing how to automate them, but ideally, this needs to be able to intercept the `LOGGER` calls in `IntegrationTest.kt`. Please advise if you want me to add tests for the cases below.

- Build the cli and then run `012_input_text.yaml`
  - Run `./gradlew maestro-cli:installDist`
  - Run `./maestro-cli/build/install/maestro/bin/maestro test maestro-test/src/test/resources/012_input_text.yaml`
  - Check for the absence of the value "secret password"
    - In the console output
    - In `~/.maestro/tests/<timestamp>/maestro.log`
    - In `~/.maestro/tests/<timestamp>/commands-(012_input_text.yaml).json` 
  - Replace "secret password" with "💸" in `012_input_text.yaml`
    - Run the flow `012_input_text.yaml` on an Android device
    - Check for the absence of the value "💸" in the same locations (console output, `maestro.log` and `commands-(012_input_text.yaml).json`)

Here are sample files with the results:

[commands-(012_input_text.yaml).json](https://github.com/user-attachments/files/16737631/commands-.012_input_text.yaml.json)
[maestro.log](https://github.com/user-attachments/files/16737632/maestro.log)
[output.txt](https://github.com/user-attachments/files/16737644/output.txt)

With "💸" replacing "secret password" on Android :

[commands-(012_input_text.yaml).json](https://github.com/user-attachments/files/16737642/commands-.012_input_text.yaml.json)
[...]

PR #3496, opened 2026-08-09
feat(inputText): add `redact` to keep secrets out of logs and reports

### LINKED ISSUE #1226: `inputText`: add a way to obscure sensitive text

Log outputs include all inputText. If a flow includes a login, the resulting maestro.log will expose that password. 

Perhaps there should be an API: `inputSensitiveText: password`

If this is the case, the value of the field should be omitted or censored (XXXXXXXXX) in logging. 

For example if I have a sensitive text of value: Password1, I currently input it using this:

```console
maestro test login.yml --env=PASSWORD=Password1
```

**test.yaml**

```yaml
...
inputText: ${PASSWORD}
...

```
and logging outputs this as:

```
[INFO ] m.cli.runner.MaestroCommandRunner - Input text ${PASSWORD} metadata CommandMetadata(numberOfRuns=null, evaluatedCommand=MaestroCommand(tapOnElement=null, tapOnPoint=null, tapOnPointV2Command=null, scrollCommand=null, swipeCommand=null, backPressCommand=null, assertCommand=null, assertConditionCommand=null, inputTextCommand=InputTextCommand(text=Password1), inputRandomTextCommand=null, launchAppCommand=null, applyConfigurationCommand=null, openLinkCommand=null, pressKeyCommand=null, eraseTextCommand=null, hideKeyboardCommand=null, takeScreenshotCommand=null, stopAppCommand=null, clearStateCommand=null, clearKeychainCommand=null, runFlowCommand=null, setLocationCommand=null, repeatCommand=null, copyTextCommand=null, pasteTextCommand=null, defineVariablesCommand=null, runScriptCommand=null, waitForAnimationToEndCommand=null, evalScriptCommand=null, mockNetworkCommand=null, scrollUntilVisible=null, travelCommand=null, assertOutgoingRequestsCommand=null, startRecordingCommand=null, stopRecordingCommand=null), logMessages=[])
[INFO ] maestro.Maestro - Inputting text: Password1
```

Perhaps we can update API and logging to look more like this:

```
maestro test login.yml --env=PASSWORD=Password1
```

**test.yaml**

```yaml
...
inputSensitiveText: ${PASSWORD}
...

```
and logging outputs this as:

```
[INFO ] m.cli.runner.MaestroCommandRunner - Input text ${PASSWORD} metadata CommandMetadata(numberOfRuns=null, evaluatedCommand=MaestroCommand(tapOnElement=null, tapOnPoint=null, tapOnPointV2Command=null, scrollCommand=null, swipeCommand=null, backPressCommand=null, assertCommand=null, assertConditionCommand=null, inputTextCommand=InputTextCommand(text=XXXX), inputRandomTextCommand=null, launchAppCommand=null, applyConfigurationCommand=null, openLinkCommand=null, pressKeyCommand=null, eraseTextCommand=null, hideKeyboardCommand=null, takeScreenshotCommand=null, stopAppCommand=null, clearStateCommand=null, clearKeychainCommand=null, runFlowCommand=null, setLocationCommand=null, repeatCommand=null, copyTextCommand=null, pasteTextCommand=null, defineVariablesCommand=null, runScriptCommand=null, waitForAnimationToEndCommand=null, evalScriptCommand=null, mockNetworkCommand=null, scrollUntilVisible=null, travelCommand=null, assertOutgoingRequestsCommand=null, startRecordingCommand=null, stopRecordingCommand=null), logMessages=[])
[INFO ] maestro.Maestro - Inputting text: XXXX
```

**Describe alternatives you've considered**

For my usecase, I am using GitHub actions and archiving the resulting failure logs from maestro. As a workaround, I can run some find and replace for any sensitive text before outputting the archive file.

Mohanad49:
## Proposed changes

Adds a `redact` option to `inputText`, so a flow that logs in stops writing its password into `maestro.log` and into the debug JSON report.

```yaml
- inputText:
    text: ${PASSWORD}
    redact: true
```

The value is replaced with `[REDACTED]` everywhere the command is *rendered* — the log line, the command description, `toString()`, and the serialized report — while the real text still reaches the driver.

This is a rebase of #1944 by @tokou, which reached review and was closed unfinished. @Fishbowler's closing note said *"If this is still of interest, please do resubmit it!"*, so this is that resubmission rather than a new proposal. The design is @bartekpacia's from the issue thread.

**One question before this is mergeable:** the issue thread suggests `mask: true`, #1944 implemented `redact: true`, and I've kept `redact` to stay close to the reviewed branch. Happy to rename it in a follow-up commit if you prefer `mask` — it's a one-line change plus fixtures.

### Three fixes on top of #1944

Rebasing surfaced problems that were in the original branch rather than caused by the 559 commits `main` has moved since:

1. **`redact: true` silently made the command optional.** `YamlInputText.parse` passed its values positionally into a constructor whose third parameter is `optional`:

   ```kotlin
   data class YamlInputText(text, label, optional, redact)
   return YamlInputText(input, label, redact)   // redact -> optional
   ```

   Both are `Boolean`, so it compiled. The effect is that a failing `inputText` step would have been skipped instead of failing the flow. Now passed by name.

2. **`getOrDefault("redact", "false") as Boolean`** threw `ClassCastException` whenever the key was absent, because the default was a `String`.

3. **The custom serializer omitted `optional`**, which every other command emits. That alone broke the existing `serialize InputTextCommand` test.

The branch also passed `redact` into `InputRandomCommand`, which has no such field — it did not compile. I've scoped this to `inputText` only, matching the issue. Happy to extend it to `inputRandom*` separately if that's wanted.

## Testing

- `./gradlew :maestro-orchestra:test :maestro-orchestra-models:test` — **449 tests, 0 failures**, on Temurin 17.
- Kept #1944's unit tests: the `redact` case in `MaestroCommandSerializationTest`, the `023_labels.yaml` reader fixture, and `012_input_text.yaml`.
- Updated the new serializer test's expected JSON to include `optional`, so it matches the shape every other command produces.

**e2e:** added `e2e/demo_app/.maestro/commands/inputText_redact.yaml`, using the existing Input/Keyboard screen, so no demo-app change was needed.

The assertion in it is the half that matters:

```yaml
- inputText:
    text: 'hunter2'
    redact: true
- assertVisible: 'hunter2'
```

Redaction must change what Maestro *reports*, never what it types. A bug that also redacted the input would make every login flow silently type `[REDACTED]` into the password field, and a test that only inspected the log would not catch it.

I couldn't run the e2e suite locally — it needs a device — so this was unverified on my side when I opened the PR. CI has since run it on **Android and iOS** and both pass.

## Issues fixed

Closes #1226. Supersedes #1944.


Mohanad49:
Rebased onto `main` — the branch had fallen behind and CI is re-running.

No rush intended, just flagging it per CONTRIBUTING's one-week guideline. Everything was green before the update: 14 checks including unit tests on Java 17 and the Android, iOS and Web e2e suites.

@Fishbowler — happy to make the `redact` → `mask` rename if that's the preference, or to split the three fixes I found in #1944 into their own commit if that reads better for review. Whichever is less work for you.

Mohanad49:
The red `Unit Test on Java 17` after the rebase is not from this PR — it's `AdbSocketTimeoutTest`, which this branch doesn't touch:

```
AdbSocketTimeoutTest > close fallback threads are bounded so a wedged transport
cannot spawn them without limit() FAILED
    com.google.common.truth.AssertionErrorWithFacts at AdbSocketTimeoutTest.kt:430
256 tests completed, 1 failed, 2 skipped
```

This PR only touches `maestro-orchestra*` and one e2e flow. The same job passed on this branch before the rebase.

I thought it might be a regression from `e008cc780` ("fetch WebView hierarchy as a serialized string"), since that lands in `DadbChromeDevToolsClient` and `AdbSocketTimeoutTest` covers the socket-close fallback for the same subsystem. That turned out to be wrong: #3510, #3511 and #3512 all sit on `6788dcc36`, which already contains `e008cc780`, and all three pass this job. So it looks like an ordinary flake — the assertion is on a thread count, which is the kind of thing a busy shared runner will trip occasionally.

I can't re-run the job from a fork. Would you mind kicking it, or I can push an empty commit if you'd rather.

Worth noting either way: because these workflows only run on pull requests, an intermittent failure in that test has no record on `main` to compare against, so it will keep surfacing as a mystery red X on whichever PR happens to catch it.

Mohanad49:
`Test on Android` is red for the same reason — a flow this PR doesn't touch:

```
[Failed] swipe (25s) (Assertion is false: "All green" is visible)
1/56 Flow Failed
```

`swipe` passed on iOS in the same run, and the other 55 passing-tagged flows passed on Android.

The flow this PR adds passed, and the log is the feature working:

```
[INFO]  Running flow inputText_redact
[INFO] Input text [REDACTED] RUNNING
[INFO] Inputting text: [REDACTED]
[INFO] Input text [REDACTED] COMPLETED
[Passed] inputText_redact (9s)
```

The literal typed into the field is `hunter2`; it appears nowhere in the output. And the flow asserts `hunter2` is visible on screen afterwards, so the redaction is confined to reporting and the driver still types the real text.

So both red checks on this PR are unrelated intermittents — `AdbSocketTimeoutTest` (a thread-count assertion) and now `swipe`. Both were green on this branch before the rebase. I'd still welcome a re-run when convenient.

Mohanad49:
Thanks for the rebase, @Fishbowler.

The one red check from it is `AdbSocketTimeoutTest`, and it's a timing race rather than anything on this branch. The failure is at `AdbSocketTimeoutTest.kt:430`:

```kotlin
assertWithMessage("the pool never saturated").that(entered.await(10, TimeUnit.SECONDS)).isTrue()
```

A 10-second latch waiting for the pool to saturate — so it fails when the runner is loaded, not when the logic is wrong. The later assertion about the cap being exceeded isn't the one that fired.

For what it's worth as evidence: #3510 passed this same job at 12:31, six minutes before this run at 12:37, and `Test on Android` went green on your rebase after failing on the previous one (that was `swipe`, also unrelated). The only `maestro-client` change here is an overload that logs a redacted string:

```kotlin
suspend fun inputText(text: String) = inputText(text, text)

suspend fun inputText(text: String, redacted: String) {
    LOGGER.info("Inputting text: $redacted")
```

Happy to rename `redact` to `mask` whenever you want it — that's still the one open question from my side.

Issue #141, opened 2022-09-07
Wait for result before Tap on point.


misterjedu:
After launching my App, I need to tap on a specific XY location on the home screen. But the view is already tapped on the splash screen, before the home screen loads. The same happens when there is a loading screen, and you need to tap on the location after the screen is loaded.

dmitry-zaitsev:
Could you perhaps share a video recording and a yaml file that you use so that we can better understand the problem?

dmitry-zaitsev:
Closing until more context is available

Issue #138, opened 2022-09-07
[Question] Contexts, Dynamics, And Test Reports


likeSo:
Hi there! I have a few questions:

- How can I read info from the current **context**? For example:
  - Read text from label A, pass the text to text field B
  - I need some data from the network
- How can I scroll or swipe specify component? such as swipe on an image view? Or double-tap it?
- How can I customize the timeout for waiting for a component to be visible?
  - My app took a long time to launch, the maestro just assert failed. I need the maestro to wait for a few seconds more time
- How can I find a component by class or type? Not all buttons have text
- Where can I find the test report?

Thank you!

likeSo:
- An element is already visible on the screen, why is the `tapOn` command still waiting for it?
- If there is a rich text component that contains a button with text, how can I tap on it?
- Why I can't tap a tab bar item by its label?
    - There is a UITabBarItem with the title "Homepage", I can not tap it by `tapOn: "Homepage"`, either `assetVisible`
- If there are so many same texts on the screen, how can I specify the index of `tapOn`?

dmitry-zaitsev:
> How can I read info from the current context?

This is not possible yet, but we have plans to add such support. Also mentioned in #106 

> How can I scroll or swipe specify component? such as swipe on an image view? Or double-tap it?

Not available yet, but will be added. Also mentioned in #55 

> How can I customize the timeout for waiting for a component to be visible?

At the moment, this is intentionally not customizeable as timeout is already very high (~10 seconds). I would like to better understand your use case - why is loading taking so much time?

> How can I find a component by class or type? Not all buttons have text

By design, Maestro is a black box testing framework and does not know about internals of the app. Think of it as a user with an impaired vision - Maestro sees the screen the same way such a user would through accessibility hints.

> Where can I find the test report?

Maestro does not generate test reports on its own yet. Would like to hear your feedback, do you see a need for such functionality?

dmitry-zaitsev:
> An element is already visible on the screen, why is the tapOn command still waiting for it?

There is a bit of a buffer between element taps to make sure the view hierarchy is stable and not going to suddenly change again.

> If there is a rich text component that contains a button with text, how can I tap on it?

A question to ask - how would a visually impaired user discover and tap on such elements? If the view is not showing up in `maestro hierarchy`, it will also not be easily accessibly to such a suer. Hence there are 2 options:

- Make the text view clickable as a whole
- Use dedicated button component instead of a rich text

[...]

PR #3500, opened 2026-08-10
Device-core validation harness: Orchestra→router over pluggable ExecutionBackend + device-core backend + mutually-exclusive driver seam

steviec:
## Why

We want to put a new device-core execution engine in front of real customer flows and see exactly where it agrees with maestro and where it doesn't — before it replaces any part of maestro's driver. The way to do that without a big-bang rewrite: carve Orchestra into a thin router over a pluggable `ExecutionBackend`, prove the relocated legacy path is behaviorally identical to stock main (a hard gate, not a code review), then run device-core through the same seam and measure the divergence per step.

This PR is the seam plus the measurement scaffolding. It is **not** a device-core rollout — device-core today implements roughly the P0 of its ROADMAP (`connect / getById / getByText / nth / inspect / tap / assertVisible`), so a solo device-core run hits a wall early on most flows by design: it hard-fails (throws a typed `BackendUnsupportedOperation`) at the first verb it can't yet serve, rather than soft-passing and drifting off the oracle. That's the point — measure fidelity on the steps it *can* run, honestly, and get a clean wall at the first one it can't. device-core is platform-equal here (iOS and Android are peers, same code path, opt-in by env var).

It's a draft because the on-device re-gate and the fidelity run are still in flight (see Verification).

## Approach

- **Phase 1 — router.** Orchestra becomes a router; every device-touching command goes through one `ExecutionBackend` seam. Above the line: flow control, variables, artifacts, reporting. Below: selector resolution, synchronization, driver calls.
- **Phase 2 — legacy backend + hard gate.** `LegacyExecutionBackend` is a verbatim relocation of today's behavior. A zero-divergence differential gate proves it produces byte-identical per-step traces to stock main across the Android corpus.
- **Phase 3 — differential harness.** `diff_traces.py` (verdict exact, element identity exact, coords ±2px; a device-core capability gap surfaces as a typed `error` in the trace — bucketed gap/infra/divergence — not a fake pass) + one consolidated runner behind a local-or-remote executor seam + `classify.py`.
- **Phase 4 — device-core backend + a mutually-exclusive driver seam.** The substance of this PR:
  - assertVisible/notVisible verdict comes from device-core's own `inspect().actionability.visible` + resolution — not a Maestro-side screen-geometry proxy (that proxy is deleted). An owed/absent visibility signal or an ambiguous/unavailable resolution throws through as ERROR rather than fabricating a verdict.
  - The same honesty generalizes past assert: **anything device-core can't yet serve — an unsupported command, a non-routable selector, an unhonorable tap/launch modifier — hard-fails with a typed `BackendUnsupportedOperation`, never a soft decline.** The old decline→PASS→continue path fabricated passes (a modified tap silently downgraded to a centered one, reported green) and, worse, let a declined *mutating* step corrupt every measurement downstream of it. Three honest exits now: gap (`BackendUnsupportedOperation` → ERROR), real divergence (`MaestroException` → FAIL), infra (`DeviceCoreUnavailable` → ERROR). `DeviceCoreUnavailable` means infra only; owed/ambiguous visibility moved to the gap bucket.
  - `DriverKind {MAESTRO, DEVICECORE}` is decided **before any provisioning**, from the statically-known platform + `MAESTRO_DEVICECORE_ASSERT` — because Android's `UiAutomation` is a singleton and maestro's driver and device-core's can't both hold it. The old decision read a live device RPC, which needed maestro's driver already open to decide whether to use maestro's driver at all; that chicken/egg is gone.
  - `backend.close()` is wired into `runFlow`'s teardown. device-core self-provisions its on-device server in `open()` and nothing was closing it — it leaked every run; now it's stopped symmetrically. Legacy's `close()` is a no-op (the session still owns `maestro`), so this is behavior-neutral for legacy.
  - Device-log / crash capture moves behind the seam as default no-op methods. Legacy delegates to `Maestro` (identical); device-core no-ops and logs the owed gap once. The seam is now the sole device path, and `maestro` is a nullable legacy handle threaded through the CLI chain (with a `requireMaestro()` guard on the legacy-only commands that genuinely need it).
  - Provisioning is `test`-scoped: `MAESTRO_DEVICECORE_ASSERT=1 maestro test` provisions device-core **only** — no `AndroidDriver`, no `UiAutomation` claim. Every other path, and every run without the env var, is exactly stock legacy.

## Verification

**Update (hard-fail + lifecycle fix + harness consolidation).** Dropped the decline machinery for the hard-fail above; wired `backend.close()` into `runFlow`'s finally (the device-core server leak); replaced the `declined` trace field with a lifecycle-derived `error{type,message}` the differential reads to bucket gap/infra/divergence; and consolidated the Python harness into one runner behind the executor seam (Remote transport folded into the seam, triplicated remote-run bash removed, committed evidence blobs and personal paths dropped). Legacy stays byte-identical by construction — the only runtime change in `runFlow` is the added `close()`, and `error` is omitted on PASS steps, so a legacy trace is byte-for-byte what it was. **516 JVM + 66 harness tests green.**

**Update (device-core integration — pinned build, four bandaids removed).** The four `maestro-device-core` fixes this seam was waiting on have landed, so the boundary workarounds are gone and device-core is pinned to one exact build (`0.1.0-6c30d8d97a45`):

- **adb serial.** device-core resolves the target serial once at `connect()` and binds every adb path — install, forward, `am instrument` — to it, fed by `--device` → the session's `instanceId` → `TargetSelector.serial`. Dropped the harness `ANDROID_SERIAL` export; `--device` is the sole multi-device disambiguation now.
- **iOS bundleId global.** iOS resolves the foreground app on-device, so `open()` no longer sets the `devicecore.ios.bundleId` process global — deleted. The app-under-test reaches device-core per-verb as an `AppId`.
- **Typed errors (decision 0011).** `executeLaunch` consumes device-core's own `DeviceEnvError` through an exhaustive `when` instead of a `catch (Exception)` catch-all: `UnsupportedOnTarget` → gap, `TransportFailure`/`OperationFailed` → infra. A launch that fails to complete is device-core unable to serve the step, never a wrong-answer FAIL.
- **git-sha pin.** The dependency moved off the floating `-SNAPSHOT` to a commit-addressed `0.1.0-<sha>`, so each build names one exact device-core jar.

`AssertVisibleVerdict` is unchanged — the assert path is device-core's hot-core read path, already sourced from `Resolution` + `EvidenceSource`; `DeviceEnvError` covers only the device-env verbs (`launchApp`/`openLink`/`stopApp`).

On-device (local emulators, no physical devices), **without `ANDROID_SERIAL` and with a second emulator attached**:
- Android Settings (deterministic): device-core provisions + runs 7/7 steps, **7 agree / 0 diverge / fidelityGreen**, pixel-identical element centers.
- GlossGenius (a real customer app): device-core provisions + launches it, **serves an id-based `tapOn` past launch**, then a clean typed wall (`BackendUnsupportedOperation`) at `waitForAnimationToEnd` — a gap ERROR, not a crash, no false divergence.
- The typed-error path itself: a real `DeviceEnvError.OperationFailed` carrying a `CliDiagnostic.renderRepro()` maps through `mapDeviceEnvError` to `DeviceCoreUnavailable`, exactly as designed.

**526 JVM + 81 harness tests green.** iOS runs locally too: `provisionBinaries` builds green on an off-pin (newer) Xcode via the drift-allow default device-core #138 wired in (a working non-canonical dylib for local mavenLocal — the canonical/pin story is device-core #130), so no toolchain block. On the simulator, device-core's **injected-dylib launchApp PASSES** (the bundleId-global removal validated end-to-end). iOS locate exists (#133) and `connect()` self-provisions the xctest runner. The live `activeApps()`/`snapshot()` read initially threw (`Resolution.Unavailable`); device-core #141 (extraction-cache fingerprint fix) resolves that — the read now succeeds. One gap remains, device-core-side: `assertVisible("General")` resolves `Absent` -> FAIL where legacy passes, consistently on iOS 16.2 and 26.5 (not flaky) — most likely the #133 foreground/SpringBoard fallback snapshotting the wrong app (device-core's `activeApps()` breadcrumb, added in #141, is where that's confirmed). Not a regression from this PR; the pin now tracks device-core `0.1.0-1c9bdd054852`.

- **Zero-divergence gate GREEN** for the legacy relocation vs stock main (Phase 2), over the ~38-flow Android replay corpus (DoorDash, crypto.com, Amazon/one-medical, Wealthsimple, and the rest) — at the pre-hard-fail HEAD. The re-run at this HEAD is pending (below).
- **Full JVM suite green** across the seam work — including the ~60 `ArtifactsGeneratorTest` cases unchanged, which is the proof the capture-behind-seam refactor didn't move legacy behavior.
- Legacy is byte-identical **by construction**: the nullable types carry the same values on the legacy path, capture delegates to the same `Maestro` calls in the same order, and device-core is dormant without the opt-in env var.

Still pending — this is why it's a draft:
- Re-run the legacy-vs-stock gate at this HEAD. The seam touched Orchestra (now the `close()` wiring too), the runners, and the session, so I want the on-device re-proof, not just the type-level argument. A 2-3 flow smoke first, escalating only if it diverges.
- Phase 5: the device-core fidelity differential — device-core solo, per reachable step, verdict + element vs the legacy oracle, hard-failing at the first owed verb, reporting per-flow reached depth. That produces the actual "where does device-core agree with maestro" number. Starting with a handful of iOS + Android flows to shake out what the harness itself still needs.

Please poke holes — especially at the mutually-exclusive driver seam (is `test`-scoping the right containment, or should device-core eligibility be modeled more explicitly?) and at the nullable-`maestro` threading.


steviec:
We're not using the router approach anymore, closing this.

PR #3503, opened 2026-08-12
fix(cli): retry upload-status polls that get no HTTP response (MA-4180)

proksh:
A dropped connection during status polling surfaces as `ApiException(statusCode = null)` and was fatal on the first occurrence, failing a green Cloud upload's CI. Now retried like the 5xx/404 cases, up to `maxPollingRetries`.

Tests: retry-then-succeed, and still-fails after budget exhausted.


pedro18x:
Good fix, and I like that one test covers retry-then-succeed and the other covers giving up.

One thing before merge. `retryCounter` is declared above the loop and nothing resets it after a poll succeeds, so `maxPollingRetries` is a budget for the whole wait rather than per failure. On a long run, five scattered drops get absorbed and the sixth still kills the CLI even though the run itself is fine, which is the same failure this PR is fixing, just later. Setting `retryCounter` back to 0 after each successful `uploadStatus` makes it behave the way the description describes. It also fixes the 500/502/404 branches, which have the same problem today.

proksh:
> Good fix, and I like that one test covers retry-then-succeed and the other covers giving up.
> 
> One thing before merge. `retryCounter` is declared above the loop and nothing resets it after a poll succeeds, so `maxPollingRetries` is a budget for the whole wait rather than per failure. On a long run, five scattered drops get absorbed and the sixth still kills the CLI even though the run itself is fine, which is the same failure this PR is fixing, just later. Setting `retryCounter` back to 0 after each successful `uploadStatus` makes it behave the way the description describes. It also fixes the 500/502/404 branches, which have the same problem today.

Nice catch, pushed the fix and test for that.

pedro18x (approved):
Looks right, the reset lands on every successful poll and the test would fail without it. LGTM.

PR #3504, opened 2026-08-12
fix(android): fetch WebView hierarchy as a serialized string to avoid CDP depth cap

proksh:
### Problem
`androidWebViewHierarchy: devtools` inspects Android WebViews over CDP. Two page-content failures were surfacing as retried `INFRA_ERROR` + triage noise (MA-4202 / MA-4205):
1. Deep external DOMs error out under CDP `returnByValue` — `-32000 "Object reference chain is too long"`.
2. When a WebView reply can't be parsed, the worker retried it 3× and alerted, even though the cause was the page's own content.

### Fix
- **Fetch the hierarchy as a JSON string in-page** (`JSON.stringify`) instead of `returnByValue`, so V8's depth cap is never hit; raise Jackson's nesting cap to 2000.
- **Classify a too-deep / too-large DOM as a test error, not infra**: a Jackson read-constraint trip, or a `StackOverflowError` from the recursive `TreeNode` bind on a small worker stack, now throws `MaestroException.WebViewInspectionFailure` → `TEST_ERROR` (in both `makeRequest` and `decodeSnapshot`) instead of a raw `IllegalStateException` the worker retries as `INFRA_ERROR`.
- A **null/empty content description** skips the WebView (native fallback) instead of hard-failing.
- **Ambiguous CDP error/event frames stay `INFRA_ERROR`** (retryable) — deliberately, pending CDP id-matching.

### Tests
Parity for deep/stringified snapshots, plus read-constraint / deep-nesting / stack-overflow / null-skip classification; `WebSocketServerStream` fixture extended to 16-bit frames.


pedro18x:
Nice fix, and the test that checks the stringified snapshot against the old shape is the right thing to pin.

One blocker. The JSON decode now happens inside the `degradeTo` block. `degradeTo` exists to skip a WebView when the connection to it breaks, so it catches `IOException`, logs a warning, and returns null. `readValue` throws `JsonProcessingException`, which is a subclass of `IOException`, so a snapshot that fails to parse now looks exactly like a broken connection: the WebView gets dropped and the run continues with only the native hierarchy. The user then hits "element not found" on every web element, with nothing in the log except a line blaming the devtools socket. `makeRequest` was written to stop that from happening: it catches `JsonProcessingException` and rethrows it as `IllegalStateException` so `degradeTo` cannot swallow it, and the decode used to sit inside that protection. Moving the decode back into `makeRequest`, or rewrapping the `readValue` call here in the same way, keeps the failure loud.

One question. The mapper does not raise Jackson's nesting limit, and Jackson 2.17.1 stops at 1000 levels by default. Each element costs two levels in the JSON (the node object plus its `children` array), so a page more than about 500 elements deep would now fail in the decode instead of in V8. Together with the point above, that failure would be silent. Do you know how deep the page was that produced the `-32000` error? If it is close to that limit, worth raising the constraint on the mapper and adding a test at that depth, otherwise the original bug may still be there in a quieter form.

proksh:
> Nice fix, and the test that checks the stringified snapshot against the old shape is the right thing to pin.
> 
> One blocker. The JSON decode now happens inside the `degradeTo` block. `degradeTo` exists to skip a WebView when the connection to it breaks, so it catches `IOException`, logs a warning, and returns null. `readValue` throws `JsonProcessingException`, which is a subclass of `IOException`, so a snapshot that fails to parse now looks exactly like a broken connection: the WebView gets dropped and the run continues with only the native hierarchy. The user then hits "element not found" on every web element, with nothing in the log except a line blaming the devtools socket. `makeRequest` was written to stop that from happening: it catches `JsonProcessingException` and rethrows it as `IllegalStateException` so `degradeTo` cannot swallow it, and the decode used to sit inside that protection. Moving the decode back into `makeRequest`, or rewrapping the `readValue` call here in the same way, keeps the failure loud.
> 
> One question. The mapper does not raise Jackson's nesting limit, and Jackson 2.17.1 stops at 1000 levels by default. Each element costs two levels in the JSON (the node object plus its `children` array), so a page more than about 500 elements deep would now fail in the decode instead of in V8. Together with the point above, that failure would be silent. Do you know how deep the page was that produced the `-32000` error? If it is close to that limit, worth raising the constraint on the mapper and adding a test at that depth, otherwise the original bug may still be there in a quieter form.

Nice push back, thanks for catching them.

Since this turned out to be a bigger issue than a one-liner, I went the other way first: added an e2e regression flow (`webView_chromeDevTools_deepDom`), a WebView whose DOM is nested 700 levels deep with an id buried at the bottom and confirmed it fails on CI before touching the fix, so we have a real reproduction on the record. It fails exactly the way you described: repeated Failed to retrieve WebView hierarchy from chrome devtools warnings, then `id: deep-dom-marker` never appears.

On your two points:

- The degradeTo swallow. Fixed.

- The depth. I reproduced it against the emulator's WebView. The page serializes to `~1025` JSON nesting levels (`~500` DOM nodes), which lands right on Jackson's default `1000` cap, so the original bug absolutely would have come back quietly in the decode. So I raised the mapper's maxNestingDepth to `2000` (`~1000 DOM` nodes, `2×` real content). One subtlety worth flagging: I measured that the recursive TreeNode bind StackOverflows at ~1300 nodes on a default-stack thread, so I deliberately kept the cap below that, an even deeper DOM now fails with a catchable exception rather than an uncatchable StackOverflowError.

With the fix in, the same e2e flow now passes. Added unit tests for both the deep-decode and the loud-failure paths too.

Ready for another look.

pedro18x (approved):
Nice!

PR #3508, opened 2026-08-13
fix(orchestra): close GraalJS engine after validating each flow

proksh:
### Problem
`WorkspaceValidator` creates a `GraalJsEngine` for every flow to evaluate its config scripts, but never closes it. Each engine owns its own GraalVM polyglot context, so validating many flows in a single long lived process leaks the contexts' compiled script state (Truffle shapes, function data, host interop `Method` caches). Memory grows without bound until the process runs out of heap.

This hits any process that validates workspaces repeatedly rather than exiting per run.

### Fix
Close each engine in a `finally` once its config scripts are evaluated, so the GraalVM context is released instead of accumulating.

### How it was found

I asked Claude to hunt for the leak by doing memory census of live pods and compare fresh vs long-running pods. What it did:

- Used `jmap -histo:live` 1 for memory census
- It took that census on a freshly-started pod and on one that had been up a long time.
- Spotted the linear-with-uptime growth. 
- Then worked with claude to reach exactly the root problem with code.

Live heap analysis of a long running process showed ~85% of the heap retained by GraalVM/GraalJS objects, scaling linearly with uptime: a fresh process held ~450k `java.lang.reflect.Method`, a 28h old one held ~9.5M, a 21x jump in lockstep across every GraalJS class. The only JS entry point on that path was `WorkspaceValidator`, whose per flow engines were never closed.

### Tests
Added `validate closes every JS engine it creates`, which injects a tracking engine factory and asserts every engine created during validation is closed. The new `jsEngineFactory` parameter defaults to `GraalJsEngine`, so callers are unaffected. Full `maestro-orchestra` suite passes (17/17 in `WorkspaceValidatorTest`).

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceValidator.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceValidator.kt
@@ -7,6 +7,7 @@ import maestro.orchestra.CompositeCommand
 import maestro.orchestra.MaestroCommand
 import maestro.orchestra.WorkspaceConfig
 import maestro.js.GraalJsEngine
+import maestro.js.JsEngine
 import maestro.orchestra.error.InvalidFlowFile
 import maestro.orchestra.error.SyntaxError as OrchestraSyntaxError
 import maestro.orchestra.error.ValidationError
@@ -59,6 +60,7 @@ object WorkspaceValidator {
         envParameters: Map<String, String>,
         includeTags: List<String>,
         excludeTags: List<String>,
+        jsEngineFactory: () -> JsEngine = ::GraalJsEngine,
     ): Result<WorkspaceValidationResult, WorkspaceValidationError> {
         return try {
             val allFlows = mutableListOf<ValidatedFlow>()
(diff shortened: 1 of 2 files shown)


pedro18x (approved):
Looks good!

amanjeetsingh150:
1. The PR looks like it would fix the issue but I also have few other questions:

- How were you able to narrow down to JS engine? Its hard to know the process from PR description
- Is this only retained leak or there are others?
- I have used hprof file in past but that was for workers, I wonder if we are able to write a skill and script around it that could help us doing "agent queries" and PRs directly using Claude.

Also it would not be system bound and can be used anywhere.

2. This second one is from Claude worth fighting against:

Why is `jsEngineFactory` a public method param? It's a test seam sitting on a published API where every caller now sees it. Orchestra already solves this as `internal val jsEngineFactory` (Orchestra.kt:151) — can we follow that and keep `validate()`'s public signature at 5 args, with the factory on an internal overload?

I haven't tried this yet but current public signature looks weird with factory since caller would think why'd i have to define things for WorkspaceValidator?

proksh:
> * How were you able to narrow down to JS engine? Its hard to know the process from PR description

I asked Claude to hunt for the leak by doing memory census of live pods and compare fresh vs long-running pods. What it did:
1. Used `jmap -histo:live 1` for memory census (Rather than pull a heavy `.hprof`, it ran a live histogram straight against the JVM inside the pod)
2. It took that census on a freshly-started pod and on one that had been up a long time. ~85% of the retained heap on the old one was GraalVM/GraalJS objects: Truffle nodes/shapes, `JSFunctionData`, and `java.lang.reflect.Method` caches.
3. Spotted the linear-with-uptime growth. ~450k `java.lang.reflect.Method fresh` → ~9.5M at 28h, a 21× jump in lockstep across every GraalJS class. GraalJS builds a reflective Method handle per polyglot context for each host method the scripts touch. Millions of retained Method, millions of live contexts never released.

Then worked with claude to reach exactly the root problem with code.

> * Is this only retained leak or there are others?

It was 85% of the problem. After the fix, compared a brand-new server and one that had run 66 hours. Nothing else keeps growing with time. The memory settles and stays flat.

> * I have used hprof file in past but that was for workers, I wonder if we are able to write a skill and script around it that could help us doing "agent queries" and PRs directly using Claude.
> Also it would not be system bound and can be used anywhere.

Make sense. I raised a PR for this - https://github.com/mobile-dev-inc/copilot/pull/2746.

> 2. This second one is from Claude worth fighting against:
> Why is `jsEngineFactory` a public method param? It's a test seam sitting on a published API where every caller now sees it. Orchestra already solves this as `internal val jsEngineFactory` (Orchestra.kt:151) — can we follow that and keep `validate()`'s public signature at 5 args, with the factory on an internal overload?
> I haven't tried this yet but current public signature looks weird with factory since caller would think why'd i have to define things for WorkspaceValidator?

Yeah true, nice catch. Raised a PR for fix here: https://github.com/mobile-dev-inc/Maestro/pull/3513.

PR #3510, opened 2026-08-14
fix(android): don't report an empty input's hint as its text

mrvissercb:
## Proposed changes

If an input has no text value, its text value is reported as the hint text (which is already available at hintText).

This fix will report it as "" if it is the hintText currently being displayed, which I believe is synonymous with the input having no value.


Fishbowler:
What's the residual risk of this breaking someone's existing flow?

mrvissercb:
> What's the residual risk of this breaking someone's existing flow?

If someone is working around this by checking `text === hintText`, I guess then their flow would break? Then they could change it to `text === ""` instead. A trivial fix to support a non-trivial gap.

In our case, AI trips over fields that say their text is valid when they're required, submissions fail, then it checks view hierarchy and sees the form is totally valid (text === hintText), checks logs and goes on wild tangents.

Fishbowler (approved):

PR #3511, opened 2026-08-14
Report invalid numeric field values as test errors

proksh:
## Problem

`index`, `point`, and `speed` can be set from JS variables. When a variable resolved to `undefined`, Maestro stringified it and did a raw numeric parse, throwing `NumberFormatException`. Because that's an unclassified runtime exception rather than a recognized flow error, a bad value in the flow surfaced as an opaque failure instead of a clear "this value is invalid" message.

## Fix

All three fields parse through one place (`NumericFields`) that raises a clear `InvalidNumericFieldValue` error instead of a raw `NumberFormatException`.

- Fields are discovered generically via a `@NumericField` annotation, so nested and composite commands are covered with no list to maintain.
- `WorkspaceValidator` runs the same checks at upload for literal values, so hard-coded mistakes fail before a device boots. Variable-resolved values are guarded at runtime by the same parser.
- A bad numeric value is treated as a syntax error: it fails the command hard, is never downgraded by `optional`, and no longer leaves the step stuck in `RUNNING`.

## Behavior changes

- **Scroll `speed` enforced to 0–100.** Out-of-range values (e.g. `speed: 500`) previously parsed and silently fell back to the default duration; they now fail with a clear error. Literals fail at upload, variable values at runtime.
- **Point bounds enforced at upload.** Hard-coded percent points like `150%,150%` and negative points like `-5,10` used to pass upload and only fail at runtime; they now fail up front. The absolute-pixel *upper* bound stays a runtime-only check (bounded by the live element).
- **`index` rejects `NaN`/`Infinity`.** Previously `NaN` silently became `0` and tapped the first element.

## Migration note

The scroll-speed 0–100 enforcement is a behavior change: flows that currently pass an out-of-range `speed` run today (the value silently defaults) but will hard-fail after this change. Anyone relying on out-of-range speed values should clamp them to 0–100. Percent-point enforcement is very unlikely to affect existing flows in practice.


maestro-orchestra-models/src/main/java/maestro/orchestra/util/NumericFields.kt
```diff
@@ -0,0 +1,116 @@
...
+        )
+
+    /** Parses `x,y` into two ints. `%` is stripped so callers keep their own range checks. */
+    fun parsePoint(raw: String): Pair<Int, Int> {
+        val coordinates = raw.replace("%", "").split(",").map { it.trim().toIntOrNull() ?: throw invalidPoint(raw) }
+        if (coordinates.size < 2) throw invalidPoint(raw)
+        return coordinates[0] to coordinates[1]
+    }
+
+    fun parseScrollSpeed(raw: String): Long = raw.toLongOrNull()
```
pedro18x: The message says "a number between 0 and 100" but only checks it parses as a `Long`, so `speed: 500` or `speed: -5` pass and feed `speedToDuration` an out-of-range value. Worth enforcing the bound so the message holds.

Minor: `pointFieldsOf`/`speedFieldsOf` are hand-synced, so a future command with a point/speed field not added here slips past this check (`index` avoids that by being generic).

proksh: Good catches, fixed both:

1. parseScrollSpeed now enforces 0..100 and throws `InvalidCommand` for out-of-range values, so speed: 500/-5 fail cleanly instead of silently hitting the default. The now-unreachable `duration < 0` fallback in speedToDuration is removed. Added bound tests.

2. Replaced the hand-synced `pointFieldsOf`/`speedFieldsOf` with a `@NumericField(kind)` annotation on the fields themselves. The reflective walk discovers them generically, same as index. New/nested commands are covered by annotating at the declaration. No list to keep in sync.

pedro18x:
The rework looks good, and I checked the annotation coverage against the old lists, nothing got lost. A few things though:

- speedToDuration now throws from evaluateScripts, which runs outside the per-command try in Orchestra, so a bad speed on an optional scrollUntilVisible kills the whole flow instead of downgrading to a warning (and the step gets stuck in RUNNING). Moving the parse to execute time like index/point fixes it.

- the opposite flip also happened: index/point parse errors are now InvalidCommand, which gets downgraded on optional commands, where before they failed the run. Might be what we want, but it changes behavior for existing flows so want to make sure it's intentional.

- heads up on the upload validation: it rejects the whole workspace on the first bad literal, including things like speed: 500 that upload and run green today. Worth collecting all errors instead of stopping at the first one, and thinking about how this lands for existing suites.

- small one: parseIndex accepts "NaN" (becomes 0, so it taps the first element) and "Infinity", worth rejecting those too.

proksh:
> The rework looks good, and I checked the annotation coverage against the old lists, nothing got lost. A few things though:
> 
> * speedToDuration now throws from evaluateScripts, which runs outside the per-command try in Orchestra, so a bad speed on an optional scrollUntilVisible kills the whole flow instead of downgrading to a warning (and the step gets stuck in RUNNING). Moving the parse to execute time like index/point fixes it.
> * the opposite flip also happened: index/point parse errors are now InvalidCommand, which gets downgraded on optional commands, where before they failed the run. Might be what we want, but it changes behavior for existing flows so want to make sure it's intentional.
> * heads up on the upload validation: it rejects the whole workspace on the first bad literal, including things like speed: 500 that upload and run green today. Worth collecting all errors instead of stopping at the first one, and thinking about how this lands for existing suites.
> * small one: parseIndex accepts "NaN" (becomes 0, so it taps the first element) and "Infinity", worth rejecting those too.

After digging in:

- **speed + index/point**: These are syntax errors, not command failures, so they should fail hard regardless of optional.  Fixed the index/point warn on optional. I made all three parsers throw a dedicated `InvalidNumericFieldValue` (subclass of `InvalidCommand`, so the worker still sees a test error) that the optional path never downgrades. Same rule for all three now.

- **upload fast-fail (whole workspace on first bad literal)**: I think stopping the upload is the right behavior, but it changes the story for suites that upload+run green today, so I'll confirm the rollout with @Fishbowler before we merge.

- **NaN/Infinity**: Fixed. `parseIndex` now rejects NaN/Infinity/-Infinity (NaN was silently becoming 0 and tapping the first element).

Regarding breaking change - we will need to notify Flightradar24, DoorDash and Blake's Company, as they are the only one using out of bound values right now.

maestro-orchestra-models/src/main/java/maestro/orchestra/util/NumericFields.kt
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    /** Static point check: [parsePoint] plus the runtime's 0..100 percent range (pixel bounds stay runtime-only). */
+    private fun validatePointLiteral(raw: String): Pair<Int, Int> {
+        val (x, y) = parsePoint(raw)
+        if (raw.contains("%") && (x !in 0..100 || y !in 0..100)) {
+            throw MaestroException.InvalidNumericFieldValue(
+                "Invalid point value \"$raw\": percentages must be between 0 and 100."
+            )
+        }
```
Fishbowler: Should this have an `else` covering both x and y being positive?
Maybe that's not this PR....

Fishbowler (approved):
Looks rational to me.
 
- I'm happy with this as a breaking change, so long as the folks we know it affects are contacted ahead of merge
- I think it would make sense that this PR includes a changelog entry in the unreleased section now that calls out what we need it to, so as not to forget it.

One nitpick below. Feel free to ignore it.

pedro18x (approved):

PR #3512, opened 2026-08-14
fix(android): confirm locale via blocking broadcast when getprop lags

proksh:
## Problem
[#3462](https://github.com/mobile-dev-inc/Maestro/pull/3462) fires the locale broadcast **detached** and confirms via `getprop persist.sys.locale`. But `getprop` can lag the reflection-based config update, so the locale *does* apply yet the poll never confirms → `SET_LOCALE_RESULT_UPDATE_CONFIGURATION_FAILED` → the cloud worker reports an `INFRA_ERROR`. This has been firing ~300× fleet-wide since #3462 shipped (2026-07-29). Infra-only — runs retry and pass on another host, no customer impact — but it's noise that trips retries/circuit-breakers.

## Fix
Keep the detached-fire + `getprop` poll as the fast path (still avoids the 120s ordered-broadcast hang #3462 fixed). When the poll doesn't confirm, fall back to the **old blocking, ordered `am broadcast`** and trust the receiver's authoritative `result=` code (re-adds `extractSetLocaleResult`). By then post-resume broadcast-queue churn has settled, so it rarely blocks.

- `result=0` → SUCCESS (fixes the false negative)
- non-zero → the receiver's real failure (genuine errors still surface)

## Tests
Assert on the applied-locale result, not on re-polling: getprop-lag + `result=0` → SUCCESS; `result=2` → failure (via the blocking broadcast).

Refs MA-4211 · regression from #3462 / MA-4105.

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -58,6 +58,8 @@ import java.util.Base64
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
 import javax.xml.parsers.DocumentBuilderFactory
 import kotlin.io.use
 
@@ -71,6 +73,7 @@ data class LocaleRetryPolicy(
     val maxAttempts: Int = 3,
     val verifyPolls: Int = 32,
     val pollIntervalMs: Long = 250L,
+    val blockingBroadcastTimeoutMs: Long = 15_000L,
 )
 
 /**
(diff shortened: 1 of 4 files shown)


maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
```diff
@@ -982,6 +990,12 @@ class AndroidDriver(
         return if (persisted.isNotEmpty()) persisted else shell("getprop ro.product.locale").trim()
     }
 
+    private fun extractSetLocaleResult(result: String): Int {
```
pedro18x, line 1001: Receiver codes (maestro-android) and `SET_LOCALE_RESULT_*` (maestro-client) are defined separately, and `RESULT_LOCALE_VALIDATION_FAILED = 3` has no constant here, so it falls to the generic "unknown exception" instead of locale-not-valid. Worth mapping 3 or sharing the codes.

Also, the blocking `am broadcast` has no timeout, so on a busy app or ANR the miss path can stall for the platform timeout (it used to return immediately). Intentional?

proksh: Good catches on both, pushed fix for both.

pedro18x:
Thanks, this looks good now. The code 3 mapping and the bounded broadcast both address what I raised. Two things I'd still fix though:

- result=0 from the blocking fallback can be a false success. The receiver's approximate-match path applies a different locale than the one requested, and it also reports success when it's missing permissions. Both of these came back as code 2 before this PR. Simplest fix is probably re-checking persist.sys.locale once after a result=0 fallback.

- on timeout, shellWithTimeout only calls shutdown(), so the am broadcast keeps running in the background: leaked thread plus an open adb stream, and the broadcast can still flip the locale after we already reported failure. There's boundedAdbCall in AdbSocketFactory that handles exactly this, or just cancel(true) with daemon threads.

proksh:
Pulling the fleet logs confirmed every hit is pure **getprop lag on a valid, non-default locale**: `en-CA` ~61%, then `it-IT`/`fr-FR`/`pt-BR`/etc., with nothing actually throwing. There are no genuine (invalid-locale / permission) failures in the population.

That undercuts this PR's core mechanism: the blocking broadcast exists to recover the receiver's authoritative result code, but here there's nothing for that code to distinguish. And the re-poll added on top could flip a genuinely-applied locale back to a failure, re-arming the exact false-negative we're trying to kill.

The simpler fix is to just wait out the late flip: after the detached retries, one longer `getprop` grace poll before declaring failure.

Thanks @pedro18x  for the review discussion that got us here. New Pr is here https://github.com/mobile-dev-inc/Maestro/pull/3516

PR #3516, opened 2026-08-19
fix(android): wait out a late locale flip with a grace poll

proksh:
## Problem

`setDeviceLocale` fires the locale broadcast detached and confirms via `getprop persist.sys.locale` within a ~24s budget (3×32×250ms). Under post-resume GMS/broadcast-queue churn the receiver runs late: the prop flips *just after* the budget, so the driver returns `UPDATE_CONFIGURATION_FAILED` for a locale that did apply → INFRA_ERROR.

Fleet data (production, arm-m2m): last 14d dominated by **en-CA ~61%**, then it-IT / fr-FR / pt-BR / fi-FI / en-GB / zh-CN — all valid, non-default locales. Pure getprop lag; nothing throws.

## Fix
Re-firing the broadcast can not cure a late-running receiver; only waiting does. After the detached retries, poll `getprop` once more over a longer **grace window** (default 120×250ms = 30s, spent only on the slow non-default path) before classifying as failure.

- No blocking/ordered broadcast (keeps the post-resume hang fix intact)
- No result-code plumbing
- Also skips the wasted trailing sleep after the final poll

### DIFF
--- a/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt
@@ -71,6 +71,7 @@ data class LocaleRetryPolicy(
     val maxAttempts: Int = 3,
     val verifyPolls: Int = 32,
     val pollIntervalMs: Long = 250L,
+    val graceVerifyPolls: Int = 120, // Waits out a late locale flip before giving up: 120 × 250ms = 30s, only on the slow path
 )
 
 /**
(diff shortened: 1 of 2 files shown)


pedro18x (approved):
This is a much better approach, removing the blocking fallback instead of patching around it. With success only reported when persist.sys.locale actually matches the target, the false-success paths from the old PR are gone, and there's no timed-out shell to leak. LGTM.

One non-blocking thought: the 30s grace window runs on every failure path too, including permanent ones like a broken receiver or a locale the receiver rejects, so a genuine failure now takes ~53s to classify instead of ~24s, with a getprop call every 250ms. Polling the grace window coarser (say 1s intervals) would get the same benefit cheaper. Fine as a follow-up.

PR #3521, opened 2026-08-20
fix(android): don't fail devtools WebView inspection on React/Vue/Angular circular DOMs

proksh:
## What

`androidWebViewHierarchy: devtools` could fail to inspect a WebView, aborting the step with:

`Failed after 4 attempts: Failed to parse DOM snapshot … Converting circular structure to JSON`

## Why

The WebView hierarchy is serialized in-page with `JSON.stringify`. React/Vue/Angular attach circular back-references to DOM nodes (e.g. `__reactFiber$…`, `__vueParentComponent`, `__ngContext__`). When such a node reaches the snapshot, `JSON.stringify` throws and the whole capture fails — surfacing as a retried infra error rather than something actionable.

## Fix

- Serialize with a cycle-safe replacer that drops live DOM nodes and breaks cycles, so the snapshot always completes. Normal and deep-DOM pages are unaffected (byte-identical output).
- If an in-page serialization error still occurs, classify it as a WebView-content (test) error instead of retrying it as infra.

## Tests

- Unit: an in-page serialization error is classified as a content failure, not infra.
- E2E (`webView_chromeDevTools_deepDom`): the demo screen now also contains a circular DOM node, so one flow covers both deep and circular DOMs. Fails without the fix, passes with it.

### DIFF
--- a/maestro-client/src/main/resources/maestro-web.js
+++ b/maestro-client/src/main/resources/maestro-web.js
@@ -79,6 +79,13 @@
 
     const isDocumentLoading = () => document.readyState !== 'complete'
 
+    // Read the identifier from the attribute, not the property: a form control named "id"/"name"/etc.
+    // clobbers node.id/node.name into a live element, which hides the real id and leaks a DOM node.
+    const attr = (node, name) => {
+        const value = node.getAttribute?.(name)
+        return typeof value === 'string' && value !== '' ? value : null
+    }
+
     const traverse = (node, includeChildren = true, iframeOffsetX = 0, iframeOffsetY = 0) => {
       if (!node || isInvalidTag(node)) return null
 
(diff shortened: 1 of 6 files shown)


maestro-client/src/main/resources/maestro-web.js
```diff
@@ -161,6 +161,21 @@
...
+        const seen = new WeakSet();
+        return (key, value) => {
+            if (typeof Node !== 'undefined' && value instanceof Node) return undefined;
+            if (value !== null && typeof value === 'object') {
+                if (seen.has(value)) return undefined;
+                seen.add(value);
+            }
+            return value;
+        };
+    }
```
Fishbowler, line 186: Is there any chance that this algorithm is too naive? This reads to me that we accept the first instance of an object, and reject all subsequent instances. Can it occur that the first instance is the reactFiber reference, and second is the "real" one? i.e. Could we accidentally fail a test because our hierarchy representation is now missing important elements?

pedro18x: The algorithm is dedupe-by-identity rather than cycle detection, and the `seen` set never gets unwound, so repeated references get dropped and not just cycles. It can't cost us hierarchy here though, since `traverse` allocates a fresh object per node so nothing in the snapshot repeats, and the fiber case can't happen either because we drop the element on the `instanceof` Node check before `JSON.stringify` ever descends into `__reactFiber$`. Where the concern does land is that Node drop: on the clobbered form case the new E2E builds, `node.id` is a live `<input>` rather than a string, so we drop it and the form comes back with no `resource-id`, meaning a flow selecting it by id would miss quietly instead of failing the capture. Worth type-guarding `node.id` / `node.name`/ `node.ariaLabel` the way `node.title` already is, or reading the attribute directly so we keep the real id.

proksh: @Fishbowler As Pedro explained, it won't cause a hierarchy loss.

Good catch about the clobbered case @pedro18x, fixed it with added test.

steviec (approved):
Looks great, merging.

PR #3522, opened 2026-08-20
Fixes CDP target selection for modern Chrome

### LINKED ISSUE #3506: Known upstream bug: inputText fails on Maestro Web (CDP/Selenium window desync)
Title: [Web] inputText fails with "Unable to locate element: id(...)" — ChromeDriver session attached to a stale/blank window handle

Body:

Environment
Maestro CLI: reproduced on both 2.8.0 and 2.6.1 (same failure regardless of CLI version)
--platform web, reproduced both headless and non-headless
macOS 26.6.1 (aarch64), Java 17 (Corretto)
Chrome 151.0.7922.137 (system Chrome; Selenium Manager auto-resolved ChromeDriver 151.0.7922.138 for it)
Reproduction
Minimal repro — happens on any page with a text input, not specific to our app. Also reproduced on a bare static HTML page with no framework at all (<input id="username">, no JS whatsoever).

url: https://example.com/login
---
- launchApp
- tapOn:
    id: "username"      # or css:, or a text selector — doesn't matter
- inputText: "hello"
tapOn/assertVisible on the same selector always succeed. inputText always fails (or hangs and eventually fails after a timeout, see below) with:

CommandFailed: no such element: Unable to locate element: {"method":"xpath","selector":"id("username")"}
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities {... goog:chromeOptions: {debuggerAddress: localhost:PORT} ...}
Root cause (confirmed)
A debug hierarchy snapshot (--debug-output) captured at the exact moment inputText fails shows "url": "data:," — a blank tab, not the page the preceding tapOn just succeeded on.
Maestro Web runs two independent automation clients at the same time: its own direct CDP channel (used for hierarchy/tap — has its own detectWindowChange logic and correctly tracks the active window/tab) and a separate Selenium/ChromeDriver session attached via debuggerAddress (used only for inputText's internal element re-lookup via a generated id(...) XPath — see CdpWebDriver.withActiveElement / createXPathFromElement in maestro-web.js).
That second (Selenium) client does not pick up the window-handle switch — it stays attached to the old/blank tab, so findElement on it naturally finds nothing.
Verified in isolation, outside Maestro entirely: plain Selenium 4.47 (webdriver.Chrome, --headless=new) against the same production page, using the exact same XPath id("username"), finds the element and successfully types via send_keys. So this is not a Chrome/ChromeDriver bug, nor an XPath id() bug — it's a Maestro-internal desync between its two clients.
Suggested fix (for maintainers)
Before inputText (and likely eraseText) does its element re-lookup, it should either (a) switch the Selenium session to the window handle the CDP client currently considers active, or (b) drop the dependency on Selenium's findElement for this step entirely and type into the already-focused element purely via CDP Input.dispatchKeyEvent, the same way tapOn already does.

Fishbowler:
Proposed changes

Recent Chrome versions, such as Chrome 151, expose an increased number of Chrome DevTools Protocol (CDP) targets, including browser UI components, extension workers, and iframes, often listed before the actual web page under test. Previously, Maestro would simply select the first available target, which led to incorrect command execution (e.g., JavaScript evaluation, screenshots, URL navigation) on the wrong target.

This change introduces a more robust CDP target selection mechanism:
- **Prioritizes the active Selenium window:** Maestro now identifies the correct target by correlating it with the `windowHandle` reported by Selenium, ensuring that both Selenium-driven and CDP-driven commands operate on the same page.
- **Filters out irrelevant targets:** Non-"page" type targets (like `browser_ui`, `service_worker`) and browser-internal URLs (`chrome://`, `chrome-extension://`) are explicitly ignored to prevent misdirection.
- **Adds CDP response timeout:** A timeout prevents indefinite hangs when CDP commands are inadvertently sent to unresponsive targets that do not return a response, such as non-visible or internal browser pages.

These changes ensure reliable interaction with web pages, especially against newer Chrome browser versions, making web tests more stable and accurate.

Testing

New unit tests have been added in `CdpWebDriverTest.kt` to thoroughly verify the `selectTarget` logic, covering various scenarios for target identification and filtering.

Issues fixed

Fixes #3506

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -246,7 +246,8 @@ jobs:
       - name: Set up Chrome
         uses: browser-actions/setup-chrome@v2
         with:
-          chrome-version: 142
+          # Kept at a pinned major rather than `stable`, so a Chrome release cannot fail unrelated PRs.
+          chrome-version: 151
 
       - name: Download artifacts
         uses: actions/download-artifact@v8
--- a/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
+++ b/maestro-client/src/main/java/maestro/drivers/CdpWebDriver.kt
@@ -1,6 +1,7 @@
 package maestro.drivers
 
 import CdpClient
+import CdpTarget
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import kotlinx.coroutines.runBlocking
 import maestro.Capability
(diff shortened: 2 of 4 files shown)


simon-gilmurray (approved):

Fishbowler:
[MA-4220](https://linear.app/mobile-dev/issue/MA-4220) is fixed by this

PR #3523, opened 2026-08-21
Add opt-in step screenshot and hierarchy capture

ayanamists:
## Why

Maestro captures artifacts for failed steps and `--analyze`, but local and CI consumers cannot request the same artifacts independently.

We are integrating Maestro with a test-reporting system that needs per-step screenshots and accessibility hierarchies without enabling AI analysis.

## Existing behavior

- Successful steps normally have no per-step artifacts.
- Failed and warned steps capture an outcome screenshot and hierarchy.
- `--analyze` captures pre-step screenshots, a full-run recording, and `screenshots/final.png`.

## Changes

This PR adds three opt-in CLI flags:

```text
--[no-]capture-step-screenshots
--[no-]capture-step-hierarchy
--capture-all-step-artifacts
```

The first two flags independently enable screenshots and hierarchies. The convenience flag enables both.

Requested artifacts are captured before each visible step:

```text
screenshots/step-003-launchApp-com.android.settings.png
screen-hierarchy/step-003-launchApp-com.android.settings.json
```

Each step artifact is referenced by its command in `commands.json`. When screenshots and hierarchies are both enabled, the pair uses the same command sequence number and filename stem.

The requested artifact types are also captured once at flow end:

```text
screenshots/final.png
screen-hierarchy/final.json
```

`final.png` already exists in the `captureFullArtifacts` path. This PR uses the same final-state model for opt-in screenshot capture and adds the corresponding `final.json` when hierarchy capture is enabled.

The final artifacts represent the flow-end boundary rather than a command. They are included in the artifact directories and manifest counts, but are not assigned to the last command in `commands.json`.

This closes the pre-step sequence: each command has its starting state, while the final artifacts record the state after the last command.

These flags only control deterministic artifact collection. They do not invoke Maestro AI analysis or an LLM.

## Compatibility

Existing behavior remains unchanged when no new flags are used.

- Failed and warned steps still capture their existing outcome artifacts.
- Negated capture flags do not disable failure artifacts.
- `--analyze` retains its pre-step screenshots, recording, and final screenshot.
- Hierarchy capture can be added to `--analyze`, but its required screenshots cannot be disabled.
- The existing `captureFullArtifacts` behavior is merged with `StepArtifactConfig`, so supplying hierarchy configuration does not remove legacy screenshots.
- Existing artifact manifest and command metadata formats are reused.

## Open question

Is `screen-hierarchy/final.json`, alongside the existing `screenshots/final.png`, the right representation for the flow-end hierarchy?

A matching documentation PR will follow once the CLI shape is agreed.

## Testing

- `./gradlew test :maestro-cli:installDist`
  - Passed: 121 Gradle tasks.
- Android physical-device test:
  - Ran `launchApp -> pressKey: Home` with `--capture-all-step-artifacts`.
  - Both visible commands produced a screenshot and hierarchy pair.
  - The flow produced an additional `final.png` and `final.json`.
  - The manifest reported three screenshots and three hierarchies.
  - The final screenshot and hierarchy both represented the launcher state.

### DIFF
--- a/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
+++ b/maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
@@ -26,7 +26,7 @@ import java.nio.file.StandardCopyOption
  * when [artifactsDir] is non-null, writes the per-flow artifact bundle
  * directly under it — see [BundleLayout] for the layout. With a null
  * [artifactsDir] (Studio's interactive runner) only the in-memory population
- * happens. [captureRunArtifacts] adds a recording and final screenshot. Step
+ * happens. [captureFullArtifacts] adds a recording and final screenshot. Step
  * screenshots and hierarchies are controlled independently by [stepArtifactConfig].
  * Failed/warned steps always capture an at-outcome screenshot plus hierarchy.
  *
@@ -42,7 +42,7 @@ import java.nio.file.StandardCopyOption
 internal class ArtifactsGenerator(
     private val artifactsDir: Path?,
     private val maestro: Maestro,
-    private val captureRunArtifacts: Boolean = false,
+    private val captureFullArtifacts: Boolean = false,
     private val onStepScreenshotCaptured: (sequenceNumber: Int, relativePath: String) -> Unit = { _, _ -> },
     private val stepArtifactConfig: StepArtifactConfig = StepArtifactConfig(),
 ) : OrchestraListener {
@@ -77,7 +77,7 @@ internal class ArtifactsGenerator(
         } catch (e: Exception) {
             logger.warn("Failed to set up artifacts directory at $artifactsDir", e)
         }
-        if (captureRunArtifacts) startFullRunRecording()
+        if (captureFullArtifacts) startFullRunRecording()
     }
 
     override fun onCommandStart(cmd: MaestroCommand, sequenceNumber: Int, depth: Int) {
@@ -132,8 +132,7 @@ internal class ArtifactsGenerator(
             }
         }
         if (artifactsDir == null || outcome is CommandOutcome.Skipped) return
-        val isFailureOutcome = outcome is CommandOutcome.Failed || outcome is CommandOutcome.Warned
-        if (!isFailureOutcome) return
+        if (outcome !is CommandOutcome.Failed && outcome !is CommandOutcome.Warned) return
         // Non-visible leaves (defineVariables/applyConfiguration) and empty commands have no screen.
         if (!StepArtifactNaming.capturesScreenshot(cmd)) return
 
@@ -179,7 +178,7 @@ internal class ArtifactsGenerator(
 
     override fun onFlowEnd() {
         // Capture the resting screen before the recording is torn down.
-        if (captureRunArtifacts) captureFinalScreenshot()
+        if (captureFullArtifacts) captureFinalScreenshot()
         stopFullRunRecording()
         val collector = collector
         if (artifactsDir != null && collector != null) {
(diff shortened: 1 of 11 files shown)


Fishbowler:
I understand stand wanting an alternative to `--analyze`.
I don't understand the before/after stuff.
Step 6's after is Step 7's before - there's little utility there. Also, isn't there already a "final" that captures the final state, making "after" redundant?

ayanamists:
> I understand stand wanting an alternative to `--analyze`. I don't understand the before/after stuff. Step 6's after is Step 7's before - there's little utility there. Also, isn't there already a "final" that captures the final state, making "after" redundant?

Thanks, this makes sense.

For context, we are integrating Maestro into an internal test-reporting product that turns each executed command into a timeline of screenshots and accessibility hierarchies. That is why we first thought about explicit pre- and post-command artifacts.

After looking at Maestro's settle behavior again, I agree that `after` is probably not needed. It does not provide a stronger guarantee for delayed async updates.

I will simplify the PR to one pre-step screenshot option, with optional hierarchy capture, while keeping the existing `--analyze` behavior.

Do you see any other concerns or changes you would recommend before I update the PR?

Fishbowler:
Perhaps there's scope for 3 flags here?

`--capture-step-hierarchy`, negatable, enables per-step hierarchy snapshots 

`--capture-step-screenshots`, negatable, enables per-step screenshots 

`--capture-all-step-artifacts`, non-negatable, convenience for the above two, plus anything added in the future.

The reason I mention negatable is because capturing hierarchy at every step is off today, but potentially could flip in the future. PicoCLI offers negatable switches, so that you can automatically offer `--no-capture-step-hierarchy`.

Should negated switches today disable failure artifacts? Probably not?

Don't forget a matching PR on the docs repo.

(So far I'm just looking at surface design, not implementation)

ayanamists:
> Perhaps there's scope for 3 flags here?
> 
> `--capture-step-hierarchy`, negatable, enables per-step hierarchy snapshots
> 
> `--capture-step-screenshots`, negatable, enables per-step screenshots
> 
> `--capture-all-step-artifacts`, non-negatable, convenience for the above two, plus anything added in the future.
> 
> The reason I mention negatable is because capturing hierarchy at every step is off today, but potentially could flip in the future. PicoCLI offers negatable switches, so that you can automatically offer `--no-capture-step-hierarchy`.
> 
> Should negated switches today disable failure artifacts? Probably not?
> 
> Don't forget a matching PR on the docs repo.
> 
> (So far I'm just looking at surface design, not implementation)

Thanks, these flags makes sense. I’ll update the PR accordingly, keep failure artifacts unaffected by the negated switches, and follow up with a matching docs PR.

maestro-orchestra/src/main/kotlin/maestro/orchestra/debug/ArtifactsGenerator.kt
```diff
@@ -97,8 +96,17 @@ internal class ArtifactsGenerator(
...
+            (stepArtifactConfig.captureScreenshots || stepArtifactConfig.captureHierarchy) &&
+            StepArtifactNaming.capturesScreenshot(cmd)
+        ) {
+            if (stepArtifactConfig.captureHierarchy && StepArtifactNaming.capturesHierarchy(cmd)) {
+                captureStepHierarchy(metadata)
+            }
+            if (stepArtifactConfig.captureScreenshots) {
+                captureStepScreenshot(metadata)
+            }
+        }
```
Fishbowler: I'm not sure I understand the outer wrapper. Why isn't this just:

```suggestion
        if (stepArtifactConfig.captureHierarchy && StepArtifactNaming.capturesHierarchy(cmd)) {
            captureStepHierarchy(metadata)
        }
        if (stepArtifactConfig.captureScreenshots && StepArtifactNaming.capturesScreenshot(cmd)) {
            captureStepScreenshot(metadata)
        }
```

ayanamists: oh, the the outer wrapper is redundant

Fishbowler:
Did a bit of reviewing, thinking and tidying. Some of it is thinking about how Artifacts should/do work. Some of it is completion (adding to the config.yaml). Some of it is putting the right bits in place so that this will work in Maestro Cloud too.

Take a look at my changes commit-by-commit and be sure you're happy with them - they'll end up being merged in your name 😁

PR #3530, opened 2026-08-24
Better web e2e

Fishbowler:
## Proposed changes

Structural:
- Move web tests to its own workspace
- Move workspace definitions to their own manifest file - far less special case stuff in `run_tests`
- Add a local webserver for serving test fixtures

Test content:
- Stop using data URIs for test cases - use local fixtures instead
- Reduce dependency on external sites remaining static - port a few to local fixtures 
- Add additional coverage (additional cross-origin iframe test, and add the tag to the textarea test)

## Testing

Works locally. Submitting to CI to see how it works there...

## Issues fixed

None, but this enables a bunch of others.


steviec (approved):
NICE

PR #3531, opened 2026-08-24
feat(cli): --android-system-image-tag selects the Android system-image variant

steviec:
## Why

Customers can't choose which Android system image their flows run on. Every emulator we create
uses `google_apis`. Some customer bugs only show up on `google_apis_playstore`, which ships real
Play Services instead of the stubs in `google_apis` — the Trackman Maps crash (MA-4173) is the one
that prompted this.

## Approach

One flag, `--android-system-image-tag <google_apis|google_apis_playstore>`, on two commands:

- **`start-device`** builds the emulator with the chosen image.
- **`cloud`** sends the choice to the backend. When the flag is unset, the request is exactly what
  it is today.

The tag field already lives on `DeviceSpec.Android` (from #3518); this PR just exposes it.

The one part worth a close look: an emulator's name now includes the tag when it isn't the default.
Without that, a playstore request would find the `google_apis` emulator of the same model and OS and
quietly run on it — the wrong image. Android emulator matching is now exact instead of a substring
check, so an untagged request doesn't match a tagged emulator whose name starts with the same text.

## Notes

- **Not on `maestro test`.** That command picks a device before it would ever create one, so the
  flag can't do anything there. It's rejected, with a test to keep it that way.
- **Creation only.** A device that's already running is used as-is.
- **Undocumented for now.** The backend and the fleet roll out first, so nothing's in the README or
  CHANGELOG yet.

## Verification

- **Unit.** The `:maestro-cli` and `:maestro-client` suites pass, including new tests that the tag
  reaches the backend request (and is absent when the flag is unset), that a playstore request
  doesn't reuse a `google_apis` emulator, and that the two names don't collide.
- **Real binary.** Created both a playstore and a `google_apis` emulator for the same model and OS
  and confirmed they stay separate, each on the right image.
- **Full stack.** Ran `maestro cloud` against a local backend and worker, with and without the flag.
  With it, the worker booted the playstore image; without it, the request was unchanged and it booted
  `google_apis`. Both flows passed.

## Companion PRs

- Worker: https://github.com/mobile-dev-inc/copilot/pull/2757
- Backend: https://github.com/mobile-dev-inc/copilot/pull/2761

Rollout order is worker → backend → this.


maestro-client/src/main/java/maestro/device/DeviceSpec.kt
```diff
@@ -32,7 +32,7 @@ enum class SystemImageTag(@JsonValue val value: String) {
     companion object {
         fun fromString(value: String): SystemImageTag {
             return entries.firstOrNull { it.value == value }
-                ?: throw IllegalArgumentException(
+                ?: throw DeviceSpecValidationException(
```
proksh, line 35: This switches fromString from `IllegalArgumentException` to `DeviceSpecValidationException` (which extends `Exception`, not IllegalArgumentException). We will need to change the exception handling on backend side.

proksh: Ahh just saw this comment https://github.com/mobile-dev-inc/copilot/pull/2761/changes#r3865913071, so its already in plan.

proksh (approved):

PR #3537, opened 2026-08-26
Propose the version and split the release skill into two checkpoints

steviec:
## Why

@Fishbowler and I ran this skill for 2.9.0 and found some ways to make the skill better. So we made them.

## Approach

- The skill proposes the version now. It reads the commits, applies a short semver rubric, and picks the bump.
- Two checkpoints, not one. Checkpoint 1 confirms the version, before anything is written. Checkpoint 2 is the go — the release runs after it.
- On go, the skill checks that required checks are green and stops if they aren't.
- The opening plan just lists what the skill will do, with the two checkpoints marked. No more per-line tags.
- Dry-run only happens when the invocation asks for it.
- Two fixes for things the old skill got wrong: the changelog casing, and a missing brew check in the verify step.

## Verification

A fresh agent read the new skill and tried it out.

- It applied the rubric to three sets of commits and got all three right: this release (minor), a fixes-only set (patch), and one that renames a command (major).
- It rendered both checkpoint messages from this release's data, and they came out right.

This is an agent reading the skill, not a real release. The new flow hasn't shipped a version yet.

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: release
-description: Use when cutting, shipping, or publishing a new Maestro version — "release 2.9.0", "cut a release", "tag vX.Y.Z", "publish the CLI", "update the changelog for the release", or any variant of "let's release maestro".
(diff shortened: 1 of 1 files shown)


pedro18x (approved):
Nice

PR #3546, opened 2026-08-31
chore(android): bump compile/target SDK to 37

proksh:
Validates Maestro against Android API 37 (`system-images;android-37.1;google_apis_ps16k;x86_64`).

## Changes
- `compileSdk`/`targetSdk` 36 → 37; driver APKs rebuilt.
- `android-37` dispatch option + ps16k image resolution in the e2e workflow.
- e2e CI: fix the real API 37 boot failure. CI's cmdline-tools 12.x can't parse the `37.1` image version and writes `target=android-0` into the AVD `.ini`, which misconfigures gfxstream and SIGABRTs surfaceflinger (`mapper.ranchu: !rcEnc->featureInfo()->hasReadColorBufferDma`, RegionSampling readback) at boot, taking system_server down so the driver's first `pm`/`am` call fails with `Broken pipe`. Fix: bump the e2e cmdline-tools to build `15859902` (avdmanager now resolves `target=android-37.1`), and set `ANDROID_AVD_HOME` and the `config.ini` append to `~/.android/avd` where the newer avdmanager creates the AVD. Verified on the 37.1 image: `target=android-37.1`, 0 surfaceflinger/ReadColorBufferDma aborts.

## Notes
- Not the 16 KB page size (the 4k `android-37.0` image crashes identically, API 36 does not) and not the driver, so no `AndroidDriver` change is included.
- Fleet follow-up (pin the worker cmdline-tools via ansible): MA-4266.


proksh:
Parking this until the copilot-side launch work is sorted; branch kept. Two e2e runs on API 37 are diagnosed (no driver bugs found; remaining fix is CI AVD feature flags). Will reopen and finish the loop afterwards.

steviec (commented):
this might be fine, but did you root cause _why_ it restarts mid-run? is this a known problem with the new system image? I wouldn't want to add a band-aid if we don't truly understand it. how did you verify it? can you reliably repro it? did you find any specific references to this issue in the Android/Google trackers or public info?

also: if this really does happen "right after boot", should this be in maestro-device?

proksh:
> this might be fine, but did you root cause _why_ it restarts mid-run? is this a known problem with the new system image? I wouldn't want to add a band-aid if we don't truly understand it. how did you verify it? can you reliably repro it? did you find any specific references to this issue in the Android/Google trackers or public info?
> 
> also: if this really does happen "right after boot", should this be in maestro-device?

You were right to flag this. I spent today root-causing it. It's a CI tooling bug, not the image and not a real transient, and the retry wasn't a proper fix.

Tried, verified it did NOT fix it: 
- emulator feature flags (`-GLDMA`,`-GLDMA2`,`-GuestAngle`)
- GPU modes (swiftshader/guest/host)
- a newer emulator (canary 37.2.8)
- the 4k image (37.0 google_apis)
All still crashed (so it's not the 16 KB page size)

Root cause: our CI ships cmdline-tools 12.x, whose avdmanager can't parse the 37.1 image version and writes `target=android-0` into the AVD `.ini`. That misconfigures gfxstream, so the guest SIGABRTs in `surfaceflinger` (`mapper.ranchu: !hasReadColorBufferDma`, RegionSampling readback), takes system_server down, and the driver's first pm/am call gets Broken pipe. Same failure mode Google documents here: https://issuetracker.google.com/issues/546200928

What finally worked (verified): bump cmdline-tools (→ build 15859902). On the same 37.1 image the AVD gets target=android-37.1 and the run shows 0 surfaceflinger/ReadColorBufferDma aborts, framework stays up. (Newer avdmanager creates the AVD under ~/.android/avd, so the AVD-home/config.ini paths were repointed there.)

For the device/cloud side: no crash exposure (that path writes the AVD .ini incl. target= itself, so it's immune). 

The one prerequisite before we run API 37 there is bumping the fleet's cmdline-tools to a current build so sdkmanager can install the 37.x image, and pinning it (today it floats to latest) so the release is reproducible.

steviec (approved):
LGTM.

PR #3550, opened 2026-08-31
Add command to verify prerequisites

ArielDemarco:
# Overview
In order to run on devices maestro, we use some dependencies. Aside of being on macOS, or using Xcode, we also need to install and use `iproxy` (from `libimobiledevice`). So, in order to simplify this (both for local dev and CI) we added a new command in the cli to do so


ArielDemarco:
Wrong fork :)

PR #3557, opened 2026-09-02
docs(skill): add bump-cli-device-support skill

proksh:
Adds `bump-cli-device-support`: a standalone Maestro-side skill for the **CLI + core-library** leg of an OS bump — making `maestro start-device` / `DeviceSpec` provision a new system image locally. This is a different concern from the on-device driver bump (`bump-android-version`, which stays untouched), and from the cloud pipeline (copilot `bump-device-support`).

## Design-neutral by intent
The skill deliberately does **not** prescribe *how* image resolution should work. That contract flips between releases (fail-fast on a missing image vs. resolve-the-package-on-the-host; `systemImageOverride` string vs. semantic tag). The skill's job is to make `start-device` provision the new image with **whatever contract the code currently uses** — so it stays correct across those redesigns. It reads the current resolution code first, then matches it.

## The loop
Reproduce the provisioning break → diagnose the layer (`DeviceSpec.kt` / `DeviceService.kt` / `DeviceCreateUtil.kt`) → fix in the CLI/core libs **with per-fix consent** → validate (`DeviceSpecTest` → boot from the branch CLI → replay real customer runs via the copilot `replay-cloud-run` skill) → **loop until green**. Durable breakage traps (`osVersion` int-parse dying on `android-37.1`, variant dirs like `google_apis_ps16k`, `--package` vs `--tag`/`--abi`) are kept as a checklist, not a prescribed fix.

Records that "supported" = starts-locally + e2e-green + cloud-bump, across three skills (this, `bump-android-version`, copilot `bump-device-support`).

Docs-only (skill markdown). No code change. Net effect vs `main`: `bump-android-version` unchanged, one new skill added.

### DIFF
--- a/.claude/skills/bump-cli-device-support/SKILL.md
+++ b/.claude/skills/bump-cli-device-support/SKILL.md
@@ -1,0 +1,103 @@
+---
+name: bump-cli-device-support
+description: Use when a new OS version or a system-image naming change breaks how the Maestro CLI + core libs provision a device locally — `maestro start-device` failing to resolve or launch a new image, `osVersion` returning 0 on a minor-versioned platform (`android-37.1`), variant image dirs (`google_apis_ps16k`), new/renamed tags, or `avdmanager` getting the wrong or absent package. NOT the on-device driver/APK bump (bump-android-version) and NOT the cloud worker/backend pipeline (copilot bump-device-support).
+---
+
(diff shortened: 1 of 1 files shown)


proksh:
Sibling of #3546 (driver bump, runs last) and #3558. Tracked in MA-4246.

PR #3558, opened 2026-09-02
feat(device): support API 37 in start-device (minor-versioned + ps16k images)

proksh:
## Why
`maestro start-device --platform android --device-os android-37` fails today. API 37 broke two assumptions the CLI made about Android system images:

1. Platforms are now minor-versioned (`android-37.1`), so `osVersion` parsed to 0.
2. There is no plain `google_apis` image for the current 37.x. Only `google_apis_ps16k` (16 KB page size) is published, so the derived `system-images;android-37;google_apis;<abi>` does not exist.

The `DeviceSpec` is the contract that travels CLI → backend → worker, so it must keep carrying intent (model, `os`, abi, optional pinned image) and stay a catalog key. This PR does not resolve or rewrite anything inside the spec. `spec.os` stays `android-37`.

## What
1. **osVersion parsing** (`DeviceSpec.kt`): `android-37.1` → 37.
2. **Respect the spec, fail loudly** (`DeviceCreateUtil.getOrCreateAndroidDevice`): the spec's own image is tried first. If it is not installed and this host's SDK offers a different image for that os, the command fails and names the exact flag:
   ```
   System image system-images;android-37;google_apis;arm64-v8a is not available for android-37. The Android SDK offers system-images;android-37.1;google_apis_ps16k;arm64-v8a instead.
   Re-run with: --device-os system-images;android-37.1;google_apis_ps16k;arm64-v8a
   ```
   An installed image never triggers an SDK lookup. A missing image the SDK does offer keeps the existing install prompt. A pinned full path is used as-is.
3. **What the SDK "offers"** (`DeviceService.resolveSystemImage` / `selectSystemImage`): newest stable minor (`37.1` over `37.0`, never `37.2-beta*`), only `google_apis`-family tags (no wear/TV/AOSP), `google_apis` over other rootable variants like `google_apis_ps16k`, never `*_playstore`. Installed images are checked before the slower `sdkmanager --list`.

API 29 to 36 are unchanged: their derived image exists, so no lookup happens.

## Tests
- `DeviceServiceTest`: selection policy against the real sdkmanager shapes for API 37.
- `DeviceCreateUtilTest`: missing image fails with the suggested `--device-os` and `spec.os` unchanged; installed image skips the SDK; missing but offered image keeps the install prompt; pinned path used as-is; existing AVD reused without lookup.
- `DeviceSpecTest`: osVersion parse.

Verified with the built CLI on an arm64 host: `android-37` fails with the message above; the suggested full path creates `Maestro_ANDROID_pixel_6_android-37.1_google_apis_ps16k` and the guest reports SDK 37 with 16 KB pages; `android-36` and `--os-version 33` behave as before.


proksh:
Sibling of #3546 (driver bump, runs last) and #3557. Tracked in MA-4246.

maestro-client/src/main/java/maestro/device/DeviceSpec.kt
steviec: I think we need to change the tag based on os version. If we use the same default tag for `osVersion >= 37`, then the resulting `systemImage` will actually be a nonexistent, broken package string. DeviceSpec is the right place to put this IMO. If you do this, you should get a valid string:
```
val spec = DeviceSpec(os="android-37.1", model="pixel_6")
spec.systemImage # => "system-images;android-37.1;google_apis_ps16k;arm64-v8a"
```

steviec (commented):
looks good, one comment

maestro-client/src/main/java/maestro/device/DeviceSpec.kt
proksh: Good catch. There's no static tag that works, though: 37.0 has google_apis, 37.1 is ps16k-only, and the newest minor keeps moving. So rather than put the tag in DeviceSpec, the spec now carries just the intent (os + family) and each host resolves the real image from its own sdkmanager.

Added PRs with this task: https://linear.app/mobile-dev/issue/MA-4250/move-system-image-resolution-out-of-devicespec-spec-carries-intent

proksh: Fixed. (Ignore above message).

steviec (approved):
Looks good!

PR #3560, opened 2026-09-03
feat(orchestra): derive the flow-command schema from the parser's own types

amanjeetsingh150:
`FlowCommandSchema` describes the Maestro flow-command surface — command names, arguments and their kinds, enum vocabularies, `swipe`'s alternative shapes, and which commands may be written bare (`- back`). It is derived by reflection from the very types the parser uses (`YamlFluentCommand`'s constructor, each `Yaml*` data class, the `stringCommands` map), so it cannot drift from what the parser accepts.

**It is an API, not a file.** Any JVM consumer already depending on `dev.mobile:maestro-orchestra` calls `FlowCommandSchema.commands()` (or `asJson()`). No URL to fetch, no path to hardcode — the surface ships in the jar and moves with the dependency version. Its first consumer is mobile-dev-inc/copilot#2776, which generates Studio's autocomplete index from it.

## What the schema publishes

- 56 commands, their arguments, kinds, required flags and enum vocabularies.
- `commonArguments` (`label`, `optional`) and `selectorArguments` — the 25 fields a `tapOn`-style selector accepts, rendered once rather than under every selector command.
- `@YamlValues` declares a vocabulary a `String`-typed field is validated against, where reflection cannot see it. Six sites: `pressKey.key`, `setOrientation.orientation`, `selector.traits`, `action`, and `setDarkMode`/`setAirplaneMode`'s value. The words are read off the enum, never written down twice.
- `@YamlRequiresOneOf` declares a rule enforced inside `toCommands`: `runFlow` and `retry` need `file` **or** `commands` and refuse both (`exclusive`), `extendedWaitUntil` needs `visible` or `notVisible` and accepts both, `addMedia` needs `files`. Without it the schema published `runFlow: {}` as valid. The names are published in written order and a consumer with no other signal takes the first, so `retry` lists `commands` first — its common form is an inline block, where `runFlow` genuinely is file-first.
- `@YamlVariant` names `swipe`'s alternative shapes in YAML terms — `byDirection`, `byCoordinates`, `byRelativeCoordinates`, `byElement` — instead of leaking `YamlSwipeElement` into a published document.
- `required` means the parser actually raises when the key is absent. That excludes optionals, nullables **and** non-nullable JVM primitives: Jackson hands the creator `0`/`false` for a missing `Int`/`Boolean` rather than failing. No command declares such a field today, so nothing published changes — but the first `val timeout: Int` to arrive would otherwise have been advertised as mandatory when it is not.
- `asJson()` carries a `version` for the document's shape, and orders variants and enum values deterministically (`sealedSubclasses` and `Class.getFields()` have no guaranteed order).

## Behaviour changes — please focus here

- **`setDarkMode` / `setAirplaneMode` honour `optional`.** It was silently dropped: the deserializers built the value with a two-argument constructor call. A flow that said `optional: true` on either command was failing the run instead of being skipped.
- **Both now reject near-miss values.** The old scalar path used `contains("enabled")`, so `setDarkMode: "is enabled now"` parsed as enabled. Case sensitivity and `${VAR}` behaviour are unchanged — neither worked before or after.
- **`openBrowser` is deleted.** It was declared on `YamlFluentCommand` with no branch in `_toCommands`, so every form of it was rejected by the parser. Nothing else in the repo referred to it.
- **The selector's `start` / `end` are deleted.** Added in "Add swipe command (#27)" in Jul 2022 when swipe reused `YamlElementSelector`, and left behind when swipe grew its own coordinate types. `toElementSelector` never read them and `ElementSelector` has no field to carry them, so Jackson bound them and the parser threw them away for three years. Invisible until the schema published them to every consumer. `tapOn: {start: "0,0"}` parsed-and-ignored before and now fails with "The property `start` is not recognized" — such a flow was never doing what its author meant. `swipe` is untouched. Worth a changelog line.
- **`_appId:` in config no longer parses.** `YamlConfig`'s parameter is named `_appId` only because the class also exposes a computed `appId` and Kotlin will not let the two share a name. That was covered with `@JsonAlias("appId")`, which makes `appId` the *alternate* spelling — so `_appId:` parsed too, which nobody intended, and anything reading the type back got the names inverted. `@JsonProperty("appId")` makes the intended name primary; `appId:` is unchanged.

The YAML words for `DarkModeValue` / `AirplaneValue` deliberately live in a `yamlValue` property rather than a `@JsonProperty` on each constant: Jackson also serializes those enums as `SetDarkModeCommand.value` on the `MaestroCommand` wire, where the constant name is what is persisted and sent between worker and backend. `YamlSetModeTest` pins that wire directly, because every YAML test passes either way. `KeyCode` has the same constraint and the same shape.

## Tests

`FlowCommandSchemaEvolutionTest` drives the derivation with synthetic `Yaml*` shapes — a command being added, an argument renamed or retyped — which is the only way to ask what happens the *next* time someone changes one. It found three bugs, fixed here: `@JsonProperty` renames were ignored, `@JsonAlias` spellings were never advertised (`launchApp: {url: …}` parses and was absent), and a stale `spelledBy` killed every `commands()` call with a nameless exception. It also pins the primitive rule against Jackson's real behaviour rather than restating it.

`RequiredClaimTest` pins the boundary against the real parser in both directions, and every one-of rule is checked against it: the names exist, none is also `required`, the empty form really is rejected, each member alone really is enough, and `exclusive` really matches. Dropping any of the four annotations fails it.

`SelectorArgumentsTest` feeds every advertised selector field back through `MaestroFlowParser` under `tapOn:`, so a field the schema offers that the parser rejects fails the build. It is what caught `start` / `end` being publishable but dead.

Known gap, pinned not hidden: `inputText.text` and `evalScript.script` are reported required, but their delegating creators read the field with `getOrDefault(name, "")`, so omitting it yields an empty string instead of an error. The schema is right, the parser is wrong; tightening it is a parsing behaviour change left for its own PR. The test fails the day it is fixed.

Verify (on the rebase onto `main` @ eae360a49):
- `./gradlew :maestro-orchestra:test` — 437 tests, 0 failures
- `./gradlew :maestro-orchestra-models:test` — 51 tests, 0 failures
- `./gradlew :maestro-test:test` — 213 tests, 0 failures
- `./gradlew :maestro-cli:compileKotlin` — passes, so the `internal`/`sealed` changes broke no callers

https://claude.ai/code/session_012naxaR7AjmbDCmckC2Y4ru
https://claude.ai/code/session_015vHfmsiT7skZdV13zcbeof
https://claude.ai/code/session_018urBBzo8pVbwhcdWVvZFKr
https://claude.ai/code/session_01BkDJX81SCvaKTW6MJECZp7


pedro18x (approved):
Went through this against the parser at the head commit and it lines up for me: the command count, the selector fields, the dropped `optional` on the set-mode map form, and `openBrowser` having no branch in the parser. The tests that check the schema against the real parser in both directions (`RequiredClaimTest`, `FlowCommandSchemaTest`) are what make this easy to trust. Nice work.

Small things, none blocking:

1. I think the "required" rule in `FlowCommandSchema` might be a bit off for primitives with no default. As far as I can tell Jackson fills those in with false or zero instead of failing. Doesn't seem to affect anything published today, so maybe just a note on the rule.

2. `setOrientation` values come out as `LANDSCAPE_LEFT` where the docs use `landscapeLeft`. `DeviceOrientation` already has a camel-case name property, so pointing `spelledBy` at that might get the documented spelling for free. Same idea as your traits note.

3. Minor: the `TempFileHandler` in the two tests never gets closed, and the e2e conformance test passes quietly if it can't find the e2e folder.

amanjeetsingh150:
Hey @pedro18x thanks for the feedback

1. For the primitives, that's true I've left them out since they're filled with Jackson
2. https://docs.maestro.dev/reference/commands-available/setorientation.md its referencing the orientation with uppercase and on traits as well the spelling doesn't matter. I can update the docs, I think it can cause confusion so its better there is consistence for the enums.
3. Addressed the temp file properly cleaning up.

Fishbowler (approved):
Looks rational to me. There's a bunch of good tidying up in here too ❤️

PR #3564, opened 2026-09-04
Fix file separators on Windows for Cloud uploads

Fishbowler:
## Proposed changes

Fixes a bug where running an individual flow that includes a runFlow to a flow in another folder could be rejected on upload

image

image

The above errors are seen in Studio, but can reproduced in CLI too. The error received depends on the flows relative position to one another, and the names involved.

Maestro creates a synthetic config.yaml in these cases, and using Windows file separators can create special escapings, like `\r`. 

This fix enforces unix-style path separators in the synthetic config.

## Testing

- Built CLI and tested a previously broken flow
- Build Studio locally using this dependency and did the same

## Issues fixed

https://maestro--dev.slack.com/archives/C041FU72T54/p1788473393390039

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceUtils.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceUtils.kt
@@ -85,8 +85,11 @@ object WorkspaceUtils {
         // Forward slashes and single quotes: on Windows, relativize() returns a
         // backslash-separated path, and backslash is a YAML escape character inside
         // double-quoted scalars (e.g. "\r" would be read as a literal carriage return).
+        // Single-quoted scalars keep backslashes literal, but need their own escape:
+        // a lone ' terminates the scalar, so every ' has to be doubled.
         val flowRelativePath = relativeTo.relativize(normalizePath(flowFile)).toString()
             .replace(File.separatorChar, '/')
+            .replace("'", "''")
         return "flows:\n  - '$flowRelativePath'\n"
     }
 
--- a/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceUtils.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceUtils.kt
@@ -1,5 +1,6 @@
 package maestro.orchestra.workspace
 
+import java.io.File
 import java.io.FileNotFoundException
 import java.net.URI
 import java.nio.file.FileSystems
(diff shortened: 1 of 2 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/workspace/WorkspaceUtils.kt
```diff
@@ -81,8 +82,12 @@ object WorkspaceUtils {
     }
 
     private fun syntheticSingleFlowConfig(flowFile: Path, relativeTo: Path): String {
+        // Forward slashes and single quotes: on Windows, relativize() returns a
+        // backslash-separated path, and backslash is a YAML escape character inside
+        // double-quoted scalars (e.g. "\r" would be read as a literal carriage return).
         val flowRelativePath = relativeTo.relativize(normalizePath(flowFile)).toString()
-        return "flows:\n  - \"$flowRelativePath\"\n"
+            .replace(File.separatorChar, '/')
+        return "flows:\n  - '$flowRelativePath'\n"
```
amanjeetsingh150, line 93: Nice catch on the Windows separators, and `File.separatorChar` is the right choice here rather than a hardcoded `'\\'` — on Unix a backslash is a legal filename character, so an unconditional replace would corrupt those names.

One issue with the move to single-quoted YAML, though: it swaps one escaping hazard for another. In single-quoted scalars `'` is the escape character and must be doubled — a lone `'` terminates the scalar. The old double-quoted form handled apostrophes correctly, so this is a regression on **every** platform, not just Windows.

`maestro cloud "Dan's flows/main.yaml"` (or anything under `client's-app/`, a `user's_login.yaml`, ...) now emits:

```yaml
flows:
  - 'Dan's flows/main.yaml'
```

Checked both forms against a YAML parser:

| config bytes | result |
|---|---|
| `flows:\n  - "Dan's flows/main.yaml"\n` | `["Dan's flows/main.yaml"]` |
| `flows:\n  - 'Dan's flows/main.yaml'\n` | `ParserError while parsing a block collection` |

Cloud-side this surfaces via `WorkspaceValidator` -> `WorkspaceExecutionPlanner.plan()` reading `/config.yaml`, i.e. the same class of upload rejection the PR is fixing.

Single quotes are still the right call (backslashes stay literal) — they just need their one escape:

```suggestion
        val flowRelativePath = relativeTo.relativize(normalizePath(flowFile)).toString()
            .replace(File.separatorChar, '/')
            .replace("'", "''")
        return "flows:\n  - '$flowRelativePath'\n"
```

The more robust alternative is to stop hand-building the YAML and let Jackson pick the quoting — `YAMLMapper().writeValueAsString(mapOf("flows" to listOf(flowRelativePath)))` — but for a single field the explicit escape is fine.

One more thought: worth adding a regression test. `WorkspaceUtilsTest` currently asserts `configContent.contains("flows/main.yaml")`, which is a substring check that passes even when the surrounding YAML is malformed — which is part of why the original Windows bug was invisible to CI. A test that actually *parses* the emitted config and round-trips a separator- and quote-hostile path would cover both this and the bug you're fixing.

Fishbowler: My Claude session highlighted the same problem with single quotes / apostrophes, and I'd decided that it was a different problem. I hadn't clocked that it was explicitly caused here...

Fishbowler: Fixed!

amanjeetsingh150 (approved):

PR #3572, opened 2026-09-08
fix(orchestra): route the assertScreenshot diff into the run bundle

amanjeetsingh150:
`assertScreenshot` failures arrive on Cloud with no diff image. Reported by Nando's UKI (upload `mupload_01kyhjn1pyfn4v0gjycgd5ysft`): the run artifacts hold the per-step screenshots, but not the one image that says *why* the assertion failed.

## Cause

`Orchestra.kt:696`

```kotlin
val diffFile = expectedFile.resolveSibling("${expectedFile.nameWithoutExtension}_diff.png")
```

`expectedFile` is the **reference** — an input, resolved from the flow's own directory. The diff is an **output**. Deriving the second from the first puts the diff in the workspace checkout.

Locally that is the folder you are already looking at, so it looks correct. On Cloud the workspace is scratch space deleted with the machine and only the artifact bundle is uploaded, so the diff is discarded. **Not a regression** — `git log -L 696,698` shows the line predates the bundle work (`0d9b257e2`), so Cloud diffs have never worked.

Two things made it worse: the diff's name derives only from the reference, so the reported flow's four failed attempts (steps 87/90/93/96) all wrote one file; and the failure message pointed at an absolute path on a machine that no longer exists.

## Not a one-off

The guard added here exists because this is a class. Prior escapes on `main`:

| Commit | What escaped |
|---|---|
| `0f86d7e0d` (#3459) | `..` paths resolving outside the run root |
| `f598a9517` (#3396) | empty recording placeholders reaching the bundle |
| `eae360a49` (#3564) | file separators on Windows for Cloud uploads |

The common shape is a command building its own output path. Today four write verbs exist (`allocate`, `allocateCommandOutput`, `adopt`, `allocateCommandArtifact`), plus a separate `validateCommandPath` you must remember to call first, plus a `?: File(...)` fallback. `assertScreenshot` used none of them.

## Change

**`Artifacts` — the one surface a command gets.** Two verbs that are deliberately not interchangeable:

```kotlin
fun file(kind: ArtifactKind, name: String): File       // produce
fun existing(kind: ArtifactKind, name: String): File?  // read back
```

`file()` owns the folder (from the kind), the extension (from the kind's `ArtifactFormat`), path validation, step attribution and the manifest entry — so every call site collapses:

```diff
-ArtifactCollector.validateCommandPath(command.path, "takeScreenshot")
-val outFile = artifactsGenerator
-    .allocateCommandArtifact(ArtifactKind.TAKE_SCREENSHOT, "${command.path}.png", "takeScreenshot")
-    ?: File("${command.path}.png")
+val outFile = artifacts.file(ArtifactKind.TAKE_SCREENSHOT, command.path)
```

It returns non-null — with no bundle the file is CWD-relative exactly as before — so no call site can reintroduce a fallback of its own.

**`artifactsDir` stops being a property on `Orchestra`.** A command body that can name the directory can work out a path outside it, which is what happened here. Commands reach artifacts only through `Artifacts`.

**The fix itself:**

```diff
-val diffFile = expectedFile.resolveSibling("${expectedFile.nameWithoutExtension}_diff.png")
+val diffFile = artifacts.file(ArtifactKind.SCREENSHOT_DIFF, "${expectedFile.nameWithoutExtension}_diff")
```

`SCREENSHOT_DIFF` gets its own `assertScreenshot/` folder and is named with the shared 1-based step index, so each retry attempt keeps its own diff.

## Behaviour changes

- **The local diff moves** from beside the reference to `<run folder>/assertScreenshot/`. Intended — local and Cloud now agree — but it is user-visible.
- **`assertScreenshot`'s second reference candidate is now confined.** It previously normalised without confinement, so `../../logs/...` could resolve *outside* the bundle and match there. Consistent with #3459's intent.

## Guards

Two, because they catch different things.

- `OrchestraArtifactFunnelTest` — a source scan of `Orchestra.kt` for `File(`, `resolveSibling(`, `Paths.get(`, `.toFile(`, failing any line not marked `// funnel-exempt: read`. After this change only two lines need the marker, both reference reads. **Adapted from the one on `ma-4167-screenshot-diff-collector`.**
- `ArtifactBoundaryTest` — runs a flow and asserts the filesystem: *a flow writes nothing outside its bundle*. This is the one that catches this ticket; the diff was written outside the bundle entirely, where a manifest-orphan check never looks.

## Verification

`./gradlew test` — **1229 tests, 0 failed**, 5 skipped, on `75e4a03de`.

## Scope

The diff is now in the manifest and uploaded, so it reaches the run archive customers download. Listing it as a *typed* artifact on the run page needs the backend to stop dropping kinds it does not recognise (`RunArtifactMapper` ends `else -> null`, which also hides `takeScreenshot` output today) — that is a separate copilot-side change.

## Relationship to MA-4167

#3484 fixes the same bug and is approved but unmerged. This takes a different route — consolidating the producer surface rather than adding a fifth allocator — while **deliberately matching its kind name, folder and step index**, so whichever lands first the other rebases cleanly. Its funnel guard and its non-null return are borrowed here with credit.

https://claude.ai/code/session_01AR1yrawifxaUgWfGXrPNfR

### DIFF
--- a/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
+++ b/maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
@@ -130,7 +130,15 @@ class DefaultFlowController : FlowController {
  */
 class Orchestra(
     private val maestro: Maestro,
-    private val artifactsDir: Path? = null,
+    /**
+     * Where this run's artifact bundle is written; null for callers that keep nothing
+     * (interactive runners), which write command output CWD-relative as before.
+     *
+     * Deliberately not a `val`: a command body that can name this directory can work out a
+     * path outside it, which is how the assertScreenshot diff came to be written into the
+     * workspace. Commands reach artifacts only through [Artifacts].
+     */
+    artifactsDir: Path? = null,
     private val captureFullArtifacts: Boolean = false,
     private val listeners: List<OrchestraListener> = emptyList(),
     private val lookupTimeoutMs: Long = 17000L,
@@ -176,6 +184,13 @@ class Orchestra(
     // artifactsDir is set and populates debugOutput either way.
     private val artifactsGenerator: ArtifactsGenerator =
         ArtifactsGenerator(artifactsDir, maestro, captureFullArtifacts, onStepScreenshotCaptured)
+
+    /**
+     * The only artifact surface a command implementation gets — deliberately not the artifacts
+     * directory itself, since a command that can name the directory can work out a path
+     * outside it.
+     */
+    private val artifacts: Artifacts = Artifacts(artifactsGenerator)
     private val effectiveListeners: List<OrchestraListener> = listOf(artifactsGenerator) + listeners
 
     private var commandSequenceCounter: Int = 0
@@ -648,9 +663,9 @@ class Orchestra(
         val path = normalizeScreenshotPath(command.path)
 
         val candidates = buildList {
-            command.flowPath?.let { add(it.resolve(path).toFile()) }
-            artifactsDir?.let { add(it.resolve(BundleLayout.TAKE_SCREENSHOT_DIR).resolve(path).normalize().toFile()) }
-            add(File(path))
+            command.flowPath?.let { add(it.resolve(path).toFile()) } // funnel-exempt: read
+            artifacts.existing(ArtifactKind.TAKE_SCREENSHOT, path)?.let { add(it) }
+            add(File(path)) // funnel-exempt: read
         }.distinctBy { it.canonicalPath }
 
         val expectedFile = candidates.firstOrNull { it.exists() }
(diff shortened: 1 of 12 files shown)


maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -130,7 +130,15 @@ class DefaultFlowController : FlowController {
...
     private val maestro: Maestro,
-    private val artifactsDir: Path? = null,
+    /**
+     * Where this run's artifact bundle is written; null for callers that keep nothing
+     * (interactive runners), which write command output CWD-relative as before.
+     *
+     * Deliberately not a `val`: a command body that can name this directory can work out a
+     * path outside it, which is how the assertScreenshot diff came to be written into the
+     * workspace. Commands reach artifacts only through [Artifacts].
+     */
```
steviec, line 140: Claude likes to take existing method and add a ton of unnecessary comments to them. Make sure you audit these and only include actually valuable ones instead of obvious ones.

maestro-orchestra/src/main/java/maestro/orchestra/Orchestra.kt
```diff
@@ -693,7 +708,10 @@ class Orchestra(
             debugMessage = "The assertScreenshot command requires a valid image file. Supported formats include PNG, JPEG, GIF, BMP, TIFF, and WBMP. The file at ${expectedFile.absolutePath} could not be read."
         )
 
-        val diffFile = expectedFile.resolveSibling("${expectedFile.nameWithoutExtension}_diff.png")
+        // The reference is an INPUT read from the workspace; the diff is this run's OUTPUT.
+        // Deriving the diff's location from the reference's is what put it outside the bundle,
+        // where Cloud discards it with the workspace.
+        val diffFile = artifacts.file(ArtifactKind.SCREENSHOT_DIFF, "${expectedFile.nameWithoutExtension}_diff")
```
steviec, line 714: The path to the diff file doesn't seem like it should be defined here, right? Don't we centrally manage the names of all output files?

PR #3573, opened 2026-09-08
Configurable JS HTTP request timeout

### LINKED ISSUE #1490: Configure HTTP client timeout
I am making a HTTP request using the HTTP client as per here:
https://maestro.mobile.dev/advanced/javascript/make-http-s-requests

This request fires an async job in our system which takes a few minutes to complete. However, Maestro is exiting with:

```
org.mozilla.javascript.WrappedException: Wrapped java.net.SocketTimeoutException: Read timed out (scripts/exec.js#5)
```

Is there any time to increase/set the timeout?

Fishbowler:
## Proposed changes

Provides env and request level configuration of timeouts for HTTP requests made from inside our JS engine.

Customers sometimes have sub-optimal test environments that require more leniency.

### Usage

```sh
export MAESTRO_JS_HTTP_TIMEOUT=60000
maestro test ./flows
```
or
```yaml
appId: com.example
env:
  MAESTRO_JS_HTTP_TIMEOUT: 60000
---
- runScript: callMyApi.js
```
or
```js
var response = http.post('http://example.com/endpoint', {
  body: JSON.stringify({foo: 'bar'}),
  timeout: 60000
})
```

## Testing

Unit + integration tests added. No e2e tests - the added tests run up to the end 

## Issues fixed

Fixes #1490

### DIFF
--- a/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
+++ b/maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
@@ -35,11 +35,12 @@ class GraalJsEngine(
     platform: String = "unknown"
 ) : JsEngine {
 
-    private val httpBinding = GraalJsHttp(httpClient)
     private val outputBinding = HashMap<String, Any>()
     private val maestroBinding = HashMap<String, Any?>()
     private val envBinding = HashMap<String, String>()
     private val envScopeStack = mutableListOf<HashMap<String, String>>()  // for scope isolation
+    private val httpBinding = GraalJsHttp(httpClient, envBinding)
+
 
     // Keys that should never be removed from context bindings
     private val permanentBindingKeys = setOf(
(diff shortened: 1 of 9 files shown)


Leland-Takamine:
Hmm I don't think this should be controlled via env var - timeout parameter makes more sense to me

Fishbowler:
@Leland-Takamine I considered the per-request timeout variable to a minimum, to solve for an endpoint that does a lot of work. I was able to speak to a user affected, and they've got a whole test env that isn't performant, so the env solves for that scenario too.

Leland-Takamine:
We should aim for minimizing our API surface area while providing flexibility to the caller. In this case a parameter is sufficient for achieving what the user needs, including being able to wrap the function in their own convenience method which passes a custom default timeout or even implements a custom env var to configure this. This is achievable without building custom logic into our core API to handle this one specific use case

PR #3585, opened 2026-09-09
EPMTSTF: iOS Device support with go-ios and preinstalled maestro-driver-ios and maestro-driver-iosUITestRunner

nevyglasVadim:
EPMTSTF: Added real device support. Added check for env variable to run without building iOS driver. Added go-ios commands.


Fishbowler:
What is EPMTSTF?

Fishbowler:
Can we talk about approach before we look at implementation? What's the approach? What is go-ios that you've mentioned? How's this been tested?

PR #3586, opened 2026-09-10
feat(ios): opt-in single-gesture double tap

iamcgn:
## Proposed changes

`doubleTapOn` on iOS is not really a double tap. It becomes `TapRepeat(2, 100ms)`, and
`Maestro.performTap` sends each tap as its own blocking HTTP request to the XCUITest runner.
The 100ms spacing never applies: the loop measures the whole blocking call, which always takes
longer than 100ms, so the computed delay comes out as zero. What separates the two taps is
however long the round trip happens to take.

On an iPhone 17 / iOS 26.5 simulator that gap measured between 337ms and 352ms across runs,
close enough to the limit that the pair only sometimes registers as a double tap.

This adds an opt-in `singleGestureTap` flag that sends both taps as one synthesized gesture
through a new `doubleTouch` route on the runner, so the interval is a fixed number instead of
whatever the network does:

```yaml
- doubleTapOn:
    id: "cell_.*"
    singleGestureTap: true
```

Nothing changes by default on any platform. Without the flag the command makes exactly the
driver calls it makes today.

Two notes on the implementation:

- The two taps have to be separate `PointerEventPath`s in one `EventRecord`. Pressing the same
  path again after lifting it is allowed and produces a four event path, but UIKit treats that
  as one continuous touch and never sets `tapCount` to 2, so no double tap recognizer fires.
- The flag is on both `TapOnElementCommand` and `TapOnPointV2Command`, since `doubleTapOn` with
  a plain `point:` parses to the second one. Covering only the element form would leave the
  coordinate form unable to opt in.

## Testing

Automated:

- Parser tests for the new YAML field: set to true, set to false, left out entirely, and on the
  `point:` form.
- Three integration tests against `FakeDriver`, covering the element form with the flag, the
  point form with the flag, and the command *without* the flag run against a driver that does
  report `ATOMIC_DOUBLE_TAP`. That last one asserts it still makes two ordinary tap calls and no
  double tap call, which is the guard against this changing default behaviour by accident.
- Existing suites for maestro-client, maestro-ios, maestro-ios-driver, maestro-orchestra and
  maestro-test all pass, including the schema evolution test, which picked up the new field
  without a hand edit.

By hand, on a booted iPhone 17 running iOS 26.5:

- Built the runner, attached it to the simulator, and drove the new route directly with curl.
- Logged `XCSynthesizedEventRecord.description` immediately before `synthesize` to confirm what
  gets built: one record, two paths, press and lift at fixed offsets, one `synthesize` per
  double tap.
- Confirmed the gesture actually registers by pointing it at a stock app that zooms on double
  tap and diffing screenshots against a no-input baseline. This is also how the single-path
  version was ruled out: it produced a correct looking four event path but did not register as
  a double tap.
- Confirmed an unregistered route returns a clean 404, so a client falling back to the existing
  two-tap loop behaves correctly.
- Ran an app flow that depends on a double tap 20 times on the same machine and app build, 10
  with the flag and 10 without. Without the flag it passed 4 of 10, with the inter-tap gap
  measured from the runner logs at 337ms to 352ms each run. With the flag it passed 10 of 10 at
  a fixed 100ms.

### DIFF
--- a/maestro-client/src/main/java/maestro/Capability.kt
+++ b/maestro-client/src/main/java/maestro/Capability.kt
@@ -2,4 +2,5 @@ package maestro
 
 enum class Capability {
     FAST_HIERARCHY,
+    ATOMIC_DOUBLE_TAP,
 }
--- a/maestro-client/src/main/java/maestro/Driver.kt
+++ b/maestro-client/src/main/java/maestro/Driver.kt
@@ -51,6 +51,9 @@ interface Driver {
 
     fun longPress(point: Point)
 
+    fun doubleTap(point: Point, intervalMs: Long): Unit =
+        throw UnsupportedOperationException("Atomic double tap is not supported by this driver")
+
     fun pressKey(code: KeyCode)
 
     fun contentDescriptor(excludeKeyboardElements: Boolean = false): TreeNode
--- a/maestro-client/src/main/java/maestro/Maestro.kt
+++ b/maestro-client/src/main/java/maestro/Maestro.kt
@@ -221,7 +221,8 @@ class Maestro(
         longPress: Boolean = false,
         appId: String? = null,
         tapRepeat: TapRepeat? = null,
-        waitToSettleTimeoutMs: Int? = null
+        waitToSettleTimeoutMs: Int? = null,
+        singleGestureTap: Boolean = false
     ) {
         LOGGER.info("Tapping on element: ${tapRepeat ?: ""} $element")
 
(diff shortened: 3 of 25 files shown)


Fishbowler:
Thanks! We're actually working on a larger change along the same theme, moving more of this functionality into the driver rather than Maestro asking the driver for primative events. Once that lands, it's gonna make this change wildly incompatible, I'm afraid, but you'll get the same value in Maestro.

Out of interest, why bother with all of the plumbing for making this optional? Why shouldn't your approach have just been the new default?

Whilst we can't move forward with this, if we were, this is the bit where I'd be nagging about the lack of e2e tests :)

iamcgn:
Oh nice, any estimate on ETA? This is blocking one of our major flows. 

I made it optional because I didn't know the rationale behind the current approach and didn't want to impose my opinion on something someone else probably gave more thought to. 

Thanks and noted on the e2e tests for the future!

PR #3591, opened 2026-09-11
fix(cli): let the OS assign the MCP viewer port

dkajiwara:
## Problem

Starting several `maestro mcp` processes at once (two Claude Code sessions reading the same `.mcp.json`) kills one of them:

```
Server stderr: | Address already in use | java.net.BindException: Address already in use
Connection failed after 10778ms (CONNECTION_CLOSED): Connection closed
```

The viewer is optional, but its bind failure escapes `McpCommand.call()` and the STDIO MCP server never starts — the client just sees `CONNECTION_CLOSED`.

## Cause

`McpViewerServer.start()` picked a port with `getFreePort()`, which opens and immediately closes a `ServerSocket` on 9999..11000 and returns the number. Processes starting in the same second all see 9999 as free; whoever wins the later `embeddedServer(...).start()` keeps it, the rest get `BindException`.

## Fix

Bind with `port = 0` and read the port back from `resolvedConnectors()`, leaving no window between picking a port and owning it. `--viewer-port` is still honoured as is, `SocketUtils.getFreePort()` stays for its other caller, and the `mcp_viewer_ready` / `open_maestro_viewer` contracts are unchanged. Only visible difference: the viewer no longer prefers 9999.

Out of scope: surviving a viewer that cannot bind at all (loopback-restricted sandboxes) — a `runCatching` fallback in `McpCommand.call()` would make a good follow-up.

## Testing

`./gradlew test detekt` green; the new `McpViewerServerTest` covers both port paths. With a local `installDist` build, 5 rounds of

```bash
for i in 1 2 3; do (maestro mcp </dev/null 2> /tmp/mcp$i.err &); done
```

before: 1 of 3 died with `BindException` in every round; after: 3 of 3 answered `initialize` in every round. Also verified that `--viewer-port 12345` binds 12345 and `open_maestro_viewer` returns `http://127.0.0.1:12345/`, that `--no-viewer` is unchanged, and that an occupied `--viewer-port` still exits with `BindException` rather than hanging.

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/mcp/viewer/McpViewerServer.kt
+++ b/maestro-cli/src/main/java/maestro/cli/mcp/viewer/McpViewerServer.kt
@@ -30,11 +30,11 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.awaitCancellation
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import maestro.cli.Dependencies
-import maestro.cli.util.getFreePort
 import maestro.device.Device
 import maestro.device.DeviceService
 import maestro.device.Platform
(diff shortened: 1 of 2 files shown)


Fishbowler:
What's the end-user value here?
If two sessions are launched at once they still can't both use the device at once.

dkajiwara:
@Fishbowler 
I ran into this myself: I keep one git worktree per device, with a Claude Code session in each, so the MCP servers genuinely run in parallel and each drives its own device. Every so often one of those sessions would come up with no Maestro tools at all — the client just showed `CONNECTION_CLOSED`, and the `BindException` was only visible in the MCP server log.

So the case isn't two sessions sharing one device; it's two sessions on two devices, where an optional HTTP viewer takes the whole MCP server down with it.

PR #3597, opened 2026-09-13
fix(record): respect --no-ansi when recording

### LINKED ISSUE #1009: --no-ansi does not work when using "maestro record"

voidstackloop:
## Proposed changes

`maestro --no-ansi record` ignored `--no-ansi` and still emitted ANSI escape sequences, because `RecordCommand` always constructed an `AnsiResultView`.

`maestro test` already handles this by selecting a `PlainTextResultView` when ANSI output is disabled (`TestCommand`). This PR does the same for `record`:

- Moves `Frame` out of `AnsiResultView` into a top-level `resultview.Frame`.
- Adds `getFrames(): List<Frame>` to the `ResultView` interface, implemented by both `AnsiResultView` (behavior unchanged) and `PlainTextResultView` (now captures cumulative plain-text frames, base64-encoded, so recordings still get text overlays).
- Updates `RecordCommand` to choose the view based on `DisableAnsiMixin.ansiEnabled`.
- Updates `VideoRenderer`, `LocalVideoRenderer`, `RemoteVideoRenderer` and `ApiClient` to use the top-level `Frame`.

## Testing

- `./gradlew :maestro-cli:compileKotlin` — builds.
- `./gradlew :maestro-cli:test --tests "maestro.cli.runner.resultview.PlainTextResultViewTest"` — passes.
- Added a unit test asserting `PlainTextResultView.getFrames()` returns base64-encoded content that contains the rendered text and no ANSI escape sequences.
- Full `./gradlew :maestro-cli:test`: 161 tests, 27 failures. The 27 failures are pre-existing/environment-related (Windows path expectations in `SessionStoreTest`, `EnvUtilsTest`, `TestDebugReporterTest`, `JUnitTestSuiteReporterTest` and `ApiClientUploadRetryTest`) and are identical on the base commit (160 tests, 27 failed). The one added test passes.

> **Does this need e2e tests?** Not applicable — this is CLI output/flag handling, covered by the unit test above; recording itself still requires a device and is exercised by the existing e2e suite.

## Issues fixed

Fixes #1009

### DIFF
--- a/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
+++ b/maestro-cli/src/main/java/maestro/cli/api/ApiClient.kt
@@ -14,7 +14,7 @@ import maestro.cli.analytics.TrialStartFailedEvent
 import maestro.cli.analytics.TrialStartPromptedEvent
 import maestro.cli.insights.AnalysisDebugFiles
 import maestro.cli.model.FlowStatus
-import maestro.cli.runner.resultview.AnsiResultView
+import maestro.cli.runner.resultview.Frame
 import maestro.cli.util.CiUtils
 import maestro.cli.util.EnvUtils
 import maestro.cli.util.PrintUtils
@@ -201,7 +201,7 @@ class ApiClient(
 
     fun render(
         screenRecording: File,
-        frames: List<AnsiResultView.Frame>,
+        frames: List<Frame>,
         progressListener: (totalBytes: Long, bytesWritten: Long) -> Unit = { _, _ -> },
     ): String {
         val baseUrl = "https://maestro-record.ngrok.io"
--- a/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
+++ b/maestro-cli/src/main/java/maestro/cli/command/RecordCommand.kt
@@ -33,6 +33,7 @@ import maestro.cli.graphics.SkiaFrameRenderer
 import maestro.cli.report.TestDebugReporter
 import maestro.cli.runner.TestRunner
 import maestro.cli.runner.resultview.AnsiResultView
+import maestro.cli.runner.resultview.PlainTextResultView
 import maestro.cli.session.MaestroSessionManager
 import maestro.cli.util.FileUtils.isWebFlow
 import maestro.orchestra.workspace.WorkspaceExecutionPlanner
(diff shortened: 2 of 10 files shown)


Fishbowler:
It's been a while since I dug into Ansi vs Plain, so I asked Claude to take a first pass.

>1. Captured frames defeat the video renderer's auto-scroll — PlainTextResultView.kt:51 (medium/high)                                                                 

>  SkiaFrameRenderer.getFocusedLineIndex() (SkiaFrameRenderer.kt:161) locates the line to keep in view by searching frame text for 🔲 then ✅ — the emoji markers AnsiResultView.status() emits. PlainTextResultView.status() emits the words PENDING/COMPLETED, so both searches return −1, focused line is always 0, and SkiaTextClipper.renderClippedText computes offsetY = 0 — no scroll. Since plain-text content is cumulative from flow start (an Ansi frame is a full redraw instead), any flow longer than the ~28-line pane renders a video whose terminal overlay is frozen on the opening lines for the whole recording. This also fires without --no-ansi whenever stdout isn't a TTY (applyCLIMixin sets ansiEnabled = stdoutIsTTY), so piped/CI record hits it too. RemoteVideoRenderer — still the default — gets the same marker-less text. Fix: keep emitting a marker the renderer recognises, or trim each captured frame to a window of recent lines. 

Sounds legitimate (although I don't know where it got "~28-line pane" from). Worth a test.

>   2. Quadratic memory from retaining a base64 copy of the whole log per frame — PlainTextResultView.kt:51 (medium)                                                     
                                                                                                                                                                       
> setState unconditionally appends Frame(..., output.toString().encodeBase64()), even when printOnce suppressed all output, so consecutive frames are usually byte-identical. MaestroCommandRunner calls refreshUi() from 8 sites, so a 1,000-command flow yields thousands of frames each holding a near-final-size copy of the log, plus base64's 4/3 inflation. TestCommand.runSingleFlow never calls getFrames(), so plain maestro test --no-ansi pays the whole cost for nothing; for record the duplicates also get serialised into the multipart frames payload in ApiClient.render. Cheap guards: snapshot only when output changed, and/or only capture when a consumer asked for frames.

Yeah, fair. Possibly the most important. Folks are already struggling with OOMs in GitHub Actions when they're running Maestro alongside an emulator on the cheap-tier runners.

>   3. Contract sanctions an empty frame list the sole consumer crashes on — ResultView.kt:11 (low)                                                                      
                                                                                                                                                                       
>  The new KDoc says views "may return an empty list", but LocalVideoRenderer.render does textFrames.lastOrNull { ... } ?: textFrames.first() (LocalVideoRenderer.kt:60) — NoSuchElementException on empty. Latent today (both impls always record ≥1 frame since TestRunner calls refreshUi() before running), but a future view written to the documented contract turns a finished recording into an unhandled exception. Either drop the allowance from the doc or make renderers tolerate it.

We'd be unlikely to _ever_ have zero frames, right?

PR #3598, opened 2026-09-14
fix(js): capture console.error and console.warn messages

### LINKED ISSUE #2055: Console.error doesn't do anything

### Steps to reproduce

Found whilst trying to repro nicely for #2001 

Repro steps
1. Clone https://github.com/Fishbowler/maestro-examples
2. `cd logging`
3. `maestro test flow.yaml`

`console.error` doesn't error, like `console.potato` would, but it doesn't create any output or add anything to the logMessages metadata in the logs.

### Actual results

Console output:

```
 ║  > Flow                                                          
 ║                                                                  
 ║    ✅   Run loggingInJS_simple.js                                
 ║         Log messages:                                            
 ║           A simple message                                       
 ║    ✅   Run loggingInJS_withEnv.js                               
 ║         Log messages:                                            
 ║           An env var - TESTDATA is potato                        
 ║    ✅   Run loggingInJS_withVar.js                               
 ║         Log messages:                                            
 ║           A variable - TESTDATA is banana                        
 ║    ✅   Run ${console.log('Hello from evalScript')}              
 ║         Log messages:                                            
 ║           Hello from evalScript
 ║    ✅   Run console.log('This will not log')
 ║    ✅   Run ${console.error('Error from evalScript')}
```

Log output:

[maestro.log](https://github.com/user-attachments/files/17006130/maestro.log)

### Expected results

Output to console and logfile

### About app

No app needed.

### About environment

Tested on an Macbook M3 Pro

### Logs

<details>
<summary>Logs</summary>

```
```

</details>

### Maestro version

1.38.1

### Anything else?

_No response_

Hensell:
## Proposed changes

Capture `console.error()` and `console.warn()` through the same callback as `console.log()`, so their output appears in command logs. This does not make a successful script fail.

## Testing

Added a regression test for mixed console messages and their order. It fails before the fix; all 28 tests in `GraalJsEngineTest` pass after it.

```sh
./gradlew :maestro-test:test --tests 'maestro.test.GraalJsEngineTest' -x :maestro-ios-driver:buildIosDriver
```

Used the checked-in iOS driver resources. No device E2E tests were run.

## Issues fixed

Fixes #2055


maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -189,6 +189,7 @@ class GraalJsEngine(
             .option("js.strict", "true")
             .logHandler(NULL_HANDLER)
             .out(outputStream)
+            .err(outputStream)
```
Fishbowler: What do you think about having a second outputStream for this, so that in the maestro.log you could differentiate them? Right now, you'd just get `JsConsole: Some text you logged`. We could prepend the call to `onLogMessage` with... something. "Warning - " maybe?

maestro-client/src/main/java/maestro/js/GraalJsEngine.kt
```diff
@@ -189,6 +189,7 @@ class GraalJsEngine(
             .option("js.strict", "true")
             .logHandler(NULL_HANDLER)
```
Fishbowler, line 194: The presence of the `NULL_HANDLER` means that the `err` stream shouldn't receive noise from the engine, like runtime warnings. That's good, but I wonder if we'll know it's importance in the future. We could add a comment? Or we can also assume that Future People will test their stuff.

Hensell:
Thanks for the feedback. I split the output into separate stdout and stderr streams, prefixing stderr messages with `Warning - ` so they are distinguishable in the flow log. I also added a comment explaining why the engine logger remains disabled and updated the regression test accordingly.

PR #3552, opened 2026-09-01
feature(cli): add `requireTags` for configs

alexv-tvsre:
## Proposed changes

Adds `requireTags`, a logical-AND counterpart to `includeTags`.

`includeTags` is an OR — a Flow is kept if it carries *any* listed tag. There's no way to say "run only Flows carrying *all* of these tags", which is what you want when tags are orthogonal dimensions (e.g. `smoke` × `android`) rather than alternatives.

Given `flow1{A,B}`, `flow2{A,C}`, `flow3{A,B,C}`:

```
includeTags: [B, C]   # OR  -> flow1, flow2, flow3
requireTags: [B, C]   # AND -> flow3
```

The filters compose rather than replace — `includeTags` selects a candidate set, `requireTags` narrows it, `excludeTags` vetoes (and still wins over both):

```kotlin
(allIncludeTags.isEmpty() || tags.any(allIncludeTags::contains))
    && (allRequireTags.isEmpty() || tags.containsAll(allRequireTags))
    && (allExcludeTags.isEmpty() || !tags.any(allExcludeTags::contains))
```

## Testing

New `TagFilterCombinationsTest`: a 30-case matrix over `flowAB{A,B}`, `flowAC{A,C}`, `flowABC{A,B,C}`, `flowUntagged{}` — each filter alone, every pairing, and all three together. 30/30 pass. 

Key cases:

| include | require | exclude | result |
|---|---|---|---|
| — | — | — | all 4 |
| `[B,C]` | — | — | AB, AC, ABC (OR, not AND) |
| — | `[B,C]` | — | ABC (AND) |
| `[C]` | `[B,C]` | — | ABC — require narrows a wider include |
| `[B]` | `[A,C]` | — | ABC — the two narrow from opposite sides |
| — | `[B]` | `[C]` | AB — exclude removes what require kept |
| `[A]` | — | `[C]` | AB — pre-existing pairing, unchanged |

Plus 6 unsatisfiable combinations asserted to raise `ValidationError` (require a tag nothing carries; exclude vetoing everything require kept; exclude contradicting include or require on the same tag), and 2 config-level cases (all three keys in `config.yaml`; config keys unioning with parameter filters).

[...]

PR #3599, opened 2026-09-14
fix(cli): omit recovered errors after successful retries

Hensell:
## Proposed changes

Only print the final flow error when the flow fails. Previously, a successful retry could leave an earlier assertion error in the debug output, which was printed even though the flow exited successfully.

Failed retries and completion hooks still report their errors.

## Testing

Added three tests using the existing fake driver: successful retry, exhausted retries, and a failing completion hook. The successful-retry test fails before the fix and passes after it.

All six tests in `TestRunnerTest` and `PlainTextResultViewTest` pass:

```sh
./gradlew :maestro-cli:test --tests 'maestro.cli.runner.TestRunnerTest' --tests 'maestro.cli.runner.resultview.PlainTextResultViewTest' -x :maestro-ios-driver:buildIosDriver
```

Used the checked-in iOS driver resources. No device E2E tests were run.

## Issues fixed

Fixes #2253

PR #3600, opened 2026-09-14
fix(orchestra): keep executionOrder working when tags exclude an ordered flow

### LINKED ISSUE #1681: [v1.36.0] "Could not find flows needed for execution in order" when including & excluding flows that are defined in an execution order
**Describe the bug**
If you include tags and exclude tags as well as having flows defined in the execution order of the config.yaml, then the `java.lang.IllegalStateException: Could not find flows needed for execution in order` is thrown:

```
java.lang.IllegalStateException: Could not find flows needed for execution in order: flow_b, flow_c
    at maestro.orchestra.workspace.ExecutionOrderPlanner.getFlowsToRunInSequence(ExecutionOrderPlanner.kt:21)
    at maestro.orchestra.workspace.WorkspaceExecutionPlanner.plan(WorkspaceExecutionPlanner.kt:103)
    at maestro.cli.command.TestCommand.call(TestCommand.kt:122)
    at maestro.cli.command.TestCommand.call(TestCommand.kt:46)
    at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
    at picocli.CommandLine.access$1200(CommandLine.java:145)
    at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
    at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
    at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
    at maestro.cli.DisableAnsiMixin$Companion.executionStrategy(DisableAnsiMixin.kt:22)
    at picocli.CommandLine.execute(CommandLine.java:2058)
    at maestro.cli.AppKt.main(App.kt:117)
```

**To Reproduce**
Flow set-up:
testFolder/
   flow_a.yaml > tags: - includeTest
   flow_b.yaml > tags: - excludeTest
   flow_c.yaml > tags: - excludeTest
   flow_d.yaml > tags: - includeTest
```
config set-up:
# NOTE: No Global Tags
executionOrder:
  continueOnFailure: true
  flowsOrder:
    - flow_a
    - flow_b
    - flow_c
    - flow_d
```

1. Start an android emulator
2. Run `maestro test --include-tags=includeTest --exclude-tags=excludeTest ` testFolder/

**Expected behavior**
TBC - However, I would expect if we are running against include & exclude tags, that we do not face a exception with the execution order because we have excluded a flow that is defined in the sequential ordering.

- Maestro version [v1.36.0] (and previous versions)
- Platform: [iOS & Android]
- Framework: [Native iOS/Android]
- Device model and OS version: [any]
- Simulator & emulator
- Host [Apple MacBook Pro M2 Max]

**Additional context**
If we want to keep an execution order e.g. `flow_a > flow_b > flow_c > flow_d` but run against different including & excluding tags, I would expect that we would maintain the execution order but omit flows from this execution order that have been excluded with the exclude tag e.g. `flow_a > ~~flow_b > flow_c >~~ flow_d`.

Mohanad49:
## Proposed changes

Fixes #1681: `executionOrder.flowsOrder` breaks when include/exclude tags filter out one of the ordered flows.

`ExecutionOrderPlanner` only receives the flows that survived tag filtering, so it can't tell a flow that was **excluded on purpose** from one that **doesn't exist**. With `flowsOrder: [flowA, flowB, flowC, flowD]`:

| Excluded by tags | Before |
|---|---|
| `flowA` (first) | `IllegalStateException: Could not find flows needed for execution in order: flowA` — the #1681 crash, also what `--include-tags=checkout` hits in the follow-up comment |
| `flowB` (middle) | no error; sequence stops at `flowA`, and **`flowC`/`flowD` silently move to the unordered pool** — so they can run before `flowA`, or keep running after it fails with `continueOnFailure: false` |

The second row isn't in the issue, but it's the more dangerous one: nothing reports it, and ordering usually exists because later flows depend on earlier ones.

This is two commits, deliberately separable.

### 1. Skip flows excluded by tags (fixes #1681)

In `WorkspaceExecutionPlanner`, flows removed by include/exclude tags drop out of `flowsOrder` before planning, so the remaining flows keep their relative order. A name that matches no flow at all is still treated as missing, exactly as today.

This commit alone fully fixes #1681.

### 2. Enforce the documented missing-flow rule

`ExecutionOrderPlannerTest` already defines the rule: missing flows at the **end** of the order are ignored, but a missing flow **before a present one** is an error. The implementation only enforced it when the *first* flow was missing — `takeWhile` took the longest present prefix, so `[flowC, flowD, flowA]` with `flowD` missing returned `[flowC]` and dropped `flowA` into the unordered pool with no error. Both existing error tests happen to have the first flow missing, which is how it slipped through.

It also compared that prefix against the de-duplicated order, so listing a flow twice (`[flowA, flowA, flowB]`) made the planner return an empty sequence — the whole order silently ignored.

The planner now checks for a gap anywhere before the last present flow, and ignores repeated names. Existing error messages are unchanged.

**This commit makes some configs fail that currently run**, so please weigh it separately — happy to drop it from this PR. The main case I'd expect in practice: an ordered flow excluded by a `flows:` glob rather than by tags. Commit 1 doesn't cover globs (that would mean parsing the config of files the glob skipped, to resolve custom `name:`s), so today that case silently demotes later flows, and with commit 2 it errors. If you'd rather globs behave like tags, I can extend commit 1 instead.

## Testing

Tests were written first and checked to fail on unmodified `main` for the stated reason:

| Test | On `main` |
|---|---|
| `025 - Execution order skips a first flow excluded by tags` | `IllegalStateException: …flowA` |
| `025 - Execution order is respected when include tags leave out an ordered flow` | `IllegalStateException: …flowA` |
| `025 - Execution order keeps later flows in sequence when a middle flow is excluded by tags` | `flowC`, `flowD` missing from sequence |
| `025 - Execution order is unchanged when no ordered flow is excluded` | passes (regression guard) |
| `if a missing flow is followed by a present one it should return an error even when the first is present` | nothing thrown |
| `if a flow is listed more than once it should keep the order of its first appearance` | `[]` |

With the changes:

- `:maestro-orchestra:test` — **443 tests, 0 failures**, including the existing `013 - Execution order is respected` and both existing error tests with their exact messages
- `WorkspaceExecutionPlannerErrorsTest` — 27/27
- `:maestro-cli:test --tests TestCommandTest` — 12/12
- Commit 1 was also run on its own: planner tests 27/27 and error tests 27/27, with only commit 2's two unit tests failing

New fixture `025_execution_order_with_tags` — numbered 025 to avoid colliding with #3552, which adds `022`–`024`. #3552 also edits `WorkspaceExecutionPlanner.kt`, in the tag-filtering block just above the lines changed here, so whichever lands second may need a small rebase.

No e2e flow added: this is planning logic that runs before any device is selected, and it's covered at the planner level.

## Issues fixed

Fixes #1681


Mohanad49:
A note on the red e2e checks, since this PR touches flow planning and that's the obvious first suspect:

- **No e2e workspace uses `executionOrder`**, so the changed path never runs there. Android planned the same **55** `demo_app` flows as `main` — #3598 and #3599 report 55/55 too.
- **Test on Android** fails only on `swipe` (`"All green" is visible`), 1 of 55 — the same flow has flaked on other PRs.
- **Test on iOS** fails on `ios-advanced-flow` in the `wikipedia` workspace (`Element not found: Get started`). Every iOS run since about 15:40 UTC today has failed there, on unrelated PRs, while this morning's runs passed. The failing runs' logs also list an iOS 26.4 simulator runtime that the passing ones don't, so a runner image update looks likely.
- **E2E gate** only reflects the two above.

Fishbowler:
There's a bunch of opinionated design in tags, and I'm not sure it's quite perfect yet.

Designing to either error on missing entries or to continue is opinionated, and I'm fine with continuing, but if we continue we need to warn loudly at the top of the run that we're doing so with some ordered flows missing.

That also applies to 2 conditions that pre-date and survive this change:
- The trailing missing flow in the order silently dropping off
- When 100% of the flows are missing, and so everything else just runs unordered

I'm wondering if we shouldn't simplify it as "missing things are wrong" and gate the subsequent behaviour behind a CLI flag or something? e.g. `--warn-on-missing-ordered flows` - use the flag to warn rather than error? If someone runs `--include-tags checkout` and their flowsOrder contains non-checkout flows, then they can manage that with `--include-tags essential,checkout`

WDYT?

Mohanad49:
Thanks for the review, and for the rebase.

Agreed on warning loudly — that part I'd do regardless of which way the rest goes. Where I'd push back slightly is on folding everything into one "missing", because I think there are two different situations behind that word:

**1. A name in `flowsOrder` that matches no flow anywhere in the workspace** — a typo, or a flow that was renamed or deleted. There's no sensible recovery: whatever the user meant, we can't run it. So: error, in *every* position.

That's your "missing things are wrong", and it closes both of the pre-existing holes you listed, because it removes the positional subtlety entirely. Today the same typo behaves three different ways depending on where it sits: error in the middle, silently dropped at the end, and "run everything unordered" if it's the only entry. Under this rule all three are one thing, and it's an error.

**2. A flow that exists but was filtered out of *this run*** — by `--include-tags` / `--exclude-tags`. Nothing is wrong here; the user asked for exactly this. Erroring makes tag filtering and `flowsOrder` mutually exclusive, which is the #1681 crash. So: continue, and warn loudly at the top of the run, naming which ordered flows were skipped and that ordering is therefore partial.

Split that way the behaviour is decidable from the workspace itself, without asking the user which mode they want — which is why I'd lean away from a flag to start with. `--warn-on-missing-ordered-flows` would only ever soften case 1, and case 1 is precisely where the error earns its keep. Easy to add later if someone turns up with a real need for the escape hatch; harder to remove once it's public API.

On `--include-tags essential,checkout` as the workaround: that's the right answer when the ordered flows are genuine dependencies (`login` before `checkout`), and I'd expect people to reach for it there. It reads worse when the ordered flows just aren't part of the selection — `flowsOrder: [login, checkout, refund]` with `--include-tags checkout` would mean re-adding `refund` to the run purely to satisfy the ordering config, which is the opposite of what the filter was for.

One limitation I should be upfront about, since it decides how clean this actually is: only *tag*-excluded flows can be resolved by name cheaply. A flow excluded by the workspace `flows:` glob is never parsed, so its custom `name:` isn't known, and it would land in case 1 and error. Two options — match glob-excluded files by filename as a fallback (which is what a flow's name defaults to anyway, so it covers most real configs and degrades to an error only for a custom `name:`), or leave glob-excluded as a hard error and say so in the message. Mild preference for the filename fallback, but I don't feel strongly, and the second is at least honest about what we can know.

If the split above sounds right, I'll push: error on unresolvable names in all positions, a warning listing the ordered flows skipped by filters, and tests covering the trailing-entry and all-entries-missing cases you raised. Say the word if you'd rather have it as error-plus-flag and I'll write it that way instead — it's your call on the surface area.

Fishbowler:
Howdy Claude. I know that tone anywhere 😂

> Nothing is wrong here; the user asked for exactly this.

Did they? I feel that's an assumption. Maybe they wanted to run the tagged flows after running the explicit 5 flows in the flowsOrder

As you say, the config-excluded files aren't currently read, and I don't think that including them for the purposes of defining the flavour of failure state is the way to go.

Mohanad49:
Yup, that was Claude tone 😂

Yeah, you're right, that was an assumption. I was reading flowsOrder as "order these, if they run", but "run these five, then whatever the tags picked" is just as valid and nothing in the config says which one was meant.

Agreed on not reading the excluded files too. Beyond the cost, it'd mean the outcome depends on how a flow got excluded (tags, we know the name; glob, we don't), and that's not a distinction anyone could predict from the outside.

So let's do it your way: drop my first commit, and anything in flowsOrder that isn't in this run is an error wherever it sits in the list, with a flag to downgrade that to a loud warning and run the rest in order. That flag will swallow typos as well as deliberate filtering, which I don't love, but I don't think the two can be told apart cheaply enough to be worth it.

One thing I'd want your call on before writing it: with error as the default, #1681 doesn't really get fixed, it just gets a clear message and a documented opt-in. Happy for that to be the resolution? If you'd rather it works out of the box, the default flips and the flag becomes the strict one instead.

Also a heads up, both conditions you spotted are pinned by existing tests (the trailing-missing one, and "if no paths are present in the sequence it should return an empty list"), so this is changing asserted behaviour rather than filling in gaps.
I still think it's the right change, just don't want to sneak that past you.

Either way the case I actually care about gets fixed: right now a middle flow being excluded drops flowC and flowD into the unordered pool silently, so they can run before flowA, or keep running after it fails with continueOnFailure: false.

If you do want "flows in flowsOrder always run, tags or not", I think that's a bigger semantic change and probably its own issue rather than something to bolt onto this.

Issue #1373, opened 2023-08-22
[v1.31.0] hideKeyboard has different behaviour based on platform


arashgm:
**Describe the bug**
hideKeyboard on iOS simulator triggers the action event of the soft keyboard which leads to different behaviour between Android and iOS , on Android it just close the keyboard without triggering the keyboard action

**Expected behavior**
hideKeyboard should only hide the keyboard not triggering the Done action , or having the exact same behaviour on both platform

**Screenshots**

**Environment information (please complete the following information):**

- Maestro version [ev1.31.0]
- Platform: [e.g. iOS 16.0, Android 12.0]
- Framework: [Flutter]
- Device model and OS version: [iPhone14 simulator][onePlus 7]
- Simulator and physical device 
- Host [Mac M2]

PR #3605, opened 2026-09-16
fix(hideKeyboard): skip the dismiss when no keyboard is visible

### LINKED ISSUE #1303: [1.30.4] Back is pressed instead of keyboard hidden if it's performed too quickly
**Describe the bug**
Let's open a screen with EditText, which is focused automatically. Then, instantly hide the on-screen keyboard with `hideKeyboard` command. Usually, it works fine - we end up with focused EditText which we can feed with text and with complete visibility of the rest of the screen. However, sometimes `hideKeyboard` performs `back` before the keyboard is fully visible which navigates the app to the previous screen, so it's kind of a race condition.

**Expected behavior**
`hideKeyboard` should ensure that the on-screen keyboard is visible before trying to hide it to reduce flakiness.

- Maestro version: [1.30.4]
- Platform: [Android 13]
- Framework: [Native Android]
- Device model and OS version: Android Emulator based on Android 13
- Host: Mac M1

mhammadjaber00:
## Proposed changes

`hideKeyboard` presses Back on Android unconditionally:

```kotlin
// AndroidDriver.kt
override fun hideKeyboard() {
    metrics.measured("operation", mapOf("command" to "hideKeyboard")) {
        shell("input keyevent 4") // 'Back', which dismisses the keyboard before handing over to navigation
        ...
```

When it runs and no IME is on screen, that Back press falls through to navigation and pops the current screen. Everything after it in the flow then runs against the wrong screen. This is #1303 — the race the reporter hit is one way in, but any `hideKeyboard` on a screen where the keyboard happens to already be down does it, which makes it bite on conditional paths and shared subflows that cannot know whether a field is focused.

`Orchestra.hideKeyboardCommand()` already asks the driver whether the keyboard is visible — but only *after* the dismiss, to raise `HideKeyboardFailure`. Its own debug message tells you to tap a static element instead, which is what people end up doing.

This PR asks the same question up front and returns early when there is nothing to dismiss:

```kotlin
private suspend fun hideKeyboardCommand(): Boolean {
    // Nothing to dismiss: on Android the dismiss is a back press, which would navigate instead
    if (!maestro.isKeyboardVisible()) {
        return true
    }

    maestro.hideKeyboard()
    ...
```

Because the guard is in `Orchestra` rather than in `AndroidDriver`, it also covers the iOS half of #1373: with no keyboard up, `hideKeyboard` no longer fires the keyboard's action event.

**One dependent change.** `AndroidDriver.isKeyboardVisible()` hardcoded Gboard's package:

```kotlin
"com.google.android.inputmethod.latin:id" in jacksonObjectMapper().writeValueAsString(root)
```

On a device with any other IME (Samsung Keyboard, SwiftKey, …) that is always `false`. Today it fails safe — the post-dismiss check just never throws. With the new early return it would fail *unsafe*: `hideKeyboard` would become a silent no-op on those devices. So this PR looks up the IME that is actually configured — via the existing `currentInputMethod()` helper — and matches against that package instead. On a Gboard device the two are the same string, so behaviour there is unchanged.

I also tried replacing the hierarchy scan with `dumpsys input_method | grep mInputShown=true`, which is the package-agnostic signal most Android harnesses use. Dropping it: on my emulator that dump had grown to ~1 MB and took **4–8 s** per call after a while. Not viable for a per-command check.

## Testing

- `./gradlew test` on Java 17 (CI's version), `ANDROID_HOME` set — green.
- New unit test `hideKeyboard is skipped when the keyboard is not visible` in `IntegrationTest`, with a `keyboardInitiallyVisible` seam on `FakeDriver`. It asserts no `Event.HideKeyboard` is recorded, i.e. nothing that could navigate. The two existing `hideKeyboard` tests are untouched and still pass — the seam defaults to `true`.
- E2E: extended `e2e/demo_app/.maestro/commands/hideKeyboard.yaml` with a second `hideKeyboard` after the keyboard is already down, then `assertVisible: 'Email'`. Before this change that second dismiss pops the Form Test screen and the assert fails. **I could not run this locally** — the Maestro device server dies at the first `deviceInfo` on my API 37 emulator, with stock 2.8.0 as well as with this branch, so it is not related to this change. The flow only touches existing selectors and `test-e2e.yaml` is `workflow_dispatch`, so it will not surprise PR CI, but it does want a dispatch run before you trust it.
- Verified separately on the emulator that the underlying premise holds: with the keyboard up, `mInputShown=true` / `mImeWindowVis=3`; after `input keyevent 4`, `false` / `0` — and the same key event with the keyboard already down navigates the app back.

## Issues fixed

Fixes #1303. Also fixes the no-keyboard case of #1373.


Fishbowler:
I love this! I can't believe that hardcoded IME has been hiding in there for so long!

We're doing some driver work behind the scenes, so can't merge this _quite_ yet. We'll either incorporate it once we're done, or fold it into the work we're doing. cc @steviec @amanjeetsingh150

